PR #4, opened 2020-10-23
C# 6.0 feature: expression bodied members

BillWagner:
Enable members to be expressions in addition to block statements.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -1616,6 +1616,7 @@ method_body
     : block
     | '=>' expression ';'
     | ';'
+    ;
 ```
 A *method_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and a valid combination of the four access modifiers ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `static` ([§15.6.3](classes.md#1563-static-and-instance-methods)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods)), `override` ([§15.6.5](classes.md#1565-override-methods)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods)), `extern` ([§15.6.8](classes.md#1568-external-methods)) and `async` ([§15.15](classes.md#1515-async-functions)) modifiers.
 
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -2502,7 +2502,7 @@ Abstract and external method declarations do not provide a method implementation
 
 The ***effective return type*** of a method is `void` if the return type is `void`, or if the method is async and the return type is `System.Threading.Tasks.Task`. Otherwise, the effective return type of a non-async method is its return type, and the effective return type of an async method with return type `System.Threading.Tasks.Task<T>` is `T`.
 
-When the effective return type of a method is `void` and the method has a block body, `return` statements ([§13.10.5](statements.md#13105-the-return-statement)) in the block are not permitted to specify an expression. If execution of the block of a void method completes normally (that is, control flows off the end of the method body), that method simply returns to its caller.
+When the effective return type of a method is `void` and the method has a block body, `return` statements ([§13.10.5](statements.md#13105-the-return-statement)) in the block shall not specify an expression. If execution of the block of a void method completes normally (that is, control flows off the end of the method body), that method simply returns to its caller.
 
 When a method has a `void` result and an expression body, the expression `E` shall be a *statement_expression*, and the body is exactly equivalent to a statment body of the form `{ E; }`.
 
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -2507,7 +2507,7 @@ When a method has a `void` result and an expression body, the expression `E` sha
 
 When the effective return type of a method is not `void` and the method has a block body, each return statement in that method's body shall specify an expression that is implicitly convertible to the effective return type. The endpoint of the method body of a value-returning method shall not be reachable. In other words, in a value-returning method with a block body, control is not permitted to flow off the end of the method body.
 
-When the effective return type of a method is not `void` and the method has an expression body, the expression shall be implicitly convertible to the effective return type, and the body is exactly equivalent to a block body of the form `{ return E; }`.
+When the effective return type of a method is not `void` and the method has an expression body, `E`, the expression shall be implicitly convertible to the effective return type, and the body is exactly equivalent to a block body of the form `{ return E; }`.
 
 > *Example*: In the following code
 > ```csharp
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -1617,6 +1617,7 @@ method_body
     | '=>' expression ';'
     ;
 ```
+
 A *method_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and a valid combination of the four access modifiers ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `static` ([§15.6.3](classes.md#1563-static-and-instance-methods)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods)), `override` ([§15.6.5](classes.md#1565-override-methods)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods)), `extern` ([§15.6.8](classes.md#1568-external-methods)) and `async` ([§15.15](classes.md#1515-async-functions)) modifiers.
 
 A declaration has a valid combination of modifiers if all of the following are true:
@@ -1640,7 +1641,7 @@ The *return_type* and each of the types referenced in the *formal_parameter_list
 
 The *method_body* is either a semicolon, a ***statement body*** or an ***expression body***. A statement body consists of a *block*, which specifies the statements to execute when the method is invoked. An expression body consists of `=>` followed by an *expression* and a semicolon, and denotes a single expression to perform when the method is invoked.
 
-For abstract and extern methods, the *method_body* consists simply of a semicolon. For partial methods the *method_body* may consist of either a semicolon, a statement body or an expression body. For all other methods, the *method_body* is either a statement body or an expression body.
+For abstract and extern methods, the *method_body* consists simply of a semicolon. For partial methods the *method_body* may consist of either a semicolon, a block body or an expression body. For all other methods, the *method_body* is either a block body or an expression body.
 
 If the *method_body* consists of a semicolon, the declaration shall not include the `async` modifier.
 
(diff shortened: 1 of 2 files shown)


standard/classes.md
```diff
@@ -1612,7 +1612,8 @@ member_name
 
 method_body
     : block
-    ;
+    | '=>' expression ';'
+    | ';'
```
Nigel-Ecma, line 1618: I think this should either be just a semicolon (the deleted 1615) or the following line should be a semicolon – the grammar rule must end in one.

standard/classes.md
```diff
@@ -1635,7 +1636,9 @@ The optional *formal_parameter_list* specifies the parameters of the method ([§
 
 The *return_type* and each of the types referenced in the *formal_parameter_list* of a method shall be at least as accessible as the method itself ([§8.5.5](basic-concepts.md#855-accessibility-constraints)).
 
-For abstract and extern methods, the *method_body* consists simply of a semicolon. For partial methods the *method_body* may consist of either a semicolon or a *block*. For all other methods, the *method_body* consists of a *block,* which specifies the statements to execute when the method is invoked.
+The *method_body* is either a semicolon, a ***statement body*** or an ***expression body***. A statement body consists of a *block*, which specifies the statements to execute when the method is invoked. An expression body consists of `=>` followed by an *expression* and a semicolon, and denotes a single expression to perform when the method is invoked.
+
+For abstract and extern methods, the *method_body* consists simply of a semicolon. For partial methods the *method_body* may consist of either a semicolon, a block body or an expression body. For all other methods, the *method_body* is either a block body or an expression body.
```
Nigel-Ecma, line 1645: "a statement body or an expression body" to match the previous para?

standard/classes.md
```diff
@@ -2490,15 +2493,19 @@ An extension method is a regular static method. In addition, where its enclosing
 
 ### 15.6.11 Method body
 
-The *method_body* of a method declaration consists of either a *block* or a semicolon.
+The *method_body* of a method declaration consists of either a block body, an expression body or a semicolon.
```
Nigel-Ecma, line 2500: Same comment as the last one and its going to occur further down as well – is it a "statement body" or a "block body"?

standard/classes.md
```diff
@@ -3437,13 +3456,17 @@ conversion_operator_declarator
...
+The *operator_body* is either a semicolon, a statement body ([§15.6.1](classes.md#1561-general)) or an expression body ([§15.6.1](classes.md#1561-general)). A statement body consists of a *block*, which specifies the statements to execute when the operator is invoked. The *block* shall conform to the rules for value-returning methods described in [§15.6.11](classes.md#15611-method-body). An expression body consists of `=>` followed by an expression and a semicolon, and denotes a single expression to perform when the operator is invoked.
+
+For `extern` operators, the *operator_body* consists simply of a semicolon. For all other operators, the *operator_body* is either a block body or an expression body.
```
Nigel-Ecma, line 3477: Looks to be the last use of "block body" in the changes which maybe should be "statement body" or vice-versa

standard/classes.md
jskeet: +1

jskeet: Given line 1639, I think the `';'` is intentional, and there should be an *extra* line of just `;` to terminate the grammar rule.

standard/classes.md
jskeet: +1

standard/classes.md
```diff
@@ -2561,7 +2573,11 @@ The *type* of a property declaration specifies the type of the property introduc
 
 The *type* of a property shall be at least as accessible as the property itself ([§8.5.5](basic-concepts.md#855-accessibility-constraints)).
 
-The *accessor_declarations*, which shall be enclosed in "`{`" and "`}`" tokens, declare the accessors ([§15.7.3](classes.md#1573-accessors)) of the property. The accessors specify the executable statements associated with reading and writing the property.
+A *property_body* may either consist of an ***accessor body*** or an expression body ([§15.6.1](classes.md#1561-general)). In an accessor body,  *accessor_declarations*, which shall be enclosed in "`{`" and "`}`" tokens, declare the accessors ([§15.7.3](classes.md#1573-accessors)) of the property. The accessors specify the executable statements associated with reading and writing the property.
+
+An expression body consisting of `=>` followed by an *expression* `E` and a semicolon is exactly equivalent to the statement body `{ get { return E; } }`, and can therefore only be used to specify getter-only properties where the result of the getter is given by a single expression.
```
jskeet: Note to self: this is correct because we're only updating to C# 6. In C# 7 of course, there can be expression-bodied setters as well.

jskeet (commented):
Looks okay to me apart from the inconsistency Nigel has spotted (and the grammar line).

standard/classes.md
jskeet, line 1618:
```suggestion
    | ';'
    ;
```

standard/classes.md
RexJaeschke: @jskeet I just looked at PR #69, which I created for v7 edits in this area, and we don't use either term there.

jskeet:
I don't think there have been any changes since the last meeting - so this meeting, let's just make sure we have concrete plans.

jskeet:
Next work for @BillWagner:

- Addressing comments
- Change statement body to block body

standard/classes.md
BillWagner: Rex has already made this change.

standard/classes.md
BillWagner: changed in both locations in this paragraph.

standard/classes.md
BillWagner: Replaced in these paragraphs, and later in the "operators" clause.

standard/classes.md
BillWagner: replaced with "statement body"

BillWagner:
@Nigel-Ecma @jskeet All changes requested have been applied in the **Classes** clause.

I did not change 12.6.1 (Expressions, anonymous functions, general). The term *body* is used in the grammar in that clause. I think we can replace it with *statement*, and replace "*block* body" with "*statement* body" in the subsequent paragraph. If we agree, I'll make that change. Otherwise this should be ready to merge and mark as complete for V6.

standard/classes.md
jskeet: It still looks wrong to me. `';'` is not a valid `method_body`, and this grammar rule doesn't have a closing semi-colon.

If what's in the change now ends up being what's in the standard, there's a problem. If you're saying this is resolved by another PR that hasn't yet been merged, I think we need to work on processes a bit more... I think we need to be able to review "the doc as it is expressed in this PR".

BillWagner: That's what I thought as well @jskeet That's how I interpreted your comment. The suggestion you made added the terminating `;` but didn't remove the `| ';'` production. I mis-interpreted the combination. Last commit removes the `;` as a valid method body.

jskeet:
Thanks Bill. I now can't find the right comment to reply to... and I'm confused about why I saw it still with `| ;` but *without* the `;` on its own... Not sure what's going on there.
It's just possible that we'll need to put it back in at some point for abstract methods and interfaces of course - I'll check that tomorrow. (If that's required, then presumably it's a bug from earlier versions.)

BillWagner:
@jskeet 

I think you're last comment is correct: The grammar should include the `;` production for both abstract and partial methods.

(On the other note about finding the conversations:  When I make the update, I 'resolve' the conversation, which hides it in the UI. You can "view resolved conversation" to see them, then you can un-resolve as well.

Nigel-Ecma:
@jskeet @BillWagner

- The conversion block body -> statement body looks good
- Agree *method_body* should have an `| ';'` alternative
- In §12.6.1 *block* can be replaced by *statement*, that's not right or what we've done in *method_body* – we've called the *block* option a "statement body". So as suggested in the subsequent para I think "block body" does become "statement body". Also in §12.6.3.15, *Finding the best common type of a set of expressions*, I think "*block* bodies" should become "statement bodies".

jskeet:
@Nigel-Ecma: I think it would be better to use "block body" everywhere that we're referring to something with a block. What's your reasoning for preferring "statement body" in some places?

Nigel-Ecma:
@jskeet @BillWagner 

No preference, I assumed I'd missed something...

Jon while your Feb 11 comment above records the decision was to go with *block body*; it includes  the note *"Statement-bodied lambda" is a somewhat-well-known-phrase*...

When I came to review I noted that the edit was the other way around so I assumed that note or something else had made the decision go the other way, something I don't recall but could have forgotten, so I just checked the edit was consistent and nothing missed.

I don't have a strong preference either way, just that it must be consistent, which it now is.

Bill?

jskeet:
Right - I think @BillWagner took things in the opposite direction of what I'd expected. (Thanks for spotting this.)

We were definitely looking to make "block body" the canonical term, as per an email I sent on Feb 11th:

> We're aware that "statement-bodied lambda expression" is a term in relatively wide use... but we believe we can be more self-explanatory and consistent if we use "block body" everywhere, effectively removing any trace of "statement body" from the standard (other than possibly adding a note).

BillWagner:
@jskeet @Nigel-Ecma 

Good catch. I did reverse that.  I'll update this PR now.

Nigel-Ecma (approved):
Now is consistent the other way around :-)

standard/classes.md
```diff
@@ -1614,8 +1614,10 @@ member_name
 
 method_body
     : block
+    | '=>' expression ';'
```
jskeet, line 1617: I'm really confused now. The semi-colon on the next line is just the "end of rule" indicator, yes? So do we actually need another line of:

```text
| ';'
```

so that a semicolon counts as a body on its own? (We describe that as being valid at line 1642.)

jskeet: (And to be clear, I suspect this was a bug before this PR, so I'm happy to punt it to a separate issue.)

standard/classes.md
```diff
@@ -2492,15 +2496,19 @@ An extension method is a regular static method. In addition, where its enclosing
...
 
 The ***effective return type*** of a method is `void` if the return type is `void`, or if the method is async and the return type is `System.Threading.Tasks.Task`. Otherwise, the effective return type of a non-async method is its return type, and the effective return type of an async method with return type `System.Threading.Tasks.Task<T>` is `T`.
 
-When the effective return type of a method is `void`, `return` statements ([§13.10.5](statements.md#13105-the-return-statement)) in that method's body are not permitted to specify an expression. If execution of the method body of a void method completes normally (that is, control flows off the end of the method body), that method simply returns to its caller.
+When the effective return type of a method is `void` and the method has a statment body, `return` statements ([§13.10.5](statements.md#13105-the-return-statement)) in the block are not permitted to specify an expression. If execution of the block of a void method completes normally (that is, control flows off the end of the method body), that method simply returns to its caller.
```
```suggestion
When the effective return type of a method is `void` and the method has a block body, `return` statements ([§13.10.5](statements.md#13105-the-return-statement)) in the block are not permitted to specify an expression. If execution of the block of a void method completes normally (that is, control flows off the end of the method body), that method simply returns to its caller.
```

jskeet: (Should "are not be permitted to" be "shall not"?)

standard/classes.md
jskeet: (I note that operator_body *does* have a `| ';'` option btw.)

jskeet (commented):
Very nearly there - just a couple of nits.

standard/classes.md
BillWagner: I think we *do* need the `|  ;` addition.  I made that change.

jskeet (approved):

PR #6, opened 2020-10-23
C# 6.0: improved overload resolution

BillWagner:
This might be a bug fix we examined in committee, but the notes do not indicate that.
It's also the only change in the area of overload resolution.

On the other hand, this does not seem to be applicable to the improved resolution introduced in C# 6.0


Nigel-Ecma (approved):
The clarification LGTM, though I'm a bit confused by the opening comment regarding whether this is all to be replaced anyway?

jskeet (approved):
To Nigel's point, I don't expect all of this to be replaced - I expect there to be further changes within overload resolution, but I expect them to be scattered around.

PR #10, opened 2020-10-23
C# 6.0 feature: nameof expressions

BillWagner:
Work done to date on nameof expressions.

Creating draft PR because of outstanding TODO item.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1122,6 +1122,7 @@ primary_no_array_creation_expression
     | checked_expression
     | unchecked_expression
     | default_value_expression
+    | nameof_expression    
     | anonymous_method_expression
     ;
 ```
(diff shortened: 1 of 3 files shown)


jskeet (commented):
I think we probably need to talk about Nigel's comments - mine are somewhat more simplistic, I think :)

jskeet:
Local variable part: it's mentioned in https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/statements.md#1362-local-variable-declarations:

> The scope of a local variable declared in a *local_variable_declaration* is the block in which the declaration occurs. It is an error to refer to a local variable in a textual position that precedes the *local_variable_declarator* of the local variable. Within the scope of a local variable, it is a compile-time error to declare another local variable or constant with the same name.

That section of the standard is quite a long way from name lookup, so I'm not sure it should really apply within `nameof`, but at least we don't have the bigger problem I thought we might have.

BillWagner:
closing in favor of #250 (which builds on this branch.

PR #11, opened 2020-10-24
C# 6.0 feature: add specification for readonly auto properties

BillWagner:
This also include the feature for auto property initializers.


BillWagner:
I resolved the conflicts today by rebasing this branch off the current V6 draft.

standard/classes.md
```diff
@@ -2855,7 +2864,11 @@ Properties can be used to delay initialization of a resource until the moment it
...
+An automatically implemented property (or ***auto-property*** for short), is a non-abstract, non-extern property with semicolon-only accessor bodies. Auto-properties shall have a get accessor and may optionally have a set accessor.
+
+When a property is specified as an automatically implemented property, a hidden backing field is automatically available for the property, and the accessors are implemented to read from and write to that backing field. The hidden backing field is inaccessible, it can be read and written only through the automatically implemented property accessors, even within the containing type. If the auto-property has no set accessor, the backing field is considered `readonly` ([§15.5.3](classes.md#1553-readonly-fields)). Just like a `readonly` field, a getter-only auto-property may also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the readonly backing field of the property.
```
BillWagner: Should the final "readonly" be in code font?

```suggestion
When a property is specified as an automatically implemented property, a hidden backing field is automatically available for the property, and the accessors are implemented to read from and write to that backing field. The hidden backing field is inaccessible, it can be read and written only through the automatically implemented property accessors, even within the containing type. If the auto-property has no set accessor, the backing field is considered `readonly` ([§15.5.3](classes.md#1553-readonly-fields)). Just like a `readonly` field, a getter-only auto-property may also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the `readonly` backing field of the property.
```

standard/variables.md
```diff
@@ -538,7 +538,7 @@ For an expression *expr* of the form:
 
 - The definite assignment state of *v* before *w* is the same as the definite assignment state of *v* before *expr*.
 - The definite assignment state of *v* before *expr_rhs* is the same as the definite assignment state of *v* after *w*.
-- If *w* is the same variable as *v*, then the definite assignment state of *v* after *expr* is definitely assigned. Otherwise, the definite assignment state of *v* after *expr* is the same as the definite assignment state of *v* after *expr_rhs*.
+- If *w* is the same variable as *v*, then the definite assignment state of *v* after *expr* is definitely assigned. Otherwise, if the assignment occurs within the instance constructor of a struct type, if *w* is a property access designating an automatically implemented property *P* on the instance being constructed and *v* is the hidden backing field of *P*, then the definite assignment state of *v* after *expr* is definitely assigned. Otherwise, the definite assignment state of *v* after *expr* is the same as the definite assignment state of *v* after *expr_rhs*.
```
BillWagner: Should this be "and" instead of ", if":

```suggestion
- If *w* is the same variable as *v*, then the definite assignment state of *v* after *expr* is definitely assigned. Otherwise, if the assignment occurs within the instance constructor of a struct type, and *w* is a property access designating an automatically implemented property *P* on the instance being constructed and *v* is the hidden backing field of *P*, then the definite assignment state of *v* after *expr* is definitely assigned. Otherwise, the definite assignment state of *v* after *expr* is the same as the definite assignment state of *v* after *expr_rhs*.
```

BillWagner (commented):
Reviewed the text. I have two questions we can resolve either by vote or on our next meeting.

Overall, this LGTM.

standard/classes.md
```diff
@@ -2575,6 +2575,15 @@ property_body
...
 
+property_initializer
+    : '=' variable_initializer ';'
+    ;
+
+property_body
+    : '{' accessor_declarations '}' property_initializer?
+    | '=>' expression ';'
+    ;
+
```
Nigel-Ecma: property_initializer & property_body have been duplicated, I'd blame git ;-)

standard/classes.md
```diff
@@ -2655,7 +2664,7 @@ The use of *accessor_modifier*s is governed by the following restrictions:
...
 
-For `abstract` and `extern` properties, the *accessor_body* for each accessor specified is simply a semicolon. A non-abstract, non-extern property may be an ***automatically implemented property***, in which case both `get` and `set` accessors shall be given, both with a semicolon body ([§15.7.4](classes.md#1574-automatically-implemented-properties)). For the accessors of any other non-abstract, non-extern property, the *accessor_body* is a *block* that specifies the statements to be executed when the corresponding accessor is invoked.
+For `abstract` and `extern` properties, the *accessor_body* for each accessor specified is simply a semicolon. A non-abstract, non-extern property may have each *accessor_body* be a semicolon, in which case it is an ***automatically implemented property*** ([§15.7.4](classes.md#1574-automatically-implemented-properties)). An automatically implemented property shall have at least a get accessor. For the accessors of any other non-abstract, non-extern property, the *accessor_body* is a *block* which specifies the statements to be executed when the corresponding accessor is invoked.
```
Nigel-Ecma: The second sentence read to me as each accessor must exist – not quite our aim here, maybe echo the first and write:

> A non-abstract, non-extern property may also have the *accessor_body* for all accessors specified be a semicolon, in which case it is an ***automatically implemented property*** ([§15.7.4](classes.md#1574-automatically-implemented-properties)).

standard/classes.md
```diff
@@ -2966,7 +2991,7 @@ A property declaration that includes both the `abstract` and `override` modifier
...
+The override declaration and the overridden base property have the same declared accessibility. In other words, an override declaration cannot change the accessibility of the base property. However, if the overridden base property is protected internal and it is declared in a different assembly than the assembly containing the override declaration then the override declaration's declared accessibility shall be protected. If the inherited property has only a single accessor (i.e., if the inherited property is read-only or write-only), the overriding property shall include only that accessor. If the inherited property includes both accessors (i.e., if the inherited property is read-write), the overriding property can include either a single accessor or both accessors.
```
Nigel-Ecma: Is the removal of "there shall be an identity conversion between the type of the overriding and the inherited property" intentional? (I've no idea whether this needs to go)

Nigel-Ecma (changes requested):

standard/classes.md
jskeet: Either that, or "read-only" not in code font.

standard/classes.md
```diff
@@ -2875,15 +2888,27 @@ When a property is specified as an automatically implemented property, a hidden
...
+> public class ReadOnlyPoint
+> {
+>     private readonly int __x;
+>     private readonly int __y;
+>     public int X { get { return __x; } }
+>     public int Y { get { return __y; } }
+>     public ReadOnlyPoint(int x, int y) { __x = x; __y = y; }
+> }
+> ```
+> The assignments to the readonly field are valid, because they occur within the constructor. *end example*
```
jskeet: Again, `readonly` field? Or read-only field?

jskeet (approved):
LGTM with the suggested changes.

standard/classes.md
BillWagner: I applied Jon's suggestion here (and where later mentioned).

standard/classes.md
BillWagner: edit per suggestion.

standard/classes.md
BillWagner: read-only field.

standard/classes.md
BillWagner: I looked over the entire clause, and it looks like that declaration was removed in error. I added it back in.

standard/classes.md
```diff
@@ -2655,7 +2655,7 @@ The use of *accessor_modifier*s is governed by the following restrictions:
...
 
-For `abstract` and `extern` properties, the *accessor_body* for each accessor specified is simply a semicolon. A non-abstract, non-extern property may be an ***automatically implemented property***, in which case both `get` and `set` accessors shall be given, both with a semicolon body ([§15.7.4](classes.md#1574-automatically-implemented-properties)). For the accessors of any other non-abstract, non-extern property, the *accessor_body* is a *block* that specifies the statements to be executed when the corresponding accessor is invoked.
+For `abstract` and `extern` properties, the *accessor_body* for each accessor specified is simply a semicolon. A non-abstract, non-extern property may also have the *accessor_body* for all accessors specified be a semicolon, in which case it is an ***automatically implemented property*** ([§15.7.4](classes.md#1574-automatically-implemented-properties)). An automatically implemented property shall have at least a get accessor. For the accessors of any other non-abstract, non-extern property, the *accessor_body* is a *block* which specifies the statements to be executed when the corresponding accessor is invoked.
```
gafter: Change from "that" to "which" doesn't look correct here.  https://www.grammarly.com/blog/which-vs-that/

```suggestion
For `abstract` and `extern` properties, the *accessor_body* for each accessor specified is simply a semicolon. A non-abstract, non-extern property may also have the *accessor_body* for all accessors specified be a semicolon, in which case it is an ***automatically implemented property*** ([§15.7.4](classes.md#1574-automatically-implemented-properties)). An automatically implemented property shall have at least a get accessor. For the accessors of any other non-abstract, non-extern property, the *accessor_body* is a *block* that specifies the statements to be executed when the corresponding accessor is invoked.
```

standard/classes.md
```diff
@@ -2875,15 +2879,27 @@ When a property is specified as an automatically implemented property, a hidden
...
-Because the backing field is inaccessible, automatically implemented read-only or write-only properties do not make sense, and are disallowed. It is however possible to set the access level of each accessor differently. Thus, the effect of a read-only property with a private backing field can be mimicked like this:
-
-```csharp
-public class ReadOnlyPoint {
-    public int X { get; private set; }
-    public int Y { get; private set; }
-    public ReadOnlyPoint(int x, int y) { X = x; Y = y; }
-}
-```
+> *Example*: The following
```
gafter:
```suggestion
> *Example*: In the following
```

standard/classes.md
```diff
@@ -2966,7 +2982,7 @@ A property declaration that includes both the `abstract` and `override` modifier
...
+The override declaration and the overridden base property have the same declared accessibility. In other words, an override declaration cannot change the accessibility of the base property. However, if the overridden base property is protected internal and it is declared in a different assembly than the assembly containing the override declaration then the override declaration's declared accessibility shall be protected. If the inherited property has only a single accessor (i.e., if the inherited property is read-only or write-only), the overriding property shall include only that accessor. If the inherited property includes both accessors (i.e., if the inherited property is read-write), the overriding property can include either a single accessor or both accessors. There shall be an identity conversion between the type of the overriding and the inherited property.
```
gafter:
```suggestion
The override declaration and the overridden base property are required to have the same declared accessibility. In other words, an override declaration may not change the accessibility of the base property. However, if the overridden base property is protected internal and it is declared in a different assembly than the assembly containing the override declaration then the override declaration's declared accessibility shall be protected. If the inherited property has only a single accessor (i.e., if the inherited property is read-only or write-only), the overriding property shall include only that accessor. If the inherited property includes both accessors (i.e., if the inherited property is read-write), the overriding property can include either a single accessor or both accessors. There shall be an identity conversion between the type of the overriding and the inherited property.
```

gafter (approved):
Looks good with a couple of minor suggested changes.

Nigel-Ecma (approved):
LGTM

jskeet:
Approved in the meeting - @BillWagner to squash and merge.

PR #12, opened 2020-10-24
C# 6.0 feature: interpolated strings

BillWagner:
Add the specification for interpolated strings.

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -35,6 +35,7 @@ The following conversions are classified as implicit conversions:
 - Identity conversions
 - Implicit numeric conversions
 - Implicit enumeration conversions
+- Implicit interpolated string conversions
 - Implicit reference conversions
 - Boxing conversions
 - Implicit dynamic conversions
@@ -88,7 +89,12 @@ There are no predefined implicit conversions to the `char` type, so values of th
 
 An implicit enumeration conversion permits a *constant-expression* ([§12.20](expressions.md#1220-constant-expressions)) with any integer type and the value zero to be converted to any *enum-type* and to any *nullable-value-type* whose underlying type is an *enum-type*. In the latter case the conversion is evaluated by converting to the underlying *enum-type* and wrapping the result ([§9.3.11](types.md#9311-nullable-value-types)).
 
-### 11.2.5 Implicit nullable conversions
+### §implicit-interpolated-string-conversions Implicit interpolated string conversions
+
+An implicit interpolated string conversion permits an *interpolated_string_expression* (§interpolated-strings) to be converted to `System.IFormattable` or `System.FormattableString` (which implements `System.IFormattable`).
+When this conversion is applied, a string value is not composed from the interpolated string. Instead an instance of `System.FormattableString` is created, as further described in §interpolated-strings.
+
+###  11.2.5 Implicit nullable conversions
 
 The implicit nullable conversions are those nullable conversions ([§11.6.1](conversions.md#1161-nullable-conversions)) derived from implicit predefined conversions.
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1152,7 +1152,7 @@ A *primary_expression* that consists of a *literal* ([§7.4.5](lexical-structure
 
 ### §interpolated-strings Interpolated strings
 
-An *interpolated_string_expression* consists of a `$` character followed by a regular or verbatim string literal, wherein placeholders, delimited by `{` and `}` characters, enclose expressions and formatting specifications. An interpolated string expression is the result of an *interpolated_string_literal* that has been broken up into individual tokens, as described in §interpolated-string-literals.
+An *interpolated_string_expression* consists of a `$` character immediately followed by a regular or verbatim string literal, wherein placeholders, delimited by `{` and `}` characters, enclose expressions and formatting specifications. An interpolated string expression is the result of an *interpolated_string_literal* that has been broken up into individual tokens, as described in §interpolated-string-literals.
 
 ```antlr
 interpolated_string_expression
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1152,7 +1152,7 @@ A *primary_expression* that consists of a *literal* ([§7.4.5](lexical-structure
 
 ### §interpolated-strings Interpolated strings
 
-An *interpolated_string_expression* consists of a `$` sign followed by a regular or verbatim string literal, wherein holes, delimited by `{` and `}`, enclose expressions and formatting specifications. An interpolated string expression is the result of an *interpolated_string_literal* that has been broken up into individual tokens, as described in §§interpolated-string-literals.
+An *interpolated_string_expression* consists of a `$` character followed by a regular or verbatim string literal, wherein placeholders, delimited by `{` and `}` characters, enclose expressions and formatting specifications. An interpolated string expression is the result of an *interpolated_string_literal* that has been broken up into individual tokens, as described in §interpolated-string-literals.
 
 ```antlr
 interpolated_string_expression
@@ -1201,7 +1201,7 @@ A formatting specification involves the use of a ***format string literal***, wh
     
 The subsequent arguments are simply the *expressions* from the *interpolations* (if any), in order.
 
-When an *interpolated_string_expression* contains multiple holes, the expressions in those holes are evaluated in textual order, from the left-most hole to the right-most hole.
+When an *interpolated_string_expression* contains multiple placeholders, the expressions in those placeholders are evaluated in textual order, from the left-most hole to the right-most hole.
 
 > *Example*: 
 > ```csharp
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1105,6 +1105,7 @@ primary_expression
 
 primary_no_array_creation_expression
     : Literal
+    | interpolated_string_expression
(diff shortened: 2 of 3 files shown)


standard/expressions.md
```diff
@@ -1140,9 +1141,80 @@ object o = (new int[3])[1];
...
 ### 12.7.2 Literals
 
+#### 12.7.2.1 General
+
 A *primary_expression* that consists of a *literal* ([§7.4.5](lexical-structure.md#745-literals)) is classified as a value.
 
-### 12.7.3 Simple names
+### §interpolated-strings Interpolated strings
+
+An *interpolated_string_expression* consists of a `$` sign followed by a regular or verbatim string literal, wherein holes, delimited by `{` and `}`, enclose expressions and formatting specifications. An interpolated string expression is the result of an *interpolated_string_literal* that has been broken up into individual tokens, as described in §§interpolated-string-literals.
```
Nigel-Ecma: In Standard-ese the `$` is probably a *character* or maybe a *prefix* (`@` is described this way), but it is not a *sign*.

"holes"? Not sure its the right word to use; Wikipedia uses the term "placeholder" – that sounds better, Swift doesn't use any term *per se* but the grammar rule is _interpolated-text_. Regardless surely it needs to be a **_holes_** at least? 

"delimited by `{` and `}` *characters*"? (emphasis indicates insertion not styling!)

Nigel-Ecma: "followed by" – in [this MS](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated) page it says:

> To identify a string literal as an interpolated string, prepend it with the $ symbol. You cannot have any white space between the $ and the " that starts a string literal.

Should this prohibition of white space be included here?

(BTW: referring to the earlier comment, "sign" has become "symbol" here, but I still think it is "character" in Standard-ese)

Nigel-Ecma: > An interpolated string expression is the result of an *interpolated_string_literal*

There seems to be some confusion here between "interpolated string expression", "*interpolated_string_expression*" and "*interpolated_string_literal*".

The two grammar production rules are identical (the rule for *interpolated_string_literal* is given in the referenced §§interpolated-string-literals). What the text appears to be saying is that the production *interpolated_string_expression* represents the result (value?) of the production *interpolated_string_literal* – which seems a bit unusual, is there any other case of one production representing the result/value of another?

Maybe we should be talking about the value of an *interpolated_string_literal*/interpolated string literal, and *interpolated_string_expression*/interpolated string expression is redundant?

(Note: other productions are duplicated between here and §§interpolated-string-literals)

standard/conversions.md
```diff
@@ -88,7 +89,12 @@ There are no predefined implicit conversions to the `char` type, so values of th
 
 An implicit enumeration conversion permits a *constant-expression* ([§12.20](expressions.md#1220-constant-expressions)) with any integer type and the value zero to be converted to any *enum-type* and to any *nullable-value-type* whose underlying type is an *enum-type*. In the latter case the conversion is evaluated by converting to the underlying *enum-type* and wrapping the result ([§9.3.11](types.md#9311-nullable-value-types)).
 
-### 11.2.5 Implicit nullable conversions
+### §implicit-interpolated-string-conversions Implicit interpolated string conversions
+
+An implicit interpolated string conversion permits an *interpolated_string_expression* (§interpolated-strings) to be converted to `System.IFormattable` or `System.FormattableString` (which implements `System.IFormattable`).
+When this conversion is applied, a string value is not composed from the interpolated string. Instead an instance of `System.FormattableString` is created, as further described in §interpolated-strings.
```
jskeet, line 95: I'm generally slightly nervous about referring to "interpolated strings" - it sounds like they're peers of strings. How would folks feel about ensuring that the standard *only* refers to "interpolated string literals" instead of bare "interpolated strings"? It's possible that I'm being unreasonably picky again.

standard/expressions.md
jskeet: Placeholder certainly works for me (and yes to characters in both places).

jskeet (commented):
(Incomplete review.)

jskeet:
Note for meeting: I don't *think* there have been changes since the last meeting, so we may wish to skip this beyond checking that we have a plan for making the changes. (The last commit claims to be 21 days ago, but I have a comment on it from 28 days ago, so I suspect it's an oddity of the repo move and/or rebasing etc.)

jskeet:
Assigned back to @BillWagner to address or respond to comments.

standard/expressions.md
BillWagner: changed to "character" in both places. Placeholder instead of holes (3 locations).

BillWagner: I added the word "immediately". I checked the text for "verbatim string literals:" [§7.4.5.6](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/lexical-structure.md#7456-string-literals) and what we have is consistent.

BillWagner: This item probably needs more discussion. Unless @gafter or @MadsTorgersen have a suggestion.

I think the result of evaluating the placeholder expressions in the interpolated_verbatim_string should result in a `string`. But, that may run foul of the recursion Neal had mentioned.

Nigel-Ecma:
In April's meeting there was a comment made that interpolated strings had been approached "wrong" in some way. In an exchange afterwards on the list Neal clarified:

> The spec and implementation identify a substring of the input - the interpolation - and then use a nested recursive instance of the parser to lex and parse it.  I think it would have been better to use conventional techniques. Parsers are after all pushdown automata and can handle recursion just fine.  The result might be a little different as we don't currently allow newlines within an interpolation unless the string is a verbatim string.  That is harder to accomplish with more conventional techniques.  But we could have the parser just accept it (newlines and all) and just report a semantic error when the start and end aren't on the same line.

We should look into doing this "right" as part of reviewing this feature, we're not looking to change the semantics but the presentation – as comments above show there are some issues/confusion over the current presentation. As we are now using Antlr as the grammar specification Neal's last point is possibly easily addressed with it's predicates.

*Note: If the presentation in the Standard and the Roslyn compiler differ there is no issue provided the described overall syntax & semantics are the same – the Standard doesn't dictate the underlying parse used by a compiler *per se*, just the syntax & semantics.*

standard/conversions.md
gafter, line 95:
```suggestion
When this conversion is applied, a string value is not composed from the interpolated string expression. Instead an instance of `System.FormattableString` is created, as further described in §interpolated-strings.
```

standard/conversions.md
```diff
@@ -88,7 +89,12 @@ There are no predefined implicit conversions to the `char` type, so values of th
 
 An implicit enumeration conversion permits a *constant-expression* ([§12.20](expressions.md#1220-constant-expressions)) with any integer type and the value zero to be converted to any *enum-type* and to any *nullable-value-type* whose underlying type is an *enum-type*. In the latter case the conversion is evaluated by converting to the underlying *enum-type* and wrapping the result ([§9.3.11](types.md#9311-nullable-value-types)).
 
-### 11.2.5 Implicit nullable conversions
+### §implicit-interpolated-string-conversions Implicit interpolated string conversions
```
gafter, line 92: I don't remember how this will get a section number eventually. What is the process?

standard/lexical-structure.md
```diff
@@ -741,7 +744,192 @@ Each string literal does not necessarily result in a new string instance. When t
...
+
+Like string literals, interpolated string literals can be either regular or verbatim. Interpolated regular string literals are delimited by `$"` and `"`, and interpolated verbatim string literals are delimited by `$@"` and `"`.
+
+Like other literals, lexical analysis of an interpolated string literal initially results in a single token, as per the grammar below. However, before syntactic analysis, the single token of an interpolated string literal is broken into several tokens for the parts of the string enclosing the holes, and the input elements occurring in the holes are lexically analyzed again. This may in turn produce more interpolated string literals to be processed, but, if lexically correct, will eventually lead to a sequence of tokens for syntactic analysis to process.
+
+```antlr
+Interpolated_String_Literal
```
gafter, line 754: I don't know why the nonterminal names are capitalized here but lowercase elsewhere.  They need to be the same.

jskeet: I strongly suspect it's due to the age of the PR - our conventions have changed over time.

standard/conversions.md
jskeet: IIRC, we merge it as-is, and then Bill's tooling runs automatically and creates a new PR with the new section numbers in. @BillWagner is that right?

BillWagner: Yes, that's correct.

@gafter Notes on how the tool resolves new sections and links to new sections is [here](https://github.com/dotnet/csharpstandard/blob/draft-v6/Contribute.md#how-to-add-or-remove-clauses)

standard/lexical-structure.md
RexJaeschke: The ANTLR convention is to use a leading uppercase letter on lexer rule names.

BillWagner:
Resolved conflicts, and this is ready for review.

Nigel-Ecma:
I've moved this PR to "Draft" as there are a few things to address before we can consider moving forward with it, I'll post something out.

BillWagner:
closing in favor of #390

PR #43, opened 2020-11-24
Create admin file "guidelines-for-spec-writing-and-formatting.md"

RexJaeschke:
Bill, this was a document I created for you back in Jan/Feb. I just came across it and thought it would be useful to add to the repo, so others can see it.


admin/guidelines-for-spec-writing-and-formatting.md
```diff
@@ -0,0 +1,10 @@
+# Guidelines to Spec Writing and Formatting
+
+1.	Put definitions of major/often-used terms up front in the Terms-and-definitions clause. Lesser terms should be italicized on their first use only in the narrative.
+1.	Don’t say the same thing normatively in more than one place; say it once and point to that place, as needed.
+1.	For requirements, use *shall* and *shall not*. Inside informative (that is, non-normative) text, **never** use *shall*, use *must* instead.
+1.	Examples and Notes marked as such are informative, and requirements *must* be clear *without* them. Ask yourself, "Does the normative text standalone if all the examples and notes are removed?"
+1.	While examples are good, keep them small, simple, and narrowly focused without using unnecessary features. A spec is *not* a tutorial!
+1.	Avoid using quoted text, as in "natural "ype." If that is a term, except for its first use it should be in Roman without the quotes. In general, such quote marks dilute the contents, especially if their use suggests, “sort of like this thingie." Rather than use an approximate description in quotes, chose the right term or other English text.
```
BillWagner: typo:

```suggestion
1.	Avoid using quoted text, as in "natural type." If that is a term, except for its first use it should be in Roman without the quotes. In general, such quote marks dilute the contents, especially if their use suggests, “sort of like this thingie." Rather than use an approximate description in quotes, chose the right term or other English text.
```

BillWagner: We don't specify fonts in Markdown. Should this just say "normal text style" instead of "Roman font"?

admin/guidelines-for-spec-writing-and-formatting.md
```diff
@@ -0,0 +1,10 @@
...
+1.	Use singular instead of plural: For example, don’t say, "Tuple types are declared using the following syntax:"; instead, say, "A tuple type is declared using the following syntax:". For the most part, the plural form is quite clear, but there can be cases where it isn’t, so it’s best to use the singular form.
+1.	Consider the case in which we have a section with *no* subsections: ## Section S; some text. Later one or more subsections are added, resulting in ## Section S; some text; ### Subsection S.1; some text; ### Subsection S.2; some text. This is *bad*, as a subsequent reference to S is ambiguous; is it to that whole section or is it to the preamble text at the start of that section? As such, you need to put the preamble text in a new subsection (which we've called "General") at the start of that section. Specifically, there should *not* be any text between a section head and its first subsection head.
```
BillWagner: Consider this description instead (it was how you described it to me, and it worked well): 
```suggestion
1.	Consider the case in which we have a section with *no* subsections: ## Section S; some text. Later one or more subsections are added, resulting in ## Section S; some text; ### Subsection S.1; some text; ### Subsection S.2; some text. This is *bad*, as a subsequent reference to S is ambiguous; is it to that whole section or is it to the preamble text at the start of that section? As such, you need to put the preamble text in a new subsection (which we've called "General") at the start of that section. Specifically, a section should contain either subsections or text, but never both.
```

BillWagner (commented):
I saw one typo, and then I had a couple questions to consider. Then, this is ready to merge.

admin/guidelines-for-spec-writing-and-formatting.md
Nigel-Ecma: Checking a few of my fonts from various foundries "regular" might be correct, I think "normal" might be a Word-ism, you also see "plain" used, and "Roman" does refer to a design style... Take your pick from the first three.

jskeet: +0.5 for regular, +1 for plain +0 for normal, -1 for Roman

admin/guidelines-for-spec-writing-and-formatting.md
```diff
@@ -0,0 +1,10 @@
+# Guidelines to Spec Writing and Formatting
+
+1.	Put definitions of major/often-used terms up front in the Terms-and-definitions clause. Lesser terms should be italicized on their first use only in the narrative.
+1.	Don’t say the same thing normatively in more than one place; say it once and point to that place, as needed.
```
jskeet:
```suggestion
1.	Don't say the same thing normatively in more than one place; say it once and point to that place, as needed.
```

In general, let's make this file ASCII...

jskeet (commented):
Should this actually be part of "Contribute.md"? I tend to think that "admin" is likely to refer to stuff that is only relevant to TC49-TG2 members.

admin/guidelines-for-spec-writing-and-formatting.md
```diff
@@ -0,0 +1,10 @@
+# Guidelines to Spec Writing and Formatting
+
+1.	Put definitions of major/often-used terms up front in the Terms-and-definitions clause. Lesser terms should be italicized on their first use only in the narrative.
```
jskeet: I believe our discussion about terms-and-definitions ended up with a broad idea that that clause would be more about meta-information about the standard (e.g. what "implementation" means) rather than about specific C# ideas.

admin/guidelines-for-spec-writing-and-formatting.md
```diff
@@ -0,0 +1,10 @@
...
+1.	Don't say the same thing normatively in more than one place; say it once and point to that place, as needed.
+1.	For requirements, use *shall* and *shall not*. Inside informative (that is, non-normative) text, **never** use *shall*, use *must* instead.
+1.	Examples and Notes marked as such are informative, and requirements *must* be clear *without* them. Ask yourself, "Does the normative text standalone if all the examples and notes are removed?"
+1.	While examples are good, keep them small, simple, and narrowly focused without using unnecessary features. A spec is *not* a tutorial!
+1.	Avoid using quoted text, as in "natural type." If that is a term, except for its first use it should be in Roman without the quotes. In general, such quote marks dilute the contents, especially if their use suggests, “sort of like this thingie." Rather than use an approximate description in quotes, chose the right term or other English text.
```
jskeet: I'm not sure what Roman means here in terms of Markdown. If it's just plain text, that would be clearer.

Additional suggestions:

- Change the curly quote at the start of "soft of like this thingie" to a straight one
- chose => choose

admin/guidelines-for-spec-writing-and-formatting.md
```diff
@@ -0,0 +1,10 @@
...
+1.	Use singular instead of plural: For example, don’t say, "Tuple types are declared using the following syntax:"; instead, say, "A tuple type is declared using the following syntax:". For the most part, the plural form is quite clear, but there can be cases where it isn’t, so it’s best to use the singular form.
```
jskeet: Three curly apostrophes in this bullet.

admin/guidelines-for-spec-writing-and-formatting.md
```diff
@@ -0,0 +1,10 @@
...
+1.	Consider the case in which we have a section with *no* subsections: ## Section S; some text. Later one or more subsections are added, resulting in ## Section S; some text; ### Subsection S.1; some text; ### Subsection S.2; some text. This is *bad*, as a subsequent reference to S is ambiguous; is it to that whole section or is it to the preamble text at the start of that section? As such, you need to put the preamble text in a new subsection (which we've called "General") at the start of that section. Specifically, a section should contain either subsections or text, but never both.
```
jskeet: I'd start this bullet with the final sentence. It's hard to start considering a case when I don't know what the area of concern is.

admin/guidelines-for-spec-writing-and-formatting.md
jskeet: (I see we've talked about Roman before. This is just another instance.)

admin/guidelines-for-spec-writing-and-formatting.md
```diff
@@ -1,10 +1,10 @@
...
-1.	Consider the case in which we have a section with *no* subsections: ## Section S; some text. Later one or more subsections are added, resulting in ## Section S; some text; ### Subsection S.1; some text; ### Subsection S.2; some text. This is *bad*, as a subsequent reference to S is ambiguous; is it to that whole section or is it to the preamble text at the start of that section? As such, you need to put the preamble text in a new subsection (which we've called "General") at the start of that section. Specifically, a section should contain either subsections or text, but never both.
+1.	Put meta-information about the standard (e.g., what "implementation" means) in the Terms-and-definitions clause. C#-specific terms should be italicized on their first use only in the narrative.
+3.	Don't say the same thing normatively in more than one place; say it once and point to that place, as needed.
```
jskeet: I'm fine with either leaving everything as "1." in the Markdown or doing the full numbering, but I'd rather we didn't have numbers that look sequential but aren't quite :)

admin/guidelines-for-spec-writing-and-formatting.md
```diff
@@ -1,10 +1,10 @@
...
+4.	For requirements, use *shall* and *shall not*. Inside informative (that is, non-normative) text, **never** use *shall*, use *must* instead.
+5.	Examples and Notes marked as such are informative, and requirements *must* be clear *without* them. Ask yourself, "Does the normative text standalone if all the examples and notes are removed?"
+6.	While examples are good, keep them small, simple, and narrowly focused without using unnecessary features. A spec is *not* a tutorial!
+7.	Avoid using quoted text, as in "natural type." If that is a term, except for its first use it should be in regular without the quotes. In general, such quote marks dilute the contents, especially if their use suggests, "sort of like this thingie." Rather than use an approximate description in quotes, choose the right term or other English text.
```
jskeet: I think "regular" needs a noun to support it here - "regular type" for example.

jskeet (approved):
Renumbering required, and the addition of one word, but otherwise I'm happy.

jskeet:
Note for meeting: I don't *think* there have been changes since the last meeting, so we may wish to skip this.

jskeet:
Assigned to @RexJaeschke to address comments; I'm happy to then provide final review.

PR #44, opened 2020-11-24
C# 7.x: out Variables

RexJaeschke:
The handling of discards in tuple deconstruction and pattern matching is deferred to the PRs that add support for tuples and pattern matching, respectively. This PR only handles the out parameter and assignment scenarios.

A couple of `> TODO` placeholders will need attention once those other PRs are created.


standard/lexical-structure.md
```diff
@@ -64,18 +64,17 @@ The productions for *simple_name* ([§11.7.4](expressions.md#1174-simple-names))
 >
 > *end example*
 
-If a sequence of tokens can be parsed (in context) as a *simple_name* ([§11.7.4](expressions.md#1174-simple-names)), *member_access* ([§11.7.6](expressions.md#1176-member-access)), or *pointer_member_access* ([§22.6.3](unsafe-code.md#2263-pointer-member-access)) ending with a *type_argument_list* ([§8.4.2](types.md#842-type-arguments)), the token immediately following the closing `>` token is examined. If it is one of
+If a sequence of tokens can be parsed (in context) as a *simple_name* ([§11.7.4](expressions.md#1174-simple-names)), *member_access* ([§11.7.6](expressions.md#1176-member-access)), or *pointer_member_access* ([§22.6.3](unsafe-code.md#2263-pointer-member-access)) ending with a *type_argument_list* ([§8.4.2](types.md#842-type-arguments)), the token immediately following the closing `>` token is examined. If it is one of the following:
```
gafter: This disambiguation is not correct.  The correct disambiguation rule appears in #61.  This rule improperly changes the meaning of `M(a < b > c)` from its current meaning (a syntactically valid chain of comparison operators) into a syntax error.  See [HERE](https://sharplab.io/#v2:CYLg1APgAgTABAQQLACgDeq5blAzDgFjgFkBDASwDsAKASk2wxWxbgA05S4BeOSgUwDu7OgG4GrdnABGPTuOaSpAYzmkFSktS4AeGXAB8cZbQ3YAvhMJaOAD1pw0cSyhepY7VExZ4cARgA2KQB7AAd+ACdSABdgiLgdag4AG34AM2iAGikI8gBzAAtoh24jVIyzLF8oQJDwqNj4gyS4cqyc/KKSsvTohRcgA) for a test that demonstrates that this is not correct.

standard/expressions.md
```diff
@@ -541,7 +541,7 @@ argument_name
 argument_value
     : expression
     | 'ref' variable_reference
-    | 'out' variable_reference
+    | 'out' local_variable_type? identifier
```
jcouv: The speclets for out vars and deconstructions relied on `declaration_expression`.
But in this doc, I see `'out' local_variable_type? identifier` and in the deconstruction [draft](https://github.com/dotnet/csharpstandard/pull/63/files#diff-d9d678d0b42d3c668ebd64159dadfd6188361121feca8fb94fddf8816433cd14R6357) I see `tuple_deconstruction_expression`.
I'm not sure what is the benefit of this structure.

I think having a `declaration_expression` that appears in multiple contexts is a good factorization. Then we don't need new syntax for deconstructions (it's `declaration_expression` and `tuple_expression`).

Tagging @cston since I'll be away.

standard/expressions.md
```diff
@@ -553,7 +553,55 @@ The *argument_value* can take one of the following forms:
...
-- The keyword `out` followed by a *variable_reference* ([§9.5](variables.md#95-variable-references)), indicating that the argument is passed as an output parameter ([§14.6.2.4](classes.md#14624-output-parameters)). A variable is considered definitely assigned ([§9.4](variables.md#94-definite-assignment)) following a function member invocation in which the variable is passed as an output parameter.
+- The keyword `out`, optionally followed by *local_variable_type* ([§12.6.2](statements.md#1262-local-variable-declarations)), followed by *identifier*, indicating that the argument is passed as an output parameter ([§14.6.2.4](classes.md#14624-output-parameters)). If *identifier* is not a discard (§discards-new-clause), the variable it designates is considered definitely assigned ([§9.4](variables.md#94-definite-assignment)) following a function member invocation in which the variable is passed as an output parameter. In this context, the variable designated by *identifier* is known as an ***out variable***.
```
BillWagner: This is where the build warnings are coming from. The discards clause is not part of this PR.

standard/lexical-structure.md
BillWagner: See https://github.com/dotnet/csharpstandard/pull/61/files#r869861306

standard/expressions.md
jcouv: Some more context on syntax shape (was discussed with LDM): https://github.com/dotnet/csharpstandard/issues/285

BillWagner: Updated to use `declaration_expression`.

Tagging @Nigel-Ecma for grammar checks.

standard/lexical-structure.md
BillWagner: Brought in the changes from #61 (which I believe can be closed without merging).

BillWagner:
@jskeet This one is now ready for the upcoming meeting.

Latest edits:

- rebase
- update all links
- remove any discard related text (that has been done in a separate PR.)

standard/expressions.md
Nigel-Ecma: *declaration_expression* is different and cannot simply be substituted in this location. Here *local_variable_type* is optional, it is not in *declaration_expression*. I'll find the change and add a suggestion there…

standard/expressions.md
```diff
@@ -549,7 +549,7 @@ argument_name
 argument_value
     : expression
     | 'ref' variable_reference
-    | 'out' variable_reference
+    | 'out' declaration_expression? identifier
```
Nigel-Ecma: While the idea of factoring out to a common shared production is not wrong *per se* in this case doing so has changed the grammar ☹️. Revert to the intermediate version:

```suggestion
    | 'out' local_variable_type? identifier
```

Nigel-Ecma (changes requested):
I haven't done a full review just looked at the grammar change as requested – unfortunately the change was wrong so needs reverting, suggestion in comment.

standard/expressions.md
jskeet: Other than generally trusting Nigel on matters of grammar, I don't have an opinion on this one - but it needs to be resolved one way or another.

standard/expressions.md
```diff
@@ -559,7 +559,55 @@ The *argument_value* can take one of the following forms:
...
+- The keyword `out`, optionally followed by *local_variable_type* ([§13.6.2](statements.md#1362-local-variable-declarations)), followed by *identifier*, indicating that the argument is passed as an output parameter ([§15.6.2.4](classes.md#15624-output-parameters)). If *identifier* is not a discard, the variable it designates is considered definitely assigned ([§9.4](variables.md#94-definite-assignment)) following a function member invocation in which the variable is passed as an output parameter. In this context, the variable designated by *identifier* is known as an ***out variable***.
+
+*identifier* denotes a new variable, an existing variable, or a discard depending on the context. Specifically,
+
+- If *local_variable_type* is present
+  - If *identifier* is `_`, it is interpreted as a typed discard
+  - Otherwise, *identifier* denotes a new local variable. It is a compile-time error if a variable by that name is currently in scope or if there is not an implicit conversion between the corresponding parameter’s type and *local_variable_type*.
```
jskeet: Two problems with this:

- It doesn't specify the direction of the required conversion (from the parameter type to the local variable type, or vice versa?)
- I suspect it actually needs to be an identity conversion.

Example:

```csharp
class Program
{
    static void Main()
    {
        // Error: Argument 1: cannot convert from 'out object' to 'out string'
        M(out object x);
    }

    static void M(out string y) => y = "";
}
```

standard/expressions.md
```diff
@@ -559,7 +559,55 @@ The *argument_value* can take one of the following forms:
...
+
+> *Example*:
+>
+> ```csharp
+> static void M(out int x) { … }
+> static void CallM1()
```
jskeet, line 576: Add a blank line between the methods?

standard/expressions.md
```diff
@@ -559,7 +559,55 @@ The *argument_value* can take one of the following forms:
...
+> {
+>     int not_;
```
jskeet: `not_` feels like an odd variable name to use for this example. Can we just call it `v` or something similarly neutral?

standard/expressions.md
```diff
@@ -559,7 +559,55 @@ The *argument_value* can take one of the following forms:
...
+>     M(out not_);      // OK: untyped existing variable
+>     M(out _);         // OK: untyped discard
+>     M(out int not_);  // error: new variable but a variable by that name already exists
+>     M(out int not_2); // OK: new variable
+>     M(out uint pu);   // error: new variable but has incompatible type
+>     M(out int _);     // OK: typed discard
+>     M(out var not_3); // OK: typed new variable
+>     M(out var _);     // OK: typed discard
```
jskeet: "implicitly typed discard" to differentiate between this and line 582? Not sure - let's discuss.

standard/expressions.md
```diff
@@ -559,7 +559,55 @@ The *argument_value* can take one of the following forms:
...
+> }
+> ```
+>
+> *end example*
+
+- If *local_variable_type* is absent
+  - If *identifier* is `_`
```
jskeet, line 593: I think we need an "otherwise" for this "if".

standard/expressions.md
```diff
@@ -559,7 +559,55 @@ The *argument_value* can take one of the following forms:
...
+>     int not_;
+>     M(out _);     // error: ref to variable before it is declared
+>     int _ = -5;   // declaration of variable _
+>     M(out _);     // OK: existing variable
+> }
+> ```
+>
+> *end example*
+
+It is also an error to reference an implicitly-typed out variable in the same argument list that immediately contains its declaration. The type of an implicitly-typed out variable is the type of the corresponding parameter in the signature of the method selected by overload resolution.
```
jskeet: "It is also an error" seems a little odd after two examples. Just "It is an error"? I think an example of this would be useful, too.

Also the second sentence of this paragraph seems unrelated to the first. Separate paragraph?

standard/expressions.md
```diff
@@ -1033,6 +1081,8 @@ Given an implicit conversion `C₁` that converts from an expression `E` to a ty
 - `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§12.6.4.6](expressions.md#12646-better-conversion-target))
 - `E` is a method group ([§12.2](expressions.md#122-expression-classifications)), `T₁` is compatible ([§20.4](delegates.md#204-delegate-compatibility)) with the single best method from the method group for conversion `C₁`, and `T₂` is not compatible with the single best method from the method group for conversion `C₂`
 
+The  conversion from an implicitly-typed out variable declaration is not considered better than any other conversion from expression.
```
jskeet: While I know what this means, it doesn't feel terribly clear - maybe we just need a reference to the "conversion from expression" section?

standard/lexical-structure.md
```diff
@@ -65,7 +65,7 @@ The productions for *simple_name* ([§12.8.4](expressions.md#1284-simple-names))
 >
 > *end example*
 
-If a sequence of tokens can be parsed (in context) as a *simple_name* ([§12.8.4](expressions.md#1284-simple-names)), *member_access* ([§12.8.7](expressions.md#1287-member-access)), or *pointer_member_access* ([§23.6.3](unsafe-code.md#2363-pointer-member-access)) ending with a *type_argument_list* ([§8.4.2](types.md#842-type-arguments)), the token immediately following the closing `>` token is examined, to see if it is
+If a sequence of tokens can be parsed (in context) as a *simple_name* ([§12.8.4](expressions.md#1284-simple-names)), *member_access* ([§12.8.7](expressions.md#1287-member-access)), or *pointer_member_access* ([§23.6.3](unsafe-code.md#2363-pointer-member-access)) ending with a *type_argument_list* ([§8.4.2](types.md#842-type-arguments)), the token immediately following the closing `>` token is examined. If it is one of the following:
```
jskeet: I don't think this really works with how the continuation is described. I'd revert this, *or* change the text at line 75.

(It's not immediately clear why this is in an `out` variable PR anyway, mind you.)

standard/expressions.md
```diff
@@ -6051,7 +6101,7 @@ assignment_operator
...
 
-The `=` operator is called the ***simple assignment operator***. It assigns the value or values of the right operand to the variable, property, indexer element or tuple elements given by the left operand. The left operand of the simple assignment operator shall not be an event access (except as described in [§15.8.2](classes.md#1582-field-like-events)). The simple assignment operator is described in [§12.21.2](expressions.md#12212-simple-assignment).
+The `=` operator is called the ***simple assignment operator***. When the left operand is not a discard (§9.2.8.1), this operator assigns the value of the right operand to the variable, property, or indexer element given by the left operand. Otherwise, the right operand is evaluated with its value going unused. The left operand of the simple assignment operator shall not be an event access (except as described in [§15.8.2](classes.md#1582-field-like-events)). The simple assignment operator is described in [§12.21.2](expressions.md#12212-simple-assignment).
```
jskeet, line 6121: Is this change and the next one intended to be in this PR? It's not really about `out` parameters... was it just something else that should have been in a discards PR?

standard/lexical-structure.md
```diff
@@ -77,7 +77,8 @@ If the following token is among this list, or an identifier in such a context, t
 > *Note*: These rules are not applied when parsing a *type_argument_list* in a *namespace_or_type_name* ([§7.8](basic-concepts.md#78-namespace-and-type-names)). *end note*
 <!-- markdownlint-disable MD028 -->
 
-<!-- markdownlint-enable MD028 -->
+> *Note*: In an *argument_value* of the form `out local_variable_type? variable_reference`, *local_variable_type* may be a generic type; for example: `F(out A<B> name)`. As the language grammar for the argument uses *variable_reference* (which is really *expression*), this context is subject to the disambiguation rule. In this case the closing `>` is followed by an identifier. *end note*
```
jskeet: This is inaccurate regardless of whether we use Nigel's suggestion - the grammar *used* to use *variable_reference* but it doesn't any more.

standard/expressions.md
BillWagner: Correct. Verified that an identity conversion is required.

standard/expressions.md
BillWagner: updated

standard/expressions.md
BillWagner: Updated. Leaving unresolved, but I believe you are correct.

standard/expressions.md
BillWagner: That's correct. This PR in its original form included both discards and out parameters. I left this change in place because it wasn't in the updated discards and tuples PR.

Let's discuss at the meeting if it should be removed.

standard/lexical-structure.md
BillWagner: That was trying to fix a merge conflict. Reverted.

standard/lexical-structure.md
BillWagner: This was also added when I was resolving merge conflicts. It could be reverted back. Let's discuss at the meeting.

standard/expressions.md
BillWagner: I'm also not sure it adds anything. Should this paragraph be removed? For discussion.

standard/lexical-structure.md
Nigel-Ecma: I don't recall being responsible for the change from *variable_reference* to *identifier* but whoever it was I believe they are correct. The change does seem to make the note a bit pointless as it is explaining that the *variable_reference* can only be an *identifier* and so the disambiguation rules will apply and this will parse correctly…

However, a revised note might be:

```suggestion
> *Note*: In an *argument_value* of the form `out local_variable_type? identifier`, *local_variable_type* may be a generic type; for example: `F(out A<B> name)`. This example would be subject to the disambiguation rule ([§6.2.5](lexical-structure.md#625-grammar-ambiguities)) and is resolved correct;y as the closing `>` is followed by an identifier. *end note*
```

(Check my x-ref syntax before accepting!)

standard/expressions.md
jskeet: Action from 2023-04-19:

```suggestion
    | 'out' declaration_expression
    | 'out' variable_reference
```

Reasoning: declaration_expression always has a local variable type, whereas it's fine to use an argument_value of `out x` if `x` is already declared, for example.

standard/expressions.md
jskeet: Action from 2023-04-19: @BillWagner to remove from this PR, but add it to the issue that @MadsTorgersen has already created for polishing on discards.

standard/lexical-structure.md
BillWagner: This xref looks good.

jskeet: Action from 2023-04-19: @BillWagner  to remove the note.

jskeet (approved):
Some simple final changes, then let's merge.

MadsTorgersen:
To the best of my analysis, this PR is fully replaced by #664.

jskeet:
Closing as obsolete.

PR #63, opened 2020-12-10
C# 7.x: Tuples

RexJaeschke:
This (Draft) PR is not quite complete.

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -81,6 +81,7 @@ In most cases, an identity conversion has no effect at runtime. However, since f
 In tuple conversions, element names are immaterial. Tuples with the same arity are identity-convertible to each other or to and from corresponding underlying `ValueTuple` types, regardless of their element names.
 
 > *Example*:
+>
 > ```csharp
 > var t = (sum: 0, count: 1);
 > 
@@ -89,15 +90,18 @@ In tuple conversions, element names are immaterial. Tuples with the same arity a
 > 
 > t2.moo = 1;
 > ```
+>
 > *end example*
 
 In the case in which an element name at one position on one side of a conversion, and the same name at a different position on the other side, the compiler shall issue a warning.
 
 > *Example*:
+>
 > ```csharp
 > (string first, string last) GetNames() { ... }
 > (string last, string first) names = GetNames(); // Oops!
 > ```
+>
 > *end example*
 
 ### 10.2.3 Implicit numeric conversions
@@ -137,7 +141,7 @@ A tuple conversion can be classified as a valid instance conversion or an extens
 
 On top of the member-wise conversions implied by implicit typing, implicit conversions between tuple types themselves are allowed.
 
-### 10.2.5 Implicit nullable conversions
+### 10.2.5 Implicit interpolated string conversions
 
 An implicit interpolated string conversion permits an *interpolated_string_expression* ([§11.7.3](expressions.md#1173-interpolated-string-expressions)) to be converted to `System.IFormattable` or `System.FormattableString` (which implements `System.IFormattable`).
 When this conversion is applied, a string value is not composed from the interpolated string. Instead an instance of `System.FormattableString` is created, as further described in [§11.7.3](expressions.md#1173-interpolated-string-expressions).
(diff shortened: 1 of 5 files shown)


RexJaeschke:
This PR adds the new section, "§deconstruction-expressions-new-clause Deconstruction expressions", which defines the grammar rule *tuple_deconstruction_expression*. However, that rule is not currently reachable, so we need to figure out where to insert it in the expressions rule hierarchy.

Nigel-Ecma:
> This PR adds the new section, "§deconstruction-expressions-new-clause Deconstruction expressions", which defines the grammar rule _tuple_deconstruction_expression_. However, that rule is not currently reachable, so we need to figure out where to insert it in the expressions rule hierarchy.

Good point! There is grammar missing, and the grammar that does exist is left recursive…

First let’s complete the grammar, but not make it reachable. Currently from the text a *tuple_deconstruction_expression* can only appear as part of a tuple deconstruction statement, so the grammar would be something like:

```ANTLR
tuple_deconstruction_statement
    : tuple_deconstruction_expression '=' expression
    ;
    
tuple_deconstruction_expression
    : '(' destination (',' destination)* ')'
    ;
    
destination
    : type? identifier
    ;
```

Left recursion is gone, and *tuple_deconstruction_expression* is reachable from * tuple_deconstruction_statement*, but the latter is not reachable yet.

Now how to make it reachable, i.e. where does it fit into the grammar? The text suggests it is a *statement_expression* however all the subclauses of that are expressions which can occur elsewhere, a *tuple_deconstruction_statement* can only be at the statement level. (Note that while a * tuple_deconstruction_statement* may only assign to existing variables unlike *assignment* it cannot be a sub-expression of another expression.)

The other obvious choices are: a subclause of *declaration_statement* – as one or more variables *may* be declared; a subclause of *embedded_statement* – on par with other statement kinds and in the hierarchy which contains *assignment*; or a subclause of *statement* itself.

Subjectively I’d lean to one on the last three and away from *statement_expression*, YMMV and I haven't checked where Roslyn puts it as someone will know!

standard/expressions.md
```diff
@@ -5862,6 +5981,117 @@ If the left operand of `a += or -=` operator is classified as an event access,
...
+    : destination ',' destination
+    | destination_list ',' destination
+    ;
+    
+destination
+    : local_variable_type? identifier
+    ;
+```
+
+Element values are copied from the source tuple to the destination(s). Each element's position is inferred from the destination position within *destination_list*. If no variable called “_” is in scope, a *destination* with *identifier* `_` indicates a discard (§discards-new-clause), and the corresponding element is discarded rather than being copied. (Discards are discussed further below.)  The destination list shall account for every element in the tuple.
```
BillWagner, line 6003: The reference to `discards-new-clause` is causing the one build warning.

BillWagner:
See #64 for remaining questions

standard/conversions.md
```diff
@@ -101,6 +127,20 @@ There are no predefined implicit conversions to the `char` type, so values of th
 
 An implicit enumeration conversion permits a *constant_expression* ([§11.20](expressions.md#1120-constant-expressions)) with any integer type and the value zero to be converted to any *enum_type* and to any *nullable_value_type* whose underlying type is an *enum_type*. In the latter case the conversion is evaluated by converting to the underlying *enum_type* and wrapping the result ([§8.3.11](types.md#8311-nullable-value-types)).
 
+### §tuple-conversions-new-clause Tuple conversions
+
+Tuple types and expressions support a variety of conversions by "lifting" conversions of the elements into overall *tuple conversion*. For the classification purpose, all element conversions are considered recursively. For example, to have an implicit conversion, all element expressions/types shall have implicit conversions to the corresponding element types.
```
gafter, line 132: Define "corresponding".

standard/expressions.md
```diff
@@ -1402,6 +1407,58 @@ Then:
 
 *end example*
 
+### §tuple-literal-expressions-new-clause Tuple literal expressions
+
+A tuple literal consists of two or more tuple literal elements, each of which is optionally named.
```
gafter, line 1412: Is the intent to specify https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.1/infer-tuple-names.md separately?

standard/types.md
```diff
@@ -403,6 +403,187 @@ Implicit conversions are available from the `null` literal to `T?` ([§10.2.7](c
...
+> *Example*:
+>
+> ```csharp
+> partial class C : IEnumerable<(string name, int age)> { ... }
+> partial class C : IEnumerable<(string fullname, int)> { ... } // Error: names must be specified and the same
+> ```
+>
+> *end example*
+
+A tuple cannot be created with the `new` operator. However, the `new` operator can be used to create and initialize an array of tuple or a nullable tuple.
```
gafter, line 500: `new Tuple<String, String>("a", "b")` is legal.  The rule is that a tuple type syntax cannot be used with the new operator.

BillWagner:
Closed in favor of #664

PR #65, opened 2020-12-10
C# v7.x: Throw Expressions

RexJaeschke:
This requires a bit more work.


RexJaeschke:
This PR adds the new section, "§throw-expression-operator-new-clause The throw expression operator", which defines the grammar rule *throw_expression*. However, that rule is not currently reachable, so we need to figure out where to insert it in the expressions rule hierarchy.

standard/expressions.md
```diff
@@ -2912,6 +2912,34 @@ These are the same transformations applied in [§6.4.3](lexical-structure.md#643
...
+```
+
+A *throw_expression* throws the value produced by evaluating the *null_coalescing_expression*, which shall denote a value of the class type `System.Exception`, a value of a class type that derives from `System.Exception` or a value of a type parameter type that has `System.Exception` (or a subclass thereof) as its effective base class. If evaluation of the expression produces `null`, a `System.NullReferenceException` shall be thrown instead.
+
+The behavior at runtime of the evaluation of a *throw expression* is the same as specified for a *throw statement* ([§12.10.6](statements.md#12106-the-throw-statement)).
+
+The type rules are as follows:
+
+- A *throw_expression* has no type.
+- A *throw_expression* is convertible to every type by an implicit conversion.
```
Nigel-Ecma: Does this need to be merged into *Expression Classifications* (currently §11.2)?

standard/expressions.md
```diff
@@ -2912,6 +2912,34 @@ These are the same transformations applied in [§6.4.3](lexical-structure.md#643
...
+
+The flow-analysis rules are as follows:
+
+- For every variable *v*, *v* is definitely assigned before the *null_coalescing_expression* of a *throw_expression* iff it is definitely assigned before the *throw_expression*.
+- For every variable *v*, *v* is definitely assigned after *throw_expression*.
```
Nigel-Ecma: This needs to be added to *Definite Assignment* (currently §9.4)

standard/expressions.md
```diff
@@ -4594,6 +4622,7 @@ The `??` operator is called the null coalescing operator.
 null_coalescing_expression
     : conditional_or_expression
     | conditional_or_expression '??' null_coalescing_expression
+    | throw_expression
```
Nigel-Ecma, line 4222: I think this allows `throw throw throw throw...` Probably should prevent that!

standard/expressions.md
```diff
@@ -2912,6 +2912,34 @@ These are the same transformations applied in [§6.4.3](lexical-structure.md#643
...
 
+### §throw-expression-operator-new-clause The throw expression operator
+
+```ANTLR
+throw_expression
+    : 'throw' null_coalescing_expression
+    ;
+```
+
+A *throw_expression* throws the value produced by evaluating the *null_coalescing_expression*, which shall denote a value of the class type `System.Exception`, a value of a class type that derives from `System.Exception` or a value of a type parameter type that has `System.Exception` (or a subclass thereof) as its effective base class. If evaluation of the expression produces `null`, a `System.NullReferenceException` shall be thrown instead.
```
gafter: This text should revert to the specification text appearing in [`12.10.6 The throw statement`](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/statements.md#12106-the-throw-statement), which says that the expression must be *implicitly convertible* to the type `System.Exception`.  This here is the old rule before it was relaxed to permit `null`.  You probably want one to refer to the other so that you don't have a whole paragraph repeated.

standard/expressions.md
gafter: Yes.

standard/expressions.md
gafter: It should be moved there.

standard/expressions.md
```diff
@@ -2912,6 +2912,34 @@ These are the same transformations applied in [§6.4.3](lexical-structure.md#643
...
+- A *throw_expression* is convertible to every type by an implicit conversion.
+
+The flow-analysis rules are as follows:
+
+- For every variable *v*, *v* is definitely assigned before the *null_coalescing_expression* of a *throw_expression* iff it is definitely assigned before the *throw_expression*.
+- For every variable *v*, *v* is definitely assigned after *throw_expression*.
+
+A *throw expression* shall only occur in the following syntactic contexts:
+
+- As the second or third operand of a ternary conditional operator (`?:`), but not both.
```
gafter: delete ", but not both".  It is prevented in C# 7 by other rules, but those other rules are relaxed in later revisions of the specification so that it becomes permitted.

standard/expressions.md
gafter: It is prevented by the text that says "A *throw expression* shall only occur in the following syntactic contexts:"

gafter (commented):
Making progress!

standard/expressions.md
Nigel-Ecma: Surely though the grammar can prevent it directly without any semantic rules?

I think (i.e. I haven't tested it through ANTLR for any issues) the following does this:

```ANTLR
null_coalescing_expression
    : conditional_or_expression
    | conditional_or_expression '??' (null_coalescing_expression | throw_expression)
    ;
```

standard/expressions.md
Nigel-Ecma: @gafter – Yup, that's what I meant (move vs. add), we concur.

standard/expressions.md
gafter: You still need the semantic rule, otherwise you could write `F(throw null)`, which the prose is intended to forbid.

Given the prose, there is no need to make the grammar more complex.

Nigel-Ecma: I'm afraid I'm going to disagree here. If the semantic rule is required for other contexts that a throw expression can occur then of course it must stay. However that in no way argues against the grammar being made more precise for *null_coalescing_expression* – the grammar is not really "more complex", adding a couple of parentheses doesn’t do that, but it is clearer.

jskeet:
Opening up for broader review - it looks like this still has quite a bit that needs doing though.

BillWagner:
> Opening up for broader review - it looks like this still has quite a bit that needs doing though.

I gave it a thorough read. I don't have any additional comments beyond what Neal and Nigel have already noted.

standard/expressions.md
```diff
@@ -2912,6 +2912,34 @@ These are the same transformations applied in [§6.4.3](lexical-structure.md#643
 
 An *anonymous_method_expression* is one of two ways of defining an anonymous function. These are further described in [§11.16](expressions.md#1116-anonymous-function-expressions).
 
+### §throw-expression-operator-new-clause The throw expression operator
```
jskeet: If `throw` is now an operator, does it need to be listed in 11.4.2?

standard/expressions.md
```diff
@@ -2912,6 +2912,34 @@ These are the same transformations applied in [§6.4.3](lexical-structure.md#643
...
+The flow-analysis rules are as follows:
+
+- For every variable *v*, *v* is definitely assigned before the *null_coalescing_expression* of a *throw_expression* iff it is definitely assigned before the *throw_expression*.
+- For every variable *v*, *v* is definitely assigned after *throw_expression*.
+
+A *throw expression* shall only occur in the following syntactic contexts:
+
+- As the second or third operand of a ternary conditional operator (`?:`), but not both.
+- As the second operand of a null coalescing operator (`??`).
+- As the body of an expression-bodied lambda or method.
```
jskeet: What about expression-bodied properties, constructors etc? (Can we just say "expression-bodied lambda or member"?)

jskeet (commented):
If both Nigel and Neal are present in today's meeting, let's talk about the grammar aspect then.

standard/expressions.md
BillWagner: My first thought is that `throw` isn't an operator:

```suggestion
### §throw-expression-operator-new-clause The throw expression
```

standard/expressions.md
Nigel-Ecma: Given @jskeet’s recent comment here are two options to improve the grammar. First:

```ANTLR
null_coalescing_expression
    : conditional_or_expression
    | conditional_or_expression '??' (null_coalescing_expression | throw_expression)
    ;
```

This avoids allowing `throw throw throw...` in the *grammar* but does introduce an “RE” rule, if that is seen as too complex then:

```ANTLR
null_coalescing_expression
    : conditional_or_expression
    | conditional_or_expression '??' null_coalescing_expression
    | conditional_or_expression '??' throw_expression
    ;
```

can be used.

 I vote for the first myself.

standard/expressions.md
jskeet: Discussed in the meeting - it has precedence, so we should list it in 11.4.2 and keep "operator" here. Precedence is the same as a null-coalescing expression.

jskeet:
Action: @gafter to apply his own suggestions. We have persuaded @Nigel-Ecma that leaving the prohibition of `throw throw throw` can be semantic.

standard/expressions.md
```diff
@@ -2912,6 +2912,34 @@ These are the same transformations applied in [§6.4.3](lexical-structure.md#643
...
+    ;
+```
+
+A *throw_expression* throws the value produced by evaluating the *null_coalescing_expression*, which shall denote a value of the class type `System.Exception`, a value of a class type that derives from `System.Exception` or a value of a type parameter type that has `System.Exception` (or a subclass thereof) as its effective base class. If evaluation of the expression produces `null`, a `System.NullReferenceException` shall be thrown instead.
+
+The behavior at runtime of the evaluation of a *throw expression* is the same as specified for a *throw statement* ([§12.10.6](statements.md#12106-the-throw-statement)).
+
+The type rules are as follows:
+
+- A *throw_expression* has no type.
```
gafter: We should edit 11.2.1 to say that "throw expression" is one of the categories.

standard/expressions.md
gafter: Leaving it in, since this is the C# 7 specification.

standard/expressions.md
```diff
@@ -150,10 +150,10 @@ The precedence of an operator is established by the definition of its associated
 > |  [§11.11](expressions.md#1111-relational-and-type-testing-operators)             | Equality                         | `==` `!=`
 > |  [§11.12](expressions.md#1112-logical-operators)             | Logical AND                      | `&`
 > |  [§11.12](expressions.md#1112-logical-operators)             | Logical XOR                      | `^`
-> |  [§11.12](expressions.md#1112-logical-operators)             | Logical OR                       | `|`
+> |  [§11.12](expressions.md#1112-logical-operators)             | Logical OR                       | `\|`
 > |  [§11.13](expressions.md#1113-conditional-logical-operators)             | Conditional AND                  | `&&`
-> |  [§11.13](expressions.md#1113-conditional-logical-operators)             | Conditional OR                   | `||`
-> |  [§11.14](expressions.md#1114-the-null-coalescing-operator)             | Null coalescing                  | `??`
+> |  [§11.13](expressions.md#1113-conditional-logical-operators)             | Conditional OR                   | `\|\|`
```
gafter: These did not render properly.

standard/expressions.md
gafter: We would also have to modify the grammar for `?:` and every function member declaration.  I don't want to complexify the spec that way.

gafter:
I think I've made all the changed that we have discussed.  This is ready for further review.

gafter:
I don't know how to fix the section numbering issues.

standard/expressions.md
BillWagner: Fixed.

BillWagner:
@jskeet 

I took another look at this while I was rebasing the V7 PRs. I think this is very close. The remaining concern is related to modifying the grammar or allowing the syntactic ambiguity and relying on the prose to disallow `throw throw throw`.

The next meeting where both @Nigel-Ecma and @gafter are present, we can resolve this and accept it with necessary recomendations.

standard/expressions.md
Nigel-Ecma: The whole v7 grammar is going need to be checked, just as was done for v6, by "someone" ;-) to check all the additions do what was intended and don't alter existing parses unintentionally, that the semantic rules fit in, etc. Let's just leave the grammar here as is for now, I don't think anything that is wrong *per se* has been noted yet, its a debate over style. "Someone" will maybe get to the overall task before the EOY...

Nigel-Ecma (commented):
Oops, my "comment" became a "review", must have hit the wrong button, nothing to add here to the comment!

standard/expressions.md
```diff
@@ -139,23 +140,23 @@ The precedence of an operator is established by the definition of its associated
...
+> |  [§11.10](expressions.md#1110-shift-operators)             | Shift                            | `<<` `>>` |
+> |  [§11.11](expressions.md#1111-relational-and-type-testing-operators)             | Relational and type-testing      | `<` `>` `<=` `>=` `is` `as` |
+> |  [§11.11](expressions.md#1111-relational-and-type-testing-operators)             | Equality                         | `==` `!=` |
+> |  [§11.12](expressions.md#1112-logical-operators)             | Logical AND                      | `&`  |
+> |  [§11.12](expressions.md#1112-logical-operators)             | Logical XOR                      | `^`  |
+> |  [§11.12](expressions.md#1112-logical-operators)             | Logical OR                       | `\|`  |
+> |  [§11.13](expressions.md#1113-conditional-logical-operators)             | Conditional AND                  | `&&`  |
+> |  [§11.13](expressions.md#1113-conditional-logical-operators)             | Conditional OR                   | `\|\|`  |
+> |  [§11.14](expressions.md#1114-the-null-coalescing-operator)             | Null coalescing                  | `??`  `throw x`  |
```
Nigel-Ecma: The first column needs and "and §11.xy" for §throw-expression-operator-new-clause. The second column could be "Null coalescing and throw expression". See "Assignment and lambda expression" below for the template to follow.

Nigel-Ecma (approved):
Just one minor fix and this is approved. I've recommended leaving the grammar as is in the debate over style.

standard/expressions.md
jskeet: @jskeet to fix and review this PR.

jskeet: (And then merge if I don't find anything else.)

jskeet: And check that it closes #66.

standard/expressions.md
```diff
@@ -2919,6 +2920,24 @@ These are the same transformations applied in [§6.4.3](lexical-structure.md#643
 
 An *anonymous_method_expression* is one of two ways of defining an anonymous function. These are further described in [§11.16](expressions.md#1116-anonymous-function-expressions).
 
+### §throw-expression-operator-new-clause The throw expression operator
```
jskeet: (New problem.)

This is currently within 11.7, so it'll become 11.7.22. The whole of 11.7 is meant to be primary expressions, and this isn't a primary expression, is it? Should we actually list it in 11.14 (the null coalescing operator) as that's the only place it's used?

@BillWagner for initial thoughts; we may need to defer to others. (Sorry, had really hoped I'd be able to merge this.)

standard/expressions.md
jskeet: Added to the table. It looks like once we've got it in the right place, it will indeed close #66.

standard/expressions.md
BillWagner: It feels wrong to put it in 11.14. 

My first thought is to insert it as the new 11.15. I agree it's not a primary expression. It's precedence is the same as the null coelescing operator, to locating it near that clause feels right. It should stay in the same location in the table in 11.4.2.

Thoughts?

jskeet: That would work for me.

@MadsTorgersen @Nigel-Ecma @gafter @ericlippert - any thoughts?

Nigel-Ecma: @BillWagner, well spotted!

Yes it is in the wrong place, I guess just a "typo" really.

Wherever it goes it does need to stay in the table in 11.4 with the first column updated to include its clause number and the second one to maybe include it by name as well; as already agreed.

As to where it goes I see two options:

Option 1: As Bill suggested add it after 11.14 as a new 11.15

Option 2: make 11.14 into a new "Null Coalescing Expressions", followed by the a grammar showing both the grammar for `null_coalescing_expression` and `throw_expression`, then have 11.14.1 be "The Null Coalescing Operator" followed by the current text of 11.14 sans the grammar, and 11.14.2 be "The Throw Expression" with the new text for throw expressions also sans its grammar. (The grammar section might need to be a "General" sub-clause, so there would be three in total, , I'll defer to @RexJaeschke on that).

Don't pick option 2 if there are plans for throw expressions to be allowed elsewhere in the language in future, otherwise pick whichever seems to fit better. I can see logic in both choices so am ambivalent.

jskeet: Throw expressions will be valid for switch expressions, so I guess a new section is better.

BillWagner: > Don't pick option 2 if there are plans for throw expressions to be allowed elsewhere in the language in future, otherwise pick whichever seems to fit better. I can see logic in both choices so am ambivalent.

Coming soon to a standard near you:  `throw` expressions in switch expressions:

```csharp
var result = switch expr
{
  0    => "nil",
  1    => "small",
  <0 => throw new ArgumentExpression("not a good value");
  _   => "more",
}
```

jskeet: I've now moved it to be a new 11.15. @BillWagner PTAL, and we can merge if you're happy.

BillWagner: @jskeet Looks like a conflict, and you haven't pushed the latest yet.

jskeet: Hmm, weird. It says it's pushed locally, so I'll need to look. Will do that tomorrow though.

BillWagner: > Hmm, weird. It says it's pushed locally, so I'll need to look. Will do that tomorrow though.

I wonder if you pushed to your fork, and not the upstream remote.

jskeet: That was my thought too, but it looks like not. Meetings etc prevent me from sorting it today - will get to it tomorrow.

jskeet: Got it. It was on a different branch in this repo. Have pushed now. PTAL @BillWagner.

standard/expressions.md
```diff
@@ -139,23 +140,23 @@ The precedence of an operator is established by the definition of its associated
...
+> |  [§11.11](expressions.md#1111-relational-and-type-testing-operators)             | Relational and type-testing      | `<` `>` `<=` `>=` `is` `as` |
+> |  [§11.11](expressions.md#1111-relational-and-type-testing-operators)             | Equality                         | `==` `!=` |
+> |  [§11.12](expressions.md#1112-logical-operators)             | Logical AND                      | `&`  |
+> |  [§11.12](expressions.md#1112-logical-operators)             | Logical XOR                      | `^`  |
+> |  [§11.12](expressions.md#1112-logical-operators)             | Logical OR                       | `\|`  |
+> |  [§11.13](expressions.md#1113-conditional-logical-operators)             | Conditional AND                  | `&&`  |
+> |  [§11.13](expressions.md#1113-conditional-logical-operators)             | Conditional OR                   | `\|\|`  |
+> |  [§11.14](expressions.md#1114-the-null-coalescing-operator) and [§throw-expression-operator-new-clause](expressions.md#throw-expression-operator-new-clause)             | Null coalescing and throw expression                  | `??`  `throw x`  |
```
BillWagner: The tool will automatically fix this if you write the following:

```suggestion
> |  [§11.14](expressions.md#1114-the-null-coalescing-operator) and §throw-expression-operator-new-clause             | Null coalescing and throw expression                  | `??`  `throw x`  |
```

BillWagner (approved):
This LGTM @jskeet 

Let's merge #628, then manage the conflicts here, and merge this.

PR #70, opened 2020-12-14
C# v7.x: async Main

RexJaeschke:


BillWagner:
C# 7 feature: converting to a draft PR until we're ready to work on V7

standard/basic-concepts.md
```diff
@@ -10,33 +10,45 @@ A program compiled as an application shall contain at least one method qualifyin
 - It shall be `static`.
 - It shall not be generic.
 - It shall be declared in a non-generic type. If the type declaring the method is a nested type, none of its enclosing types may be generic.
-- It shall not have the `async` modifier.
+- It may have the `async` modifier provided the method's return type is `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`.
+- The return type shall be `void`, `int`, `System.Threading.Tasks.Task`, or `System.Threading.Tasks.Task<int>`.
 - The return type shall be `void` or `int`.
```
KalleOlaviNiemitalo: This line should be deleted as the previous one supersedes it.

standard/basic-concepts.md
```diff
@@ -10,33 +10,45 @@ A program compiled as an application shall contain at least one method qualifyin
...
 
-If the return type of the entry point method is `void`, reaching the right brace (`}`) that terminates that method, or executing a `return` statement that has no expression, results in a termination status code of `0`. If the entry point method terminates due to an exception ([§20.4](exceptions.md#204-how-exceptions-are-handled)), the exit code is implementation-specific. Additionally, the implementation may provide alternative APIs for specifying the exit code.
+If the return type of the entry point method is `void` or `System.Threading.Tasks.Task`, reaching the right brace (`}`) that terminates that method, or executing a `return` statement that has no expression, results in a termination status code of `0`. If the entry point method terminates due to an exception ([§20.4](exceptions.md#204-how-exceptions-are-handled)), the exit code is implementation-specific. Additionally, the implementation may provide alternative APIs for specifying the exit code.
```
KalleOlaviNiemitalo: OK, it does not matter what this paragraph says about a method whose return type is `System.Threading.Tasks.Task` but which does not have the `async` modifier. If the final right brace of such a method looks reachable to the compiler, or the method contains `return;`, then it cannot even be compiled, so the runtime behavior does not matter.

jskeet:
I'm going to try to review this before next week's meeting.

standard/basic-concepts.md
```diff
@@ -10,33 +10,45 @@ A program compiled as an application shall contain at least one method qualifyin
...
 
 Ordinarily, the declared accessibility ([§7.5.2](basic-concepts.md#752-declared-accessibility)) of a method is determined by the access modifiers ([§14.3.6](classes.md#1436-access-modifiers)) specified in its declaration, and similarly the declared accessibility of a type is determined by the access modifiers specified in its declaration. In order for a given method of a given type to be callable, both the type and the member shall be accessible. However, the application entry point is a special case. Specifically, the execution environment can access the application’s entry point regardless of its declared accessibility and regardless of the declared accessibility of its enclosing type declarations.
 
+When the entry point method has a return type of `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`, the compiler synthesizes an actual entry-point method that calls the corresponding `Main` method, as follows:
+
+- `static Task Main()` results in the compiler emitting the equivalent of  
+   `private static void $GeneratedMain() => Main().GetAwaiter().GetResult();`
```
KalleOlaviNiemitalo: `await` can use `GetAwaiter` and `GetResult` extension methods, or methods with parameters that have default values. Does the compiler have to do the same kind of lookups when it generates the entry point? It would not matter with the standard library, though.

BillWagner: If I correctly understand your question, no.  An `async` `Main` method must return either `Task` or `Task<int>`. These aren't considered valid entry points:

```csharp
public async void Main() {}
public async ValueTask<int> Main() {}
```

The compiler will consider those valid methods, but not valid entry points.

standard/basic-concepts.md
jskeet: Done.

standard/basic-concepts.md
jskeet: I'd argue that's covered elsewhere, and that this paragraph describes the behavior of code that *does* compile.

However, I do wonder whether we can just simplify it to say that if the return type is `void` and the method completes without an exception, or if the return type is `Task` and the returned task completes successfully (we can wordsmith that to be consistent with async elsewhere) then the termination status code is 0.

(It's actually not clear to me whether the termination status code is always 0 if the Main method completes but there are other non-daemon threads - one of those could call Environment.Exit with a non-zero status code for example.)

standard/basic-concepts.md
jskeet: I wonder whether we should call that out in a note.

jskeet (commented):
This looks mostly fine, but I'd like to discuss how we talk about the termination status code.

standard/basic-concepts.md
KalleOlaviNiemitalo: I meant a valid entry point like this:

```C#
public static async System.Threading.Tasks.Task Main() {}
```

if the System.Threading.Tasks.Task type does not have a GetAwaiter() member but there is instead an accessible extension method with that name, as permitted in [§C.3](https://github.com/dotnet/csharpstandard/blob/cabbf59f114db57172fd87c9429cdcc313aa716e/standard/standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271). Does the compiler have to find that extension method when it generates the real entry point? If so, I think it should use the `using` directives that are in effect for the async Main method.

KalleOlaviNiemitalo: "compiler emitting the equivalent" does not say _where_  the compiler emits this. If it emits near the definition of the async Main method, then the `using` directives there would naturally affect the `GetAwaiter` and `GetResult` lookups. But if the compiler emits in another part of a partial class, then the `using` directives might not take effect.

standard/basic-concepts.md
```diff
@@ -10,33 +10,44 @@ A program compiled as an application shall contain at least one method qualifyin
...
```
KalleOlaviNiemitalo: Roslyn apparently generates the actual entry-point method also if the source code has `static extern System.Threading.Tasks.Task Main();`. OK, the wording allows that behavior.

standard/basic-concepts.md
```diff
@@ -10,33 +10,44 @@ A program compiled as an application shall contain at least one method qualifyin
...
 
 Ordinarily, the declared accessibility ([§7.5.2](basic-concepts.md#752-declared-accessibility)) of a method is determined by the access modifiers ([§14.3.6](classes.md#1436-access-modifiers)) specified in its declaration, and similarly the declared accessibility of a type is determined by the access modifiers specified in its declaration. In order for a given method of a given type to be callable, both the type and the member shall be accessible. However, the application entry point is a special case. Specifically, the execution environment can access the application’s entry point regardless of its declared accessibility and regardless of the declared accessibility of its enclosing type declarations.
 
+When the entry point method has a return type of `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`, the compiler synthesizes an actual entry-point method that calls the corresponding `Main` method, as follows:
```
KalleOlaviNiemitalo: It's a bit confusing that "actual entry point" can return Task in "an external mechanism may be used to specify which method is deemed to be the actual entry point for the application", but then "actual entry-point method" cannot return Task in "the compiler synthesizes an actual entry-point method".

standard/basic-concepts.md
jskeet: Jon to work out some text for this. We can either clarify that code is generated *as if* with the using directives, but without that affecting any other program text - or just that calls to the extension methods are generated (e.g. explicitly).

standard/basic-concepts.md
jskeet: Agreed, we should be consistent about what "actual" means.

standard/basic-concepts.md
jskeet: Also it might be expression-bodied, therefore not have a `}`

standard/basic-concepts.md
KalleOlaviNiemitalo: If GetAwaiter or GetResult has an optional parameter with a CallerMemberNameAttribute, does that use "Main"?

standard/basic-concepts.md
KalleOlaviNiemitalo: "synchronous entry point" perhaps?

standard/basic-concepts.md
BillWagner: I think the signatures and implementation of the generated method should be an informative note. One normative point that should be covered is that the generated entrypoint method must not exit until the `Task` returned by the `async Main` method has completed or faulted.

If we want to require that `GetAwaiter().GetResult()` is used to avoid any deadlocks, that's fine with me as well.

standard/basic-concepts.md
```diff
@@ -10,33 +10,44 @@ A program compiled as an application shall contain at least one method qualifyin
...
 
-If the return type of the entry point method is `void`, reaching the right brace (`}`) that terminates that method, or executing a `return` statement that has no expression, results in a termination status code of `0`. If the entry point method terminates due to an exception ([§20.4](exceptions.md#204-how-exceptions-are-handled)), the exit code is implementation-specific. Additionally, the implementation may provide alternative APIs for specifying the exit code.
+If the return type of the entry point method is `void` or `System.Threading.Tasks.Task`, reaching the right brace (`}`) that terminates that method, or executing a `return` statement that has no expression, results in a termination status code of `0`. If the entry point method terminates due to an exception ([§20.4](exceptions.md#204-how-exceptions-are-handled)), the exit code is implementation-specific. Additionally, the implementation may provide alternative APIs for specifying the exit code.
```
BillWagner: Maybe consider this wording:

```suggestion
If the return type of the entry point method is `void` or `System.Threading.Tasks.Task`, reaching the endpoint of the method, or executing a `return` statement that has no expression, results in a termination status code of `0`. If the entry point method terminates due to an exception ([§20.4](exceptions.md#204-how-exceptions-are-handled)), the exit code is implementation-specific. Additionally, the implementation may provide alternative APIs for specifying the exit code.
```

BillWagner (commented):
I had one or two suggestions. Once we resolve the discussion on the compiler generated entry points, this is ready, IMO.

jskeet:
Agreed with Bill's proposal to make this less normative, which side-steps some issues. Jon to propose tweaked wording, Bill will review.

BillWagner:
@jskeet 

Looks like this one got dropped. It was last assigned to you for wordsmithing. I can pick up the baton for updating the text if you don't have time for it.

jskeet:
I've done a bit more work on this in a fifth commit, but I think we should discuss other areas - the idea that the entry point can return `Task<int>` is pervasive for the rest, and I think it would be useful if we could describe it just in terms of the what the synthetic entry point does instead.

standard/basic-concepts.md
```diff
@@ -10,33 +10,48 @@ A program compiled as an application shall contain at least one method qualifyin
...
 
 Ordinarily, the declared accessibility ([§7.5.2](basic-concepts.md#752-declared-accessibility)) of a method is determined by the access modifiers ([§14.3.6](classes.md#1436-access-modifiers)) specified in its declaration, and similarly the declared accessibility of a type is determined by the access modifiers specified in its declaration. In order for a given method of a given type to be callable, both the type and the member shall be accessible. However, the application entry point is a special case. Specifically, the execution environment can access the application’s entry point regardless of its declared accessibility and regardless of the declared accessibility of its enclosing type declarations.
 
+When the entry point method has a return type of `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`, the compiler synthesizes a synchronous entry-point method that calls the corresponding `Main` method. The synthetic method has parameters and return types based on the `Main` method:
```
BillWagner: synthetic -> synthesized (see other instances as well):

```suggestion
When the entry point method has a return type of `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`, the compiler synthesizes a synchronous entry-point method that calls the corresponding `Main` method. The synthesized method has parameters and return types based on the `Main` method:
```

standard/basic-concepts.md
```diff
@@ -10,33 +10,48 @@ A program compiled as an application shall contain at least one method qualifyin
...
+> *Note*: Methods with the `async` modifier must have exactly one of the two return types specified above in order to qualify as an entry point. An `async void` method, or an `async` method returning a different awaitable type such as `ValueTask` or `ValueTask<int>` does not qualify as an entry point. *end note*
+
+If more than one method qualifying as an entry point is declared within a program, an external mechanism may be used to specify which method is deemed to be the actual entry point for the application. By default, a qualifying method having a return type of `int` or `void` is selected over one having a return type of `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`. It is a compile-time error for a program to be compiled as an application without exactly one entry point. A program compiled as a class library may contain methods that would qualify as application entry points, but the resulting library has no entry point.
```
jskeet: Note: probably clarify this "preference" when distinguishing between kinds of entry point.

BillWagner:
closing and reopening to re-run all actions.

standard/basic-concepts.md
jskeet: Done.

standard/basic-concepts.md
jskeet: I've avoided these issues by a) introducing the concept of an effective entry point (which is always synchronous); b) talking about execution completing without an exception rather than talking about reaching a `}`.

standard/basic-concepts.md
jskeet: I've got very different wording now.

standard/basic-concepts.md
jskeet: I'm now not sure this is required. Let's discuss.

standard/basic-concepts.md
```diff
@@ -10,33 +10,50 @@ A program compiled as an application shall contain at least one method qualifyin
...
+
+Execution of the synthesized method proceeds as follows:
+
+- The synthesized method calls the `Main` method, passing its `string[]` parameter value as an argument if the `Main` method has such a parameter.
+- If the `Main` method throws an exception, the exception is propagated by the synthesized method.
+- Otherwise, the synthesized entry point waits for the returned task to complete, calling `GetAwaiter().GetResult()` on the task, using either the parameterless instance method or the extension method described by [§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271). If the task fails, `GetResult()` will throw an exception, and this exception is propagated by the synthesized method.
+- For a `Main` method with a return type of `System.Threading.Tasks.Task<int>`, if the task completes successfully, the `int` value returned by `GetResult()` is returned from the synthesized method.
+
+The ***effective entry point*** of an application is the entry point declared within the program, or the synthesized method if one is required as described above. The return type of the effective entry point is therefore always `void` or `int`.
```
jskeet, line 37: This effective entry point concept is new, and makes the rest of the text simpler. (Basically the effective entry point is always synchronous.)

I was considering introducing another term, *declared entry point*, as the `Main` method that actually has source code. Do folks think that would be useful?

(Of course, all of this will need to change for top-level statements...)

standard/basic-concepts.md
```diff
@@ -10,33 +10,50 @@ A program compiled as an application shall contain at least one method qualifyin
...
 
 ## 7.2 Application termination
 
 ***Application termination*** returns control to the execution environment.
 
-If the return type of the application’s entry point method is `int`, the value returned serves as the application’s ***termination status code***. The purpose of this code is to allow communication of success or failure to the execution environment.
+If the return type of the application’s effective entry point method is `int` and execution completes without resulting in an exception, the value of the `int` returned serves as the application’s ***termination status code***. The purpose of this code is to allow communication of success or failure to the execution environment. If the return type of the effective entry point method is `void` and execution completes without resulting in an exception, the termination status code is `0`.
```
jskeet, line 54: This is rewritten to move away from lexical aspects of method completion, to semantic aspects.

If there's better wording than "execution completes without resulting in an exception" I'd love to use it. (In Java there's the concept of "completes normally" - maybe we should introduce that?)

jskeet (commented):
This is now ready for another review - I believe it handles everything.

standard/basic-concepts.md
```diff
@@ -10,33 +10,50 @@ A program compiled as an application shall contain at least one method qualifyin
...
+> *Note*: Methods with the `async` modifier must have exactly one of the two return types specified above in order to qualify as an entry point. An `async void` method, or an `async` method returning a different awaitable type such as `ValueTask` or `ValueTask<int>` does not qualify as an entry point. *end note*
+
+If more than one method qualifying as an entry point is declared within a program, an external mechanism may be used to specify which method is deemed to be the actual entry point for the application. By default, a qualifying method having a return type of `int` or `void` is selected over one having a return type of `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`. It is a compile-time error for a program to be compiled as an application without exactly one entry point. A program compiled as a class library may contain methods that would qualify as application entry points, but the resulting library has no entry point.
```
BillWagner: Consider this sentence instead:

```suggestion
If more than one method qualifying as an entry point is declared within a program, an external mechanism may be used to specify which method is deemed to be the actual entry point for the application. If a qualifying method having a return type of `int` or `void` is found, any qualifying method having a return type of `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>` is not considered an entry point method. It is a compile-time error for a program to be compiled as an application without exactly one entry point. A program compiled as a class library may contain methods that would qualify as application entry points, but the resulting library has no entry point.
```

BillWagner (approved):
:shipit:

jskeet:
Agreed on the call.

PR #84, opened 2020-12-16
Initial README for the tools directory

jskeet:
This is extremely basic, but better than nothing as a starting point.


BillWagner (approved):
Thanks for the ping @jskeet 

This LGTM.

I'll :shipit: now

PR #104, opened 2020-12-27
C# v7.x: Local functions

RexJaeschke:


standard/statements.md
```diff
@@ -352,6 +355,63 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+A *local_function_declaration* may include one `async` ([§15.15](classes.md#1515-async-functions)) modifier and one `unsafe` ([§23.1](unsafe_code.md#231-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a task type  ([§15.15.1](classes.md#15151-general)).
+
+A local function is declared at block scope, and that function may capture variables from the enclosing scope. Each call of the function requires captured variables to be definitely assigned before the call. The compiler shall determine which variables are definitely assigned on return.
+
+A local function may be called from a lexical point prior to its definition. However, it is a compile-time error for the function to be declared lexically prior to the declaration of a variable it wishes to capture.
+
+It is a compile-time error for a local function to declare a parameter or local variable with the same name as one declared in the enclosing scope.
+
+Local function declaration statements are always reachable.
```
KalleOlaviNiemitalo:
```C#
class C {
    int F() {
        if (true) {
            void G() {}
            return 1;
        } else {
            void H() {}
            return 2;
        }
    }
}
```

Is there a rule that says `return 1;` is reachable and `return 2;` is not? According to [13.3.2 Statement lists](https://github.com/dotnet/csharpstandard/blob/c86c9b0f464572b060dfa6688968237a4d23ad9e/standard/statements.md#1332-statement-lists), it depends on whether "The end point of the preceding statement is reachable." But if both local function declaration statements are reachable, then I don't see why the end point of `void G() {}` would be reachable and that of `void H() {}` would not.

standard/basic-concepts.md
```diff
@@ -62,8 +62,8 @@ There are several different types of declaration spaces, as described in the fol
...
+-   Each method declaration, property declaration, property accessor declaration, indexer declaration, indexer accessor declaration, operator declaration, instance constructor declaration, anonymous function, and local function creates a new declaration space called a ***local variable declaration space***. Names are introduced into this declaration space through formal parameters (*fixed_parameter*s and *parameter_array*s) and *type_parameter*s. The set accessor for a property or an indexer introduces the valuename as a formal parameter. The body of the function member, anonymous function, or local function, if any, is considered to be nested within the local variable declaration space. It is an error for a local variable declaration space and a nested local variable declaration space to contain elements with the same name. Thus, within a nested declaration space it is not possible to declare a local variable or constant with the same name as a local variable or constant in an enclosing declaration space. It is possible for two declaration spaces to contain elements with the same name as long as neither declaration space contains the other.
```
KalleOlaviNiemitalo: Perhaps this should say ``the `value` name`` rather than `the valuename`. That was not introduced in this PR though.

standard/statements.md
```diff
@@ -414,6 +417,65 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+local_function_body
+    : block
+    | '=>' expression ';'
+    ;
+```
+
+Unless specified otherwise below, the semantics of all grammar elements is the same as for *method_declaration* ([§14.6.1](classes.md#1461-general), read in the context of a local function instead of a method.
+
+A *local_function_declaration* may include one `async` ([§14.15](classes.md#1415-async-functions)) modifier and one `unsafe` ([§22.1](unsafe-code.md#221-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a task type  ([§14.15.1](classes.md#14151-general)).
```
BillWagner: This will likely need to be revisited in light of generalized async return types. (`ValueTask<T>` works, among others).

standard/statements.md
```diff
@@ -414,6 +417,65 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+A local function is declared at block scope, and that function may capture variables from the enclosing scope. Each call of the function requires captured variables to be definitely assigned before the call. The compiler shall determine which variables are definitely assigned on return.
```
BillWagner: Adding @RikkiGibson

I found some interesting bits here, some of which don't apply until C# 8 (the reason I tagged Rikki).  Consider this example:

```csharp
public class C
{
    private string field;
    public C()
    {
        string local;
        F();
        String s = field;
        field = local;        
        void F()
        {
            this.field = "assigned";
            local = "assigned";
        }
    }
}
```

The compiler determines that the local function `F` does definitely assign `local` (`field` is definitely assigned to `null`). However, the nullable analysis doesn't set the null state of `field` or `local` to *not null* after `F()` is called. That makes sense the nullable reference spec doesn't specify that the compiler determines the null state of variables when a local function is called, but it seems odd to me. (That question doesn't need to be answered until we get to C# 8, so not blocking this PR).

BillWagner (commented):
This looks great @RexJaeschke  Should we mark this ready for review?

I've got one note for how local functions interacts with generalized async return types. 

I've also got one second question that won't need to be addressed until C# 8.

standard/statements.md
RikkiGibson: It is definitely a strange corner of our flow analysis. We made a deliberate decision to *not* "flow out" nullable state changes from local functions. They only "flow in"--i.e. when you have multiple call sites, the compiler determines a "merged worst state" to use as the initial state to analyze the local function. This was due to the performance cost and complexity of definite assignment analysis versus nullable analysis.

standard/statements.md
BillWagner: There isn't such a rule right now. The compiler determines that H() is reachable.

BillWagner: Discussed in the committee meeting:

```suggestion
Local function bodies are always reachable. The endpoint of a local function is reachable if the beginning point of the local function is reachable.
```

standard/statements.md
BillWagner: The statement "Each call of the function requires captured variables to be definitely assigned before the call." is incorrect. Captured variables must only be definitely assigned before it is *read* by the body of the local function.

BillWagner: Add informative examples here as well.

BillWagner (commented):
Add notes from committee meeting.

BillWagner:
This is now ready for final review.

@jskeet Should I tag others for review?

gafter:
A lot of the work on this feature was done by @agocke.  It would be good to get his feedback.

agocke:
I'll try to take a look. It looks like the definite assignment section needs work. Unlike lambdas, local functions use a more precise definite assignment when not converted to delegates. I'll try to write up some appropriate spec language for this. Overview of the differences is at https://commentout.com/localfunc.html

RexJaeschke:
@BillWagner I think the header and modifiers rules needs tweaking:

We have

```ANTLR
local_function_header
    : local_function_modifiers? return_type identifier type_parameter_list?
        ( formal_parameter_list? ) type_parameter_constraints_clause*
    ;
local_function_modifiers
    : 'async'
    | 'unsafe'
    ;
```

That allows a local function to have no modifiers, `async`, or `unsafe`, but it does *not* allow `async unsafe`, which I think is a valid combination. [In V8, we'll add `static`, and in V9, `extern`.]

Following the corresponding pattern for modifiers on methods ([14.16.1](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1461-general)), the rule should be renamed `local_function_modifier` (singular), with 0 or more occurrences allowed, with both rules changed, as follows:

```ANTLR
local_function_header
    : local_function_modifier* return_type identifier type_parameter_list?
        ( formal_parameter_list? ) type_parameter_constraints_clause*
    ;
local_function_modifier
    : 'async'
    | 'unsafe'
    ;
```

RexJaeschke:
Another grammar problem!

Not surprisingly, the header rule uses the same rules `type_parameter_list` and `formal_parameter_list` as `method_header` in [14.16.1](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#1461-general), as follows:. 

```ANTLR
local_function_header
    : local_function_modifiers? return_type identifier type_parameter_list?
        ( formal_parameter_list? ) type_parameter_constraints_clause*
    ;
```

However, these two rules **allow** attributes, yet it isn't until V9 that support for attributes is added to type parameters and formal parameters in local functions (as well as to local functions themselves)). So, for V7 and V8, we need to come up with a stripped-down set of these two rules that *disallow* attributes.

Here's my proposal for that (with admittedly long, but clear rule names):

```ANTLR
local_function_header
    : local_function_modifier* return_type identifier type_parameter_list_with_no_attributes?
        ( formal_parameter_list_with_no_attributes? ) type_parameter_constraints_clause*
    ;

local_function_modifier
    : 'async'
    | 'unsafe'
    ;

type_parameter_list_with_no_attributes
    : '<' type_parameters_with_no_attributes '>'
    ;

type_parameters_with_no_attributes
    : type_parameter
    | type_parameters_with_no_attributes ',' type_parameter
    ;

formal_parameter_list_with_no_attributes
    : fixed_parameters_with_no_attributes
    | fixed_parameters_with_no_attributes ',' parameter_array_with_no_attributes
    | parameter_array_with_no_attributes
    ;

fixed_parameters_with_no_attributes
    : fixed_parameter_with_no_attributes (',' fixed_parameter_with_no_attributes)*
    ;

fixed_parameter_with_no_attributes
    : parameter_modifier? type identifier default_argument?
    ;

parameter_array_with_no_attributes
    : 'params' array_type identifier
    ;
```

**This grammar includes the suggestion from my [earlier comment](https://github.com/dotnet/csharpstandard/pull/104#issuecomment-1126378608).**

gafter:
Rather than modifying the syntax, I suggest adding a semantic rule that attributes are not permitted on local functions.  Then, later, we can just remove that one sentence.

RexJaeschke:
Yes, @gafter, you are quite right; adding a constraint is a far less intrusive way of dealing with this.

I suggest that after the following text

> A *local_function_declaration* may include one `async` ([§14.15](classes.md#1415-async-functions)) modifier and one `unsafe` ([§22.1](unsafe-code.md#221-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a task type  ([§14.15.1](classes.md#14151-general)).

we add

>  It is a compile-time error for *type_parameter_list* or *formal_parameter_list* to contain *attributes*.

However, we'll still need the grammar change w.r.t the modifiers, as I mentioned earlier.

RexJaeschke:
Another grammar issue. We have the following

```ANTLR
local_function_body
    : block
    | '=>' expression ';'
    ;
```

It is likely that this was originally cloned from *method_body* ([14.6.1](standard/classes.md#1461-general), with the ` | ';'` rule omitted, as external local-function support wasn't added until V9). However, in V6, we added another rule to *method_body*, as follows:

```ANTLR
    | '=>' null_conditional_invocation_expression ';'
```

Should that rule also be added to *local_function_body*? If so, then Nigel's corresponding grammar note from the discussion of *method_body* should be adapted for here, as follows:

> Grammar note: When recognising a *local_function_body* if both the *null_conditional_invocation_expression* and *expression* alternatives are applicable then the former shall be chosen.

standard/statements.md
```diff
@@ -414,6 +417,100 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+A *local_function_declaration* may include one `async` ([§14.15](classes.md#1415-async-functions)) modifier and one `unsafe` ([§22.1](unsafe-code.md#221-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a task type  ([§14.15.1](classes.md#14151-general)).
+
+A local function is declared at block scope, and that function may capture variables from the enclosing scope. Captured variables must be definitely assigned before read by the body of the local function in each call to the function. The compiler shall determine which variables are definitely assigned on return.
+
+> *Example*: The following example demonstrates definite assignment for captured variables in local functions. If a local function reads a captured variable before writing it, the captured variable must be definitely assigned before calling the local function. The local function `F1` reads `s` without assigning it. It is an error if `F1` is called before `s` is definitely assigned. `F2` assigns `i` before reading it. It may be called before `i` is definitely assigned. Furthermore, `F1` may be called after `F2` because `i` is definitely assigned in `F2`.
```
jskeet: This part:

> Furthermore, `F1` may be called after `F2` because `i` is definitely assigned in `F2`.

... doesn't make much sense as `F1` doesn't read `i`. (I suspect at some point both functions used the same variable.)

standard/statements.md
```diff
@@ -414,6 +417,100 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+>         Console.WriteLine(i);
+>     }
+> }
+> ```
+>
+> *end example*
+
+A local function may be called from a lexical point prior to its definition. However, it is a compile-time error for the function to be declared lexically prior to the declaration of a variable it wishes to capture.
+
+It is a compile-time error for a local function to declare a parameter or local variable with the same name as one declared in the enclosing scope.
```
jskeet, line 488: I don't think this restriction is actually correct... or at least it isn't now. The code below is valid:

```csharp
using System;

class Program
{
    static void Main()
    {
        int x = 10;
        Local1("hello");
        Local2();
        
        void Local1(string x)
        {
            Console.WriteLine(x);
        }
        
        void Local2()
        {
            string x = "hello";
            Console.WriteLine(x);
        }
    }
}
```

Is this something that changed after C# 7?

standard/statements.md
```diff
@@ -414,6 +417,100 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+Local function bodies are always reachable. The endpoint of a local function is reachable if the beginning point of the local function is reachable.
+
+> *Example*: There are two common use cases for local functions: public iterator methods and public async methods. In iterator methods, any exceptions are observed only when calling code that enumerates the returned sequence. In async methods, any exceptions are only observed when the returned Task is awaited. The following example demonstrates separating parameter validation from the iterator implementation using a local function:
+>
+> ```csharp
+> public static IEnumerable<char> AlphabetSubset(char start, char end)
+> {
+>     if (start < 'a' || start > 'z')
```
jskeet, line 452: Let's add braces to this example, and a black line after the return statement.

standard/statements.md
```diff
@@ -414,6 +417,100 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+>
+> *end example*
+
+A local function may be called from a lexical point prior to its definition. However, it is a compile-time error for the function to be declared lexically prior to the declaration of a variable it wishes to capture.
+
+It is a compile-time error for a local function to declare a parameter or local variable with the same name as one declared in the enclosing scope.
+
+Local function bodies are always reachable. The endpoint of a local function is reachable if the beginning point of the local function is reachable.
+
+> *Example*: There are two common use cases for local functions: public iterator methods and public async methods. In iterator methods, any exceptions are observed only when calling code that enumerates the returned sequence. In async methods, any exceptions are only observed when the returned Task is awaited. The following example demonstrates separating parameter validation from the iterator implementation using a local function:
```
jskeet: Normally "example" refers to the preceding text - this is more of a "general use case" - in particular it's not about reachability. Maybe the example should be right at the start of the section, under the first sentence of description?

standard/statements.md
BillWagner: updated.

standard/statements.md
BillWagner: Yes. I just verified this. Your example doesn't compiler in C# 7.3, but does in 8.0. I wrote #590 to capture this.

standard/statements.md
BillWagner: Added brackets. I think I moved it to the right location (after the grammar).

BillWagner:
> Another grammar issue. We have the following
> 
> ```antlr
> local_function_body
>     : block
>     | '=>' expression ';'
>     ;
> ```
> 
> It is likely that this was originally cloned from _method_body_ ([14.6.1](standard/classes.md#1461-general), with the ` | ';'` rule omitted, as external local-function support wasn't added until V9). However, in V6, we added another rule to _method_body_, as follows:
> 
> ```antlr
>     | '=>' null_conditional_invocation_expression ';'
> ```
> 
> Should that rule also be added to _local_function_body_? If so, then Nigel's corresponding grammar note from the discussion of _method_body_ should be adapted for here, as follows:
> 
> > Grammar note: When recognising a _local_function_body_ if both the _null_conditional_invocation_expression_ and _expression_ alternatives are applicable then the former shall be chosen.

I verified that yes, the `null_conditional_invocation_expression` is a valid expression for a local function. I made these edits.

BillWagner:
I've updated this based on all the feedback *except* the definite assignment question.

Should we make a separate PR that updates [9.4](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/variables.md#94-definite-assignment) by adding a clause for assignment in local functions? Or should we put those updates in the statements section?

/cc @agocke

agocke:
Let me try to write down some simplified wording and see if you can massage it into spec language.

Local functions are analyzed in the context of their parent method. There are two new control flow paths that matter for local functions: calls and delegate conversions. Definite assignment for the body of each local function is defined separately for each call site. At each invocation, variables captured by the local function are considered definitely assigned if they were definitely assigned at the point of call. A control flow path also exists to the local function body at this point and is considered reachable. After a call to the local function, captured variables that were definitely assigned at every control point leaving the function (`return` statements, `yield` statements, `await` statements) are considered definitely assigned after the call location.

Delegate conversions have a control flow path to the local function body and captured variables are also considered assigned for the body if they are assigned before the conversion, but variables assigned by the local function are not considered assigned after the conversion.

Note: the above implies that bodies are re-analyzed for definite assignment at every local function invocation or delegate invocation. That's one way to achieve the required result, but for performance the compiler doesn't actually do so. The implementation should be equivalent, however.

BillWagner:
Meeting discussion item:

I can take Andy's text above and apply it to the standard. I'll propose that the definite assignment rules should be a new sub-clause of [9.4.4](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/variables.md#944-precise-rules-for-determining-definite-assignment), "Precise rules for definite assignment". 

@jskeet Can we discuss at today's meeting?

BillWagner:
Discussed in meeting:

- Add one or two sub-clauses under 9.4.4. to describe the definite assignment rules for local functions. Request another review after that's been added.

agocke:
Some other things about local functions that may be relevant:

- Unlike methods, they cannot be overloaded
- All calls to local functions must be static resolved. That means that the following is illegal:
```C#
using System;
public class C {
    public void M() {
        dynamic d = null;
        Local(d);
        void Local(params int[] xs) { }
    }
}
```
- `unsafe` uses the parent lexical scope, so `unsafe` on a parent method propagates to the nested local function
- `async` does *not* use the parent lexical scope, so the following is illegal
```C#
    public async Task M()
    {
        await Local();
        Task Local()
        {
            await Task.Delay(0);
        }
    }
```
- Local functions can be used to observe that the _scope_ and _lifetime_ of variables sometimes differ in C#. For example,
   ```C#
    public void M(Object o) {
        switch (o)
        {
            case 1:
                Local();
                break;
            case int x:
                void Local() { x = 1; }
                break;
        }
    }
   ```
   In this example the scope of the pattern variable `x` is the case clause it's in, while the lifetime is the whole switch statement

BillWagner:
Thanks @agocke 

Comments related to the edits I'm making now:

> Some other things about local functions that may be relevant:
> 
> * Unlike methods, they cannot be overloaded

Added to the new clause in statements.

> * All calls to local functions must be static resolved. That means that the following is illegal:
> 
> ```cs
> using System;
> public class C {
>     public void M() {
>         dynamic d = null;
>         Local(d);
>         void Local(params int[] xs) { }
>     }
> }
> ```

Added to clause in statements.

> * `unsafe` uses the parent lexical scope, so `unsafe` on a parent method propagates to the nested local function
> * `async` does _not_ use the parent lexical scope, so the following is illegal
> 
> ```cs
>     public async Task M()
>     {
>         await Local();
>         Task Local()
>         {
>             await Task.Delay(0);
>         }
>     }
> ```

Also added.

> * Local functions can be used to observe that the _scope_ and _lifetime_ of variables sometimes differ in C#. For example,
>   ```cs
>    public void M(Object o) {
>        switch (o)
>        {
>            case 1:
>                Local();
>                break;
>            case int x:
>                void Local() { x = 1; }
>                break;
>        }
>    }
>   ```
>       
>   In this example the scope of the pattern variable `x` is the case clause it's in, while the lifetime is the whole switch statement

I don't think this is needed in the standard. (It's interesting though.)

BillWagner:
@jskeet I've incorporated Andy's comments and description for definite assignment in local functions. I don't think there's enough time for everyone to review before our next meeting, but I'd like to put this one up for the following meeting.

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+>     {
+>         for (var c = start; c < end; c++)
+>             yield return c;
+>     }
+> }
+> ```
+>
+> *end example*
+
+Unless specified otherwise below, the semantics of all grammar elements is the same as for *method_declaration* ([§14.6.1](classes.md#1461-general), read in the context of a local function instead of a method.
```
Nigel-Ecma: Missing paren

```suggestion
Unless specified otherwise below, the semantics of all grammar elements is the same as for *method_declaration* ([§14.6.1](classes.md#1461-general)), read in the context of a local function instead of a method.
```

standard/variables.md
```diff
@@ -802,6 +802,63 @@ For a *lambda_expression* or *anonymous_method_expression* *expr* with a body (e
...
+
+Delegate conversions have a control flow path to the local function body. Captured variables are definitely assigned for the body if they are definitely assigned before the conversion. Variables assigned by the local function are not considered assigned after the conversion.
+
+> *note:* the above implies that bodies are re-analyzed for definite assignment at every local function invocation or delegate conversion. Compilers are not required to re-analyze the body of a local function at each invocation or delegate conversion. The implementation must produce results equivalent to that description.
```
Nigel-Ecma: Typo:

```suggestion
> *Note:* the above implies that bodies are re-analyzed for definite assignment at every local function invocation or delegate conversion. Compilers are not required to re-analyze the body of a local function at each invocation or delegate conversion. The implementation must produce results equivalent to that description.
```

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+local_function_body
+    : block
+    | '=>' expression ';'
+    | '=>' null_conditional_invocation_expression ';'
+    ;
+```
+
+Grammar note: When recognising a *local_function_body* if both the *null_conditional_invocation_expression* and *expression* alternatives are applicable then the former shall be chosen.
+
+> *Example*: There are two common use cases for local functions: public iterator methods and public async methods. In iterator methods, any exceptions are observed only when calling code that enumerates the returned sequence. In async methods, any exceptions are only observed when the returned Task is awaited. The following example demonstrates separating parameter validation from the iterator implementation using a local function:
```
Nigel-Ecma: Why *public* iterator/async, wouldn't the same benefits apply to those with other accessibilities, and I missing something?

Nigel-Ecma (commented):
A couple of typos and a question...

standard/statements.md
BillWagner: Good point. Let's remove "public":

```suggestion
> *Example*: There are two common use cases for local functions: iterator methods and async methods. In iterator methods, any exceptions are observed only when calling code that enumerates the returned sequence. In async methods, any exceptions are only observed when the returned Task is awaited. The following example demonstrates separating parameter validation from the iterator implementation using a local function:
```

Nigel-Ecma (approved):

Nigel-Ecma (approved):

standard/basic-concepts.md
```diff
@@ -64,8 +64,9 @@ There are several different types of declaration spaces, as described in the fol
...
+- Each method declaration, property declaration, property accessor declaration, indexer declaration, indexer accessor declaration, operator declaration, instance constructor declaration, anonymous function, and local function creates a new declaration space called a ***local variable declaration space***. Names are introduced into this declaration space through formal parameters (*fixed_parameter*s and *parameter_array*s) and *type_parameter*s. The set accessor for a property or an indexer introduces the name `value` as a formal parameter. The body of the function member, anonymous function, or local function, if any, is considered to be nested within the local variable declaration space. It is an error for a local variable declaration space and a nested local variable declaration space to contain elements with the same name. Thus, within a nested declaration space it is not possible to declare a local variable or constant with the same name as a local variable or constant in an enclosing declaration space. It is possible for two declaration spaces to contain elements with the same name as long as neither declaration space contains the other.
```
jskeet, line 67: This code appears to break this restriction, for both a parameter and a local variable declared in a local function. (Ignore values etc - I adapted a piece of code I happened to be looking at.)

```csharp
using System;

class Test
{
    static void Main()
    {
        var f = float.MaxValue;
        var d = 10;

        Method("hello");
        Console.WriteLine(d+f);

        string Method(string f)
        {
            int d = 20;
            return d.ToString() + f;
        }

    }
}
```

standard/statements.md
```diff
@@ -270,15 +270,18 @@ In addition to the reachability provided by normal flow of control, a labeled st
 
 ### 12.6.1 General
 
-A *declaration_statement* declares a local variable or constant. Declaration statements are permitted in blocks, but are not permitted as embedded statements.
+A *declaration_statement* declares a local variable, local constant, or local function. Declaration statements are permitted in blocks, but are not permitted as embedded statements.
```
jskeet, line 273: I can't remember offhand what an embedded statement is, which means another reader might not either - maybe link to it?

jskeet: Ah, it's saying that a declaration statement doesn't count as an embedded statement. Makes more sense.

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+A *local_function_declaration* may include one `async` ([§14.15](classes.md#1415-async-functions)) modifier and one `unsafe` ([§22.1](unsafe-code.md#221-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a task type  ([§14.15.1](classes.md#14151-general)). The `unsafe` modifier uses the containing lexical scope. The `async` modifier does not use the containing lexical scope. It is a compile-time error for *type_parameter_list* or *formal_parameter_list* to contain *attributes*.
+
+A local function is declared at block scope, and that function may capture variables from the enclosing scope. Captured variables must be definitely assigned before read by the body of the local function in each call to the function. The compiler shall determine which variables are definitely assigned on return (§definite-assignment-rules-for-local-function).
+
+A local function may be called from a lexical point prior to its definition. However, it is a compile-time error for the function to be declared lexically prior to the declaration of a variable it wishes to capture.
```
gafter: Change "it wishes to capture" to "used in the local function".

standard/variables.md
```diff
@@ -802,6 +802,63 @@ For a *lambda_expression* or *anonymous_method_expression* *expr* with a body (e
 >
 > *end example*
 
+#### §definite-assignment-rules-for-local-function Rules for variables in local functions
+
+Local functions are analyzed in the context of their parent method. There are two control flow paths that matter for local functions: function calls and delegate conversions.
+
+Definite assignment for the body of each local function is defined separately for each call site. At each invocation, variables captured by the local function are considered definitely assigned if they were definitely assigned at the point of call. A control flow path also exists to the local function body at this point and is considered reachable. After a call to the local function, captured variables that were definitely assigned at every control point leaving the function (`return` statements, `yield` statements, `await` statements) are considered definitely assigned after the call location.
```
ericlippert: await statements, or expressions?

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+    ;
+
+local_function_body
+    : block
+    | '=>' expression ';'
+    | '=>' null_conditional_invocation_expression ';'
+    ;
+```
+
+Grammar note: When recognising a *local_function_body* if both the *null_conditional_invocation_expression* and *expression* alternatives are applicable then the former shall be chosen.
```
jskeet: Can we reorder the options in the grammar then? In particular it's odd for the former in the prose to be the latter in the grammar, as well as making more sense (to me anyway) as "first match wins".

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+The *identifier* of a *local_function_declaration* must be unique in its declared block scope. In other words, overloaded *local_function_declaration*s are not allowed.
+
+A *local_function_declaration* may include one `async` ([§14.15](classes.md#1415-async-functions)) modifier and one `unsafe` ([§22.1](unsafe-code.md#221-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a task type  ([§14.15.1](classes.md#14151-general)). The `unsafe` modifier uses the containing lexical scope. The `async` modifier does not use the containing lexical scope. It is a compile-time error for *type_parameter_list* or *formal_parameter_list* to contain *attributes*.
+
+A local function is declared at block scope, and that function may capture variables from the enclosing scope. Captured variables must be definitely assigned before read by the body of the local function in each call to the function. The compiler shall determine which variables are definitely assigned on return (§definite-assignment-rules-for-local-function).
```
ericlippert: Is "before read" grammatical?

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+>     {
+>         throw new ArgumentOutOfRangeException(paramName: nameof(end), message: "end must be a letter");
+>     }
+>     if (end <= start)
+>     {
+>         throw new ArgumentException($"{nameof(end)} must be greater than {nameof(start)}");
+>     }
+>     return alphabetSubsetImplementation();
+>
+>     IEnumerable<char> alphabetSubsetImplementation()
```
jskeet: I normally write local functions to follow normal .NET method naming conventions, so this would be AlphabetSubsetImplementation. What do the MS docs normally do?

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+>     {
+>         for (var c = start; c < end; c++)
```
jskeet, line 468: Let's add braces around this :)

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+>             yield return c;
+>     }
+> }
+> ```
+>
+> *end example*
+
+Unless specified otherwise below, the semantics of all grammar elements is the same as for *method_declaration* ([§14.6.1](classes.md#1461-general)), read in the context of a local function instead of a method.
+
+The *identifier* of a *local_function_declaration* must be unique in its declared block scope. In other words, overloaded *local_function_declaration*s are not allowed.
```
jskeet: And it can't be the same as any variables declared in the scope either?

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+A *local_function_declaration* may include one `async` ([§14.15](classes.md#1415-async-functions)) modifier and one `unsafe` ([§22.1](unsafe-code.md#221-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a task type  ([§14.15.1](classes.md#14151-general)). The `unsafe` modifier uses the containing lexical scope. The `async` modifier does not use the containing lexical scope. It is a compile-time error for *type_parameter_list* or *formal_parameter_list* to contain *attributes*.
```
jskeet: Nit: double space after "a task type"

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+A local function is declared at block scope, and that function may capture variables from the enclosing scope. Captured variables must be definitely assigned before read by the body of the local function in each call to the function. The compiler shall determine which variables are definitely assigned on return (§definite-assignment-rules-for-local-function).
+
+A local function may be called from a lexical point prior to its definition. However, it is a compile-time error for the function to be declared lexically prior to the declaration of a variable it wishes to capture.
+
+It is a compile-time error for a local function to declare a parameter or local variable with the same name as one declared in the enclosing scope.
+
+Local function bodies are always reachable. The endpoint of a local function is reachable if the beginning point of the local function is reachable.
```
```suggestion
Local function bodies are always reachable. The endpoint of a local function declaration is reachable if the beginning point of the local function declaration is reachable.
```

standard/statements.md
jskeet: "before read" => "before being read"?

standard/statements.md
ericlippert: This might be more clear with an example.

standard/statements.md
jskeet: Ha - looks like I've noticed this a second time above...

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+Calls to local functions must be statically resolved.
```
gafter: Not sure what this means.  Should this be something said in the section describing the invocation expression, or the simple name expression?

standard/statements.md
jskeet: What's the difference between the local function body being reachable and the beginning point of the local function being reachable? (I suspect I know, but this seems unnecessarily cryptic.)

standard/statements.md
jskeet: Is this all about prohibiting type inference using dynamic types? Or is there more to it?

standard/statements.md
gafter: "Captured" is not defined.

standard/unsafe-code.md
```diff
@@ -18,11 +18,12 @@ An implementation that does not support unsafe code is required to diagnose any
...
+The unsafe features of C# are available only in unsafe contexts. An unsafe context is introduced by including an `unsafe` modifier in the declaration of a type, member, or local function, or by employing an *unsafe_statement*:
 
 - A declaration of a class, struct, interface, or delegate may include an `unsafe` modifier, in which case, the entire textual extent of that type declaration (including the body of the class, struct, or interface) is considered an unsafe context.
   > *Note*: If the *type_declaration* is partial, only that part is an unsafe context. *end note*
 - A declaration of a field, method, property, event, indexer, operator, instance constructor, finalizer, or static constructor may include an `unsafe` modifier, in which case, the entire textual extent of that member declaration is considered an unsafe context.
+- A declaration of a local function may include an `unsafe` modifier, in which case, the entire textual extent of that local function is considered an unsafe context.
```
jskeet: Any reason not to just include "local function" in the list of declarations above?

standard/variables.md
gafter: "captured" is not defined.

standard/statements.md
gafter: Having said that, it is already a restriction in the language that a variable cannot be used except after its point of declaration.

standard/basic-concepts.md
jskeet: This is covered in #590.

standard/statements.md
gafter: Perhaps would be more clear to say it may not be dynamically resolved?

standard/basic-concepts.md
BillWagner: WIth C# 7.3, the above program generates two errors:

```
1>error CS0136: A local or parameter named 'd' cannot be declared in this scope because that name is used in an enclosing local scope to define a local or parameter
1>error CS0136: A local or parameter named 'f' cannot be declared in this scope because that name is used in an enclosing local scope to define a local or parameter
```

In C# 8, and later, this compiles.

standard/statements.md
jskeet: We have two examples of what it prohibits:

- Use of `dynamic` for `params` which may change whether it's expanded or not
- Use of `dynamic` for type inference

(There may be others.)

BillWagner: See Andy's second bullet in https://github.com/dotnet/csharpstandard/pull/104#issuecomment-1150455447.

If the argument to a local function is dynamic, the function to be called must be resolved at compile time, not runtime.

standard/statements.md
jskeet: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions is inconsistent - `nthFactorial` and `GetOddSequenceEnumerator` are both used as examples...

standard/statements.md
```diff
@@ -414,6 +417,78 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+>     if (end <= start)
+>     {
+>         throw new ArgumentException($"{nameof(end)} must be greater than {nameof(start)}");
+>     }
+>     return alphabetSubsetImplementation();
+>
+>     IEnumerable<char> alphabetSubsetImplementation()
+>     {
+>         for (var c = start; c < end; c++)
+>             yield return c;
```
```suggestion
>         for (var c = start; c < end; c++)
>         {
>             yield return c;
>         }
```

standard/statements.md
```suggestion
A local function is declared at block scope, and that function may capture variables from the enclosing scope. Captured variables must be definitely assigned before being read by the body of the local function in each call to the function. The compiler shall determine which variables are definitely assigned on return (§definite-assignment-rules-for-local-function).
```

standard/statements.md
```diff
@@ -414,6 +417,80 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+> }
+> ```
+>
+> *end example*
+
+Unless specified otherwise below, the semantics of all grammar elements is the same as for *method_declaration* ([§14.6.1](classes.md#1461-general)), read in the context of a local function instead of a method.
+
+The *identifier* of a *local_function_declaration* must be unique in its declared block scope. In other words, overloaded *local_function_declaration*s are not allowed.
+
+A *local_function_declaration* may include one `async` ([§14.15](classes.md#1415-async-functions)) modifier and one `unsafe` ([§22.1](unsafe-code.md#221-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a task type  ([§14.15.1](classes.md#14151-general)). The `unsafe` modifier uses the containing lexical scope. The `async` modifier does not use the containing lexical scope. It is a compile-time error for *type_parameter_list* or *formal_parameter_list* to contain *attributes*.
```
```suggestion
A *local_function_declaration* may include one `async` ([§14.15](classes.md#1415-async-functions)) modifier and one `unsafe` ([§22.1](unsafe-code.md#221-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a task type ([§14.15.1](classes.md#14151-general)). The `unsafe` modifier uses the containing lexical scope. The `async` modifier does not use the containing lexical scope. It is a compile-time error for *type_parameter_list* or *formal_parameter_list* to contain *attributes*.
```

standard/statements.md
jskeet: (Not a change for local functions, so resolving.)

standard/statements.md
jskeet: https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/classes.md#146-methods has null_conditional_invocation_expression first with the same text as is here.

But there's no example there either.

Suggest we add an example in 14.6 and refer to it from here.

jskeet: 11.7.9 gives a little more motivation, but still doesn't have an expression-bodied member example.

standard/statements.md
jskeet: Except that in C# 8 (at least) it's valid to declare a local variable within a local function, where the name of the local variable is also the name of a local function in the enclosing scope.

standard/statements.md
```diff
@@ -414,6 +417,80 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+>         }
+>     }
+> }
+> ```
+>
+> *end example*
+
+Unless specified otherwise below, the semantics of all grammar elements is the same as for *method_declaration* ([§14.6.1](classes.md#1461-general)), read in the context of a local function instead of a method.
+
+The *identifier* of a *local_function_declaration* must be unique in its declared block scope. In other words, overloaded *local_function_declaration*s are not allowed.
```
```suggestion
The *identifier* of a *local_function_declaration* must be unique in its declared block scope. One consequence of this is that overloaded *local_function_declaration*s are not allowed.
```

standard/variables.md
BillWagner: should be "await expressions"

jskeet: It's defined for anonymous function expressions in https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#111662-captured-outer-variables.

We could potentially leave it here without definition, but raise an issue to extract the concept of a captured variable to a common place and reference it from 11.16.6.2 and here.

standard/statements.md
BillWagner: That restriction is in 7.7.

BillWagner: Edited per https://github.com/dotnet/csharpstandard/pull/104#discussion_r965276993 and added a reference to §7.7

standard/statements.md
BillWagner: This discussion is captured in #621 

Edit here to add a link to §14.6.1.

standard/statements.md
BillWagner: reworded the sentence.

standard/statements.md
BillWagner: See #621

standard/statements.md
BillWagner: I'd prefer `camelCaseLocalFunctionNames()`, because they are local in scope.  If you agree 👍 this comment. If you'd prefer `PascalCaseLocalFunctionsNames()`, because they are methods, 👎 this comment.

standard/statements.md
BillWagner: Replaced with the following text:

> If the argument to a local function is dynamic, the function to be called must be resolved at compile time, not runtime.

standard/variables.md
BillWagner: See #622

No change at this time. This should be addressed as part of #622

standard/statements.md
BillWagner: See #622

standard/statements.md
BillWagner: The second sentence has been reworded. Instead of "In other words", it says "One consequence of this is that..."

Because variables are *identifiers*, the text is valid for C# 7 and later.

standard/unsafe-code.md
BillWagner: I wrote this bullet to match the existing bullet items.

standard/statements.md
BillWagner: The discussion is here: https://github.com/dotnet/csharpstandard/pull/104#discussion_r616112191

I believe it is to have the correct rules on the reachability of any statement following a local function body. Maybe @gafter can suggest more clear language?

BillWagner:
@jskeet 

I've updated this PR based on all comments from the last meeting. There are two open conversations to resolve. The first, we need to pick a consensus. The second, we need to agree on the wording.

Can we add this to the next meeting agenda?

jskeet:
@BillWagner: Will do :)

standard/statements.md
KalleOlaviNiemitalo: An example could help

```C#
class C {
    int M() {
        // L();
        return 1;
        int L() { return 2; }
        return 3;
    }
}
```

Because of the statement `return 1;`, the beginning point of the local function L is not reachable. Consequently, the ending point of L is not reachable, and the statement `return 3;` is not reachable. The statement `return 2;` in the body of L is considered reachable even if L is not called.

standard/statements.md
BillWagner: Resolution from the committee: Let's use `PascalCaseLocalFunctionNames()`

BillWagner:
@jskeet Check the final commit, and when you're ready, let's :shipit:

standard/statements.md
```diff
@@ -414,6 +417,100 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+The *identifier* of a *local_function_declaration* must be unique in its declared block scope. One consequence of this is that overloaded *local_function_declaration*s are not allowed.
+
+A *local_function_declaration* may include one `async` ([§14.15](classes.md#1415-async-functions)) modifier and one `unsafe` ([§22.1](unsafe-code.md#221-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a task type ([§14.15.1](classes.md#14151-general)). The `unsafe` modifier uses the containing lexical scope. The `async` modifier does not use the containing lexical scope. It is a compile-time error for *type_parameter_list* or *formal_parameter_list* to contain *attributes*.
+
+A local function is declared at block scope, and that function may capture variables from the enclosing scope. It is a compile time error if a captured variables is read by the body of the local function if it is not definitely assigned before each call to the function. The compiler shall determine which variables are definitely assigned on return (§definite-assignment-rules-for-local-function).
```
jskeet: Nit: "a captured variables" => "a captured variable".

jskeet: And possibly change "if it is not" to "but is not". (It's a little odd to have the two "ifs" in the same sentence.) That could be a later change.

standard/statements.md
```diff
@@ -414,6 +417,100 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+
+A local function may be called from a lexical point prior to its definition. However, it is a compile-time error for the function to be declared lexically prior to the declaration of a variable used in the local function (§7.7).
```
jskeet: definition => declaration? (We have local function declarations, not local function definitions.)

standard/statements.md
jskeet: "compile time" => "compile-time"

jskeet (approved):
Just a few nits left; happy for it to be merged after that.

PR #201, opened 2021-01-13
update licenses

BillWagner:
Per committee meeting, and discussion with .NET Foundation executive director, we're using the Creative Commons v4 attribution license, and the MIT license for tools code.


jskeet (approved):
This is fine - I think it would be good to clarify the position in Contribute.md or README.md to refer to these files.

jskeet (approved):

PR #202, opened 2021-01-14
Update v6-feature-tracker.md

RexJaeschke:
Added a Status column to the table, so can see which features have been completed.


RexJaeschke:
This appears to be obsolete.

Issue #169, opened 2017-04-24
12.7.7.2 and 12.18.2 Array access evaluation order for assignments


jskeet:
13.7.7.2 (array access) suggests that evaluating an array access expression performs nullity and index checks.

13.18.2 (simple assignment) states for an assignment `x = y`:

> - If x is classified as a variable:
>   - x is evaluated to produce the variable.
>   - y is evaluated and, if required, converted to the type of x through an implicit conversion (§12.2).

That all sounds fine, but my understanding of reality is that the nullity check and array index check are only performed *after* `y` is evaluated. Sample code to show that...

**Null check comes after RHS**

```csharp
using System;

public class Test
{
    static void Main()
    {
        int[] array = null;
        array[10] = SideEffect(); // Prints "Side effect" then throws
    }

    private static int SideEffect()
    {
        Console.WriteLine("Side effect");
        return 10;
    }
}
```

**Index check comes after RHS**

```csharp
using System;

public class Test
{
    static void Main()
    {
        int[] array = new int[5];
        array[10] = SideEffect(); // Prints "Side effect" then throws
    }

    private static int SideEffect()
    {
        Console.WriteLine("Side effect");
        return 10;
    }
}
```

I have very vague recollections of this being properly specified somewhere, but I can't easily find it - and assuming it *does* exist, I think we should have a reference to it within at least one of these sections.

Assigning to Mads initially as he may well know what's going on...

jskeet:
Assigned to @MadsTorgersen to consider before the next meeting.

MadsTorgersen:
Jon is right.

According to [12.7.7.1 Element access](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#12771-general) `a[i]` is an array access when `a` is of an array type.

In [12.7.7.2 Array access](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#12772-array-access) such an array access is classified as a variable, and it is described in detail how to compute the location referenced by the variable.

In [12.18.2 Simple assignment](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#12182-simple-assignment) it is clear that the variable on the left-hand side (the array access) is computed before the right-hand side:

> - If `x` is classified as a variable:
>   - `x` is evaluated to produce the variable.
>   - `y` is evaluated and, if required, converted to the type of `x` through an implicit conversion ([§11.2](conversions.md#112-implicit-conversions)).
[...]

PR #69, opened 2020-12-14
C# 7.x: Adding more support for expression-bodied members

RexJaeschke:

Issue #66, opened 2020-12-10
v7.x Q re throw expressions


RexJaeschke:
In Draft PR https://github.com/ECMA-TC49-TG2/csharpstandard/pull/65, I've added almost all the edits to support the addition of throw expressions. Here's what's missing:

1. I've added the new section "The throw expression operator" to the the Expressions clause, but it needs to be "plugged into" the grammar, so it has precedence. (Also, for now, I've put this new section under primary expressions, but that might not be correct.)

KalleOlaviNiemitalo:
The interaction of <https://github.com/dotnet/csharpstandard/pull/65> and <https://github.com/dotnet/csharpstandard/pull/213> does not allow a throw expression as an operand of a ref conditional expression:

```csharp
ref int F(bool b, ref int i) => ref b ? ref i : ref throw null;
```

because "Both *expression*s shall be variables having the same type" and "A *throw_expression* has no type." I suppose this restriction is by design.

BillWagner:
See PR #65 

Note that instead of "primary", the throw expression precedence should be the same as the null coalescing question.

jskeet:
@KalleOlaviNiemitalo: It's not clear to me from your comment whether this is how the compiler actually behaves, or whether this is an area you think this needs a spec change. I want to close this issue when #65 is merged, so please could you file a new issue if you believe we need to do more work on this?

KalleOlaviNiemitalo:
@jskeet the standard text matches compiler behaviour and I think the behaviour is reasonable so no action is required.

BillWagner:
@jskeet PR #65 has been merged. I think that the questions raised here have been answered.

Can we close this as fixed in #65

jskeet:
Yes, I think so.

PR #213, opened 2021-01-24
C# 7.x: ref locals and returns

RexJaeschke:
This (Draft) PR is not quite complete.


RexJaeschke:
Completed edits. Removed Draft status.

KalleOlaviNiemitalo:
Missing:

* `=> ref` in _method_body_ and similar. However, the draft-v7 branch does not even allow `=>` there yet; <https://github.com/dotnet/csharpstandard/pull/4> was merged to draft-v6 but those changes are not in draft-v7. <https://github.com/dotnet/csharpstandard/pull/69> does not add them, either.

* ref return type in _property_declaration_ and _indexer_declaration_. Roslyn allows these, and at least System.Span\<T> and System.Span\<T>.Enumerator have public members like that.

* ref return type in _operator_declaration_. AFAIK, Roslyn does not allow this, so perhaps you'll decide to leave this out, even though I don't foresee any specification problem in allowing this in _unary_operator_declarator_ and _binary_operator_declarator_. In contrast, _conversion_operator_declarator_ converting to a reference seems like it would require changes elsewhere in the standard.

KalleOlaviNiemitalo:
As I wrote in <https://github.com/dotnet/csharpstandard/issues/169#issuecomment-874671673>, the standard should describe the runtime array element type checks required for ref locals and returns. Moreover, it should describe how those checks are skipped when `ref readonly` is used. See also <https://github.com/dotnet/csharpstandard/issues/264#issuecomment-874738023> regarding type checks for `in` parameters.

333fred:
One thing I don't see in this spec, @RexJaeschke, is how to handle a case like https://github.com/dotnet/roslyn/issues/53113 (forgive me if it's already subsumed by other parts of the spec; if it is, I couldn't find what does). Based on how we handle `M(ref *(int*)0)`, it seems like such code should be allowed and should not dereference the pointer, but I didn't see anything in the ref locals section talking about it.

BillWagner:
Converted to draft status (C# 7 feature)

KalleOlaviNiemitalo:
If <https://github.com/dotnet/csharpstandard/pull/104> is merged before this, I suppose this will need to add `=> ref` for local functions, too.

standard/expressions.md
```diff
@@ -4637,6 +4640,18 @@ The conditional operator is right-associative, meaning that operations are group
 
 The first operand of the `?:` operator shall be an expression that can be implicitly converted to `bool`, or an expression of a type that implements `operator true`. If neither of these requirements is satisfied, a compile-time error occurs.
 
+If `ref` is present:
+Both *expression*s shall be variables having the same type, and that type is the type of the result.
```
KalleOlaviNiemitalo: OK.  Because `throw new Exception()` is not a variable, it cannot be used with `ref` in `b ? ref i : ref throw new Exception()`, which might otherwise be allowed by <https://github.com/dotnet/csharpstandard/pull/65>.

KalleOlaviNiemitalo: I had already noted the same in <https://github.com/dotnet/csharpstandard/issues/66#issuecomment-872961242>.

gafter:
This still needs to safety rules to be incorporated. See https://github.com/dotnet/csharpstandard/issues/334 and https://github.com/dotnet/csharplang/blob/standard-proposals/proposals/csharp-7.2/span-safety.md#draft-language-specification . Those safety rules are for ref return, ref locals, and ref structs.

standard/expressions.md
BillWagner: updated

BillWagner:
> One thing I don't see in this spec, @RexJaeschke, is how to handle a case like [dotnet/roslyn#53113](https://github.com/dotnet/roslyn/issues/53113) (forgive me if it's already subsumed by other parts of the spec; if it is, I couldn't find what does). Based on how we handle `M(ref *(int*)0)`, it seems like such code should be allowed and should not dereference the pointer, but I didn't see anything in the ref locals section talking about it.

I added language for this in the ref assignment clause.

BillWagner:
> If #104 is merged before this, I suppose this will need to add `=> ref` for local functions, too.

Done. The only changed seems to be the grammar

BillWagner:
@jskeet 

This one is now ready for final review and merge.

The remaining section renumbering issues are links to the ref safety rules that are in #742

standard/classes.md
```diff
@@ -1969,7 +1969,11 @@ A declaration has a valid combination of modifiers if all of the following are t
 - If the declaration includes the `sealed` modifier, then the declaration also includes the `override` modifier.
 - If the declaration includes the `partial` modifier, then it does not include any of the following modifiers: new, `public`, `protected`, `internal`, `private`, `virtual`, `sealed`, `override`, `abstract`, or `extern`.
 
-The *return_type* of a method declaration specifies the type of the value computed and returned by the method. The *return_type* is `void` if the method does not return a value. If the declaration includes the `partial` modifier, then the return type shall be `void` ([§15.6.9](classes.md#1569-partial-methods)). If the declaration includes the `async` modifier then the return type shall be `void` or a *task type* ([§15.15.1](classes.md#15151-general)).
+It is a compile-time error to have both `ref` and a *return_type* of `void`.
+
+If `ref` is present, the method ***returns-by-ref*** ; otherwise, if *return_type* is `void`, the method ***returns-no-value***; otherwise, the method ***returns-by-value***.
```
```suggestion
If `ref` is present, the method ***returns-by-ref***; otherwise, if *return_type* is `void`, the method ***returns-no-value***; otherwise, the method ***returns-by-value***.
```

standard/classes.md
```diff
@@ -1969,7 +1969,11 @@ A declaration has a valid combination of modifiers if all of the following are t
...
+
+The *return_type* of a method declaration specifies the type of the result, if any, returned by the method. A returns-no-value method does not return a value. A returns-by-ref method returns a *variable_reference* (§9.5), that is optionally read-only. A returns-by-value method returns a value. If the declaration includes the `partial` modifier, then *return_type* shall be `void` ([§15.6.9](classes.md#1569-partial-methods)). If the declaration includes the `async` modifier then *return_type* shall be `void` or a *task type* ([§15.15.1](classes.md#15151-general)).
```
jskeet: I'd expect something at the end of here prohibiting a returns-by-ref method from having the `async` modifier.

standard/classes.md
```diff
@@ -2994,11 +2998,11 @@ The ***effective return type*** of a method is `void` if the return type is `voi
 
 When the effective return type of a method is `void` and the method has a block body, `return` statements ([§13.10.5](statements.md#13105-the-return-statement)) in the block shall not specify an expression. If execution of the block of a void method completes normally (that is, control flows off the end of the method body), that method simply returns to its caller.
 
-When the effective return type of a method is `void` and the method has an expression body, the expression `E` shall be a *statement_expression*, and the body is exactly equivalent to a statment body of the form `{ E; }`.
+When the effective return type of a method is `void` and the method has an expression body, the expression `E` shall be a *statement_expression*, and the body is exactly equivalent to a statement body of the form `{ E; }`.
```
```suggestion
When the effective return type of a method is `void` and the method has an expression body, the expression `E` shall be a *statement_expression*, and the body is exactly equivalent to a block body of the form `{ E; }`.
```

(Looks like we missed this when we changed "statement body" to "block body" before - presumably due to the typo.)

standard/classes.md
```diff
@@ -2994,11 +2998,11 @@ The ***effective return type*** of a method is `void` if the return type is `voi
...
+For a returns-by-value method ([§15.6.1](classes.md#1561-general)), each return statement in that method's body shall specify an expression that is implicitly convertible to the effective return type. The endpoint of the method body shall not be reachable. In other words, control is not permitted to flow off the end of the method body.
 
-When the effective return type of a method is not `void` and the method has an expression body, `E`, the expression shall be implicitly convertible to the effective return type, and the body is exactly equivalent to a block body of the form `{ return E; }`.
+For a returns-by-ref method ([§15.6.1](classes.md#1561-general)), each return statement in that method's body shall specify an expression whose type is that of the effective return type, and has a *ref_safe_scope* of *calling_method* (§ref-span-safety-escape-scopes). The endpoint of the method body shall not be reachable. In other words, control is not permitted to flow off the end of the method body.
```
jskeet: I wonder whether it would be clearer to have a new paragraph with the method endpoint not being reachable for both returns-by-value and returns-by-ref, rather than having this duplication.

standard/classes.md
```diff
@@ -3077,7 +3081,9 @@ A *property_declaration* may include a set of *attributes* ([§22](attributes.md
 
 Property declarations are subject to the same rules as method declarations ([§15.6](classes.md#156-methods)) with regard to valid combinations of modifiers.
 
-The *type* of a property declaration specifies the type of the property introduced by the declaration, and the *member_name* ([§15.6.1](classes.md#1561-general)) specifies the name of the property. Unless the property is an explicit interface member implementation, the *member_name* is simply an *identifier*. For an explicit interface member implementation ([§18.6.2](interfaces.md#1862-explicit-interface-member-implementations)), the *member_name* consists of an *interface_type* followed by a “`.`” and an *identifier*.
+The *type* of a prop declaration specifies the type of the result, if any, returned by the method. A returns-no-value method does not return a value. A returns-by-ref method returns a *variable_reference* (§9.5), that is optionally read-only.
```
jskeet: This paragraph looks like it's not finished - "prop" should be "property", and it then starts talking about a method. Presumably this is meant to be about a ref or ref-readonly property. Given the paragraph below, should this just be deleted?

standard/delegates.md
```diff
@@ -41,7 +40,11 @@ The `public`, `protected`, `internal`, and `private` modifiers control the acces
 
 The delegate’s type name is *identifier*.
 
-The optional *formal_parameter_list* specifies the parameters of the delegate, and *return_type* indicates the return type of the delegate.
+As with methods ([§15.6.1](classes.md#1561-general)), if `ref` is present, the delegate returns-by-ref; otherwise, if *return_type* is `void`, the delegate returns-no-value; otherwise, the delegate returns-by-value.
+
+The optional *formal_parameter_list* specifies the parameters of the delegate.
+
+*return_type* indicates the return type of the delegate. The optional `ref` indicates that a reference to a *variable_reference* (§9.5) is returned, with the optional `readonly` indicating that that variable is read-only.
```
jskeet: Just "indicates that a *variable_reference* is returned"? That would be consistent with the method details given above - I don't think we talk about a reference to a *variable_reference* elsewhere.

standard/delegates.md
```diff
@@ -83,10 +86,11 @@ Except for instantiation, any operation that can be applied to a class or class
...
 
+- `D` and `M` both have a *return_type* preceded by `ref readonly`, both have a *return_type* preceded by `ref` only, or both have a *return_type* not preceded by `ref`.
 - `D` and `M` have the same number of parameters, and each parameter in `D` has the same `ref` or `out` modifiers as the corresponding parameter in `M`.
 - For each value parameter (a parameter with no `ref` or `out` modifier), an identity conversion ([§10.2.2](conversions.md#1022-identity-conversion)) or implicit reference conversion ([§10.2.8](conversions.md#1028-implicit-reference-conversions)) exists from the parameter type in `D` to the corresponding parameter type in `M`.
 - For each `ref` or `out` parameter, the parameter type in `D` is the same as the parameter type in `M`.
-- An identity or implicit reference conversion exists from the return type of `M` to the return type of `D`.
+- If `D` and `M` are returns-by-value ([§15.6.1](classes.md#1561-general), [§20.2](delegates.md#202-delegate-declarations)), an identity or implicit reference conversion exists from the return type of `M` to the return type of `D`.
```
jskeet, line 92: I suspect we need another item here about compatibility of return-types for returns-by-ref. Otherwise `delegate ref int X()` and `delegate ref string Y()` are compatible with each other.

standard/delegates.md
```diff
@@ -83,10 +86,11 @@ Except for instantiation, any operation that can be applied to a class or class
 
 A method or delegate type `M` is ***compatible*** with a delegate type `D` if all of the following are true:
 
+- `D` and `M` both have a *return_type* preceded by `ref readonly`, both have a *return_type* preceded by `ref` only, or both have a *return_type* not preceded by `ref`.
```
jskeet: I suspect what we *actually* want is that their "return mode" is the same *and* if that mode is returns-by-ref then they're both ref readonly or just ref.

At the moment, I think this list would get confused when considering `void X()` and `int X()`. (They're not *both* returns-by-value, so the final item arguably doesn't need to be checked...)

standard/expressions.md
```diff
@@ -1737,8 +1737,9 @@ The optional *argument_list* ([§12.6.2](expressions.md#1262-argument-lists)) pr
...
+- Otherwise, if the *invocation_expression* invokes a returns-by-ref method ([§15.6.1](classes.md#1561-general)) or a returns-by-ref delegate, the result is a variable with an associated type of the return type of the method or delegate. If the invocation is of an instance method, and the receiver is of a class type `T`, the associated type is picked from the first declaration or override of the method found when starting with `T` and searching through its base classes.
+- Otherwise, the *invocation_expression* invokes a returns-by-val method ([§15.6.1](classes.md#1561-general)) or returns-by-val delegate, and the result is a value, with an associated type of the return type of the method or delegate. If the invocation is of an instance method, and the receiver is of a class type `T`, the associated type is picked from the first declaration or override of the method found when starting with `T` and searching through its base classes.
```
```suggestion
- Otherwise, the *invocation_expression* invokes a returns-by-value method ([§15.6.1](classes.md#1561-general)) or returns-by-value delegate, and the result is a value, with an associated type of the return type of the method or delegate. If the invocation is of an instance method, and the receiver is of a class type `T`, the associated type is picked from the first declaration or override of the method found when starting with `T` and searching through its base classes.
```

standard/expressions.md
```diff
@@ -4562,7 +4565,19 @@ The conditional operator is right-associative, meaning that operations are group
 
 The first operand of the `?:` operator shall be an expression that can be implicitly converted to `bool`, or an expression of a type that implements `operator true`. If neither of these requirements is satisfied, a compile-time error occurs.
 
-The second and third operands, `x` and `y`, of the `?:` operator control the type of the conditional expression. If both `x` and `y` are *default_literal*s ([§12.8.20](expressions.md#12820-default-value-expressions)), a compile-time error occurs.
+If `ref` is present:
+Both *expression*s shall be variables having the same type, and that type is the type of the result.
```
jskeet: How about different types but with an identity conversion?

```csharp
dynamic d = "";
object o = "";
bool condition = true;
ref object x = ref (condition ? ref o : ref d);
```

standard/expressions.md
```diff
@@ -4562,7 +4565,19 @@ The conditional operator is right-associative, meaning that operations are group
...
+The result is a variable, which is writeable if both *expression* variables are writeable.
+The run-time processing of a ref conditional expression of the form `b ? ref x : ref y` consists of the following steps:
+
+- First, `b` is evaluated, and the `bool` value of `b` is determined:
+  - If an implicit conversion from the type of `b` to `bool` exists, then this implicit conversion is performed to produce a `bool` value.
+  - Otherwise, the `operator true` defined by the type of `b` is invoked to produce a `bool` value.
+- If the `bool` value produced by the step above is `true`, then `x` is evaluated and the resulting reference becomes the result of the conditional expression.
+- Otherwise, `y` is evaluated and the resulting reference becomes the result of the conditional expression.
```
jskeet: "resulting variable reference" here and above just to make it really clear?

standard/expressions.md
```diff
@@ -4562,7 +4565,19 @@ The conditional operator is right-associative, meaning that operations are group
...
+
+If `ref` is not present:
+The second and third operands, `x` and `y`, of the `?:` operator control the type of the conditional expression.
```
```suggestion
If `ref` is not present, the second and third operands, `x` and `y`, of the `?:` operator control the type of the conditional expression:
```

jskeet: And I think we're missing the "If both `x` and `y` are *default_literal*s ([§12.8.20](expressions.md#12820-default-value-expressions)), a compile-time error occurs." that we had before.

standard/expressions.md
```diff
@@ -6195,6 +6212,55 @@ When a property or indexer declared in a *struct_type* is the target of an assig
 >
 > *end example*
 
+### §ref-assignment-new-clause Ref assignment
+
+The left operand shall be an expression that binds to a ref local variable, a ref parameter (other than `this`), or an out parameter. The right operand shall be an expression that yields a *variable_reference* designating a value of the same type as the left operand.
```
jskeet, line 6228: I think it would be good to have an introductory sentence to give context. Just something as simple as "The `= ref` operator is known as the *ref assignment* operator." Just something to ease the reader back into the context.

standard/expressions.md
```diff
@@ -6195,6 +6212,55 @@ When a property or indexer declared in a *struct_type* is the target of an assig
...
+
+It is a compile time error if the *ref_safe_scope* of the left operand is wider than the *ref_safe_scope* of the left operand (§ref-span-safety-escape-scopes).
+
+If the left operand is a writeable ref (i.e., it designates anything other than a `ref readonly` local or  `in` parameter), then the right operand shall be a writeable *variable_reference*. If the right operand variable is writeable, the left operand may be declared `ref` or `ref readonly`.
+
+The operation makes the left operand an alias of the right operand variable. The alias may be made read-only even if the right operand variable is writeable.
+
+The ref assignment operator yields a *variable_reference* of the assigned type. It is writeable if the left operand is writeable.
+
+The ref assignment operator shall not evaluate the value of right hand operand, only its reference.
```
jskeet: I think this needs more precision. Consider:

```csharp
static void Main()
{
    int a = 0;
    ref int b = ref M(ref a);
}

static ref int M(ref int y) => ref y;
```

Here the second line of `Main` *does* evaluate the right operand, in that it calls the method. It doesn't then evaluate the variable reference return by `M`.

I suspect we can basically get away with deleting this line entirely - other than *potentially* for the note below.

standard/expressions.md
```diff
@@ -6195,6 +6212,55 @@ When a property or indexer declared in a *struct_type* is the target of an assig
...
+> r1 = ref M3();        // error; M3 returns a ref readonly, which r1 cannot honor
+> ref readonly r2 = …;
+> r2 = ref M2();        // OK; makes an alias, adding read-only protection
+> r2 = ref M3();        // OK; makes an alias and honors the read-only
+> r2 = ref (r1 = ref M2());  // OK; r1 is an alias to a writable variable,
+>                       // r2 is an alias (with read-only access) to the same variable
+> // OK, r3 is a null reference.
+> ref int r3 = ref System.Runtime.CompilerServices.Unsafe.NullRef<int>();
+> // OK, both r3 and f4 are null references
+> ref int r4 = ref r4;
```
jskeet: Should this be `ref int r4 = ref r3;`?

standard/expressions.md
```diff
@@ -6195,6 +6212,55 @@ When a property or indexer declared in a *struct_type* is the target of an assig
...
+<!-- markdownlint-enable MD028 -->
+> *Example*: Here are some examples of using `= ref`:
+>
+> ```csharp
+> public static int M1() { … }
+> public static ref int M2() { … }
+> public static ref uint M2u() { … }
+> public static ref readonly int M3() { … }
+> 
+> ref r1 = …;
```
jskeet: Don't we need a type here? Also, this is going to be hard to turn into a compilable example due to the "...".

Maybe we should make the first "OK" example the local variable initializer for both r1 and r2.

standard/statements.md
```diff
@@ -1029,12 +1035,16 @@ The `foreach` statement enumerates the elements of a collection, executing an em
...
+The *local_variable_type* and *Identifier* of a foreach statement declare the ***iteration variable*** of the statement. If the `var` identifier is given as the *local_variable_type*, and no type named `var` is in scope, the iteration variable is said to be an ***implicitly typed iteration variable***, and its type is taken to be the element type of the `foreach` statement, as specified below.
+
+If *foreach_statement* contains both or neither `ref` and `readonly`, the iteration variable denotes a variable that is treated as read-only. Otherwise, if *foreach_statement* contains `ref` without `readonly`, the iteration variable denotes a variable that shall be writable.
+
+The iteration variable corresponds to a local variable with a scope that extends over the embedded statement. During execution of a `foreach` statement, the iteration variable represents the collection element for which an iteration is currently being performed. A compile-time error occurs if the embedded statement attempts to modify an iteration variable that denotes a read-only variable (via assignment or the `++` and `--` operators) or pass that iteration variable as a `ref` or `out` parameter.
```
```suggestion
The iteration variable corresponds to a local variable with a scope that extends over the embedded statement. During execution of a `foreach` statement, the iteration variable represents the collection element for which an iteration is currently being performed. If the iteration variable denotes a read-only variable, a compile-time error occurs if the embedded statement attempts to modify it (via assignment or the `++` and `--` operators) or pass it as a `ref` or `out` parameter.
```

(Personally I find that much clearer - otherwise I want it to be "passes" instead of "pass", *and* the part about it being read-only doesn't look like it applies to the passing part.)

standard/statements.md
```diff
@@ -1386,25 +1396,38 @@ Because a `goto` statement unconditionally transfers control elsewhere, the end
...
+
+It is a compile-time error to use a return-by-ref from a method declared as being returns-no-value or returns-by-value.
+
+It is a compile-time error to use a return-by-value from a method declared as being returns-no-value or returns-by-ref.
+
+It is a compile-time error to use a return-by-ref if *expression* is not a *variable_reference* or is a reference to a variable whose *ref_safe_scope* is not *calling_method* (§ref-span-safety-escape-scopes).
+
+It is a compile-time error to use a return-by-ref from a method declared with the *method_modifier* `async`.
+
+A function member is said to ***compute a value*** if it is a method with a returns-by-value method ([§15.6.11](classes.md#15611-method-body)), the `get` accessor of a property or indexer, or a user-defined operator. Function members that are returns-no-value do not compute a value or are methods with the effective return type `void`, `set` accessors of properties and indexers, `add` and `remove` accessors of event, instance constructors, static constructors and finalizers. Function members that are returns-by-ref do not compute a value.
```
jskeet: What about the `get` accessor of a ref property? (Basically I suspect this should be a returns-by-value property or returns-by-value indexer.)

standard/statements.md
```diff
@@ -1386,25 +1396,38 @@ Because a `goto` statement unconditionally transfers control elsewhere, the end
...
 
-Within a function member, a `return` statement with no expression can only be used if the function member does not compute a value. Within a function member, a `return` statement with an expression can only be used if the function member computes a value. Where the `return` statement includes an expression, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) shall exist from the type of the expression to the effective return type of the containing function member.
+For a return-by-value, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) shall exist from the type of *expression* to the effective return type ([§15.6.11](classes.md#15611-method-body)) of the containing function member. For a return-by-ref, the type of *expression* shall be the same as the effective return type of the containing function member.
```
jskeet: Identity conversions in the returns-by-ref case?

standard/classes.md
BillWagner: Removed. I think you're right.

standard/expressions.md
BillWagner: Updated.

Added text for implicitly typed variables (which prefers dynamic).

standard/expressions.md
BillWagner: I made that change. I think I like it. 

Worth discussion in our meeting:  Should `= ref` be described as a modifier on the right operand of an assignment, or a new operator?

standard/expressions.md
BillWagner: done

standard/statements.md
BillWagner: added that clarification

jskeet (commented):
Action from 2023-04-19: @BillWagner to work out what's happened and get the PR back to a sensible state.

standard/delegates.md
jskeet:
```suggestion
*return_type* indicates the return type of the delegate. The optional `ref` indicates that a *variable reference* (§9.5) is returned, with the optional `readonly` indicating that that variable is read-only.
```

standard/delegates.md
jskeet: Action 2023-04-19: @BillWagner to fix offline.

standard/delegates.md
jskeet: Action 2023-04-19: @BillWagner to fix.

standard/expressions.md
jskeet: Action 2023-04-19: It's a different operator. @BillWagner  to check that it's listed a) in the table of all operators; b) in the grammar.

standard/expressions.md
jskeet: Action 2023-04-19: This is misworded: the RHS *is* evaluated, to a storage location. But the resulting storage location is not read. (So if the storage location is null - which we're going to look at or excise - we *do not* get an NRE.)

standard/expressions.md
```diff
@@ -6195,6 +6211,58 @@ When a property or indexer declared in a *struct_type* is the target of an assig
...
+
+If the left operand is a writeable ref (i.e., it designates anything other than a `ref readonly` local or  `in` parameter), then the right operand shall be a writeable *variable_reference*. If the right operand variable is writeable, the left operand may be declared `ref` or `ref readonly`.
+
+The operation makes the left operand an alias of the right operand variable. The alias may be made read-only even if the right operand variable is writeable.
+
+The ref assignment operator yields a *variable_reference* of the assigned type. It is writeable if the left operand is writeable.
+
+The ref assignment operator shall not evaluate the value of right hand operand, only its reference.
+
+> *Note*: The previous rule allows assigning a null reference to another reference variable. *end note*
```
jskeet: Action 2023-04-19: Delete this.

standard/expressions.md
```diff
@@ -6195,6 +6211,58 @@ When a property or indexer declared in a *struct_type* is the target of an assig
...
+> r1 = ref M3();        // error; M3 returns a ref readonly, which r1 cannot honor
+> ref readonly r2 = …;
+> r2 = ref M2();        // OK; makes an alias, adding read-only protection
+> r2 = ref M3();        // OK; makes an alias and honors the read-only
+> r2 = ref (r1 = ref M2());  // OK; r1 is an alias to a writable variable,
+>                       // r2 is an alias (with read-only access) to the same variable
+> // OK, r3 is a null reference.
+> ref int r3 = ref System.Runtime.CompilerServices.Unsafe.NullRef<int>();
+> // OK, both r3 and f4 are null references
+> ref int r4 = ref r3;
```
```suggestion
```

(Delete everything about nulls.)

jskeet (commented):
Nearly there, but another round of review required for some of the more fiddly bits.

gafter:
Need grammar rules for `= ref` and `... ? ref ... : ref ...` operators.

standard/delegates.md
BillWagner: This one is fine, because it specifies returns-by-value.

Move the first bullet to last, and make it only for returns-by-ref.

standard/delegates.md
BillWagner: Moved to the last bullet, specify only returns-by-ref, and specify an identity conversion between the return types.

standard/expressions.md
BillWagner: reworded

standard/expressions.md
BillWagner: verified and updated.

jskeet (approved):

standard/expressions.md
```diff
@@ -6195,6 +6213,50 @@ When a property or indexer declared in a *struct_type* is the target of an assig
 >
 > *end example*
 
+### §ref-assignment-new-clause Ref assignment
```
BillWagner, line 6224: Note to update before merging:  Once #742 is merged, the new clause for `ref` safety should link to this location.

standard/expressions.md
```diff
@@ -6195,6 +6213,50 @@ When a property or indexer declared in a *struct_type* is the target of an assig
...
+> r1 = ref M1();        // Error; M1 returns a value, not a reference
+> r1 = ref M2();        // OK; makes an alias
+> r1 = ref M2u();       // Error; lhs and rhs have different types
+> r1 = ref M3();        // error; M3 returns a ref readonly, which r1 cannot honor
+> ref readonly r2 = …;
+> r2 = ref M2();        // OK; makes an alias, adding read-only protection
+> r2 = ref M3();        // OK; makes an alias and honors the read-only
+> r2 = ref (r1 = ref M2());  // OK; r1 is an alias to a writable variable,
+>                       // r2 is an alias (with read-only access) to the same variable
+>
```
jskeet: Remove blank line

standard/expressions.md
```diff
@@ -4562,7 +4565,20 @@ The conditional operator is right-associative, meaning that operations are group
 
 The first operand of the `?:` operator shall be an expression that can be implicitly converted to `bool`, or an expression of a type that implements `operator true`. If neither of these requirements is satisfied, a compile-time error occurs.
 
-The second and third operands, `x` and `y`, of the `?:` operator control the type of the conditional expression. If both `x` and `y` are *default_literal*s ([§12.8.20](expressions.md#12820-default-value-expressions)), a compile-time error occurs.
+If `ref` is present:
```
BillWagner, line 4574: @jskeet to make a suggestion for a note on how this relates to `= ref`

BillWagner: @jskeet 

I made a note here in the next commit. That way, we don't lose this discussion. We can update the text of the note via a suggestion or PR after I merge these into a single feature branch.

standard/expressions.md
BillWagner: It was already referenced. I moved it up 3 paragraphs.

standard/expressions.md
KalleOlaviNiemitalo: `i = b ? ref x : throw null;` is discussed in <https://github.com/dotnet/csharplang/discussions/919>.

PR #216, opened 2021-01-26
C# 7.x: non-trailing named args

RexJaeschke:


BillWagner:
Converting to draft until we're ready for C# 7

standard/expressions.md
```diff
@@ -579,6 +579,8 @@ The corresponding parameters for function member arguments are established as fo
 - For properties, when invoking the get accessor there are no arguments. When invoking the set accessor, the expression specified as the right operand of the assignment operator corresponds to the implicit value parameter of the set accessor declaration.
 - For user-defined unary operators (including conversions), the single operand corresponds to the single parameter of the operator declaration.
 - For user-defined binary operators, the left operand corresponds to the first parameter, and the right operand corresponds to the second parameter of the operator declaration.
+- An unnamed argument corresponds to no parameter when it is after an out-of-position named argument or a named params argument.
+  > *Note*: This prevents `void M(bool a = true, bool b = true, bool c = true, );` being invoked by `M(c: false, valueB);`. The first argument is used out-of-position (the argument is used in first position, but the parameter named `c` is in third position), so the following arguments should be named. In other words, non-trailing named arguments are only allowed when the name and the position result in finding the same corresponding parameter. *end note*
```
KalleOlaviNiemitalo: This seems to have an excess comma after `bool c = true`.

BillWagner: Removing extra comma:

```suggestion
  > *Note*: This prevents `void M(bool a = true, bool b = true, bool c = true);` being invoked by `M(c: false, valueB);`. The first argument is used out-of-position (the argument is used in first position, but the parameter named `c` is in third position), so the following arguments should be named. In other words, non-trailing named arguments are only allowed when the name and the position result in finding the same corresponding parameter. *end note*
```

BillWagner (commented):
Adding suggestion

BillWagner:
@RexJaeschke I think this is ready. I'll remove the draft status so it's in my queue to review.

standard/expressions.md
jskeet: Resolve the trailing space as well? (And we should commit the suggestion before resolving the conversation.)

standard/expressions.md
```diff
@@ -581,6 +581,8 @@ The corresponding parameters for function member arguments are established as fo
 - For properties, when invoking the get accessor there are no arguments. When invoking the set accessor, the expression specified as the right operand of the assignment operator corresponds to the implicit value parameter of the set accessor declaration.
 - For user-defined unary operators (including conversions), the single operand corresponds to the single parameter of the operator declaration.
 - For user-defined binary operators, the left operand corresponds to the first parameter, and the right operand corresponds to the second parameter of the operator declaration.
+- An unnamed argument corresponds to no parameter when it is after an out-of-position named argument or a named params argument.
```
```suggestion
- An unnamed argument corresponds to no parameter when it is after an out-of-position named argument or a named argument that corresponds to a parameter array.
```

standard/expressions.md
```diff
@@ -581,6 +581,8 @@ The corresponding parameters for function member arguments are established as fo
...
+  > *Note*: This prevents `void M(bool a = true, bool b = true, bool c = true, );` being invoked by `M(c: false, valueB);`. The first argument is used out-of-position (the argument is used in first position, but the parameter named `c` is in third position), so the following arguments should be named. In other words, non-trailing named arguments are only allowed when the name and the position result in finding the same corresponding parameter. *end note*
```
jskeet: Would "following arguments" be better as "subsequent arguments" or "arguments following it"? Otherwise it sounds like they may be "arguments following this piece of text in the spec".

jskeet (commented):
A couple of comments that should be discussed, but I think we should be able to get them resolved and get the PR merged in our next meeting.

standard/expressions.md
BillWagner: I would go with "subsequent"

Nigel-Ecma: I would stick with "following" rather than "subsequent" as the latter can be used to cover following *in time* e.g. "a subsequent event"

Nigel-Ecma (approved):
It is short and clear, though I wonder why it was added [*] I think we can approve it.

Oh, and I think we should stick with "following"

[*] It doesn't seem to add much of benefit. It would have made more sense to me if `M(b: false, cValue)` was supported, i.e. accept the defaults before `b` and give values for `b` onwards.

jskeet:
> It doesn't seem to add much of benefit

I've definitely found it useful in real code. It's not that unusual to have a set of arguments that are already in positional order, but you want to clarify two arguments of the same type, without specifying the names of *all* the arguments from there onwards.

standard/expressions.md
BillWagner: edited suggestion to remove the trailing space as well.

standard/expressions.md
BillWagner: My opinion isn't strong. I'm fine with "following"

jskeet (approved):
Agreed at 2022-11-30 meeting.

standard/expressions.md
jskeet: Right, let's leave it as is for now.

BillWagner:
closing and reopening

PR #219, opened 2021-02-05
C# 7.x: in parameter mode

### LINKED ISSUE #264: Information about "in" function arguments missing
The `Argument Lists` section at https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#argument-lists references the `ref` and `out` parameter modifiers but is missing information on the newer `in` modifier which passes an item by reference but as read-only. The `in` modifier is described in the keywords section here: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/in-parameter-modifier

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e783650c-e063-e268-316e-564362cb1e37
* Version Independent ID: aa9f4996-cab4-0bed-6452-213342453785
* Content: [Expressions - C# language specification](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#argument-lists)
* Content Source: [spec/expressions.md](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md)
* Product: **dotnet-csharp**
* Technology: **csharp-spec**
* GitHub Login: @dotnet-bot
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 52212](https://dev.azure.com/msft-skilling/Content/_workitems/edit/52212)

### LINKED ISSUE #296: ref-safe-to-escape rule for in parameter
Looking at

https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md#method-invocation

> - For each `in` parameter of the method, if there is a corresponding expression that is an lvalue, its _ref-safe-to-escape_, otherwise the nearest enclosing scope 

This is only correct if the argument has an identity conversion to the type of the parameter. For any other type of conversion, I think it should be "nearest enclosing scope" because a copy will be made.

---
[Associated WorkItem - 52211](https://dev.azure.com/msft-skilling/Content/_workitems/edit/52211)

RexJaeschke:
Almost all the edits support the addition of the `in` parameter mode.

Support for the v7.2 feature, `ref` with `this` (as well as `in` with `this`) in an extension method was also added.

Fixes #264 
Fixes #296


standard/expressions.md
```diff
@@ -586,8 +588,18 @@ The corresponding parameters for function member arguments are established as fo
...
+     > *Example*: Given the following declarations and method calls:
+     > ```csharp
+     > public static void M1(in int p1) { … }
+     > int i = 10;
+     > M1(i);         // transformed to a temporary input argument
+     > M1(i + 5);     // transformed to a temporary input argument
+     > ```
+     > In each method call, an unnamed `int` is created, initialized with the argument’s value, and then passed as an input argument. See [§12.6.4.2](expressions.md#12642-applicable-function-member) and §better-argument-passing-mode-new-clause. *end example*  
+-   For an input, output, or reference parameter, the variable reference is evaluated and the resulting storage location becomes the storage location represented by the parameter in the function member invocation. If the variable reference given as an input, output, or reference is an array element of a *reference_type*, a run-time check is performed to ensure that the element type of the array is identical to the type of the parameter. If this check fails, a `System.ArrayTypeMismatchException` is thrown.
```
KalleOlaviNiemitalo: The run-time check should not be performed for an input parameter. See <https://github.com/dotnet/csharpstandard/issues/264#issuecomment-874738023>.

standard/basic-concepts.md
```diff
@@ -504,9 +503,9 @@ Signatures are the enabling mechanism for ***overloading*** of members in classe
...
 
-Although `out` and `ref` parameter modifiers are considered part of a signature, members declared in a single type cannot differ in signature solely by `ref` and `out`. A compile-time error occurs if two members are declared in the same type with signatures that would be the same if all parameters in both methods with `out` modifiers were changed to `ref` modifiers. For other purposes of signature matching (e.g., hiding or overriding), `ref` and `out` are considered part of the signature and do not match each other.
+Although `in`, `out`, and `ref` parameter modifiers are considered part of a signature, members declared in a single type cannot differ in signature solely by `in`, `out`, and `ref`. A compile-time error occurs if two members are declared in the same type with signatures that would be the same if all parameters in both methods with `out` modifiers were changed to `ref` modifiers. For other purposes of signature matching (e.g., hiding or overriding), `in`, `out`, and `ref` are considered part of the signature and do not match each other.
```
KalleOlaviNiemitalo: Needs to mention `in` in _"all parameters in both methods with `out` modifiers"_

standard/classes.md
```diff
@@ -1993,9 +1993,9 @@ parameter_array
...
+A *fixed_parameter* consists of an optional set of *attributes* ([§21](attributes.md#21-attributes)); an optional `in`, `out,` `ref`, or `this` modifier; a *type*; an *identifier*; and an optional *default_argument*. Each *fixed_parameter* declares a parameter of the given type with the given name. The `this` modifier designates the method as an extension method and is only allowed on the first parameter of a static method in a non-generic, non-nested static class. Extension methods are further described in [§14.6.10](classes.md#14610-extension-methods). A *fixed_parameter* with a *default_argument* is known as an ***optional parameter***, whereas a *fixed_parameter* without a *default_argument* is a ***required parameter***. A required parameter may not appear after an optional parameter in a *formal_parameter_list*.
```
KalleOlaviNiemitalo: Seems `this` should be separated from the "an optional […] modifier" list if `ref this` is allowed in C# 7.2.

standard/classes.md
```diff
@@ -2049,6 +2050,20 @@ When a formal parameter is a value parameter, the corresponding argument in a me
...
+
+When a formal parameter is an input parameter, the corresponding argument in a method invocation shall consist of either the keyword `in` followed by a *variable_reference* (§input-parameters-new-clause) of the same type as the formal parameter, or an *expression* for which an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from that argument expression to the type of the corresponding parameter. A variable shall be definitely assigned before it can be passed as an input parameter.
+
+It is a compile-time error to modify the value of an input parameter.
+
+Within a method, an input parameter is always considered definitely assigned.
+
+A method declared as an iterator ([§14.14](classes.md#1414-iterators)) may not have input parameters.
+
+In a method that takes reference parameters, it is possible for multiple names to represent the same storage location.
```
KalleOlaviNiemitalo: Should this say "input parameters"?

standard/variables.md
```diff
@@ -12,7 +12,7 @@ As described in the following subclauses, variables are either ***initially assi
 
 ### 9.2.1 General
 
-C# defines seven categories of variables: static variables, instance variables, array elements, value parameters, reference parameters, output parameters, and local variables. The subclauses that follow describe each of these categories.
+C# defines seven categories of variables: static variables, instance variables, array elements, value parameters, input parameters, reference parameters, output parameters, and local variables. The subclauses that follow describe each of these categories.
```
KalleOlaviNiemitalo: Eight categories

standard/variables.md
```diff
@@ -110,6 +108,19 @@ The following definite assignment rules apply to output parameters.
...
+A parameter declared with an `in` modifier is an ***input parameter***.
+
+An input parameter does not create a new storage location. Instead, an input parameter represents the same storage location as the variable given as the argument in the function member, anonymous function, or local function invocation. Thus, the value of an input parameter is always the same as the underlying variable.
+
+The following definite assignment rules apply to input parameters.
+
+- A variable shall be definitely assigned ([§9.4](variables.md#94-definite-assignment)) before it can be passed as an input parameter in a function member or delegate invocation.
+- Within a function member, anonymous function, or local function an input parameter is considered initially assigned.
+
+For a `struct` type, within an instance method or instance accessor ([§11.2.1](expressions.md#1121-general)) or instance constructor with a constructor initializer, the `this` keyword behaves exactly as an input parameter of the struct type ([§11.7.12](expressions.md#11712-this-access)).
```
KalleOlaviNiemitalo: What does this paragraph mean; can you show an example?

standard/basic-concepts.md
BillWagner: It may be more succinct to use the concepts of "pass by reference" and "pass by value". Methods can be overloaded if they differ using pass by reference and pass by value. However, it's a compiler error if the only difference is in which modifier is used for pass by reference.

I'll tag other locations where this change could be used.

standard/basic-concepts.md
```diff
@@ -533,9 +532,7 @@ The types `object` and `dynamic` are not distinguished when comparing signatures
...
->
-> *end example*
+> Note that any `in`, `out`, and `ref` parameter modifiers ([§14.6.2](classes.md#1462-method-parameters)) are part of a signature. Thus, `F(int)`, `F(in int)`, `F(out int)` , and `F(ref int)` are all unique signatures. However, `F(in int)`, `F(out int)` , and `F(ref int)` cannot be declared within the same interface because their signatures differ solely by `in`, `out`, and `ref`. Also, note that the return type and the `params` modifier are not part of a signature, so it is not possible to overload solely based on return type or on the inclusion or exclusion of the `params` modifier. As such, the declarations of the methods `F(int)` and `F(params string[])` identified above, result in a compile-time error. *end example*
```
BillWagner: Same comment as above.

standard/classes.md
```diff
@@ -741,8 +741,7 @@ A *class_declaration* creates a new declaration space ([§7.3](basic-concepts.md
 > *Note*: Since the fully qualified name of a type declaration encodes the number of type parameters, two distinct types may share the  same name as long as they have different number of type parameters. *end note*
 
 - The name of a constant, field, property, or event shall differ from the names of all other members declared in the same class.
-
-- The name of a method shall differ from the names of all other non-methods declared in the same class. In addition, the signature ([§7.6](basic-concepts.md#76-signatures-and-overloading)) of a method shall differ from the signatures of all other methods declared in the same class, and two methods declared in the same class shall not have signatures that differ solely by `ref` and `out`.
+- The name of a method shall differ from the names of all other non-methods declared in the same class. In addition, the signature ([§7.6](basic-concepts.md#76-signatures-and-overloading)) of a method shall differ from the signatures of all other methods declared in the same class, and two methods declared in the same class shall not have signatures that differ solely by `in`, `out`, and `ref`.
```
BillWagner, line 779: same as above.

standard/classes.md
```diff
@@ -1945,7 +1944,7 @@ If the *method_body* consists of a semicolon, the declaration shall not include
...
 
-The name of a method shall differ from the names of all other non-methods declared in the same class. In addition, the signature of a method shall differ from the signatures of all other methods declared in the same class, and two methods declared in the same class may not have signatures that differ solely by `ref` and `out`.
+The name of a method shall differ from the names of all other non-methods declared in the same class. In addition, the signature of a method shall differ from the signatures of all other methods declared in the same class, and two methods declared in the same class may not have signatures that differ solely by `in`, `out`, and `ref`.
```
BillWagner, line 1991: same as above.

standard/expressions.md
```diff
@@ -584,8 +586,24 @@ The corresponding parameters for function member arguments are established as fo
...
+     > int i = 10;
+     > M1(i);         // transformed to a temporary input argument
+     > M1(i + 5);     // transformed to a temporary input argument
+     > ```
+     >
+     > In each method call, an unnamed `int` is created, initialized with the argument’s value, and then passed as an input argument. See [§11.6.4.2](expressions.md#11642-applicable-function-member) and §better-argument-passing-mode-new-clause.
+     >
+     > *end example*
+
+- For an input, output, or reference parameter, the variable reference is evaluated and the resulting storage location becomes the storage location represented by the parameter in the function member invocation. If the variable reference given as an input, output, or reference is an array element of a *reference_type*, a run-time check is performed to ensure that the element type of the array is identical to the type of the parameter. If this check fails, a `System.ArrayTypeMismatchException` is thrown.
```
BillWagner: This could also be simpler by using "pass by reference" terminology.

standard/expressions.md
```diff
@@ -620,7 +638,7 @@ The expressions of an argument list are always evaluated in textual order.
 >
 > *end example*
 
-The array co-variance rules ([§16.6](arrays.md#166-array-covariance)) permit a value of an array type `A[]` to be a reference to an instance of an array type `B[]`, provided an implicit reference conversion exists from `B` to `A`. Because of these rules, when an array element of a *reference_type* is passed as a reference or output parameter, a run-time check is required to ensure that the actual element type of the array is *identical* to that of the parameter.
+The array co-variance rules ([§16.6](arrays.md#166-array-covariance)) permit a value of an array type `A[]` to be a reference to an instance of an array type `B[]`, provided an implicit reference conversion exists from `B` to `A`. Because of these rules, when an array element of a *reference_type* is passed as an input, output, or reference parameter, a run-time check is required to ensure that the actual element type of the array is *identical* to that of the parameter.
```
BillWagner: same point: "passed by reference" might be simpler.

standard/expressions.md
```diff
@@ -964,10 +985,34 @@ A function member is said to be an ***applicable function member*** with respect
...
-- Otherwise, the expanded form is applicable if for each argument in `A` the parameter-passing mode of the argument is identical to the parameter-passing mode of the corresponding parameter, and
-  - for a fixed value parameter or a value parameter created by the expansion, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from the argument expression to the type of the corresponding parameter, or
-  - for a `ref` or `out` parameter, the type of the argument expression is identical to the type of the corresponding parameter.
-
+- Otherwise, the expanded form is applicable if for each argument in `A`, one of the following is true:
+  - the parameter-passing mode of the argument is identical to the parameter-passing mode of the corresponding parameter, and
+    - for a fixed value parameter or a value parameter created by the expansion, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from the argument expression to the type of the corresponding parameter, or
+    - for `in`, `out`, or `ref` parameter, the type of the argument expression is identical to the type of the corresponding parameter.
```
BillWagner, line 1013: could be simpler to use pass by reference.

standard/expressions.md
```diff
@@ -995,6 +1040,19 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
 - Otherwise if one member is a non-lifted operator and the other is a lifted operator, the non-lifted one is better.
 - If neither function member was found to be better, and all parameters of `Mᵥ` have a corresponding argument whereas default arguments need to be substituted for at least one optional parameter in `Mₓ`, then `Mᵥ` is better than `Mₓ`. Otherwise, no function member is better.
 
+#### §better-argument-passing-mode-new-clause Better argument-passing mode
+
+It is permitted to have corresponding parameters in two overloaded methods differ only by parameter-passing mode provided one and only one of the two parameters has value-passing mode, as follows:
```
BillWagner: See? You're already introducing the concept of pass by value here.

standard/interfaces.md
```diff
@@ -232,7 +232,7 @@ All interface members implicitly have public access. It is a compile-time error
...
 
 - The name of a type parameter in the *type_parameter_list* of an interface declaration shall differ from the names of all other type parameters in the same *type_parameter_list* and shall differ from the names of all members of the interface.
-- The name of a method shall differ from the names of all properties and events declared in the same interface. In addition, the signature ([§7.6](basic-concepts.md#76-signatures-and-overloading)) of a method shall differ from the signatures of all other methods declared in the same interface, and two methods declared in the same interface may not have signatures that differ solely by `ref` and `out`.
+- The name of a method shall differ from the names of all properties and events declared in the same interface. In addition, the signature ([§7.6](basic-concepts.md#76-signatures-and-overloading)) of a method shall differ from the signatures of all other methods declared in the same interface, and two methods declared in the same interface may not have signatures that differ solely by `in`, `out`, and `ref`.
```
BillWagner, line 239: same comment

standard/variables.md
```diff
@@ -72,7 +70,7 @@ For the purpose of definite assignment checking, an array element is considered
 
 ### 9.2.5 Value parameters
 
-A parameter declared without a `ref` or `out` modifier is a ***value parameter***.
+A parameter declared without an `in`, `out`, or `ref` modifier is a ***value parameter***.
```
BillWagner, line 74: After this clause would be where to introduce the concept of pass-by-reference to encompass *reference parameters*, *output parameters* and *input parameters*.

gafter:
Please ensure https://github.com/dotnet/csharplang/issues/945 is reflected in this specification.

standard/basic-concepts.md
BillWagner: I tried different phrasing, and it seemed better to just add `in` as @KalleOlaviNiemitalo recommended.

The main reason is that there isn't great concise wording, because `in` parameters can be transformed and passed by value.

standard/variables.md
BillWagner: I removed this paragraph, but made the correct updates to §12.8.13

BillWagner:
> Please ensure [dotnet/csharplang#945](https://github.com/dotnet/csharplang/issues/945) is reflected in this specification.

I believe that rule is covered here: https://github.com/dotnet/csharpstandard/pull/219/files#diff-d9d678d0b42d3c668ebd64159dadfd6188361121feca8fb94fddf8816433cd14R1073-R1085

standard/classes.md
BillWagner: updated.

standard/expressions.md
BillWagner: Updated

BillWagner:
@jskeet This one is now ready for review and merge.

I've rebased, updated all links, and responded to all remaining feedback.

Note for @gafter I believe https://github.com/dotnet/csharplang/issues/945 is incorporated in this spec. The dates of comments and commits make it hard for me to be sure. Can you review?

standard/basic-concepts.md
```diff
@@ -522,9 +522,8 @@ The following accessibility constraints exist:
...
+- The signature of a method consists of the name of the method, the number of type parameters, and the type and parameter-passing mode of each of its formal parameters, considered in the order left to right. For these purposes, any type parameter of the method that occurs in the type of a formal parameter is identified not by its name, but by its ordinal position in the type parameter list of the method. The signature of a method specifically does not include the return type, parameter names, type parameter names, type parameter constraints, the `params` or `this` parameter modifiers, nor whether parameters are required or optional.
+- The signature of an instance constructor consists of the type and parameter-passing mode of each of its formal parameters, considered in the order left to right. The signature of an instance constructor specifically does not include the `params` modifier that may be specified for the right-most parameter.
```
jskeet, line 526: Nor whether the parameters are required or optional? (Probably no need to change that in this PR, but we may want to file an issue. Let's discuss in the meeting.)

standard/variables.md
```diff
@@ -109,7 +107,18 @@ The following definite-assignment rules apply to output parameters.
...
+A parameter declared with an `in` modifier is an ***input parameter***.
+
+An input parameter does not create a new storage location. Instead, an input parameter represents the same storage location as the variable given as the argument in the function member, anonymous function, or local function invocation. Thus, the value of an input parameter is always the same as the underlying variable.
+
+The following definite assignment rules apply to input parameters.
+
+- A variable shall be definitely assigned ([§9.4](variables.md#94-definite-assignment)) before it can be passed as an input parameter in a function member or delegate invocation.
+- Within a function member, anonymous function, or local function an input parameter is considered initially assigned.
+
+For a `struct` type, within an instance method or instance accessor ([§12.2.1](expressions.md#1221-general)) or instance constructor with a constructor initializer, the `this` keyword behaves exactly as an input parameter of the struct type ([§12.8.13](expressions.md#12813-this-access)).
```
jskeet: I thought that within a `struct` instance method, `this` behaves exactly as an *ref* parameter. (But not for async/iterator methods.)

In particular, you can assign to `this` within a struct method, but you can't assign to input parameters.

Perhaps just remove this paragraph entirely?

standard/conversions.md
```diff
@@ -920,9 +920,10 @@ An implicit conversion exists from a method group ([§12.2](expressions.md#122-e
...
 
 - A single method `M` is selected corresponding to a method invocation ([§12.8.9.2](expressions.md#12892-method-invocations)) of the form `E(A)`, with the following modifications:
-  - The argument list `A` is a list of expressions, each classified as a variable and with the type and modifier (`ref` or `out`) of the corresponding parameter in the *formal_parameter_list* of `D` — excepting parameters of type `dynamic`, where the corresponding expression has the type `object` instead of `dynamic`.
+  - The argument list `A` is a list of expressions, each classified as a variable and with the type and modifier (`in`, `out`, or `ref`) of the corresponding parameter in the *formal_parameter_list* of `D` --- excepting parameters of type `dynamic`, where the corresponding expression has the type `object` instead of `dynamic`.
```
jskeet: Revert the change from — to ---?

standard/conversions.md
```diff
@@ -920,9 +920,10 @@ An implicit conversion exists from a method group ([§12.2](expressions.md#122-e
...
   - The candidate methods considered are only those methods that are applicable in their normal form and do not omit any optional parameters ([§12.6.4.2](expressions.md#12642-applicable-function-member)). Thus, candidate methods are ignored if they are applicable only in their expanded form, or if one or more of their optional parameters do not have a corresponding parameter in `D`.
 - A conversion is considered to exist if the algorithm of [§12.8.9.2](expressions.md#12892-method-invocations) produces a single best method `M` having the same number of parameters as `D`.
+- Even if the conversion exists, a compile-time error occurs if the selected method `M` is not compatible ([§20.4](delegates.md#204-delegate-compatibility)) with the delegate type `D`.
```
jskeet: Was this left over from a previous PR? I thought we now didn't need this "conversion exists but the result isn't compatible" part. I may be confused though. (It doesn't obviously belong in *this* PR, mind you.)

standard/delegates.md
```diff
@@ -83,9 +83,10 @@ Except for instantiation, any operation that can be applied to a class or class
...
 
-- `D` and `M` have the same number of parameters, and each parameter in `D` has the same `ref` or `out` modifiers as the corresponding parameter in `M`.
-- For each value parameter (a parameter with no `ref` or `out` modifier), an identity conversion ([§10.2.2](conversions.md#1022-identity-conversion)) or implicit reference conversion ([§10.2.8](conversions.md#1028-implicit-reference-conversions)) exists from the parameter type in `D` to the corresponding parameter type in `M`.
-- For each `ref` or `out` parameter, the parameter type in `D` is the same as the parameter type in `M`.
+- `D` and `M` have the same number of parameters, and each parameter in `D` has the same `in`, `out`, or `ref` modifiers as the corresponding parameter in `M`.
+- For each value parameter, an identity conversion ([§10.2.2](conversions.md#1022-identity-conversion)) or implicit reference conversion ([§10.2.8](conversions.md#1028-implicit-reference-conversions)) exists from the parameter type in `D` to the corresponding parameter type in `M`.
+
+- For each `in`, `out`, or `ref` parameter, the parameter type in `D` is the same as the parameter type in `M`.
```
jskeet: I suspect identity conversions are okay here.

jskeet: (I have a bit of deja vu here... did we have this conversation before, and determine that "the same" actually means "the same or there's an identity conversion"?)

standard/expressions.md
```diff
@@ -594,8 +596,24 @@ The corresponding parameters for function member arguments are established as fo
...
+- For a value parameter, if the parameter’s passing mode is value
+  - the argument expression is evaluated and an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) to the corresponding parameter type is performed. The resulting value becomes the initial value of the value parameter in the function member invocation.
+  - otherwise, the parameter’s passing mode is input, in which case, an unnamed *variable_reference* is created with the same type as that of the corresponding parameter. The argument expression is evaluated and an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) to the corresponding parameter type is performed. The resulting value becomes the initial value of the unnamed *variable_reference* whose storage location becomes the storage location represented by the input parameter in the function member invocation.
```
jskeet: We might want an example where this is distinctly odd, because an argument "passed by value" really isn't...

```csharp
class Program
{
    static void Main()
    {
        int x = 0;
        M(x, ref x);
    }

    static void M(in int a, ref int b)
    {
        Console.WriteLine(a); // Prints 0
        b = 5;
        Console.WriteLine(a); // Prints 5
    }
}
```

jskeet: I'm not suggesting we do that in this PR though.

jskeet: Note that calling `M(x + 0, ref x)` prints 0 twice. I wonder how many tools would propose removing the "+ 0" as pointless...

standard/expressions.md
```diff
@@ -679,7 +697,7 @@ When a function member with a parameter array is invoked in its expanded form wi
 >
 > *end example*
 
-When arguments are omitted from a function member with corresponding optional parameters, the default arguments of the function member declaration are implicitly passed.
+When arguments are omitted from a function member with corresponding optional parameters, the default arguments of the function member declaration are implicitly passed.  (This can involve the creation of an unnamed *variable_reference*, as described above.)
```
jskeet: Nit: single space after period.

standard/expressions.md
```diff
@@ -594,8 +596,24 @@ The corresponding parameters for function member arguments are established as fo
...
+     > *Example*: Given the following declarations and method calls:
+     >
+     > ```csharp
+     > public static void M1(in int p1) { … }
+     > int i = 10;
+     > M1(i);         // transformed to a temporary input argument
+     > M1(i + 5);     // transformed to a temporary input argument
+     > ```
+     >
+     > In each method call, an unnamed `int` is created, initialized with the argument’s value, and then passed as an input argument. See [§12.6.4.2](expressions.md#12642-applicable-function-member) and §better-argument-passing-mode-new-clause.
```
```suggestion
     > In each method call, an unnamed `int` variable is created, initialized with the argument’s value, and then passed as an input argument. See [§12.6.4.2](expressions.md#12642-applicable-function-member) and §better-argument-passing-mode-new-clause.
```

standard/expressions.md
```diff
@@ -737,7 +755,10 @@ For each of the method arguments `Eᵢ`:
...
-- Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is a reference ([§15.6.2.3](classes.md#15623-reference-parameters)) or output ([§15.6.2.4](classes.md#15624-output-parameters)) parameter then an *exact inference* ([§12.6.3.9](expressions.md#12639-exact-inferences)) is made *from* `U` *to* `Tᵢ`.
+
+- If `Eᵢ` is an anonymous function, an *explicit parameter type inference* ([§12.6.3.8](expressions.md#12638-explicit-parameter-type-inferences)) is made *from* `Eᵢ` *to* `Tᵢ`
+- Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is a value parameter ([§15.6.2.2](classes.md#15622-value-parameters)) then a *lower-bound inference* ([§12.6.3.10](expressions.md#126310-lower-bound-inferences)) is made *from* `U` *to* `Tᵢ`.
+- Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is an input (§method-input-parameters-new-clause), reference ([§15.6.2.3](classes.md#15623-reference-parameters)), or output ([§15.6.2.4](classes.md#15624-output-parameters)) parameter then an *exact inference* ([§12.6.3.9](expressions.md#12639-exact-inferences)) is made *from* `U` *to* `Tᵢ`.
```
```suggestion
- Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is an input parameter (§method-input-parameters-new-clause), reference parameter ([§15.6.2.3](classes.md#15623-reference-parameters)), or output parameter ([§15.6.2.4](classes.md#15624-output-parameters)) then an *exact inference* ([§12.6.3.9](expressions.md#12639-exact-inferences)) is made *from* `U` *to* `Tᵢ`.
```

(Feel free to reject - this just feels clearer to me.)

standard/expressions.md
```diff
@@ -1025,6 +1070,19 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
 - Otherwise if one member is a non-lifted operator and the other is a lifted operator, the non-lifted one is better.
 - If neither function member was found to be better, and all parameters of `Mᵥ` have a corresponding argument whereas default arguments need to be substituted for at least one optional parameter in `Mₓ`, then `Mᵥ` is better than `Mₓ`. Otherwise, no function member is better.
 
+#### §better-argument-passing-mode-new-clause Better argument-passing mode
+
+It is permitted to have corresponding parameters in two overloaded methods differ only by parameter-passing mode provided one and only one of the two parameters has value-passing mode, as follows:
```
jskeet: I don't think "only one" is required here, given that we're saying "they differ only in parameter-passing mode". They can't differ only in parameter passing mode and both have a value-passing mode :)

standard/expressions.md
```diff
@@ -2064,9 +2122,9 @@ A *this_access* is permitted only in the *block* of an instance constructor, an
...
 - When `this` is used in a *primary_expression* within an instance constructor of a struct, it is classified as a variable. The type of the variable is the instance type ([§15.3.2](classes.md#1532-the-instance-type)) of the struct within which the usage occurs, and the variable represents the struct being constructed.
   - If the constructor declaration has no constructor initializer, the `this` variable behaves exactly the same as an `out` parameter of the struct type. In particular, this means that the variable shall be definitely assigned in every execution path of the instance constructor.
-  - Otherwise, the `this` variable behaves exactly the same as a `ref` parameter of the struct type. In particular, this means that the variable is considered initially assigned.
+  - Otherwise, the `this` variable behaves exactly the same as a `in` parameter of the struct type. In particular, this means that the variable is considered initially assigned, and can't be reassigned.
```
jskeet: Unless this changes in C# 8+, this doesn't appear to be the case:

```csharp
struct Test
{
    Test(int t) : this()
    {
        // Constructor with a constructor initializer, but still able to reassign this afterwards
        this = new Test("h");
    }

    Test(string x)
    {
    }
}
```

standard/expressions.md
```diff
@@ -2064,9 +2122,9 @@ A *this_access* is permitted only in the *block* of an instance constructor, an
...
 - When `this` is used in a *primary_expression* within an instance method or instance accessor of a struct, it is classified as a variable. The type of the variable is the instance type ([§15.3.2](classes.md#1532-the-instance-type)) of the struct within which the usage occurs.
-  - If the method or accessor is not an iterator ([§15.14](classes.md#1514-iterators)) or async function ([§15.15](classes.md#1515-async-functions)), the `this` variable represents the struct for which the method or accessor was invoked, and behaves exactly the same as a `ref` parameter of the struct type.
+  - If the method or accessor is not an iterator ([§15.14](classes.md#1514-iterators)) or async function ([§15.15](classes.md#1515-async-functions)), the `this` variable represents the struct for which the method or accessor was invoked, and behaves exactly the same as a `in` parameter of the struct type.
```
jskeet: Again, that doesn't appear to be the case.

```csharp
class Program
{
    static void Main()
    {
        Test t = new Test(5);
        t.ChangeMe(10);
        Console.WriteLine(t.value); // Prints 10
    }
}

struct Test
{
    public int value;

    public Test(int value)
    {
        this.value = value;
    }

    public void ChangeMe(int value)
    {
        this = new Test(value);
    }
}
```

standard/variables.md
jskeet: (This is somewhat discussed above.)

standard/conversions.md
BillWagner: I think that's a merge conflict issue. All of these were quite old and the merging gave me headaches. I'll remove.

standard/delegates.md
BillWagner: We did, in each of the PRs for `in`, `out`, and `ref` struct tpes.

standard/expressions.md
BillWagner: I think you're right. Let's discuss, but I think this change should be reverted.

standard/expressions.md
BillWagner: Also agreed.

standard/variables.md
BillWagner: All of these (except the constructor note) should only apply to `readonly` structs. (The `this` instance can be reassigned in a ctor on a readonly struct)

standard/expressions.md
Nigel-Ecma: The language here is wrong, we don't "create" (instances of) grammar productions. The semantics is that a temporary variable/storage location is created and it is that which is assigned to and then passed. Meeting in 10min so no wordsmithing at this time!

Nigel-Ecma: @jskeet  – "nice" example of aliasing, that should certainly go in at least a Note and maybe as an Example

standard/basic-concepts.md
jskeet: Action from 2023-04-19: @BillWagner to assign an issue to @jskeet; no change required in this PR.

standard/basic-concepts.md
```diff
@@ -537,9 +536,9 @@ Signatures are the enabling mechanism for ***overloading*** of members in classe
...
 
-Although `out` and `ref` parameter modifiers are considered part of a signature, members declared in a single type cannot differ in signature solely by `ref` and `out`. A compile-time error occurs if two members are declared in the same type with signatures that would be the same if all parameters in both methods with `out` modifiers were changed to `ref` modifiers. For other purposes of signature matching (e.g., hiding or overriding), `ref` and `out` are considered part of the signature and do not match each other.
+Although `in`, `out`, and `ref` parameter modifiers are considered part of a signature, members declared in a single type cannot differ in signature solely by `in`, `out`, and `ref`. A compile-time error occurs if two members are declared in the same type with signatures that would be the same if all parameters in both methods with `out` modifiers were changed to `ref` or `in` modifiers. For other purposes of signature matching (e.g., hiding or overriding), `in`, `out`, and `ref` are considered part of the signature and do not match each other.
```
jskeet:
```suggestion
Although `in`, `out`, and `ref` parameter modifiers are considered part of a signature, members declared in a single type cannot differ in signature solely by `in`, `out`, and `ref`. A compile-time error occurs if two members are declared in the same type with signatures that would be the same if all parameters in both methods with `out` or `in` modifiers were changed to `ref` modifiers. For other purposes of signature matching (e.g., hiding or overriding), `in`, `out`, and `ref` are considered part of the signature and do not match each other.
```

jskeet: Action from 2023-04-19: accept this suggestion :)

standard/expressions.md
jskeet: Action from 2023-04-19: @BillWagner to log the need for a note about odd aliasing later, and assign to @jskeet.

See #782

gafter, line 601:
```suggestion
  - otherwise, the parameter’s passing mode is input, in which case, an unnamed *variable_reference* is created with the same type as that of the corresponding parameter. The argument expression is evaluated and an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) to the corresponding parameter type is performed. The resulting value becomes the initial value of the unnamed *variable_reference*. That storage location is represented by the input parameter in the function member invocation.
```

jskeet: Addressing Nigel's comment of "The language here is wrong, we don't "create" (instances of) grammar productions.":

```suggestion
  - otherwise, the parameter’s passing mode is input, in which case, a storage location is created with the same type as that of the corresponding parameter. The argument expression is evaluated and an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) to the corresponding parameter type is performed. The resulting value is stored within that storage location. That storage location is represented by the input parameter in the function member invocation.
```

Action from 2023-04-19: Accept this suggestion, we think...

standard/expressions.md
```diff
@@ -679,7 +697,7 @@ When a function member with a parameter array is invoked in its expanded form wi
 >
 > *end example*
 
-When arguments are omitted from a function member with corresponding optional parameters, the default arguments of the function member declaration are implicitly passed.
+When arguments are omitted from a function member with corresponding optional parameters, the default arguments of the function member declaration are implicitly passed. (This can involve the creation of an unnamed *variable_reference*, as described above.)
```
```suggestion
When arguments are omitted from a function member with corresponding optional parameters, the default arguments of the function member declaration are implicitly passed. (This can involve the creation of a storage location, as described above.)
```

jskeet: Action from 2023-04-19: Accept this suggestion to be consistent with the previous one.

standard/expressions.md
```diff
@@ -737,7 +755,10 @@ For each of the method arguments `Eᵢ`:
 
 - If `Eᵢ` is an anonymous function, an *explicit parameter type inference* ([§12.6.3.8](expressions.md#12638-explicit-parameter-type-inferences)) is made *from* `Eᵢ` *to* `Tᵢ`
 - Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is a value parameter ([§15.6.2.2](classes.md#15622-value-parameters)) then a *lower-bound inference* ([§12.6.3.10](expressions.md#126310-lower-bound-inferences)) is made *from* `U` *to* `Tᵢ`.
-- Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is a reference ([§15.6.2.3](classes.md#15623-reference-parameters)) or output ([§15.6.2.4](classes.md#15624-output-parameters)) parameter then an *exact inference* ([§12.6.3.9](expressions.md#12639-exact-inferences)) is made *from* `U` *to* `Tᵢ`.
+
```
jskeet: Action from 2023-04-19: delete this blank line and subsequent to lines, which are duplicates of the ones above.

standard/expressions.md
```diff
@@ -737,7 +755,10 @@ For each of the method arguments `Eᵢ`:
...
+- If `Eᵢ` is an anonymous function, an *explicit parameter type inference* ([§12.6.3.8](expressions.md#12638-explicit-parameter-type-inferences)) is made *from* `Eᵢ` *to* `Tᵢ`
+- Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is a value parameter ([§15.6.2.2](classes.md#15622-value-parameters)) then a *lower-bound inference* ([§12.6.3.10](expressions.md#126310-lower-bound-inferences)) is made *from* `U` *to* `Tᵢ`.
+- Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is an input parameter (§method-input-parameters-new-clause), reference parameter ([§15.6.2.3](classes.md#15623-reference-parameters)), or output parameter ([§15.6.2.4](classes.md#15624-output-parameters)) then an *exact inference* ([§12.6.3.9](expressions.md#12639-exact-inferences)) is made *from* `U` *to* `Tᵢ`.
```
jskeet: Action from 2023-04-19: This isn't right. If `Eᵢ` has the `in` modifier, and `xᵢ` is an input parameter then there's an exact inference. If `xᵢ` is an input parameter but `Eᵢ` does *not* have the `in` modifier, there's just a lower bound inference from `U` to `Tᵢ`. (This isn't in the form a suggestion yet, just trying to express what's wrong.)

standard/expressions.md
```diff
@@ -984,11 +1005,35 @@ A function member is said to be an ***applicable function member*** with respect
...
+> public static void M2(in int p1) { … }
+> 
+> M2(in i);   // M2(in int) is applicable
+> M2(i);      // M2(in int) is applicable
+> M2(i + 5);  // M2(in int) is applicable
+> ```
+>
+> *end example*
+  
+#### 11.6.4.3 Better function member
```
BillWagner: This got broken in the rebase:

```suggestion
#### 12.6.4.3 Better function member
```

standard/delegates.md
```diff
@@ -83,9 +83,10 @@ Except for instantiation, any operation that can be applied to a class or class
 
 A method or delegate type `M` is ***compatible*** with a delegate type `D` if all of the following are true:
 
-- `D` and `M` have the same number of parameters, and each parameter in `D` has the same `ref` or `out` modifiers as the corresponding parameter in `M`.
-- For each value parameter (a parameter with no `ref` or `out` modifier), an identity conversion ([§10.2.2](conversions.md#1022-identity-conversion)) or implicit reference conversion ([§10.2.8](conversions.md#1028-implicit-reference-conversions)) exists from the parameter type in `D` to the corresponding parameter type in `M`.
-- For each `ref` or `out` parameter, the parameter type in `D` is the same as the parameter type in `M`.
+- `D` and `M` have the same number of parameters, and each parameter in `D` has the same `in`, `out`, or `ref` modifiers as the corresponding parameter in `M`.
+- For each value parameter, an identity conversion ([§10.2.2](conversions.md#1022-identity-conversion)) or implicit reference conversion ([§10.2.8](conversions.md#1028-implicit-reference-conversions)) exists from the parameter type in `D` to the corresponding parameter type in `M`.
+
```
gafter: Delete this line I think

standard/expressions.md
jskeet: Action 2023-04-19: this needs to be for members of readonly structs, other than the constructor.
@BillWagner to fix.

standard/expressions.md
jskeet: Action 2023-04-19: As above, this is a struct vs readonly struct difference.

standard/variables.md
jskeet: Action 2023-04-19: Another one for readonly structs.

standard/delegates.md
jskeet: Action 2023-04-19: Agreed.

BillWagner: delete this line:

```suggestion
```

standard/expressions.md
jskeet: @BillWagner to wordsmith; worth getting this one checked.

jskeet (commented):
Result 2023-04-19: nearly there, but not quite ready for "approve, Bill to fix then merge" - the type inference part probably needs review.

standard/variables.md
BillWagner: I deleted this instance. Under reference parameters, (9.2.6), this requirement is stated correctly (`this` behaves as a reference parameter)

standard/expressions.md
BillWagner: Wordsmithed into 2 bullets. Leaving resolved to review updates.

KalleOlaviNiemitalo:
If I understand correctly, this PR would require all three calls to choose `Fun(in int)`, but that's not [what happens at sharplab.io](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUHjAAgGFCBvPQqw1TANhoBZCAxaACgEspDvheAlOUrVRtAJzsARGyhcefAVIEBuEVQC+edTXpNWHADYB7KAHNChoRVyixmSTKOmzS1Tq35b1Wg1TMAWQBDbnZrHVFZdkxVQgB6OMIAYyDDQwBnAzluXihgAW1vOyz2djgATygggFtOJIEYlXjElLTMqJNzAq9iqijOEEIyypq6htiEwgAlaGBOaoBTACFuOAWwAFEADySFgAc50w88DSA===):

```csharp
using System;

class C {
    static void Fun(in int i) {
        Console.WriteLine("Fun(in int)");
    }

    static void Fun(long l) {
        Console.WriteLine("Fun(long)");
    }

    static void Main() {
        Fun(1); // calls Fun(in int)

        Fun((dynamic)1); // calls Fun(long)

        Fun(i: (dynamic)1); // RuntimeBinderException
    }
}
```

KalleOlaviNiemitalo:
[Another variant](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUHjAAgGFCBvPQqw1TANhoBZCAxaACgEspDvheYvKPwBWASnKVq02gE52AIjZQuPPoL5iFYgNxSqAXzz6a9Jqw5gApgDMh/ToIA2AeygBzQuMm5pMzPJKlrb2zm7uWromRvi+1LQMqMwAsgCG3OwSFHF+9ryEALyEAAx6sblUyuzWdo4luoQA9I2EAMapTk4AzhYqNaGErh5iJtJV/XXscACeUKkAtpytYsUNzW0d3b2qA5qj1OtV3AKEU7MLSytrLVZgYC5gJADKABwAzHSM0XgGQA==).  Here, I think both calls should go to Fun(ref int, long).

```csharp
using System;

class C {
    static void Fun(in int i, int j) {
        Console.WriteLine("Fun(in int, int)");
    }

    static void Fun(ref int i, long j) {
        Console.WriteLine("Fun(ref int, long)");
    }

    static void Main() {
        int i = 0;

        Fun(ref i, 0); // calls Fun(ref int, long)
        Fun(ref i, (dynamic)0); // calls Fun(in int, int)
        // Fun(in i, (dynamic)0); // error CS8364
    }
}
```

It seems to me that this mismatch should be considered a bug in the runtime binder implementation, rather than something to be changed in this PR.  Whether the bug will ever be fixed is a different matter.

standard/expressions.md
```diff
@@ -753,12 +753,10 @@ Type inference takes place in phases. Each phase will try to infer type argument
...
-
 - If `Eᵢ` is an anonymous function, an *explicit parameter type inference* ([§12.6.3.8](expressions.md#12638-explicit-parameter-type-inferences)) is made *from* `Eᵢ` *to* `Tᵢ`
 - Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is a value parameter ([§15.6.2.2](classes.md#15622-value-parameters)) then a *lower-bound inference* ([§12.6.3.10](expressions.md#126310-lower-bound-inferences)) is made *from* `U` *to* `Tᵢ`.
 - Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is an input parameter (§method-input-parameters-new-clause), reference parameter ([§15.6.2.3](classes.md#15623-reference-parameters)), or output parameter ([§15.6.2.4](classes.md#15624-output-parameters)) then an *exact inference* ([§12.6.3.9](expressions.md#12639-exact-inferences)) is made *from* `U` *to* `Tᵢ`.
+- Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is an input parameter (§method-input-parameters-new-clause) then a *lower bound inference* ([§12.6.3.10](expressions.md#126310-lower-bound-inferences)) is made *from* `U` *to* `Tᵢ`. A new storage location is created following the rules of §12.6.2.2 on corresponding parameters.
```
jskeet, line 761: Was the intention to remove "an input parameter" from the bullet above? Otherwise, I can't see how we end up in this "Otherwise" as it's covered by the preceding one.

standard/expressions.md
```diff
@@ -1033,8 +1031,6 @@ For a function member that includes a parameter array, if the function member is
 >
 > *end example*
   
-#### 12.6.4.3 Better function member
```
jskeet: Gosh, do we never refer to "better function member" now?

jskeet: Hmm - we do, on line 994. Do we really want to remove this subhead?

jskeet (commented):
I think I may be a bit confused by this in general, but I hope the comments are helpful.

standard/expressions.md
BillWagner: Yes. Thanks. Great catch.

standard/expressions.md
BillWagner: In a previous rebase this subhead moved below the following bullet list. The following bullets should be part of "The second phase", which starts on line 764.

BillWagner:
> If I understand correctly, this PR would require all three calls to choose `Fun(in int)`, but that's not [what happens at sharplab.io](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUHjAAgGFCBvPQqw1TANhoBZCAxaACgEspDvheAlOUrVRtAJzsARGyhcefAVIEBuEVQC+edTXpNWHADYB7KAHNChoRVyixmSTKOmzS1Tq35b1Wg1TMAWQBDbnZrHVFZdkxVQgB6OMIAYyDDQwBnAzluXihgAW1vOyz2djgATygggFtOJIEYlXjElLTMqJNzAq9iqijOEEIyypq6htiEwgAlaGBOaoBTACFuOAWwAFEADySFgAc50w88DSA===):
> 
> ```cs
> using System;
> 
> class C {
>     static void Fun(in int i) {
>         Console.WriteLine("Fun(in int)");
>     }
> 
>     static void Fun(long l) {
>         Console.WriteLine("Fun(long)");
>     }
> 
>     static void Main() {
>         Fun(1); // calls Fun(in int)
> 
>         Fun((dynamic)1); // calls Fun(long)
> 
>         Fun(i: (dynamic)1); // RuntimeBinderException
>     }
> }
> ```

Great catch!  I added that when an argument is `dynamic`, there must be an identity conversion to the parameter type if passed as an `in` parameter.

BillWagner:
> It seems to me that this mismatch should be considered a bug in the runtime binder implementation, rather than something to be changed in this PR. Whether the bug will ever be fixed is a different matter.

See /dotnet/runtime#26614. Yes, it's a runtime bug.

BillWagner:
> [Another variant](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYCMBYAUHjAAgGFCBvPQqw1TANhoBZCAxaACgEspDvheYvKPwBWASnKVq02gE52AIjZQuPPoL5iFYgNxSqAXzz6a9Jqw5gApgDMh/ToIA2AeygBzQuMm5pMzPJKlrb2zm7uWromRvi+1LQMqMwAsgCG3OwSFHF+9ryEALyEAAx6sblUyuzWdo4luoQA9I2EAMapTk4AzhYqNaGErh5iJtJV/XXscACeUKkAtpytYsUNzW0d3b2qA5qj1OtV3AKEU7MLSytrLVZgYC5gJADKABwAzHSM0XgGQA==). Here, I think both calls should go to Fun(ref int, long).
> 
> ```cs
> using System;
> 
> class C {
>     static void Fun(in int i, int j) {
>         Console.WriteLine("Fun(in int, int)");
>     }
> 
>     static void Fun(ref int i, long j) {
>         Console.WriteLine("Fun(ref int, long)");
>     }
> 
>     static void Main() {
>         int i = 0;
> 
>         Fun(ref i, 0); // calls Fun(ref int, long)
>         Fun(ref i, (dynamic)0); // calls Fun(in int, int)
>         // Fun(in i, (dynamic)0); // error CS8364
>     }
> }
> ```
> 
> It seems to me that this mismatch should be considered a bug in the runtime binder implementation, rather than something to be changed in this PR. Whether the bug will ever be fixed is a different matter.

This one took me a long time to work through, but I think the behavior is correct and consistent with this PR.

`Fun(ref i, 0)` calls `Fun(ref int, long)` because the constant `0` can be converted to either `int` or `long` (both conversions are equally good. Therefore, the `ref` modifier takes precedence.

`Fun(ref i, (dynamic)0)` calls `Fun(in int, int)` because the implicit conversion of a dynamic expression to `int` is a better conversion than the conversion to `long` ([§12.6.4.4](https://learn.microsoft.com/dotnet/csharp/language-reference/language-specification/expressions#12644-better-conversion-from-expression). Therefore, `Fun(in int i, int j)` is the better overload.

How did you come to a different conclusion?

KalleOlaviNiemitalo:
> `Fun(ref i, (dynamic)0)` calls `Fun(in int, int)` because the implicit conversion of a dynamic expression to `int` is a better conversion than the conversion to `long` ([§12.6.4.4](https://learn.microsoft.com/dotnet/csharp/language-reference/language-specification/expressions#12644-better-conversion-from-expression). Therefore, `Fun(in int i, int j)` is the better overload.

The runtime binder should not choose `Fun(in int i, int j)`  for the `Fun(ref i, (dynamic)0)` invocation because the `ref i` argument is not in the format required by the `in int i` input parameter, according to this: <https://github.com/dotnet/csharpstandard/blob/41e3c1229cfbc34e69bd3ffb5f39f4cbd12341ca/standard/classes.md?plain=1#L2098>

BillWagner:
> > `Fun(ref i, (dynamic)0)` calls `Fun(in int, int)` because the implicit conversion of a dynamic expression to `int` is a better conversion than the conversion to `long` ([§12.6.4.4](https://learn.microsoft.com/dotnet/csharp/language-reference/language-specification/expressions#12644-better-conversion-from-expression). Therefore, `Fun(in int i, int j)` is the better overload.
> 
> The runtime binder should not choose `Fun(in int i, int j)` for the `Fun(ref i, (dynamic)0)` invocation because the `ref i` argument is not in the format required by the `in int i` input parameter, according to this:
> 
> https://github.com/dotnet/csharpstandard/blob/41e3c1229cfbc34e69bd3ffb5f39f4cbd12341ca/standard/classes.md?plain=1#L2098

This is a roslyn bug.  Logged as dotnet/roslyn#12649

KalleOlaviNiemitalo:
> Logged as dotnet/roslyn#12649

That's about the `lock` statement and type parameters.  I don't see how it relates to `dynamic` or input parameters.

BillWagner:
> > Logged as [dotnet/roslyn#12649](https://github.com/dotnet/roslyn/issues/12649)
> 
> That's about the `lock` statement and type parameters. I don't see how it relates to `dynamic` or input parameters.

Thanks. Here's the correct link: https://github.com/dotnet/runtime/issues/86118

standard/variables.md
```diff
@@ -109,7 +107,18 @@ The following definite-assignment rules apply to output parameters.
 - Within a function member or anonymous function, an output parameter is considered initially unassigned.
 - Every output parameter of a function member, anonymous function, or local function shall be definitely assigned ([§9.4](variables.md#94-definite-assignment)) before the function member, anonymous function, or local function returns normally.
 
-Within an instance constructor of a struct type, the `this` keyword behaves exactly as an output or reference parameter of the struct type, depending on whether the constructor declaration includes a constructor initializer ([§12.8.13](expressions.md#12813-this-access)).
+### §input-parameters-new-clause Input parameters
+
+A parameter declared with an `in` modifier is an ***input parameter***.
+
+An input parameter does not create a new storage location. Instead, an input parameter represents the same storage location as the variable given as the argument in the function member, anonymous function, or local function invocation. Thus, the value of an input parameter is always the same as the underlying variable.
```
Nigel-Ecma: With the introduction of reference variables (#742) where *referent* is defined and potentially changeable (by `= ref`) we need to review the language used (pass-by-)reference parameters and check for consistency/compatibility.

standard/classes.md
```diff
@@ -2033,9 +2033,9 @@ parameter_array
...
+A *fixed_parameter* consists of an optional set of *attributes* ([§22](attributes.md#22-attributes)); an optional `in`, `out,` `ref` modifier; a *type*; an *identifier*; and an optional *default_argument*. Each *fixed_parameter* declares a parameter of the given type with the given name. The `this` modifier designates the method as an extension method and is only allowed on the first parameter of a static method in a non-generic, non-nested static class. If the parameter is a `struct` type or a type parameter constrained to a `struct`, the `this` modifier may be combined with either the `ref` or `in` modifier, but not the `out` modifier. Extension methods are further described in [§15.6.10](classes.md#15610-extension-methods). A *fixed_parameter* with a *default_argument* is known as an ***optional parameter***, whereas a *fixed_parameter* without a *default_argument* is a ***required parameter***. A required parameter may not appear after an optional parameter in a *formal_parameter_list*.
```
gafter:
```suggestion
A *fixed_parameter* consists of an optional set of *attributes* ([§22](attributes.md#22-attributes)); an optional `in`, `out`, `ref` modifier; a *type*; an *identifier*; and an optional *default_argument*. Each *fixed_parameter* declares a parameter of the given type with the given name. The `this` modifier designates the method as an extension method and is only allowed on the first parameter of a static method in a non-generic, non-nested static class. If the parameter is a `struct` type or a type parameter constrained to a `struct`, the `this` modifier may be combined with either the `ref` or `in` modifier, but not the `out` modifier. Extension methods are further described in [§15.6.10](classes.md#15610-extension-methods). A *fixed_parameter* with a *default_argument* is known as an ***optional parameter***, whereas a *fixed_parameter* without a *default_argument* is a ***required parameter***. A required parameter may not appear after an optional parameter in a *formal_parameter_list*.
```

standard/variables.md
BillWagner: Use the language of "referrant" rather than the language on "doesn't allocate storage" or "uses the same storage"

Issue #37, opened 2020-11-20
ANTLR: Deciding on how far to go with this


RexJaeschke:
On 2020-10-03, I circulated the following paper: [ANTLR and the C# Spec.docx](https://github.com/ECMA-TC49-TG2/csharpstandard/files/5575949/ANTLR.and.the.C.Spec.docx). In that, I discussed using the ANTLR grammar notation and the non-trivial amount of work I thought it would take to get the grammar to validate using the ANTLR toolkit.

Neal has reviewed that paper and his feedback follows in a comment below, with a proposal on how we might go forward. Once we discuss (and probably agree with) his suggestion, I'll update this comment by adding the remaining questions and to keep track of our decisions on those questions .

Here are the ANTLR-related questions and their status:

Proposal # | Status | Description
------------ | ------- | -------------
1 | Yes it should | Decide whether our grammar should be complete enough to validate (3 Rex's paper). **Note that *validation* here means "get through the ANTLR tool without error," which doesn't necessarily mean that any parser generated will be complete/correct.** This is especially true for semantic predicates.
2 | Resolved, PRs #208, #209  | Change case of lexer rule names (5.1 Rex's paper)
3 | Resolved, PR #261 | Expanding lexer rule alternatives of the form <...> (3 Rex's paper)
4 | Resolved, PR #225 | Limited use of token names (5.3 Rex's paper)
5 | In-progress | Resolve Mutual Left-Recursion issue (15 Rex's paper)
6 | Resolved, PR #233 | How to handle the unsafe extensions
7 | Resolved | Marking lexer helper rules with `fragment` 
8 | Resolved | Ultimately, revise 7.2, "Grammars" in lexical-structure.md to say what we do/don't do w.r.t full grammar validation

**When you add a comment to this issue, please limit it to a single Proposal # and say which Proposal # that is.**

Any decisions we make that require changes to the grammar notation will not only have to be made to the v5 grammar we're starting with in v6, but also possibly to the v6 (and v7) PRs in waiting.

Nigel-Ecma:
I've taken a brief look at ANTLR and I beleive the rules Rex quotes above can be written:

```
Not_Slash_Or_Asterisk
    : ~[*/] // Any Unicode character except / or *
    ;

Letter_Character
    : [\p{Lu}\p{Ll}\p{Lt}\p{Lm}\p{Lo}\p{Nl}] // A Unicode Character of classes Lu, Ll, Lt, Lm, Lo, or Nl
[...]

PR #225, opened 2021-02-17
Resolve ANTLR warnings by defining named lexer tokens

RexJaeschke:
This PR addresses Issue #37, Proposal 4 (https://github.com/ECMA-TC49-TG2/csharpstandard/issues/37#issuecomment-767636955). Specifically, it

* Adds definitions for 6 named lexer tokens, and then refers to them in various lexer rule alternatives
* Adds some ANTLR-related commentary
* Defines *New_Line* as a strict superset of *New_Line_Character*
* Fixes an error in *Single_Regular_String_Literal_Character*
* Changes *identfier* => *Identifier* in expressions.md
* Changes *Right_shift* => *Right_Shift* in classes.md

For now this is a Draft PR, as I want to explore the idea of replacing the use of certain terminals in the syntactic grammar with the 6 new named lexer tokens (for example, `'default'` in a switch statement becomes `DEFAULT`).   

I explored that and decided to *not* go down that path, as explained in https://github.com/ECMA-TC49-TG2/csharpstandard/issues/37#issuecomment-780859388.

### DIFF
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -20,7 +20,7 @@ Conforming implementations shall accept Unicode compilation units encoded with t
 
 ### 7.2.1 General
 
-This specification presents the syntax of the C# programming language using two grammars. The ***lexical grammar*** ([§7.2.2](lexical-structure.md#722-grammar-notation)) defines how Unicode characters are combined to form line terminators, white space, comments, tokens, and pre-processing directives. The ***syntactic grammar*** ([§7.2.4](lexical-structure.md#724-syntactic-grammar)) defines how the tokens resulting from the lexical grammar are combined to form C# programs.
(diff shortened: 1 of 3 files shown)


standard/lexical-structure.md
```diff
@@ -84,6 +88,19 @@ then the *type_argument_list* is retained as part of the *simple_name*, *member_
...
+```ANTLR
+DEFAULT  : 'default' ;
+NULL     : 'null' ;
+TRUE     : 'true' ;
+FALSE    : 'false' ;
+ASTERISK : '*' ;
+SLASH    : '/' ;
+```
+
+Although these are lexer rules, these names are spelled in all-uppercase letters to distinguish them from ordinary lexer rule names. These rules are also written on a single line.
```
Nigel-Ecma: I think "These rules are also written on a single line." is redundant and better omitted.

Nigel-Ecma (approved):
I'm somewhat ambivalent on the calling out of a few tokens as lexical rules and the use of uppercase for them, I'm not sure it enhances the grammar or that it harms it either; however the changes seem correct and I'm therefore happy to approve. I've suggested one line is redundant and can therefore be removed.

jskeet (approved):
I agree with Nigel's comment about the extraneous line.

I believe there was a good reason for the all-caps lexer tokens, although I can't now remember what it was :)

jskeet:
Note for meeting: I don't *think* there have been changes since the last meeting, so we may wish to skip this.

Issue #67, opened 2020-12-11
v7.x Target-typed "default" literal


RexJaeschke:
@gafter, I'm assigning this to you in the hope you can give me early feedback on my initial thoughts on how to write a proposal for this feature, so I can get started. Thanks in advance! Of course, input from other members is also welcome.

Proposal https://github.com/dotnet/csharplang/blob/standard-proposals/proposals/csharp-7.1/target-typed-default.md allows the use of `default` on its own (rather than `default(T)`) in contexts where the type can be inferred.

The "Detailed design" part of the proposal suggests introducing a new kind of expression, the *default literal*. At a first/second read, that suggests a separate section in expressions.md. However, the "Summary" part of the proposal states

> The target-typed default feature is a shorter form variation of the default(T) operator, which allows the type to be omitted. Its type is inferred by target-typing instead. Aside from that, it behaves like default(T).

Here's my thinking:

Rather than add a new section, what we already have in [12.7.15 Default value expressions](https://github.com/ECMA-TC49-TG2/csharpstandard/blob/draft-v7/standard/expressions.md#12715-default-value-expressions) can be extended to handle this new spelling of a default value. For example, the existing grammar could be extended thus:

```antlr
default_value_expression
    : 'default' '(' type ')'
    | default_literal
    ;

default_literal
    : 'default'
    ;
```

rather than the more succinct

```antlr
default_value_expression
    : 'default' ('(' type ')')?
    ;
```

[By adding the grammatically unnecessary rule *default_literal*, we give a name to that alternative that we can refer to in the narrative.]

That said, I see an advantage to also having a name for the existing alternative; perhaps we could do something like the following instead:

```antlr
default_value_expression
    : explictly_typed_default
    | default_literal
    ;

explictly_typed_default
    : 'default' '(' type ')'
    ;

default_literal
    : 'default'
    ;
```

Using this approach, I think the changes to this section are minimal.

The only other edits I see are

1. To conversion.md, add a new section *default literal conversion*
1. To expressions.md deal with the possibility of `default == default` (if that is not already covered)

I'll write some tests to figure out what support is given by the `is` and `as` operators.

PR #236, opened 2021-03-14
C# 7.x: Default literals

RexJaeschke:
This PR implements the V7 proposal discussed in Issue #67.


BillWagner:
converting to draft: C# 7 feature

standard/expressions.md
```diff
@@ -2817,17 +2817,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
 
 ### 11.7.19 Default value expressions
 
-A default value expression is used to obtain the default value ([§9.3](variables.md#93-default-values)) of a type. Typically a default value expression is used for type parameters, since it might not be known if the type parameter is a value type or a reference type. (No conversion exists from the `null` literal ([§6.4.5.7](lexical-structure.md#6457-the-null-literal)) to a type parameter unless the type parameter is known to be a reference type ([§8.2](types.md#82-reference-types)).)
+A default value expression is used to obtain the default value ([§9.3](variables.md#93-default-values)) of an explicit or inferred type. Typically, a default value expression is used for type parameters, since it might not be known if the type parameter is a value type or a reference type. (No conversion exists from the `null` literal ([§6.4.5.7](lexical-structure.md#6457-the-null-literal)) to a type parameter unless the type parameter is known to be a reference type ([§8.2](types.md#82-reference-types)).)
```
jskeet: Since the introduction of the default literal, I'm not sure that this "typically" is correct. For example, I often see it used for default parameter values, especially for CancellationToken:

```csharp
public void DoSomethingAsync(CancellationToken cancellationToken = default)
```

No type parameters involved. Would "One notable use" be a suitable substitution here? And should everything after the first sentence be in a note?

standard/expressions.md
```diff
@@ -2817,17 +2817,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
...
     : 'default' '(' type ')'
     ;
+
+default_literal
+    : 'default'
+    ;
 ```
 
-If the *type* in a *default_value_expression* evaluates at run-time to a reference type, the result is `null` converted to that type. If the *type* in a *default_value_expression* evaluates at run-time to a value type, the result is the value type’s default value ([§8.3.3](types.md#833-default-constructors)).
+*default_literal* allows the type to be inferred by target-typing.
```
jskeet: I don't think "target-typing" is a phrase we use anywhere else. Should we drop this sentence entirely? I don't see anything about *how* the type of a default literal is inferred, but maybe that's already covered in general? (I'm genuinely unsure about that.)

standard/expressions.md
```diff
@@ -2817,17 +2817,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
...
+
+If the explicit or inferred *type* in a *default_value_expression* evaluates at run-time to a reference type, or in an unsafe context, to a pointer type ([§22.3](unsafe-code.md#223-pointer-types)), the result is `null` converted to that type. If the explicit or inferred *type* in a *default_value_expression* evaluates at run-time to a value type, the result is the value type's default value ([§8.3.3](types.md#833-default-constructors)).
```
jskeet: I don't think `*type*` should be in italics here (either time), given that it's only applicable in one branch of *default_value_expression*.

standard/expressions.md
```diff
@@ -2817,17 +2817,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
...
 
-A *default_value_expression* is a constant expression ([§11.20](expressions.md#1120-constant-expressions)) if *type* is a reference type or a type parameter that is known to be a reference type ([§8.2](types.md#82-reference-types)). In addition, a *default_value_expression* is a constant expression if the type is one of the following value types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type.
+A *default_value_expression* is a constant expression ([§11.20](expressions.md#1120-constant-expressions)) if explicit or inferred *type* is a reference type or a type parameter that is known to be a reference type ([§8.2](types.md#82-reference-types)). In addition, a *default_value_expression* is a constant expression if the type is one of the following value types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type.
```
```suggestion
A *default_value_expression* is a constant expression ([§11.20](expressions.md#1120-constant-expressions)) if the explicit or inferred type is a reference type or a type parameter that is known to be a reference type ([§8.2](types.md#82-reference-types)). In addition, a *default_value_expression* is a constant expression if the type is one of the following value types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type.
```

standard/expressions.md
jskeet, line 2839:
```suggestion
If the explicit or inferred type in a *default_value_expression* evaluates at run-time to a reference type, or in an unsafe context, to a pointer type ([§22.3](unsafe-code.md#223-pointer-types)), the result is `null` converted to that type. If the explicit or inferred type in a *default_value_expression* evaluates at run-time to a value type, the result is the value type's default value ([§8.3.3](types.md#833-default-constructors)).
```

jskeet: This duplicates information in 9.3, which is entitled "default values" but is about default values of variables.

I wonder whether it might be worth refactoring to have "default value of a type" somewhere in clause 8 (types) and then refer to that from both here and 9.3. That can come later though.

jskeet (commented):
A bunch of comments - we should discuss this at the next meeting.

standard/expressions.md
BillWagner: I agree. It's not normative. §10.2.12 and §10.3.8 have normative text that there isn't a conversion from `null` to a type parameter that isn't known to be a reference type.

Nigel-Ecma: I don't think the phrasing "an explicit or inferred type" is correct here (and elsewhere), we should follow the phraseology used for `null`. So something along the lines of "obtain the default value of a [target] type" depending on the context.

standard/expressions.md
```diff
@@ -2819,17 +2819,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
...
     : 'default' '(' type ')'
     ;
+
+default_literal
+    : 'default'
+    ;
 ```
 
-If the *type* in a *default_value_expression* evaluates at run-time to a reference type, the result is `null` converted to that type. If the *type* in a *default_value_expression* evaluates at run-time to a value type, the result is the value type’s default value ([§8.3.3](types.md#833-default-constructors)).
+*default_literal* allows the type to be inferred by target-typing.
```
Nigel-Ecma: Agree with @jskeet, "target-typing" is not used in the Standard. I think we should describe *default_literal* in the same way as a *null_literal* is described, i.e. replace this line with:

> A `default_literal` represents a default value (§9.3). It does not have a type, but can be converted to any type
or nullable through a default literal conversion (§default-literal-new-clause)

standard/expressions.md
Nigel-Ecma: Agree with @jskeet. This duplicates information in §9.3 and in the first line of this clause. Just remove the whole paragraph. If we need to talk about *runtime type* either incorporate that into the first line of the clause or replace this paragraph with one saying the value produced depends on the target type at runtime – but that might be implicit elsewhere?

standard/variables.md
```diff
@@ -160,6 +160,7 @@ The default value of a variable depends on the type of the variable and is deter
 
 - For a variable of a *value_type*, the default value is the same as the value computed by the *value_type*’s default constructor ([§8.3.3](types.md#833-default-constructors)).
 - For a variable of a *reference_type*, the default value is `null`.
+- In an unsafe context, for a variable of a *pointer_type*, the default value is `null`.
```
Nigel-Ecma, line 163: Is "In an unsafe context" needed here as pointer types only occur in unsafe contexts? If its not this and preceding line become:

> - For a variable of a *reference_type* or *pointer_type*, the default value is `null`.

standard/expressions.md
```diff
@@ -2819,17 +2819,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
...
 
-A *default_value_expression* is a constant expression ([§11.21](expressions.md#1121-constant-expressions)) if *type* is a reference type or a type parameter that is known to be a reference type ([§8.2](types.md#82-reference-types)). In addition, a *default_value_expression* is a constant expression if the type is one of the following value types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type.
+A *default_value_expression* is a constant expression ([§11.21](expressions.md#1121-constant-expressions)) if explicit or inferred *type* is a reference type or a type parameter that is known to be a reference type ([§8.2](types.md#82-reference-types)). In addition, a *default_value_expression* is a constant expression if the type is one of the following value types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type.
```
Nigel-Ecma: Use of "explicit or inferred" to remove. The "In addition" phrasing seems wrong, I'd be tempted to just use a single list:

> A *default_value_expression* is a constant expression ([§11.21](expressions.md#1121-constant-expressions)) if the type is:
> - a reference type or a type parameter that is known to be a reference type ([§8.2](types.md#82-reference-types));
> - one of the following value types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,`; or
> - any enumeration type.

Nigel-Ecma (dismissed):
Some of the phraseology doesn't match the Standard (it presumable comes from the source material) so needs changing. Further I think we should describe default similar to null for consistency.

The restrictions on the use of *default_literal* as the argument to unary & binary operators, as described in https://github.com/dotnet/csharplang/blob/main/meetings/2017/LDM-2017-05-31.md#default-in-operators, do not appear to be in this Standard. Where they ever added to the spec or implemented?

standard/expressions.md
jskeet: That looks good to me. Will discuss at the meeting later "today" (UTC). @Nigel-Ecma: I know you can't make the meeting, but I'm going to assume you're happy with your own suggestion.

standard/variables.md
jskeet: And potentially this should move to the unsafe section of the spec anyway?

standard/expressions.md
BillWagner: I agree with Nigel's point on "inferred type". When I read that, I think it would allow this:

```csharp
var i = default;
```

That's not allowed. I think the following suggestion on "conversion to any type" would be better: it would work for both known types and type parameters.

standard/expressions.md
```diff
@@ -2819,17 +2819,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
...
+
+default_literal
+    : 'default'
+    ;
 ```
 
-If the *type* in a *default_value_expression* evaluates at run-time to a reference type, the result is `null` converted to that type. If the *type* in a *default_value_expression* evaluates at run-time to a value type, the result is the value type’s default value ([§8.3.3](types.md#833-default-constructors)).
+*default_literal* allows the type to be inferred by target-typing.
+
+If the explicit or inferred *type* in a *default_value_expression* evaluates at run-time to a reference type, or in an unsafe context, to a pointer type ([§22.3](unsafe-code.md#223-pointer-types)), the result is `null` converted to that type. If the explicit or inferred *type* in a *default_value_expression* evaluates at run-time to a value type, the result is the value type's default value ([§8.3.3](types.md#833-default-constructors)).
```
BillWagner, line 2841: Instead of "inferred type", I would use the conversion language proposed earlier. Also, I don't think this happens *at run time* so I removed that phrase.

```suggestion
If the *default_value_expression* is converted to a reference type, or in an unsafe context, to a pointer type ([§22.3](unsafe-code.md#223-pointer-types)), the result is `null` converted to that type. If the *default_value_expression* is converted to a value type, the result is the value type's default value ([§8.3.3](types.md#833-default-constructors)).
```

BillWagner (approved):
A few nits, but this is getting very close.

standard/expressions.md
Nigel-Ecma: @jskeet – in this rare case I do agree with myself ;-)

BillWagner:
closing and reopening to trigger actions.

standard/expressions.md
```diff
@@ -2819,17 +2819,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
 
 ### 11.7.19 Default value expressions
 
-A default value expression is used to obtain the default value ([§9.3](variables.md#93-default-values)) of a type. Typically a default value expression is used for type parameters, since it might not be known if the type parameter is a value type or a reference type. (No conversion exists from the `null` literal ([§6.4.5.7](lexical-structure.md#6457-the-null-literal)) to a type parameter unless the type parameter is known to be a reference type ([§8.2](types.md#82-reference-types)).)
+A default value expression is used to obtain the default value ([§9.3](variables.md#93-default-values)) of an explicit or inferred type. Typically, a default value expression is used for type parameters, since it might not be known if the type parameter is a value type or a reference type. (No conversion exists from the `null` literal ([§6.4.5.7](lexical-structure.md#6457-the-null-literal)) to a type parameter unless the type parameter is known to be a reference type ([§8.2](types.md#82-reference-types)).)
```
```suggestion
A default value expression is used to obtain the default value ([§9.3](variables.md#93-default-values)) of a type.
```

standard/expressions.md
```suggestion
A *default_literal* represents a default value (§9.3). It does not have a type, but can be converted to any type through a default literal conversion (§default-literal-new-clause).
```

jskeet: @Nigel-Ecma: Removed the nullable part as "a type" is fine. (In the null literal it's to a reference type or nullable value type; we don't need the restriction here.)

standard/expressions.md
```diff
@@ -2819,17 +2819,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
...
+
+default_literal
+    : 'default'
+    ;
 ```
 
-If the *type* in a *default_value_expression* evaluates at run-time to a reference type, the result is `null` converted to that type. If the *type* in a *default_value_expression* evaluates at run-time to a value type, the result is the value type’s default value ([§8.3.3](types.md#833-default-constructors)).
+A *default_literal* represents a default value (§9.3). It does not have a type, but can be converted to any type through a default literal conversion (§default-literal-new-clause).
+
+If the explicit or inferred *type* in a *default_value_expression* evaluates at run-time to a reference type, or in an unsafe context, to a pointer type ([§22.3](unsafe-code.md#223-pointer-types)), the result is `null` converted to that type. If the explicit or inferred *type* in a *default_value_expression* evaluates at run-time to a value type, the result is the value type's default value ([§8.3.3](types.md#833-default-constructors)).
```
```suggestion
The result of a *default_value_expression* is the default ([§9.3](variables.md#93-default-values)) of the explicit type in an *explictly_typed_default*, or the target type of the conversion for a *default_value_expression*.
```

standard/expressions.md
jskeet: Doing this in a suggestion. (Runtime type is basically about type parameters, but doesn't need to be called out if we're referring to 9.3.)

standard/expressions.md
```diff
@@ -2819,17 +2819,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
...
 
-A *default_value_expression* is a constant expression ([§11.21](expressions.md#1121-constant-expressions)) if *type* is a reference type or a type parameter that is known to be a reference type ([§8.2](types.md#82-reference-types)). In addition, a *default_value_expression* is a constant expression if the type is one of the following value types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type.
+A *default_value_expression* is a constant expression ([§11.21](expressions.md#1121-constant-expressions)) if explicit or inferred *type* is a reference type or a type parameter that is known to be a reference type ([§8.2](types.md#82-reference-types)). In addition, a *default_value_expression* is a constant expression if the type is one of the following value types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type.
```
```suggestion
A *default_value_expression* is a constant expression ([§11.21](expressions.md#1121-constant-expressions)) if the type is one of:

- a reference type
- a type parameter that is known to be a reference type ([§8.2](types.md#82-reference-types));
- one of the following value types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,`; or
- any enumeration type.
```

standard/expressions.md
jskeet: Resolving this comment having added it as a suggestion.

standard/expressions.md
```diff
@@ -2819,17 +2819,28 @@ The `unchecked` operator is convenient when writing constants of the signed inte
...
     : 'default' '(' type ')'
     ;
+
+default_literal
+    : 'default'
+    ;
 ```
 
-If the *type* in a *default_value_expression* evaluates at run-time to a reference type, the result is `null` converted to that type. If the *type* in a *default_value_expression* evaluates at run-time to a value type, the result is the value type’s default value ([§8.3.3](types.md#833-default-constructors)).
+A *default_literal* represents a default value (§9.3). It does not have a type, but can be converted to any type through a default literal conversion (§default-literal-new-clause).
```
```suggestion
A *default_literal* represents a default value ([§9.3](variables.md#93-default-values)). It does not have a type, but can be converted to any type through a default literal conversion (§default-literal-new-clause).
```

jskeet (approved):

standard/variables.md
jskeet: I'll be creating a new issue to say "let's move all pointer stuff to section 22" - and then a PR to do this.

BillWagner (approved):
This LGTM (per Nov 30th meeting)

PR #238, opened 2021-03-15
C# 7.x: Add initializer list to `stackalloc`

RexJaeschke:
I cloned a bunch of text from array initializers, omitting the ability to have multidimensional array allocation and arrays of arrays.

A very long time after I first created this PR, I discovered that other functionality was added earlier than V7.3. Starting in V7.1, `stackalloc` itself can appear *outside* of unsafe code, *provided* it is used in the context of a `Span` or `ReadOnlySpan`. As a result, I smuggled into this PR (via its branch) more edits. They make `stackalloc` a primary operator, and almost all of the semantics of it are moved from unsafe.md to expressions.md.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -3065,13 +3065,12 @@ stackalloc_initializer
     ;
 
 stackalloc_initializer_elements
-    : '{' stackalloc_initializer_element_list? '}'
-    | '{' stackalloc_initializer_element_list ',' '}'
-    ;
+     : '{' stackalloc_initializer_element_list '}'
+     ;
 
 stackalloc_initializer_element_list
-    : stackalloc_element_initializer (',' stackalloc_element_initializer)*
-    ;
+     : stackalloc_element_initializer (',' stackalloc_element_initializer)* ','?
+     ;
     
 stackalloc_element_initializer
     : expression
@@ -3082,13 +3081,13 @@ The *unmanaged_type* ([§8.8](types.md#88-unmanaged-types)) indicates the type o
 
 If *unmanaged_type* is omitted, it is inferred from the corresponding *stackalloc_initializer_elements*. If *expression* is omitted from *stackalloc_initializer*, it is inferred to be the number of *stackalloc_element_initializer*s in the corresponding *stackalloc_initializer_elements*.
 
-When a *stackalloc_initializer* includes both *expression* and *stackalloc_initializer_elements* the number of elements in that *stackalloc_initializer_elements* shall match the value of *expression*.
+When a *stackalloc_initializer* includes both *expression* and *stackalloc_initializer_elements*, the *expression* shall be a *constant_expression* and the number of elements in that *stackalloc_initializer_elements* shall match the value of *expression*.
 
 A stack allocation initializer of the form `stackalloc T[E]` requires `T` to be an *unmanaged_type* and `E` to be an expression implicitly convertible to type `int`. The operator allocates `E * sizeof(T)` bytes from the call stack, and the resulting type and value are determined, as follows:
 
 - If *unmanaged_type* is a *pointer_type* ([§22.3](unsafe-code.md#223-pointer-types)) or if *stackalloc_initializer* is used in a context that requires a *pointer_type*, the result is a pointer, of type `T*`, to the newly allocated block. If the context cannot be inferred, a pointer context is assumed. As pointer contexts require unsafe code, see §stack-allocation for more information.
 - Otherwise, the result has type `System.Span<T>` and maps to the newly allocated block. Apart from being used as the initializer of a local variable, this result shall be permitted in the following contexts only:
-  - The right operand of an *assignment_operator* that is not embedded in some larger expression
+  - The right operand of a simple assignment expression (§11.19.2) that is not embedded in some larger expression
   - The second and third operands in a *conditional_expression*
 
 If `E` is a negative value, then the behavior is undefined. If `E` is zero, then no allocation is made, and the value returned is implementation-defined. If there is not enough memory available to allocate a block of the given size, a `System.StackOverflowException`  is thrown.
@@ -3112,7 +3111,7 @@ Except for the `stackalloc` operator, C# provides no predefined constructs for m
 > Span<int> spn2 = stackalloc int[3] { -10, -15, -30 };   // memory initialized
 > Span<int> spn3 = stackalloc[] { 11, 12, 13 };           // type int is inferred
 > var spn4 = stackalloc[] { 11, 12, 13 };                 // error; need unsafe mode as can't infer context
-> Span<long> spn5 = stackalloc[] { 11, 12, 13 };          // error; no conversion from int to Span<long>
+> Span<long> spn5 = stackalloc[] { 11, 12, 13 };          // error; no conversion from Span<int> to Span<long>
 > Span<long> spn6 = stackalloc[] { 11, 12L, 13 };         // converts 11 and 13, and returns Span<long>
 > Span<long> spn7 = stackalloc long[] { 11, 12, 13 };     // converts all and returns Span<long>
 > ReadOnlySpan<int> spn8 = stackalloc int[] { 10, 22, 30 }; // implicit conversion of Span<T>
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -3091,7 +3091,7 @@ When *stackalloc_initializer_elements* is present, the *stackalloc_initializer_e
 
 Access via an instance of `System.Span<T>` to the elements of an allocated block is range checked.
 
-Stack allocation initializers are not permitted in `catch` or `finally` blocks ([§12.11](statements.md#1211-the-try-statement)).
+Stack allocation initializers are not permitted in `catch` or `finally` blocks ([§13.11](statements.md#1311-the-try-statement)).
 
 > *Note*: Stack allocation initializers are allowed in `async` methods, but their return value can't be assigned. Neither pointers nor `ref struct` types, like `Span<T>` are allowed in `async` methods. *end note*
 <!-- markdownlint-disable MD028 -->
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -3106,15 +3106,15 @@ Except for the `stackalloc` operator, C# provides no predefined constructs for m
 > *Example*:
 >
 > ```csharp
-> Span<int> spn1 = stackalloc int[3];                     // memory uninitialized
-> Span<int> spn2 = stackalloc int[3] { -10, -15, -30 };   // memory initialized
-> Span<int> spn3 = stackalloc[] { 11, 12, 13 };           // type int is inferred
+> Span<int> span1 = stackalloc int[3];                     // memory uninitialized
+> Span<int> span2 = stackalloc int[3] { -10, -15, -30 };   // memory initialized
+> Span<int> span3 = stackalloc[] { 11, 12, 13 };           // type int is inferred
 > var spn4 = stackalloc[] { 11, 12, 13 };                 // error; result is int*, not allowed in a safe context
-> Span<long> spn5 = stackalloc[] { 11, 12, 13 };          // error; no conversion from Span<int> to Span<long>
-> Span<long> spn6 = stackalloc[] { 11, 12L, 13 };         // converts 11 and 13, and returns Span<long>
-> Span<long> spn7 = stackalloc long[] { 11, 12, 13 };     // converts all and returns Span<long>
-> ReadOnlySpan<int> spn8 = stackalloc int[] { 10, 22, 30 }; // implicit conversion of Span<T>
-> Widget<double> spn9 = stackalloc double[] { 1.2, 5.6 }; // implicit conversion of Span<T>
+> Span<long> span5 = stackalloc[] { 11, 12, 13 };          // error; no conversion from Span<int> to Span<long>
+> Span<long> span6 = stackalloc[] { 11, 12L, 13 };         // converts 11 and 13, and returns Span<long>
+> Span<long> span7 = stackalloc long[] { 11, 12, 13 };     // converts all and returns Span<long>
+> ReadOnlySpan<int> span8 = stackalloc int[] { 10, 22, 30 }; // implicit conversion of Span<T>
+> Widget<double> span9 = stackalloc double[] { 1.2, 5.6 }; // implicit conversion of Span<T>
 > 
 > public class Widget<T>
 > {
@@ -3122,7 +3122,7 @@ Except for the `stackalloc` operator, C# provides no predefined constructs for m
 > }
 > ```
 >
-> In the case of `spn8`, `stackalloc` results in a `Span<int>`, which is converted by an implicit operator to `ReadOnlySpan<int>`. Similarly, for `spn9`, the resulting `Span<double>` is converted to the user-defined type `Widget<double> using the conversion, as shown.
+> In the case of `span8`, `stackalloc` results in a `Span<int>`, which is converted by an implicit operator to `ReadOnlySpan<int>`. Similarly, for `span9`, the resulting `Span<double>` is converted to the user-defined type `Widget<double> using the conversion, as shown.
 > *end example*
 
 ### 12.8.21 Nameof expressions
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -147,7 +147,7 @@ The precedence of an operator is established by the definition of its associated
 >
 > |  **Subclause**      | **Category**                     | **Operators**                                          |
 > |  -----------------  | -------------------------------  | -------------------------------------------------------|
-> |  [§12.8](expressions.md#128-primary-expressions)              | Primary                          | `x.y` `x?.y` `f(x)` `a[x]` `a?[x]` `x++` `x--` `new` `typeof` `default` `checked` `unchecked` `delegate`  |
+> |  [§12.8](expressions.md#128-primary-expressions)              | Primary                          | `x.y` `x?.y` `f(x)` `a[x]` `a?[x]` `x++` `x--` `new` `typeof` `default` `checked` `unchecked` `delegate` `stackalloc`  |
 > |  [§12.9](expressions.md#129-unary-operators)              | Unary                            | `+` `-` `!` `~` `++x` `--x` `(T)x` `await x` |
 > |  [§12.10](expressions.md#1210-arithmetic-operators)              | Multiplicative                   | `*` `/` `%` |
 > |  [§12.10](expressions.md#1210-arithmetic-operators)              | Additive                         | `+` `-` |
@@ -1280,6 +1280,7 @@ primary_no_array_creation_expression
     | anonymous_method_expression
     | pointer_member_access     // unsafe code support
     | pointer_element_access    // unsafe code support
+    | stackalloc_initializer
     ;
 ```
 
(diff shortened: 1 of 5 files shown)


RexJaeschke:
There is one unresolved issue remaining logged on Teams: The C# standard currently states, "If there is not enough memory available to allocate a block of the given size, a `System.StackOverflowException` is thrown." However, this is not true. My tests show that the program terminates with the message “Stack overflow.” written to the console, and exit code -1073741571 (which might be arbitrary). My try/catch clauses for `System.StackOverflowException` and `System.Exception` are ignored. Any ideas on how I should spec the behavior I observe?

standard/expressions.md
```diff
@@ -2370,6 +2371,77 @@ A *default_value_expression* is a constant expression ([§12.20](expressions.md#
...
+
+A stack allocation initializer of the form `stackalloc T[E]` requires `T` to be an *unmanaged_type* and `E` to be an expression implicitly convertible to type `int`. The operator allocates `E * sizeof(T)` bytes from the call stack, and the resulting type and value are determined, as follows:
+
+- If *unmanaged_type* is a *pointer_type* ([§23.3](unsafe-code.md#233-pointer-types)) or if *stackalloc_initializer* is used in a context that requires a *pointer_type*, the result is a pointer, of type `T*`, to the newly allocated block. If the context cannot be inferred, a pointer context is assumed. As pointer contexts require unsafe code, see [§23.3](unsafe-code.md#239-stack-allocation) for more information.
+- Otherwise, the result has type `System.Span<T>` and maps to the newly allocated block. Apart from being used as the initializer of a local variable, this result shall be permitted in the following contexts only:
+  - The right operand of an *assignment_operator* that is not embedded in some larger expression
+  - The second and third operands in a *conditional_expression*
```
KalleOlaviNiemitalo: Seems like this would be invalid then:

```C#
using System;
public class C {
    public void M() {
        _ = ((Span<byte>)stackalloc byte[1])[0];
    }
}
```

standard/expressions.md
```diff
@@ -2370,6 +2371,77 @@ A *default_value_expression* is a constant expression ([§12.20](expressions.md#
...
+stackalloc_element_initializer
+    : expression
+    ;
+```
+
+The *unmanaged_type* ([§9.8](types.md#98-unmanaged-types)) indicates the type of the items that will be stored in the newly allocated location, and the *expression* indicates the number of these items. Taken together, these specify the required allocation size. As the size of a stack allocation cannot be negative, it is a compile-time error to specify the number of items as a *constant_expression* that evaluates to a negative value.
+
+If *unmanaged_type* is omitted, it is inferred from the corresponding *stackalloc_initializer_elements*. If *expression* is omitted from *stackalloc_initializer*, it is inferred to be the number of *stackalloc_element_initializer*s in the corresponding *stackalloc_initializer_elements*.
+
+When a *stackalloc_initializer* includes both *expression* and *stackalloc_initializer_elements* the number of elements in that *stackalloc_initializer_elements* shall match the value of *expression*.
```
KalleOlaviNiemitalo: Should *expression* be required to be a constant expression in that case, so that the compiler can check that the length matches? Thus making the following invalid even if `i` would be 1 at run time.

```C#
using System;
public class C {
    public void M(int i) {
        Span<byte> s = stackalloc byte[i] {1};
    }
}
```

KalleOlaviNiemitalo:
> My try/catch clauses for System.StackOverflowException and System.Exception are ignored.

I'm not sure whether the runtime considers a stack overflow a *corrupted state exception* as described in [HandleProcessCorruptedStateExceptionsAttribute](https://docs.microsoft.com/dotnet/api/system.runtime.exceptionservices.handleprocesscorruptedstateexceptionsattribute?view=netframework-4.8).

KalleOlaviNiemitalo:
Also <https://github.com/dotnet/runtime/issues/40302> for catchable stack overflow.

standard/expressions.md
KalleOlaviNiemitalo: Oh right, it is invalid in C# 7, but valid in C# 8.

KalleOlaviNiemitalo: The proposed wording would allow the following *stackalloc_initializer* because it is the second operand in a *conditional_expression*, yet the compiler rejects it in C# 7.

```C#
using System;
public class C {
    public void M() {
        _ = ((Span<byte>)(true ? stackalloc byte[1] : default(Span<byte>)))[0];
    }
}
```

standard/expressions.md
KalleOlaviNiemitalo: [12.7.11.5 Array creation expressions](https://github.com/dotnet/csharpstandard/blob/cf0e08d15a437a1773e6da213358de6fba1a7ee6/standard/expressions.md#127115-array-creation-expressions) states a similar requirement:

<https://github.com/dotnet/csharpstandard/blob/cf0e08d15a437a1773e6da213358de6fba1a7ee6/standard/expressions.md#L1924>

standard/expressions.md
KalleOlaviNiemitalo: *assignment_operator* includes compound assignment operators, so the wording would allow this as well:

```C#
using System;
public class C {
    public static C operator +(C c, Span<byte> s) { return c; }
    public static void M(C c) {
        c += stackalloc byte[1];
    }
}
```

But the compiler rejects it in C# 7.

KalleOlaviNiemitalo:
The behavior on stack overflow could be implementation-defined, constrained such that if the implementation throws a catchable exception, then it must be a StackOverflowException. An implementation that throws e.g. ThreadAbortException or OutOfMemoryException instead, and lets a C# `catch` clause catch this exception, would not be conforming. I think requiring implementors to document the behavior would not be too burdensome.

standard/expressions.md
```diff
@@ -2370,6 +2371,77 @@ A *default_value_expression* is a constant expression ([§12.20](expressions.md#
...
+
+Except for the `stackalloc` operator, C# provides no predefined constructs for managing non-garbage collected memory. Such services are typically provided by supporting class libraries or imported directly from the underlying operating system.
+
+> *Example*:
+> ```csharp
+> Span<int> spn1 = stackalloc int[3];                     // memory uninitialized
+> Span<int> spn2 = stackalloc int[3] { -10, -15, -30 };   // memory initialized
+> Span<int> spn3 = stackalloc[] { 11, 12, 13 };           // type int is inferred
+> var spn4 = stackalloc[] { 11, 12, 13 };                 // error; need unsafe mode as can't infer context
+> Span<long> spn5 = stackalloc[] { 11, 12, 13 };          // error; no conversion from int to Span<long>
```
KalleOlaviNiemitalo: no conversion from *Span\<int\>* to Span\<long\>

standard/expressions.md
```diff
@@ -2370,6 +2371,77 @@ A *default_value_expression* is a constant expression ([§12.20](expressions.md#
...
+
+When a *stackalloc_initializer* includes both *expression* and *stackalloc_initializer_elements* the number of elements in that *stackalloc_initializer_elements* shall match the value of *expression*.
+
+A stack allocation initializer of the form `stackalloc T[E]` requires `T` to be an *unmanaged_type* and `E` to be an expression implicitly convertible to type `int`. The operator allocates `E * sizeof(T)` bytes from the call stack, and the resulting type and value are determined, as follows:
+
+- If *unmanaged_type* is a *pointer_type* ([§23.3](unsafe-code.md#233-pointer-types)) or if *stackalloc_initializer* is used in a context that requires a *pointer_type*, the result is a pointer, of type `T*`, to the newly allocated block. If the context cannot be inferred, a pointer context is assumed. As pointer contexts require unsafe code, see [§23.3](unsafe-code.md#239-stack-allocation) for more information.
```
KalleOlaviNiemitalo: It is not clear what "a context that requires a *pointer_type*" means here:

```C#
using System;

class Widget
{
    public static unsafe implicit operator Widget(void* p) => null;
}

static class Class1
{
    static unsafe void S()
    {
        Widget w = stackalloc byte[1];
    }
}
```

The compiler chooses to return `byte*` from `stackalloc` and convert that to `void*`. However, the context does not seem to require a *pointer_type* as the context would accept `Span<int>`. Does the standard explain why the compiler does not choose `Span<byte>` and then report that it cannot be converted to `Span<int>`?

standard/expressions.md
```diff
@@ -2370,6 +2371,77 @@ A *default_value_expression* is a constant expression ([§12.20](expressions.md#
...
+- Otherwise, the result has type `System.Span<T>` and maps to the newly allocated block. Apart from being used as the initializer of a local variable, this result shall be permitted in the following contexts only:
+  - The right operand of an *assignment_operator* that is not embedded in some larger expression
+  - The second and third operands in a *conditional_expression*
+
+If `E` is a negative value, then the behavior is undefined. If `E` is zero, then no allocation is made, and the value returned is implementation-defined. If there is not enough memory available to allocate a block of the given size, a `System.StackOverflowException`  is thrown.
+
+When *stackalloc_initializer_elements* is present, the *stackalloc_initializer_elelement_list* shall consist of a sequence of expressions, each having an implicit conversion to *unmanaged_type* ([§11.2](conversions.md#112-implicit-conversions)). The expressions initialize elements in the allocated memory in increasing order, starting with the element at index zero. In the absence of a *stackalloc_initializer_elements*, the content of the newly allocated memory is undefined.
```
KalleOlaviNiemitalo: *stackalloc_initializer_element_list*, not *stackalloc_initializer_elelement_list*

standard/expressions.md
KalleOlaviNiemitalo: If *stackalloc_initializer* can be used in *assignment_operator* or *conditional_expression*, then that needs some changes in the grammar too. So far, *stackalloc_initializer* is referenced only from *local_variable_initializer*.

standard/expressions.md
```diff
@@ -2370,6 +2371,77 @@ A *default_value_expression* is a constant expression ([§12.20](expressions.md#
...
+
+Access via an instance of `System.Span<T>` to the elements of an allocated block is range checked.
+
+Stack allocation initializers are not permitted in `catch` or `finally` blocks ([§13.11](statements.md#1311-the-try-statement)).
```
KalleOlaviNiemitalo, line 3096: Where else they are not permitted:

- `stackalloc` returning a pointer in an `async` method is largely disallowed because §12.8.8.1 does not allow *await_expression*  in an unsafe context. If the `async` method does not contain any *await_expression*, then the pointer is not much of a problem. <https://github.com/dotnet/csharpstandard/blob/cf0e08d15a437a1773e6da213358de6fba1a7ee6/standard/expressions.md#L2562>
- `stackalloc` returning a pointer in an iterator is disallowed because §13.3.1 because does not allow an unsafe context in an iterator block. <https://github.com/dotnet/csharpstandard/blob/cf0e08d15a437a1773e6da213358de6fba1a7ee6/standard/statements.md#L135>
- `stackalloc` returning a Span is in an `async` method or in an iterator is disallowed by <https://github.com/dotnet/csharpstandard/pull/333> §ref-modifier-new-clause. <https://github.com/dotnet/csharpstandard/blob/a2492999cae55edc25ba5f46f5ef83a3a3ae8b91/standard/structs.md#L69-L70>

Might be worth adding a note.

standard/expressions.md
Nigel-Ecma: In similar vein to @KalleOlaviNiemitalo’s observation should `{ }` be a valid *stackalloc_initializer_elements* as allowed by the grammar? Indeed the grammar allows `stackalloc [] {}`.

Maybe:

```[ANTLR]
stackalloc_initializer_elements
     : '{' stackalloc_initializer_element_list '}'
     ;
stackalloc_initializer_element_list
     : stackalloc_element_initializer (',' stackalloc_element_initializer)* ','?
     ;

File: standard/expressions.md (hunk 1 above)
Comment (Nigel-Ecma) [reply standard/expressions.md]: > Oh right, it is invalid in C# 7, but valid in C# 8.

C# 7 seems the wiser choice here, why is it valid in C# 8?

Comment (Nigel-Ecma) [reply standard/expressions.md]: > The proposed wording would allow the following _stackalloc_initializer_ because it is the second operand in a _conditional_expression_, yet the compiler rejects it in C# 7.
> 
> ```cs
> using System;
> public class C {
>     public void M() {
>         _ = ((Span<byte>)(true ? stackalloc byte[1] : default(Span<byte>)))[0];
>     }
> }
> ```

Was the text meant to imply “The second __*and*__ third operands in a”, i.e. neither or both branches? One could imagine a reason for that restriction.

**Review by Nigel-Ecma — CHANGES_REQUESTED** [review]
As @KalleOlaviNiemitalo’s comments demonstrate this needs some work.

File: standard/expressions.md (hunk 2 above)
Comment (BillWagner) [reply standard/expressions.md]: updated per these three comments.

File: standard/expressions.md (hunk 3 above)
Comment (BillWagner) [reply standard/expressions.md]: fixed

File: standard/expressions.md (hunk 1 above)
Comment (BillWagner) [reply standard/expressions.md]: Admittedly, my head hurts now. 

The previous example fails because the `stackalloc` is part of a "nested expression". The following variation compiles in C# 7.3:

```csharp
_ = true ? stackalloc byte[1] : default(Span<byte>);
```

I'm asking for a definition of "nested expression". It doesn't occur in the standard, and I can't find the term used in LDM notes.

BillWagner (commented):
One outstanding question, but I've addressed the other comments.

standard/expressions.md
BillWagner: This answer turns out to simplify the PR somewhat. In C# 7.3, a *stack_initializer* can be used only as a *local_variable_initializer*. In that context, the result is `T*`, with an implicit conversation to `Span<T>` for use in safe contexts. 

That means the `stackalloc` initializer can be used as *either or both* the 2nd and 3rd operands of a *conditional_expression*.

standard/expressions.md
BillWagner: Note added.

standard/expressions.md
```diff
@@ -2906,6 +2907,77 @@ A *default_value_expression* is a constant expression ([§11.21](expressions.md#
...
+
+When *stackalloc_initializer_elements* is present, the *stackalloc_initializer_element_list* shall consist of a sequence of expressions, each having an implicit conversion to *unmanaged_type* ([§10.2](conversions.md#102-implicit-conversions)). The expressions initialize elements in the allocated memory in increasing order, starting with the element at index zero. In the absence of a *stackalloc_initializer_elements*, the content of the newly allocated memory is undefined.
+
+Access via an instance of `System.Span<T>` to the elements of an allocated block is range checked.
+
+Stack allocation initializers are not permitted in `catch` or `finally` blocks ([§12.11](statements.md#1211-the-try-statement)).
+
+> *Note*: Stack allocation initializers are allowed in `async` methods, but their return value can't be assigned. Neither pointers nor `ref struct` types, like `Span<T>` are allowed in `async` methods. *end note*
```
BillWagner, line 3098: This is because the following is legal (if not particularly useful):

```csharp
public async Task<int> M4()
{
    _ = stackalloc int[3];
    await Task.Delay(1000);
    return 5;
}
```

But, this isn't legal:

```csharp
public async Task<int> M4()
{
    Span<int> storage = stackalloc int[3]; // error can't declare a ref struct in an async method
    await Task.Delay(1000);
    return 5;
}
```

BillWagner:
@jskeet 

This one should be ready for a final review. I don't think we have time to add more to the agenda for tomorrow, but can we add it to the list for review / approval asynchronously between this meeting and the next?

jskeet:
Just found this again - will review today/tomorrow.

standard/expressions.md
```diff
@@ -2906,6 +2907,77 @@ A *default_value_expression* is a constant expression ([§11.21](expressions.md#
...
+     ;
+    
+stackalloc_element_initializer
+    : expression
+    ;
+```
+
+The *unmanaged_type* ([§8.8](types.md#88-unmanaged-types)) indicates the type of the items that will be stored in the newly allocated location, and the *expression* indicates the number of these items. Taken together, these specify the required allocation size. As the size of a stack allocation cannot be negative, it is a compile-time error to specify the number of items as a *constant_expression* that evaluates to a negative value.
+
+If *unmanaged_type* is omitted, it is inferred from the corresponding *stackalloc_initializer_elements*. If *expression* is omitted from *stackalloc_initializer*, it is inferred to be the number of *stackalloc_element_initializer*s in the corresponding *stackalloc_initializer_elements*.
```
jskeet: We should say *how* it's inferred (and that the inferred type must be an unmanaged type - although that part is sort of present in line 2943).
That can come in a fast-follow PR though.

standard/expressions.md
```diff
@@ -2906,6 +2907,77 @@ A *default_value_expression* is a constant expression ([§11.21](expressions.md#
...
+
+Stack allocation initializers are not permitted in `catch` or `finally` blocks ([§12.11](statements.md#1211-the-try-statement)).
+
+> *Note*: Stack allocation initializers are allowed in `async` methods, but their return value can't be assigned. Neither pointers nor `ref struct` types, like `Span<T>` are allowed in `async` methods. *end note*
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+> *Note*: There is no way to explicitly free memory allocated using `stackalloc`. *end note*
+
+All stack-allocated memory blocks created during the execution of a function member are automatically discarded when that function member returns.
```
jskeet, line 3104: Not until then? What happens if it's in a loop?

```csharp
for (int i = 0; i < 10_000; i++)
{
    Span<int> spn1 = stackalloc int[10];
}
```

According to this, we'll end up allocating space for 100,000 ints on the stack... is that actually the case? Or is the stack space reclaimed at the end of the loop body?

standard/expressions.md
```diff
@@ -2906,6 +2907,77 @@ A *default_value_expression* is a constant expression ([§11.21](expressions.md#
...
+
+Except for the `stackalloc` operator, C# provides no predefined constructs for managing non-garbage collected memory. Such services are typically provided by supporting class libraries or imported directly from the underlying operating system.
+
+> *Example*:
+>
+> ```csharp
+> Span<int> spn1 = stackalloc int[3];                     // memory uninitialized
```
jskeet: Can we make these `span1` etc instead of `spn1`? Removing the 'a' seems a little pointless

standard/expressions.md
```diff
@@ -2906,6 +2907,77 @@ A *default_value_expression* is a constant expression ([§11.21](expressions.md#
...
+> ReadOnlySpan<int> spn8 = stackalloc int[] { 10, 22, 30 }; // implicit conversion of Span<T>
+> Widget<double> spn9 = stackalloc double[] { 1.2, 5.6 }; // implicit conversion of Span<T>
+> 
+> public class Widget<T>
+> {
+>     public static implicit operator Widget<T>(Span<double> sp) { return null; }
+> }
+> ```
+>
+> In the case of `spn8`, `stackalloc` results in a `Span<int>`, which is converted by an implicit operator to `ReadOnlySpan<int>`. Similarly, for `spn9`, the resulting `Span<double>` is converted to the user-defined type `Widget<double> using the conversion, as shown.
```
jskeet: Note that if we change the names above, they need to be handled here too.

standard/expressions.md
jskeet: Okay - but it would presumably be feasible to call a method using the span as an argument, and store the result of that method in a variable, assuming it's got an appropriate return type?

standard/portability-issues.md
```diff
@@ -10,14 +10,14 @@ This annex collects some information about portability that appears in this spec
 
 The behavior is undefined in the following circumstances:
 
-1. The behavior of the enclosing async function when an awaiter’s implementation of the interface methods `INotifyCompletion.OnCompleted` and `ICriticalNotifyCompletion.UnsafeOnCompleted` does not cause the resumption delegate to be invoked at most once ([§11.8.8.4](expressions.md#11884-run-time-evaluation-of-await-expressions)).
+1. The behavior of the enclosing async function when an awaiter's implementation of the interface methods `INotifyCompletion.OnCompleted` and `ICriticalNotifyCompletion.UnsafeOnCompleted` does not cause the resumption delegate to be invoked at most once ([§11.8.8.4](expressions.md#11884-run-time-evaluation-of-await-expressions)).
```
jskeet: Not sure why this has been changed... but IIRC we have tooling to "recurl" quotes anyway, right?

standard/portability-issues.md
```diff
@@ -10,14 +10,14 @@ This annex collects some information about portability that appears in this spec
...
 1. Passing pointers as `ref` or `out` parameters ([§22.3](unsafe-code.md#223-pointer-types)).
 1. When dereferencing the result of converting one pointer type to another and the resulting pointer is not correctly aligned for the pointed-to type. ([§22.5.1](unsafe-code.md#2251-general)).
 1. When the unary `*` operator is applied to a pointer containing an invalid value ([§22.6.2](unsafe-code.md#2262-pointer-indirection)).
 1. When a pointer is subscripted to access an out-of-bounds element ([§22.6.4](unsafe-code.md#2264-pointer-element-access)).
 1. Modifying objects of managed type through fixed pointers ([§22.7](unsafe-code.md#227-the-fixed-statement)).
-1. The content of memory newly allocated by `stackalloc` ([§22.9](unsafe-code.md#229-stack-allocation)).
-1. Attempting to allocate a negative number of items using `stackalloc`([§22.9](unsafe-code.md#229-stack-allocation)).
+1. The content of memory newly allocated by `stackalloc` (§stack-allocation).
```
jskeet, line 19: These will be fixed automatically, right?

standard/unsafe-code.md
```diff
@@ -1079,8 +1073,26 @@ All stack-allocated memory blocks created during the execution of a function mem
 >
 > a `stackalloc` initializer is used in the `IntToString` method to allocate a buffer of 16 characters on the stack. The buffer is automatically discarded when the method returns.
 >
+> Note, however, that `IntToString` can be rewritten in safe mode; that is, without using pointers, as follows:
+>
+> ```csharp
+> public static string IntToString(int value)
+> {
+>     if (value == int.MinValue) return "-2147483648";    // this value has no positive equivalent
```
jskeet: Can we put this in braces?

standard/unsafe-code.md
```diff
@@ -1079,8 +1073,26 @@ All stack-allocated memory blocks created during the execution of a function mem
...
+>     int n = value >= 0 ? value : -value;
+>     Span<char> buffer = stackalloc char[16];
+>     int idx = 16;
+>     do
+>     {
+>         buffer[--idx] = (char)(n % 10 + '0');
+>         n /= 10;
+>     } while (n != 0);
+>     if (value < 0) buffer[--idx] = '-';
```
jskeet: And this.

standard/unsafe-code.md
```diff
@@ -1079,8 +1073,26 @@ All stack-allocated memory blocks created during the execution of a function mem
...
+>     return buffer.Slice(idx).ToString();
+> }
+>
+> ```
+>
 > *end example*
 
-Except for the `stackalloc` operator, C# provides no predefined constructs for managing non-garbage collected memory. Such services are typically provided by supporting class libraries or imported directly from the underlying operating system.
```
jskeet, line 1116: Do we definitely want to remove this?

standard/expressions.md
KalleOlaviNiemitalo: Yes, that will allocate 100,000 ints before it frees any. Even if `stackalloc` is in a loop, the memory is freed only when the function returns.

ECMA-335 §III.3.47 "localloc – allocate space in the local dynamic memory pool":

> When the current method returns, the local memory pool is available for reuse.

There is an analyzer. [CA2014: Do not use stackalloc in loops](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2014?view=vs-2022)

standard/expressions.md
BillWagner: I'll recommend a second PR. The same issue exists in [§16.7](https://learn.microsoft.com/dotnet/csharp/language-reference/language-specification/arrays#167-array-initializers) Array initializers. That clause also states that the array element type can be inferred, but doesn't say how.

Non normative language: My first tests indicate that the compiler determines the natural type of each element, and then determines a common type that each element's natural type can be converted to. Note that (for non-stack alloc arrays), `new object()` would ensure the best type is `object[]`.

standard/expressions.md
BillWagner: Yes, this is legal:

```csharp
async Task<int> M4()
{
    int r = M5(stackalloc int[3]);
    await Task.Delay(1000);
    return 5;
}

int M5(Span<int> arg)
{
    return arg.Length;
}
```

BillWagner (commented):
responding to feedback.

standard/expressions.md
jskeet: Yup, I'm fine to defer that.

standard/expressions.md
jskeet: Perhaps (in a follow-up PR) it would be worth given that as an example. Otherwise it looks somewhat useless...

standard/expressions.md
jskeet: Thanks for the clarification :)

standard/unsafe-code.md
jskeet: Ah, it's moved - that's fine.

standard/portability-issues.md
BillWagner: Yes, we do.

standard/portability-issues.md
BillWagner: Yes

standard/expressions.md
```diff
@@ -1220,6 +1220,7 @@ primary_no_array_creation_expression
     | anonymous_method_expression
     | pointer_member_access     // unsafe code support
     | pointer_element_access    // unsafe code support
+    | stackalloc_initializer
```
Nigel-Ecma: *stackalloc_initalizer* should not be added here as it is not a general expression. In *__Stack allocation__* below it is stated it is a *local_variable_initializer* and in *__13.6.2 Local variable declarations__* it has been added to this production.

```suggestion
```

standard/expressions.md
Nigel-Ecma: Is it legal?

According to text a *stackalloc_initializer* is only valid as a *local_variable_initializer*. It is not a general expression in a similar way that *array_initializer* is not, both are included in the grammar in the places they are allowed.

My other comment *removes* it from *primary_no_array_creation_expression* precisely for this reason. (If it does need to be an expression then we need to check for any grammar gotchas as there would be two routes to it from *local_variable_initializer*)

Nigel-Ecma (changes requested):
There is some contradiction in the text as to where a *stackalloc_initializer* is valid. I've resolved this one way – following the approach used for *array_initializer*, that is not making it an expression and only including it in the places it is allowed (which is just one for *stackalloc_initializer*). We need to agree that my resolution is the correct one and gives the intended semantics.

standard/expressions.md
```diff
@@ -2989,6 +2990,77 @@ A *default_value_expression* is a constant expression ([§12.23](expressions.md#
 - one of the following value types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,`; or
 - any enumeration type.
 
+### §stack-allocation Stack allocation
```
BillWagner, line 3057: "call stack" needs to be defined in this section.

standard/statements.md
```diff
@@ -319,7 +319,7 @@ local_variable_declarator
 local_variable_initializer
     : expression
     | array_initializer
-    | stackalloc_initializer    // unsafe code support
+    | stackalloc_initializer
```
jskeet: Remove line 326, as it's no longer only in unsafe code.

standard/unsafe-code.md
```diff
@@ -1110,8 +1104,32 @@ All stack-allocated memory blocks created during the execution of a function mem
 >
 > a `stackalloc` initializer is used in the `IntToString` method to allocate a buffer of 16 characters on the stack. The buffer is automatically discarded when the method returns.
```
jskeet: This will need to be "`stackalloc` expression"

BillWagner:
Discussion on 5/17 meeting:

- `stackalloc` is an expression, with an optional initializer
- search for all uses of `stackalloc` to remove any remaining references to unsafe code.
- define stack in this PR
- the result of a stackalloc expression must obey the ref safety rules

standard/unsafe-code.md
jskeet: Actually, this is optional as in this particular case it *is* an initializer.

standard/expressions.md
BillWagner: Decision from 5/17: the stackalloc_initializer should stay here because it is a general expression. It should be removed from 13.6.2.

standard/expressions.md
BillWagner: Decision from 5/17 meeting: `stackalloc` is an expression. It should not be restricted to a *local_variable_initializer*.

PR #239, opened 2021-03-15
C# 7.x: accessing fixed fields without pinning

RexJaeschke:
While I think my solution allows the newer behavior, it might be too permissive.

See 'The main “challenge discussion in https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.3/indexing-movable-fixed-fields.md.

### DIFF
--- a/standard/unsafe-code.md
+++ b/standard/unsafe-code.md
@@ -962,7 +962,6 @@ In a member access of the form `E.I`, if `E` is of a struct type and a member lo
 
 - If the expression `E.I` does not occur in an unsafe context, a compile-time error occurs.
 - If `E` is classified as a value, a compile-time error occurs.
-- Otherwise, if `E` is a moveable variable ([§22.4](unsafe-code.md#224-fixed-and-moveable-variables)) and the expression `E.I` is not a *fixed_pointer_initializer* ([§22.7](unsafe-code.md#227-the-fixed-statement)), a compile-time error occurs.
 - Otherwise, `E` references a fixed variable and the result of the expression is a pointer to the first element of the fixed-size buffer member `I` in `E`. The result is of type `S*`, where S is the element type of `I`, and is classified as a value.
 
 The subsequent elements of the fixed-size buffer can be accessed using pointer operations from the first element. Unlike access to arrays, access to the elements of a fixed-size buffer is an unsafe operation and is not range checked.
--- a/standard/unsafe-code.md
+++ b/standard/unsafe-code.md
@@ -954,20 +954,20 @@ A fixed-size buffer declaration that declares multiple fixed-size buffers is equ
 
 Member lookup ([§11.5](expressions.md#115-member-lookup)) of a fixed-size buffer member proceeds exactly like member lookup of a field.
 
-A fixed-size buffer can be referenced in an expression using a *simple_name* ([§11.7.4](expressions.md#1174-simple-names)) or a *member_access* ([§11.7.6](expressions.md#1176-member-access)).
+A fixed-size buffer can be referenced in an expression using a *simple_name* ([§11.7.4](expressions.md#1174-simple-names)), a *member_access* ([§11.7.6](expressions.md#1176-member-access)), or an *element_access* (§11.7.10).
 
 When a fixed-size buffer member is referenced as a simple name, the effect is the same as a member access of the form `this.I`, where `I` is the fixed-size buffer member.
 
-In a member access of the form `E.I`, if `E` is of a struct type and a member lookup of `I` in that struct type identifies a fixed-size member, then `E.I` is evaluated an classified as follows:
+In a member access of the form `E.I`, if `E` is of a struct type and a member lookup of `I` in that struct type identifies a fixed-size member, then `E.I` is evaluated and classified as follows:
 
 - If the expression `E.I` does not occur in an unsafe context, a compile-time error occurs.
 - If `E` is classified as a value, a compile-time error occurs.
-- Otherwise, if `E` is a moveable variable ([§22.4](unsafe-code.md#224-fixed-and-moveable-variables)), the expression `E.I` must be one of the following, or a compile-time error occurs:
-  - A *fixed_pointer_initializer* ([§22.7](unsafe-code.md#227-the-fixed-statement))
-  - An *pointer_element_access* (§22.6.4) expression where `I` is *fixed-size buffer* (§22.8).
+- Otherwise, if `E` is a moveable variable ([§22.4](unsafe-code.md#224-fixed-and-moveable-variables)) and the expression `E.I` is not a *fixed_pointer_initializer* ([§22.7](unsafe-code.md#227-the-fixed-statement)), a compile-time error occurs.
 - Otherwise, `E` references a fixed variable and the result of the expression is a pointer to the first element of the fixed-size buffer member `I` in `E`. The result is of type `S*`, where S is the element type of `I`, and is classified as a value.
 
 The subsequent elements of the fixed-size buffer can be accessed using pointer operations from the first element. Unlike access to arrays, access to the elements of a fixed-size buffer is an unsafe operation and is not range checked.
+
+In an element access of the form `E[I]`, if `E` is a fixed-sized buffer, and `I` is implicitly convertible to `int`, then `E[I]` is evaluated exactly as the *pointer_element_access* (§22.6.4), `P[I]`, where `P` is the address of the fixed-sized buffer, `E`.
 
 > *Example*: The following declares and uses a struct with a fixed-size buffer member.
 >
--- a/standard/unsafe-code.md
+++ b/standard/unsafe-code.md
@@ -964,7 +964,7 @@ In a member access of the form `E.I` where `E.` may be the implicit `this.`, if 
 - If `E` is classified as a value, a compile-time error occurs.
 - Otherwise, if `E` is a moveable variable (§22.4) then:
   - If the expression `E.I` is a *fixed_pointer_initializer* (§22.7), then the result of the expression is a pointer to the first element of the fixed size buffer member `I` in `E`.
-  - Otherwise if the expression `E.I` is a *primary_no_array_creation_expression* within an *element_access* expression of the form `E.I[J]` then then the result of the expression is the result of the *primary_no_array_creation_expression* (§11.7.10.1).
+  - Otherwise if the expression `E.I` is a *primary_no_array_creation_expression* (§11.7.10.1) within an *element_access* (§11.7.10) of the form `E.I[J]`, then the result of `E.I` is a pointer, `P`, to the first element of the fixed size buffer member `I` in `E`, and the enclosing *element_access* is then evaluated as the *pointer_element_access* (§22.6.4) `P[J]`.
   - Otherwise a compile-time error occurs.
 - Otherwise, `E` references a fixed variable and the result of the expression is a pointer to the first element of the fixed-size buffer member `I` in `E`. The result is of type `S*`, where S is the element type of `I`, and is classified as a value.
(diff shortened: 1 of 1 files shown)


BillWagner:
converted to draft: C# 7 feature.

jskeet:
This isn't an area I'm particularly expert in, but it does look like this is too permissive.

In particular, I think it would imply that the second example in the [feature doc](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.3/indexing-movable-fixed-fields.md) should be valid - and it's not. The feature only adds the ability for the *first* example to become valid without a `fixed` statement.

I honestly don't know how we'd allow the indexing part at the moment. One for discussion, certainly.

KalleOlaviNiemitalo:
Current compiler behaviour ([SharpLab](https://sharplab.io/#v2:CYLg1APgrgdgzgQwGYFMAEcAuAnKBjTNAYTQG8BYAKDRrSQEsAPFYNemQgIyiQG0BOALoBuKrTQABACxoAsgAoAlGTHjaAfTQBeNJgAW9OADpufAGwi0AeitoA8gGlaqtWk075+wyZ6LeF4WtbFGxsAHtsABo0GBQWODomFgxMBEwUAFsUDhc0AF8qPKA===)):

```csharp
unsafe struct C {
    fixed int buf[9];
    void M() {
        _ = this.buf[6]; // OK   
        _ = (this.buf)[6]; // error, needs fixed statement
    }
}
```

So this seems to require a new rule that specifically allows the syntax for indexing a fixed-size buffer.

KalleOlaviNiemitalo:
Fixed-size buffers cannot be nested

```csharp
unsafe struct C {
    public fixed int buf[9];
}

unsafe struct D {
    fixed C buf2[5]; // error, C not a fixed-size primitive type
    void M() {
        _ = (this.buf2[1].buf)[3];
    }
}
```

so the new rule need not state that the movability of `this.buf2[1].buf` depends on the movability of `this.buf2[1]` because such a construct cannot be defined anyway.

BillWagner:
For discussion in the meeting. I think this works. (I can't make a suggestion on the PR, because the only change is a deletion):

Instead of deleting the single line, modify it as follows:

- Otherwise, if `E` is a moveable variable ([§22.4](unsafe-code.md#224-fixed-and-moveable-variables)), the expression `E.I` must be one of the following, or a compile-time error occurs:
    - A *fixed_pointer_initializer* ([§22.7](unsafe-code.md#227-the-fixed-statement))
    - An *element_access* (§11.710) expression where `I` is *fixed-size buffer* (§22.8).

KalleOlaviNiemitalo:
@BillWagner problems with that wording:

- `E.I` does not match the _element_access_ syntax defined in §11.7.10.1. Instead you could say `E.I` is the _primary_no_array_creation_expression_ in _element_access_. It would be even better to reference the _pointer_element_access_ syntax in §22.6.4.
- The last list item "Otherwise, `E` references a fixed variable and the result of the expression is a pointer […]" is not reached so you need to specify the semantics of the `E.I[E0]` expression somewhere else. The type conversion of E0 apparently matches §22.6.4 but I'm not sure whether fixed-size buffer element access should be described as involving a pointer type.

BillWagner:
Update based on above comment:

For discussion in the meeting. I think this works. (I can't make a suggestion on the PR, because the only change is a deletion):

Instead of deleting the single line, modify it as follows:

- Otherwise, if `E` is a moveable variable ([§22.4](unsafe-code.md#224-fixed-and-moveable-variables)), the expression `E.I` must be one of the following, or a compile-time error occurs:
    - A *fixed_pointer_initializer* ([§22.7](unsafe-code.md#227-the-fixed-statement))
    - An *pointer_element_access* (§22.6.4) expression where `I` is *fixed-size buffer* (§22.8).
-  Otherwise, `E` references a fixed variable and the result of the expression is a pointer to the first element of the fixed-size buffer member `I` in `E`. The result is of type `S*`, where S is the element type of `I`, and is classified as a value.

Note that the last bullet item is in the spec now. I added it to the comment because I thought its presence addresses @KalleOlaviNiemitalo 's second bullet point above. (Namely, if `E` isn't a moveable variable, this bullet applies).

BillWagner:
ping @jskeet for Meeting agenda

jskeet:
@BillWagner: Not quite sure what you mean - it's got the "meeting: discuss" label, so I'd expect us to just find it that way. Anything else you wanted me to do?

BillWagner:
> @BillWagner: Not quite sure what you mean - it's got the "meeting: discuss" label, so I'd expect us to just find it that way. Anything else you wanted me to do?

Nope. Just hoping we can resolve this one today.

BillWagner:
Results from committee meeting:

We think the latest change is close, but may need refinement.

@MadsTorgersen and @jskeet tagged for review

@KalleOlaviNiemitalo Can you add a comment on the confused emjoi above?

KalleOlaviNiemitalo:
@BillWagner, `E.I` (where `E` is an expression and `I` is an identifier) cannot match the _pointer_element_access_ syntax either, although it can match the _primary_no_array_creation_expression_ part of that. So the change in the proposed wording does not address the first bullet point in <https://github.com/dotnet/csharpstandard/pull/239#issuecomment-1411533158>.

[§22.6.4 Pointer element access](https://github.com/dotnet/csharpstandard/blob/02d7b266318ad760a7a2d88f8dcb7638f3af9a84/standard/unsafe-code.md#2264-pointer-element-access) defines the syntax like so:

```ANTLR
pointer_element_access
    : primary_no_array_creation_expression '[' expression ']'
    ;
```

i.e. it would have to end with ']', which `E.I` cannot do.

BillWagner:
Ping @jskeet @MadsTorgersen 

I've updated this to restore the previous language for member access, and add new text for *element_access* (`E[I]`), that follows the rules for *pointer_element_access*. This should be ready for a final review and (hopefully) merging.

standard/unsafe-code.md
```diff
@@ -967,6 +967,8 @@ In a member access of the form `E.I`, if `E` is of a struct type and a member lo
 
 The subsequent elements of the fixed-size buffer can be accessed using pointer operations from the first element. Unlike access to arrays, access to the elements of a fixed-size buffer is an unsafe operation and is not range checked.
 
+In an element access of the form `E[I]`, if `E` is a fixed-sized buffer, and `I` is implicitly convertible to `int`, then `E[I]` is evaluated exactly as the *pointer_element_access* (§22.6.4), `P[I]`, where `P` is the address of the fixed-sized buffer, `E`.
```
jskeet: I thought I'd added a review comment earlier, but apparently not. Do we need some of the same restrictions as in member access, but for element access, e.g. `E` not being a value, and occurring in an unsafe context?

BillWagner: I don't think so. If `E` is a *fixed-size buffer*, it can't be a value. Also, fixed size buffers can only occur in unsafe code. 

I think it would be redundant.

Nigel-Ecma: I *think* this is in the wrong place, the change is restricted to expressions of the form `E.I[J]` – compare the two examples in the proposal document.

I suggest where this special case/feature needs to be added is in the third bullet above (line 965) which becomes something following the *outline*:

- Otherwise, if `E` is a moveable variable ([§22.4](unsafe-code.md#224-fixed-and-moveable-variables)) then:
  - if the expression `E.I` is a *fixed_pointer_initializer* ([§22.7](unsafe-code.md#227-the-fixed-statement)) everything is good
  - otherwise if the expression `E.I` is a *primary_no_array_creation_expression* within an *element_access* expression of the form `E.I[J] then yada yada;
  - otherwise a compile-time error occurs.

I said it was an *outline*!

Does this capture this feature?

Nigel-Ecma (changes requested):
I think this is too permissive as is, it should only capture the case of `E.I[J]`.

standard/unsafe-code.md
Nigel-Ecma: Probably should have said that in `E.I[J]` the `E.` can of course be the implicit `this.`

BillWagner: Great suggestion @Nigel-Ecma 

I updated based on your comments.

standard/unsafe-code.md
```diff
@@ -958,17 +958,18 @@ A fixed-size buffer can be referenced in an expression using a *simple_name* ([
...
+In a member access of the form `E.I` where `E.` may be the implicit `this.`, if `E` is of a struct type and a member lookup of `I` in that struct type identifies a fixed-size member, then `E.I` is evaluated and classified as follows:
 
 - If the expression `E.I` does not occur in an unsafe context, a compile-time error occurs.
 - If `E` is classified as a value, a compile-time error occurs.
-- Otherwise, if `E` is a moveable variable ([§22.4](unsafe-code.md#224-fixed-and-moveable-variables)) and the expression `E.I` is not a *fixed_pointer_initializer* ([§22.7](unsafe-code.md#227-the-fixed-statement)), a compile-time error occurs.
+- Otherwise, if `E` is a moveable variable (§22.4) then:
+  - If the expression `E.I` is a *fixed_pointer_initializer* (§22.7), then the result of the expression is a pointer to the first element of the fixed size buffer member `I` in `E`.
+  - Otherwise if the expression `E.I` is a *primary_no_array_creation_expression* within an *element_access* expression of the form `E.I[J]` then then the result of the expression is the result of the *primary_no_array_creation_expression* (§11.7.10.1).
```
Nigel-Ecma: Maybe a bit more expansion of my "yada, yada" as there is nothing now about being evaluated like a *pointer_element_access* – or was this equivalence more  illustrative?

If not then maybe:

> - Otherwise if the expression `E.I` is a *primary_no_array_creation_expression* (§11.7.10.1) within an *element_access* (§11.7.10) of the form `E.I[J]`, then the result of `E.I` is a pointer, `P`, to the first element of the fixed size buffer member `I` in `E`, and the enclosing *element_access* is then evaluated as the *pointer_element_access* (§22.6.4) `P[J]`.

BillWagner: Yes! It is a *primary_no_array_creation_expression* that is a *pointer_element_access*.

BillWagner:
**Current status as of March 20**:

I think I've addressed all remaining comments from Nigel and Jon. I didn't resolve the conversation from Jon above. It would be good to have additional opinions here.

jskeet (approved):
Just to be clear, this is two new features:

- `x.Buffer[index]`
- `&x.Buffer`

... neither of which would have been valid before? Assuming that's the case, this is fine.

BillWagner:
> Just to be clear, this is two new features:
> 
> * `x.Buffer[index]`
> * `&x.Buffer`
> 
> ... neither of which would have been valid before? Assuming that's the case, this is fine.

Only the first is valid without `fixed`. The second returns a pointer, so requires fixed.  The two features that now don't require pinning are:

```
var n = x.Buffer[index];
x.Buffer[inces] = n + 2; // Using the element in the buffer as a variable
```

jskeet:
> > Just to be clear, this is two new features:
> > 
> > * `x.Buffer[index]`
> > * `&x.Buffer`
> > 
> > ... neither of which would have been valid before? Assuming that's the case, this is fine.
> 
> Only the first is valid without `fixed`. The second returns a pointer, so requires fixed.

Okay - so leaving aside pinning, was the second valid before now? I'm only mentioning it because it looks like it's new text.

BillWagner:
> Okay - so leaving aside pinning, was the second valid before now? I'm only mentioning it because it looks like it's new text.

Yes, that's correct.

MadsTorgersen (approved):
Looks good!

Nigel-Ecma (approved):
LGTM

PR #240, opened 2021-03-16
C# 7.x: Support for Pattern-Based `fixed` Statements

### LINKED ISSUE #241: v7.x Q's. re "Pattern-Based `fixed` Statements"
I have some questions before I can complete Draft PR https://github.com/ECMA-TC49-TG2/csharpstandard/pull/240.

## Q1. `GetPinnableReference` returning a `ref readonly` variable

In the PR, the example I provided for the user-defined class `C` has the following member:

```csharp
public ref int GetPinnableReference() => ref _value;
```

However, if appropriate, the return type can also contain `readonly`, as follows:

```csharp
public ref readonly int GetPinnableReference() => ref _value;
```

Given the following `fixed` statement (which works for both cases above)

```csharp
fixed (int* p = c1) { ... }
```

how can the readonly-ness of the `ref int` returned in the second case be preserved? As far as I can tell, it can't, and attempts to store through `*p` are allowed! What am I missing?

BTW, `String.GetPinnableReference` returns `ref **readonly** char`.

## Q2. The Empty Case

In the existing text for handling arrays, strings, and such, mention is made of what happens if a string is `null` or empty, an array is `null` or empty, and such. Do we need to say anything about this possibility in our new, general, user-defined type case?

I experimented with returning `null` from a class like `C`, but `null` can't be returned by ref. Then I stumbled on the class type `System.Runtime.CompilerServices.Unsafe` (which allows one to do magical things with refs) and its method `NullRef<T>`. However, I'm guessing that we *don't* want to mention this type/method. How might one return a null-like indicator?

---
[Associated WorkItem - 52134](https://dev.azure.com/msft-skilling/Content/_workitems/edit/52134)

RexJaeschke:
See feature spec here: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.3/pattern-based-fixed.md

Fixes #241


standard/unsafe-code.md
```diff
@@ -681,6 +681,7 @@ It is an error to use a captured local variable ([§11.16.6.2](expressions.md#11
...
 - An expression of type `string`, provided the type `char*` is implicitly convertible to the pointer type given in the `fixed` statement. In this case, the initializer computes the address of the first character in the string, and the entire string is guaranteed to remain at a fixed address for the duration of the `fixed` statement. The behavior of the `fixed` statement is implementation-defined if the string expression is `null`.
+- An expression of type other than *array_type* or `string`, provided the type implements an accessible method named `GetPinnableReference`, that method has no parameters, that method returns a `ref` or `ref readonly` variable of an *unmanaged_type* `T`, and `T*` is implicitly convertible to the pointer type given in the `fixed` statement. In this case, the initializer computes the address of the returned variable, and that variable is guaranteed to remain at a fixed address for the duration of the `fixed` statement.
```
KalleOlaviNiemitalo: How about `GetPinnableReference` extension methods? The wording "the type implements" doesn't seem to cover those. [SharpLab](https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgZgATIEzoMLoG8BYAKHXMwzADth0BJAblIF9TS1MBGANkxwCihUhUrc+AJwCmAM3Q06AcSnAACjWoBDAEYQpAJVlTp1AMZSAFMAAWYAM550pgJToAvAD500uaYB0TCIUQeScyLzoAK7UdpoyUpgALOgAshb4LiHCZKIUMmAAHlIw6BYKAFToAA7uTq4EbDnkjY1AA=)

I'm surprised that additional parameters with default values, or with `params`, are not allowed; but that seems to match what the compiler does.

BillWagner:
See questions in #241

BillWagner:
@jskeet 

This one is small. I believe it's now ready for review.

standard/unsafe-code.md
Nigel-Ecma: I'm missing something here, the original proposal lists under drawbacks:

> `GetPinnableReference` is intended to be used only in `fixed`, but nothing prevents its use in safe code, so implementor must keep that in mind.

So why isn't it declared with the `unsafe` modifier?

standard/unsafe-code.md
```diff
@@ -704,6 +704,7 @@ It is an error to use a captured local variable ([§11.17.6.2](expressions.md#11
...
+- An expression of type other than *array_type* or `string`, provided the type there exists an accessible method or accessible extension method named `GetPinnableReference`, that method has no parameters, that method returns a `ref` or `ref readonly` variable of an *unmanaged_type* `T`, and `T*` is implicitly convertible to the pointer type given in the `fixed` statement. The `GetPinnableReference`. In this case, the initializer computes the address of the returned variable, and that variable is guaranteed to remain at a fixed address for the duration of the `fixed` statement. A `GetPinnableReference()` method can be used by the `fixed` statement when overload resolution (§11.6.4) produces exactly one function member and that function member satisfies the preceding conditions. The `GetPinnableReference` method should return a reference to an address equal to zero, such as that returned from `System.Runtime.CompilerServices.Unsafe.NullRef<T>()` when there is no data to pin.
```
Nigel-Ecma: Looks like this has been mangled:
- "provided the type there exists an accessible method"
- "The `GetPinnableReference`."

Include the signature `ref [readonly] T GetPinnableReference()` as in the proposal rather than just describe it?

Nigel-Ecma (changes requested):
Given my first thought was "where is the `unsafe` modifier" I might be way off on this one!

Some of the text appears to have been mangled, see specific comment, hence the "request changes".

standard/unsafe-code.md
BillWagner: Fixed.

standard/unsafe-code.md
BillWagner: I checked with the compiler team on this comment. 

It's not meant to provide any normative statement. `GetPinnableReference` may not do anything `unsafe`, so doesn't need to be marked `unsafe`. The fact that a caller can misuse the API isn't a language issue.

We could put a warning in a note if you think it's wise.

Nigel-Ecma: So we *could* add `unsafe` to avoid the "drawback" but that would be an "abuse" of the modifier... hmmm... I think we just leave it as is.

standard/unsafe-code.md
```diff
@@ -864,6 +865,32 @@ A `char*` value produced by fixing a string instance always points to a null-ter
...
+> *Example*: The following code shows a *fixed_pointer_initializer* with an expression of type other than *array_type* or `string`:
+>
+> ```csharp
+> public class C
+> {
+>     private int _value;
+>     public C(int value) => _value = value;
+>     public ref int GetPinnableReference() => ref _value;
+> }
+> public class Test
```
jskeet, line 881: Nit: blank line between classes.

standard/unsafe-code.md
```diff
@@ -864,6 +865,32 @@ A `char*` value produced by fixing a string instance always points to a null-ter
...
+>         C c = new C(10);
+>         fixed (int* p = c)
+>         {
+>             // ...
+>         }
+>     }
+> }
+> ```
+>
+> Type `C` has an accessible `GetPinnableReference` method with the correct signature. In the `fixed` statement, the `ref int` returned from that method when it is called on `c` is used to initialize the `int*` pointer `p`. *end example*
```
jskeet, line 894: Is this example actually safe though? Is it doing pinning via the magic of ref? (I can't remember the terminology.)
If this is functional but not useful, that's fine - if the example is actively dangerous (because the object could be moved) then we should call that out.

jskeet (approved):
Can I assume Nigel's requested change was done? I don't see what the request is now.

BillWagner:
> Can I assume Nigel's requested change was done? I don't see what the request is now.

Yes

standard/unsafe-code.md
BillWagner: Adding @jcouv for verification.

My understanding is that it is safe (not actively dangerous). The code *does* pin the memory for `c`, which makes accessing `C._value` safe.

jskeet: SGTM. (It might be worth discussing whether we have or need any text in the standard that describes the impact of returning a field by reference.)

BillWagner: > SGTM. (It might be worth discussing whether we have or need any text in the standard that describes the impact of returning a field by reference.)

We will. That's the subject of #213 (I'll be looking at updating it tomorrow)

jskeet (approved):

Nigel-Ecma (approved):
LGTM

standard/standard-library.md
```diff
@@ -355,6 +355,11 @@ namespace System.Runtime.CompilerServices
     {
         public IndexerNameAttribute(String indexerName);
     }
+
+    public static class Unsafe
+    {
+        public static T NullRef<T>();
```
KalleOlaviNiemitalo, line 361: NullRef\<T\>() should return `ref T` rather than plain `T`.

Nigel-Ecma: Can you please add a PR for this?

KalleOlaviNiemitalo: No — I have not made any pull requests in the dotnet organisation and don't intend to start now.

jskeet: I can pick this up early next week.

Nigel-Ecma: @KalleOlaviNiemitalo – No problem, I've added one. Thanks for spotting this.

PR #244, opened 2021-03-17
C# 7.x: enhanced generic constraints

RexJaeschke:
Add support for the following kinds of primary constraints: `unmanaged`, `System.Enum`, and `System.Delegate`.


BillWagner:
converting to draft: C# 7 feature

standard/classes.md
```diff
@@ -412,23 +413,31 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
 > *Note*: The `System.Nullable<T>` type specifies the non-nullable value type constraint for `T`. Thus, recursively constructed types of the forms `T??` and `Nullable<Nullable<T>>` are prohibited. *end note*
 
-Pointer types are never allowed to be type arguments and are not considered to satisfy either the reference type or value type constraints.
+The token `unmanaged` is neither a keyword nor a contextual keyword. When it is encountered, it will either:
+- Bind to a type named `unmanaged`
+- Bind to no type, in which case, it is interpreted as the unmanaged type constraint.
+
+The unmanaged type constraint specifies that a type argument used for the type parameter shall be a non-nullable unmanaged type.
+
+Pointer types are never allowed to be type arguments and are not considered to satisfy either the `reference` type or `value` type constraints.
```
KalleOlaviNiemitalo: `` `reference` `` and `` `value` `` markup seems inappropriate here because, unlike `unmanaged`, those words do not have specific meanings in C# source code.

standard/classes.md
```diff
@@ -412,23 +413,31 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
 
 The list of constraints given in a `where` clause can include any of the following components, in this order: a single primary constraint, one or more secondary constraints, and the constructor constraint, `new()`.
 
-A primary constraint can be a class type or the ***reference type constraint*** `class` or the ***value type constraint*** `struct`. A secondary constraint can be a *type_parameter* or *interface_type*.
+A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, or the ***unmanaged type constraint*** `unmanaged`.
```
BillWagner:
```suggestion
A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, or the ***unmanaged type constraint*** `unmanaged`.
```

standard/classes.md
```diff
@@ -412,23 +413,31 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
 
 The value type constraint specifies that a type argument used for the type parameter shall be a non-nullable value type. All non-nullable struct types, enum types, and type parameters having the value type constraint satisfy this constraint. Note that although classified as a value type, a nullable value type ([§8.3.11](types.md#8311-nullable-value-types)) does not satisfy the value type constraint. A type parameter having the value type constraint shall not also have the *constructor_constraint*, although it may be used as a type argument for another type parameter with a *constructor_constraint*.
 
 > *Note*: The `System.Nullable<T>` type specifies the non-nullable value type constraint for `T`. Thus, recursively constructed types of the forms `T??` and `Nullable<Nullable<T>>` are prohibited. *end note*
 
-Pointer types are never allowed to be type arguments and are not considered to satisfy either the reference type or value type constraints.
+The token `unmanaged` is neither a keyword nor a contextual keyword. When it is encountered, it will either:
```
BillWagner:
```suggestion
The token `unmanaged` is neither a keyword nor a contextual keyword. When it is encountered, it will either:

```

standard/classes.md
```diff
@@ -412,23 +413,32 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
```
Nigel-Ecma: This might be inconsistent with the approach taken for `nameof` (§11.7.20) which is a contextual keyword (§6.4.4) and hence an *identifier* (§6.4.3) – this might stem from different ways the MS docs and the Standard spec contextual keywords and identifiers.

If we follow the approach for `nameof` then `unmanaged` would be added as an alternative to `contextual_keyword` and lines 426-429 could be replaced by a paragraph echoing the one for resolving `nameof` in §11.7.20, e.g. (written by copying the `nameof` para and editing it!):

> Because `unmanaged` is not a keyword, in *primary_constraint* the unmanaged constraint is always syntactically ambiguous with *class_type*. For compatibility reasons, if a name lookup (§11.7.4) of the name `unmanaged` succeeds it is treated as a `class_type`. Otherwise it is treated as the unmanaged constraint.

standard/classes.md
Nigel-Ecma: I think those are both typos, also doesn't `unmanaged` need to be added? So we get:

> Pointer types are never allowed to be type arguments and are not considered to satisfy either the `class` type, `struct`, or `unmanaged` type constraints.

standard/classes.md
```diff
@@ -412,23 +413,32 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
+Pointer types are never allowed to be type arguments and are not considered to satisfy either the `reference` type or `value` type constraints.
 
 If a constraint is a class type, an interface type, or a type parameter, that type specifies a minimal “base type” that every type argument used for that type parameter shall support. Whenever a constructed type or generic method is used, the type argument is checked against the constraints on the type parameter at compile-time. The type argument supplied shall satisfy the conditions described in [§8.4.5](types.md#845-satisfying-constraints).
 
 A *class_type* constraint shall satisfy the following rules:
 
 - The type shall be a class type.
 - The type shall not be `sealed`.
-- The type shall not be one of the following types: `System.Array`, `System.Delegate`, `System.Enum`, or `System.ValueType`.
+- The type shall not be one of the following types: `System.Array` or `System.ValueType`.
```
Nigel-Ecma, line 459: Do we highlight anywhere that `System.Delegate` and `System.Enum` are now allowed (as of v7.3)? Should we? (It seems a notable change from v<7.3)

Nigel-Ecma (changes requested):
A few things to look at, the consistency in approach/description being a key one.

standard/classes.md
jskeet: Agreed - this is also similar to `var` (and `dynamic`, although I now can't see where that's actually specified).

It definitely sounds like the description in https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/lexical-structure.md#644-keywords - if there's a reason why it's *not* a contextual keyword, I'd like to hear it and we can include that in the standard.

standard/classes.md
jskeet: I agree that it should say `class` and `struct`, but as pointers are unmanaged types, Nigel's suggestion reads a little oddly. How about something like:

> Pointer types are never allowed to be type arguments, and don't satisfy any type constraints, even `unmanaged`, despite being unmanaged types.

standard/classes.md
jskeet: It would be good to include it in a list of changes since v6, but I don't think the relaxation warrants any other highlighting.

standard/classes.md
```diff
@@ -431,23 +432,32 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
 
 > *Note*: The `System.Nullable<T>` type specifies the non-nullable value type constraint for `T`. Thus, recursively constructed types of the forms `T??` and `Nullable<Nullable<T>>` are prohibited. *end note*
 
-Pointer types are never allowed to be type arguments and are not considered to satisfy either the reference type or value type constraints.
+The token `unmanaged` is neither a keyword nor a contextual keyword. When it is encountered, it will either:
+
+- Bind to a type named `unmanaged`
+- Bind to no type, in which case, it is interpreted as the unmanaged type constraint.
+
+The unmanaged type constraint specifies that a type argument used for the type parameter shall be a non-nullable unmanaged type.
```
```suggestion
The unmanaged type constraint specifies that a type argument used for the type parameter shall be a non-nullable unmanaged type ([§8.8](types.md#88-unmanaged-types)).
```

jskeet (commented):
Another one that could do with discussion at the meeting, I think.

standard/classes.md
Nigel-Ecma: As contextual keywords are just a subset of identifiers I can see no reason why `unmanaged` should not be one.

standard/classes.md
```diff
@@ -431,23 +432,32 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
 
 The value type constraint specifies that a type argument used for the type parameter shall be a non-nullable value type. All non-nullable struct types, enum types, and type parameters having the value type constraint satisfy this constraint. Note that although classified as a value type, a nullable value type ([§8.3.11](types.md#8311-nullable-value-types)) does not satisfy the value type constraint. A type parameter having the value type constraint shall not also have the *constructor_constraint*, although it may be used as a type argument for another type parameter with a *constructor_constraint*.
 
 > *Note*: The `System.Nullable<T>` type specifies the non-nullable value type constraint for `T`. Thus, recursively constructed types of the forms `T??` and `Nullable<Nullable<T>>` are prohibited. *end note*
 
-Pointer types are never allowed to be type arguments and are not considered to satisfy either the reference type or value type constraints.
+The token `unmanaged` is neither a keyword nor a contextual keyword. When it is encountered, it will either:
```
Nigel-Ecma:
```suggestion
Because `unmanaged` is not a keyword, in *primary_constraint* the unmanaged constraint is always syntactically ambiguous with *class_type*. For compatibility reasons, if a name lookup (§11.7.4) of the name `unmanaged` succeeds it is treated as a `class_type`. Otherwise it is treated as the unmanaged constraint.
```

The following suggestions to delete lines 446-448 are part of this, but are not included in the review for some reason...

standard/classes.md
```diff
@@ -431,23 +432,32 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
+
```
Nigel-Ecma, line 448:
```suggestion
```

standard/classes.md
```diff
@@ -431,23 +432,32 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
+- Bind to a type named `unmanaged`
```
Nigel-Ecma:
```suggestion
```

standard/classes.md
```diff
@@ -431,23 +432,32 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
+- Bind to no type, in which case, it is interpreted as the unmanaged type constraint.
```
Nigel-Ecma:
```suggestion
```

standard/classes.md
```diff
@@ -485,6 +495,8 @@ If the `where` clause for a type parameter includes a constructor constraint (wh
 
 It is a compile-time error for *type_parameter_constraints* having a *primary_constraint* of `struct` to also have a *constructor_constraint*.
```
Nigel-Ecma: The following added line is a bit repetitious, maybe:

```suggestion
It is a compile-time error for *type_parameter_constraints* having a *primary_constraint* of `struct` or `unmanaged` to also have a *constructor_constraint*.
```

and deleted the added line?

Nigel-Ecma (changes requested):
I've thumbed up Jon's suggestions, made my previous suggestion into a formal one (which took me 4 suggestions, couldn't figure out how to do a single multiline one), and added a new one to avoid repetition.

If the re-wording suggestion is adopted then `unamanaged` needs to be added to `contextual_keyword` in §6.4.4 Keywords

standard/classes.md
RexJaeschke: On the TG2 call of 2023-02-01, we agreed with this approach.

standard/classes.md
```diff
@@ -431,23 +432,32 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
 
-Pointer types are never allowed to be type arguments and are not considered to satisfy either the reference type or value type constraints.
+The token `unmanaged` is neither a keyword nor a contextual keyword. When it is encountered, it will either:
+
+- Bind to a type named `unmanaged`
+- Bind to no type, in which case, it is interpreted as the unmanaged type constraint.
+
+The unmanaged type constraint specifies that a type argument used for the type parameter shall be a non-nullable unmanaged type.
+
+Pointer types are never allowed to be type arguments and are not considered to satisfy either the `reference` type or `value` type constraints.
```
BillWagner:
```suggestion
Pointer types are never allowed to be type arguments, and don't satisfy any type constraints, even unmanaged, despite being unmanaged types.
```

standard/classes.md
RexJaeschke: In the TG2 call, we agreed to Nigel's 4-part change.

standard/classes.md
RexJaeschke: On the TG2 call of 2023-02-01, Rex will add this to the list of "what's new in V7."

standard/classes.md
RexJaeschke: ON the TG2 call of 2023-02-01, we agreed to this.

BillWagner (approved):
Add one set of typos to fix.

standard/classes.md
```diff
@@ -483,7 +490,9 @@ It is valid for `S` to have the value type constraint and `T` to have the refer
 
 If the `where` clause for a type parameter includes a constructor constraint (which has the form `new()`), it is possible to use the `new` operator to create instances of the type ([§11.7.15.2](expressions.md#117152-object-creation-expressions)). Any type argument used for a type parameter with a constructor constraint shall be a value type, a non-abstract class having a public parameterless constructor, or a type parameter having the value type constraint or constructor constraint.
 
-It is a compile-time error for *type_parameter_constraints* having a *primary_constraint* of `struct` to also have a *constructor_constraint*.
+It is a compile-time error for *type_parameter_constraints* having a *primary_constraint* of `struct` or `unmanaged` to also have a *constructor_constraint*.
+
+It is a compile-time error for *type_parameter_constraints* having a *primary_constraint* of `unmanaged` to also have a *constructor_constraint*.
```
BillWagner: delete these:

```suggestion
```

BillWagner:
@Nigel-Ecma Can you approve the changes on this PR? I think Rex and I have addressed all your comments, or accepted your suggesitons.

standard/classes.md
RexJaeschke: Done!

RexJaeschke:
Per Nigel's request, add `unmanaged` to `contextual_keyword` in 6.4.4.

Nigel-Ecma (approved):
looks good :-)

PR #250, opened 2021-04-01
nameof expressions alternative

jskeet:
This is an alternative to #10, which builds on that.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1122,6 +1122,7 @@ primary_no_array_creation_expression
     | checked_expression
     | unchecked_expression
     | default_value_expression
+    | nameof_expression    
     | anonymous_method_expression
     ;
 ```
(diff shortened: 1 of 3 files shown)


Nigel-Ecma (approved):
LGTM, drops some questionable (to me) bits from the MS source :-)

standard/expressions.md
```diff
@@ -2314,7 +2315,88 @@ If the *type* in a *default_value_expression* evaluates at run-time to a referen
...
+
+> *Example*: The following illustrates the results of various `nameof` expressions, assuming a generic type `List<T>` declared within the `System.Collections.Generic` namespace:
+> ```csharp
+> using System.Collections.Generic;
+> 
+> using TestAlias = System.String;
+> 
+> class Program {
+>    static void Main() {
+>       var point = (x: 3, y: 4);
```
KalleOlaviNiemitalo, line 2363: Tuple types are not part of C# 6.

jskeet: Well spotted, thanks! I'll leave this in *right now*, but before merging I'll add a commit to remove it, then create a PR to re-add it for C# 7. (We should potentially show the use of `point.Item1` as well.)

PR #251, opened 2021-04-05
Feature: An alternative approach to handling "null conditional" ops

Nigel-Ecma:
Following on from the discussion on #7 at the last meeting this is a draft PR adding the null conditional operations into the grammar at the same level as their respective unconditional siblings.

Some of this is guesswork/invention...

On the invention side in particular note that three terminals: `?.`, `?(` and `?[`; have been added to *Operator_Or_Punctuator*.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/csharpstandard?pullRequest=251) <br/>All CLA requirements met.

standard/expressions.md
```diff
@@ -1275,6 +1278,36 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
...
+
+A *null_conditional_member_access* consists of a *primary_expression* followed by a "`?.`" token, followed by an *Identifier*, optionally followed by a *type_argument_list*.
+
+```ANTLR
+null_conditional_member_access
+    : primary_expression '?.' Identifier type_argument_list?
+    ;
+```
+
+A  *null_conditional_member_access* expression `E` is of the form `P?.I`. Let `T` be the type of the *member_access*  expression `P.I`. The meaning of `E` is determined as follows:
```
jskeet: Super-nit - is there a double space after `*member_access*`?

standard/expressions.md
```diff
@@ -1275,6 +1278,36 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
...
+
+- If `T` is a type parameter that is not known to be a reference type or a non-nullable value type, a compile-time error occurs.
+- If `T` is a non-nullable value type, then the type of `E` is `T?`, and the meaning of `E` is the same as the meaning of:
+	> ```csharp
+	> ((object)P == null) ? (T?)null : P.I
+	> ```
+  Except that `P` is evaluated only once.
+- Otherwise the type of `E` is `T`, and the meaning of `E` is the same as the meaning of:
```
jskeet, line 1318: Any reason the indentation here isn't the same as in the lines above?

standard/expressions.md
```diff
@@ -1459,6 +1492,45 @@ The run-time processing of a delegate invocation of the form `D(A)`, where `D`
 
 See [§20.6](delegates.md#206-delegate-invocation) for details of multiple invocation lists without parameters.
 
+### §null-conditional-invocation-expression Null Conditional Invocation Expression
```
jskeet, line 1516: Is this actually implemented? I've never seen it before. Is the expectation that it should make something like the code below compile?

```csharp
Action action = null;
action?();
```

That *doesn't* compile, whereas `action?.Invoke();` does.

Perhaps this was a part of the feature proposal that was dropped but not removed from the text?

standard/expressions.md
```diff
@@ -1275,6 +1278,36 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
...
+    : primary_expression '?.' Identifier type_argument_list?
+    ;
+```
+
+A  *null_conditional_member_access* expression `E` is of the form `P?.I`. Let `T` be the type of the *member_access*  expression `P.I`. The meaning of `E` is determined as follows:
+
+- If `T` is a type parameter that is not known to be a reference type or a non-nullable value type, a compile-time error occurs.
+- If `T` is a non-nullable value type, then the type of `E` is `T?`, and the meaning of `E` is the same as the meaning of:
+	> ```csharp
+	> ((object)P == null) ? (T?)null : P.I
```
jskeet: Minor note for the future - I'd love to replace the `((object)P == null)` part with `P is null`. We can only do that when pattern matching is implemented, but I think it would be good to do so when we can.

I'm tempted to file a reminder issue... any thoughts?

standard/lexical-structure.md
```diff
@@ -755,7 +755,7 @@ Operator_Or_Punctuator
     | '+'  | '-'  | '*'  | '/'  | '%'   | '&'  | '|'  | '^'  | '!'  | '~'
     | '='  | '<'  | '>'  | '?'  | '??'  | '::' | '++' | '--' | '&&' | '||'
     | '->' | '==' | '!=' | '<=' | '>='  | '+=' | '-=' | '*=' | '/=' | '%='
-    | '&=' | '|=' | '^=' | '<<' | '<<=' | '=>'
+    | '&=' | '|=' | '^=' | '<<' | '<<=' | '=>' | '?.' | '?[' | '?('
```
jskeet: If my suspicions are correct, we'll be able to remove `?(`.

standard/expressions.md
jskeet: Note: it's possible that we need a *null_conditional_invocation_expression* as well as *null_conditional_member_access* in order for `action?.Invoke()` (or any other sort of invocation via a null conditional operator) to work - I'm not immediately clear whether that counts as a *null_conditional_member_access* where the result is a method group that is then invoked. (Brain not quite working right now.)

standard/statements.md
```diff
@@ -361,6 +361,7 @@ expression_statement
 
 statement_expression
     : invocation_expression
+    | null_conditional_invocation_expression
```
jskeet: Again, we may be able to remove this... I'm not sure. (I suspect I need to look at invocation_expression more closely.)

jskeet (commented):
I definitely prefer this approach over the one in #7 - thanks very much for the work here, Nigel.

As you can tell, I'm slightly confused about how null conditional invocation expressions work - see comments for details.

standard/expressions.md
```diff
@@ -1275,6 +1278,36 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
 > ```
 > Within the `A` class, those occurrences of the Color identifier that reference the Color type are delimited by `**`, and those that reference the Color field are not. *end example*
 
+### §null-conditional-member-access Null Conditional Member Access
+
+A *null_conditional_member_access* consists of a *primary_expression* followed by a "`?.`" token, followed by an *Identifier*, optionally followed by a *type_argument_list*.
```
gafter: I don't believe `?.` is one token.  See https://sharplab.io/#v2:EYLgtghgzgLgpgJwDQxASwDZICYgNQA+AAgEwCMAsAFBEDMABKfQML0De19XjDRALPQCyACiJkADPSgBKdp24KAbhAT0M9ALxSA/PQB0AGTgA7AOYwAFgG55XAL7U7QA

standard/expressions.md
```diff
@@ -1459,6 +1492,45 @@ The run-time processing of a delegate invocation of the form `D(A)`, where `D`
 
 See [§20.6](delegates.md#206-delegate-invocation) for details of multiple invocation lists without parameters.
 
+### §null-conditional-invocation-expression Null Conditional Invocation Expression
+
+A *null_conditional_invocation_expression* consists of a *primary_expression* followed by a "`?(`" token, followed by an optional *argument_list*, followed by a "`)`" token.
```
gafter: The `?` and `(` are separate tokens.

standard/expressions.md
Nigel-Ecma: I didn't originally include as I was using MS's C#6 Language Reference as the informal spec to work form, then I noticed it in PR #7 so added it assuming I'd missed it elsewhere. Following your query I did I wider search and found a reference on SO that it doesn't exist due to too many parsing issues. I'll remove it.

standard/expressions.md
gafter: But I think it isn't implemented because of an ambiguity with the `?:` expression.

standard/expressions.md
Nigel-Ecma: Not a super-nit if the MD parser doesn't remove them. I'll take a look an remove any I find.

jskeet:
Nigel to have a look at whether this PR's approach makes sense, after we had more discussion about why the grammar in #7 was expressed that way.

standard/expressions.md
Nigel-Ecma: Do you mean indenting the code within the bullet? That seems correct to me, does the Standard not do this? Or are you referring to something else?

standard/expressions.md
Nigel-Ecma: I think it is worth noting regardless of whether we go with PR #7 or this one.

standard/lexical-structure.md
Nigel-Ecma: Agree

standard/statements.md
Nigel-Ecma: Agree

standard/expressions.md
Nigel-Ecma: Hmmm... Is there a good reason to allow spaces or is this an artefact of the way the compiler added the token? The token `??` already exists (as do other multi-symbol tokens where a prefix is also a token), does the compiler allow a space? Historically there was the change to break up right shift (assignment) due to the conflict when generics were introduced, but there is no such conflict between `?`, `??`, `?.` and `?[` is there?

If it is technically a breaking change to make it a single token I really hope it doesn't effect anyone (we might never know of course as they could be too embarrassed to admit it!)

I'd encourage we go for the single token, but its not a hill I'll die on! What do folks think? (The question applies to both PR #7 and this one).

standard/expressions.md
Nigel-Ecma: I'll "resolve conversation" this one as it's dealt with by the preceding thread (both operators should be handled the same way!)

Nigel-Ecma:
> Nigel to have a look at whether this PR's approach makes sense, after we had more discussion about why the grammar in #7 was expressed that way.

Before I look into this properly can I just confirm my understanding of our meeting conversation:

- Both PR #7 and #251 accept the same language
- They *may* (that is part of the question to resolve) have *equivalent* semantics.
- PR #7 *may* allow constructs which are disallowed for the non-conditional sibling operators, which *might* be an issue (part of the question to resolve)
- PR #7 produces a parse tree which is easier/more logical/<insert> for the compiler writer
- PR #251 aims at producing a description more logical to the programmer, if such is possible (part of the question to resolve)

Note: In the second bullet I don't say the "same" as #251 might be described as "null propagating" where a null-conditional operator supplies a null to another null-conditional operator, etc., which by-description results in multiple "is this null" tests but any compiler would likely short circuit. So the semantics might not the same but be equivalent – just a different way of describing the same result.

If that summary is either correct, or once it is corrected, I'll look into whether the approach in #7 is required and if not which of #7 & #251 might be more desirable.

standard/expressions.md
jskeet: I suspect the output may be fine in each case, but it's a bit odd to see the inconsistency. Compare lines 1295 and 1300... I can't see why they'd have different numbers of spaces before the `>`

Nigel-Ecma:
## Withdrawing Draft PR

To handle null conditional operators correctly the operators `.`, `?.`, `[]` and `?[]` need to combine similarly to the way the operators `&`, `&&`, `|` and `||` do. The approach in this draft does not achieve that as it lacks the priority stratification enforced by the productions for these existing operators.

Unfortunately this still leaves the concerns with PR #7, including the more recently noted differences in what is permitted on the LHS of `?[]` compared to `[]` and whether they are significant.

Are they addressable? More work to do...

Nigel-Ecma:
The grammar has been extended so that null-conditional access operations "capture" any immediately following unconditional access operations. This should match the desired semantics, keep the description of these operators in the same part of the grammar as their unconditional siblings. Changes to statement expressions are not included as they feel like a wart that may have a better solution, but they could follow the same as #7 if there is no solution found. @MadsTorgersen is this the intended semantics?

standard/expressions.md
```diff
@@ -1279,11 +1279,16 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
 
 ### §null-conditional-member-access Null Conditional Member Access
 
-A *null_conditional_member_access* consists of a *primary_expression* followed by a "`?.`" token, followed by an *Identifier*, optionally followed by a *type_argument_list*.
+A *null_conditional_member_access* consists of a *primary_expression* followed by a "`?.`" token, followed by an *Identifier* with an optional *type_argument_list*, followed by zero or more *unconditional_access_part*s.
 
 ```ANTLR
 null_conditional_member_access
-    : primary_expression '?.' Identifier type_argument_list?
+    : primary_expression '?.' Identifier type_argument_list? unconditional_access_part*
```
jskeet: "unconditional_access_part" sounds like something that will be evaluated unconditionally, and I don't *think* that's what we're looking for, is it?

jskeet (commented):
I'm not attempting to assess whether this is equivalent to #7 or not. I still prefer this approach to organizing the feature over the approach in #7 if we can satisfy ourselves that it's equivalent.

standard/expressions.md
Nigel-Ecma: I agree its not a great name, but probably better than my other thought of `null_faulting_access`, over to the wordsmiths!

standard/expressions.md
```diff
@@ -1275,6 +1277,42 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
 > ```
 > Within the `A` class, those occurrences of the Color identifier that reference the Color type are delimited by `**`, and those that reference the Color field are not. *end example*
 
+### §null-conditional-member-access Null Conditional Member Access
+
+A *null_conditional_member_access* consists of a *primary_expression* followed by a "`?.`" token, followed by an *Identifier* with an optional *type_argument_list*, followed by zero or more *unconditional_access_part*s.
```
gafter: I believe `?.` is not a token.  It is a pair of tokens.  You could put a space between them in a program.

standard/expressions.md
gafter: `?.` is a pair of tokens.

standard/expressions.md
```diff
@@ -1513,6 +1551,36 @@ The binding-time processing of an indexer access of the form `P[A]`, where `P` i
 
 Depending on the context in which it is used, an indexer access causes invocation of either the *get_accessor* or the *set_accessor* of the indexer. If the indexer access is the target of an assignment, the *set_accessor* is invoked to assign a new value ([§12.18.2](expressions.md#12182-simple-assignment)). In all other cases, the *get_accessor* is invoked to obtain the current value ([§12.2.2](expressions.md#1222-values-of-expressions)).
 
+### §null-conditional-element-access Null Conditional Element Access
+
+A *null_conditional_element_access* consists of a *primary_no_array_creation_expression* followed by a "`?[`" token, followed by an *argument_list*, followed by a "`]`" token, followed by zero or more *unconditional_access_part*s.
```
gafter: `?[` is a pair of tokens.

Nigel-Ecma:
A summary of the key differences & changes compared to the original MS:

- The null conditional operators were added to *unary_expression* rather than alongside their non-conditional siblings in *primary_expression* in the MS approach. Here they are added to *primary_expression*.
- The MS approach captured all member/element accesses *including* conditional ones which follow an initial null conditional operation. Here only non-conditional ones are captured, this simplifies the description of the semantics. The overall semantics remains the same.
  - As part of the simplification a *null_conditional_invocation_expression* is introduced which syntactically is a struct subset of *null_conditional_member/element_access* – those that end in an invocation.
  - The *null_conditional_member/element_access* descriptions have the semantic restriction that the expressions must **not** be of type `void`/classified as nothing and reference *null_conditional_invocation_expression* for those that are `void`. This separates out the semantics for value producing expressions, defined as having the same meaning as a if/then/else, and non-value producing ones, defined using an if/then, into different clauses rather than a longer set of nested bullets. (The meanings follow the same code as in previous approach.)
  - *Note:* In doing this PR I noticed that in  the Standard there is sometimes reference to use of something in a "statement expression" and in others "statement expression or lambda body". I wonder whether some/all of the former should be the latter... I haven't checked, but I have included *null_conditional_invocation_expression* in *anonymous_method_body*.
- Also added a *null_conditional_projection_initializer*, a subset of *null-conditional-member-access", which does not include capturing adjacent accesses/invocations. As the name suggests it is a projection initializer for anonymous objects.

To test the grammar is parsing as expected I use a simple test file, `nullConditional.cs`. *(Note: the grammar does take advantage of ANTLR's top-to-bottom rule ordering so the above mentioned subset productions get picked over the full set productions.)* This file along with the graphic output of TestRig/grun as `nullConditional.png` is attached as a ZIP (sorry, Github barfed on the PNG and suggested I ZIP it). *(Note: the grammar fed to ANTLR has undergone mutual left recursion removal in *primary_expression* so you might spot "missing" nodes in the PNG due to this.)*

This is hopefully #251 pretty much done, please comment & compare with #7.

[nullConditional.zip](https://github.com/dotnet/csharpstandard/files/6567702/nullConditional.zip)

standard/expressions.md
jskeet: Filed #328. Resolving.

standard/expressions.md
```diff
@@ -1280,6 +1282,50 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
 > ```
 > Within the `A` class, those occurrences of the Color identifier that reference the Color type are delimited by `**`, and those that reference the Color field are not. *end example*
 
+### §null-conditional-member-access Null Conditional Member Access
+
+A *null_conditional_member_access* is a conditional relation of *member_access* ([§12.7.5](expressions.md#member-access)) and  it is a binding time error if the result type is `void`. For a null conditional expression where the result type may be `void` see ([](expressions.md#null-conditional-invocation-expression)).
```
jskeet: What does "conditional relation" mean here? (Is it broadly "conditional equivalent to"?) Not objecting to the term - just checking whether it's expected to have a precise technical meaning, or whether it's just descriptive.

standard/expressions.md
```diff
@@ -1280,6 +1282,50 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
...
+
+A *null_conditional_member_access* consists of a *primary_expression* followed by the two tokens "`?`" and "`.`", followed by an *Identifier* with an optional *type_argument_list*, followed by zero or more *captured_access*es.
+
+```ANTLR
+null_conditional_member_access
+    : primary_expression '?' '.' Identifier type_argument_list? captured_access*
+    ;
+    
+captured_access
```
jskeet: Is "captured access" a new concept here? If so, might we want to *consider* naming it something else to avoid it being confused with variable capture within anonymous functions?

standard/expressions.md
```diff
@@ -1280,6 +1282,50 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
...
+   > ```
+  Except that `P` is evaluated only once.
+
+*Note*: A *null_conditional_member_access*  is left-associative ([§12.4.2](expressions.md#1242-operator-precedence-and-associativity)) so in an expression of the form:
+> ```csharp
+> P?.I₀?.I₁
+> ```
+then if `P` evaluates to `null` neither `I₀` or `I₁` are evaluated. The same is true if an expression is a sequence of *null_conditional_member_access* or *null_conditional_element_access* §null-conditional-element-access operations. *end note*
+
+A *null_conditional_projection_initializer* is a subset of *null_conditional_member_access* and has the same semantics. It only occurs as a projection initializer in an anonymous object creation expression ([12.7.11.7](expressions.md#anonymous_object_creation_expression)).
```
jskeet: Describing one thing (that isn't a set) as a subset of another thing doesn't quite sit right. I think I know what you mean... let's see if we can come up with something better in the meeting (or you can convince me that this is fine, of course).

standard/expressions.md
```diff
@@ -1464,6 +1510,25 @@ The run-time processing of a delegate invocation of the form `D(A)`, where `D`
 
 See [§20.6](delegates.md#206-delegate-invocation) for details of multiple invocation lists without parameters.
 
+### §null-conditional-invocation Null Conditional Invocation Expression
+
+A *null_conditional_invocation_expression* is syntactically either a *null_conditional_member_access* ([§12.7.6](expressions.md#null_conditional_member_access)) or *null_conditional_element_access* ([§12.7.6](expressions.md#null_conditional_element_access)) where the final *captured_access* is an invocation expression ([§12.7.6](expressions.md#inovation_expression)).
+
+Unlike the syntactically equivalent *null_conditional_member_access* or *null_conditional_element_access* a *null_conditional_invocation_expression* may be classified as nothing and only occurs as part of a *statement_expression* or *anonymous_method_body*.
```
```suggestion
Unlike the syntactically equivalent *null_conditional_member_access* or *null_conditional_element_access*, a *null_conditional_invocation_expression* may be classified as nothing and only occurs as part of a *statement_expression* or *anonymous_method_body*.
```

(This is just adding a comma for clarity - we've got three things in a row, and I think it's helpful to emphasize that the first two belong together, and the third is separate.)

standard/expressions.md
```diff
@@ -1464,6 +1510,25 @@ The run-time processing of a delegate invocation of the form `D(A)`, where `D`
...
+
+```ANTLR
+null_conditional_invocation_expression
+    : null_conditional_member_access '(' argument_list? ')'
+    | null_conditional_element_access '(' argument_list? ')'
+    ;
+```
+
+A  *null_conditional_invocation_expression* expression `E` is of the form `P?I`; where `I` is the remainder of the syntactically equivalent *null_conditional_member_access* or *null_conditional_element_access*, `I` will therefore start with `.` or `[`. Let `PI` signify the concatention of `P` and `I`. The meaning of `E` is the same as the meaning of:
```
jskeet: I wonder whether something other than `I` would be useful here... when I see `I` in an example, I expect it to be an identifier, rather than something beginning with `.` or `[`. Perhaps `A` for access?

standard/expressions.md
```diff
@@ -1464,6 +1510,25 @@ The run-time processing of a delegate invocation of the form `D(A)`, where `D`
...
+> ```csharp
+> if ((object)P == null) ? PI
```
jskeet: Unless I'm mis-reading it, we don't want the `?` here.

standard/expressions.md
```diff
@@ -1518,6 +1583,39 @@ The binding-time processing of an indexer access of the form `P[A]`, where `P` i
...
+A *null_conditional_element_access* consists of a *primary_no_array_creation_expression* followed by the two tokens "`?`" and "`[`", followed by an *argument_list*, followed by a "`]`" token, followed by zero or more *captured_access*es.
+
+```ANTLR
+null_conditional_element_access
+    : primary_no_array_creation_expression '?' '[' argument_list ']' captured_access*
+    ;
+```
+
+A *null_conditional_element_access* is a conditional relation of *element_access* ([§12.7.7](expressions.md#element-access)) and it is a binding time error if the result type is `void`. For a null conditional expression where the result type may be `void` see ([](expressions.md#null-conditional-invocation-expression)).
+
```
jskeet, line 1617: Remove one of these blank lines?

jskeet (commented):
A few more thoughts, but I definitely like this approach. Will discuss later.

standard/expressions.md
Nigel-Ecma: I thought the connection between *null_conditional_member_access* & *member_access*, and similarly for the element access pair, was worth noting. That left the wording... I was going to write "sibling" but for some reason ended up with "relation". It is purely descriptive and I welcome better wordsmithing!

standard/expressions.md
Nigel-Ecma: This is a new concept, the language doesn't really have any other operators which behave quite in the way the null -conditional ones do – they "capture" following operations (#7 captures more than #251, it is one of the differences, but that isn't fundamental *per se* and both have to find a way to describe the behaviour). After some iterations I came up with "captured access", I didn't consider the comparison with "variable capture"...

"[null] conditional access" didn't seem right as the accesses themselves are not conditional [and they will fault on null], they are just the subject of another conditional – as the explanation of the semantics in terms of an equivalent conditional (`?`/':` or `if`) shows. Captured, Controlled, Effected, Impacted, Subject? I just noticed  I went with the first alphabetically (pure happenstance) ;-)

Another case where wordsmithing input is welcome!

standard/expressions.md
BillWagner: Here's how the tool will recognize the link to the new clause you've added:

```suggestion
A *null_conditional_member_access* is a conditional relation of *member_access* ([§12.7.5](expressions.md#member-access)) and  it is a binding time error if the result type is `void`. For a null conditional expression where the result type may be `void` see (§null-conditional-invocation).
```

standard/expressions.md
Nigel-Ecma: Changed.

standard/expressions.md
Nigel-Ecma: Changed

standard/expressions.md
Nigel-Ecma: Changed

standard/expressions.md
Nigel-Ecma: Changed

standard/expressions.md
Nigel-Ecma: @jskeet – You can think of rules as defining a set, but how about:

```suggestion
A *null_conditional_projection_initializer* is a restriction of *null_conditional_member_access* and has the same semantics. It only occurs as a projection initializer in an anonymous object creation expression ([12.7.11.7](expressions.md#anonymous_object_creation_expression)).
```

or "a restricted version of"? "Restrict" as the latter allows everything of the former and some more.

standard/expressions.md
Nigel-Ecma: It was probably down to my tab setting be 3 and bullet indentation just being 2. I pushed them back 1 space, no changed in output

standard/expressions.md
Nigel-Ecma: Changed to "conditional version", its not exactly "equivalent" (due to the capturing) but I can see "relation" might be confusing. I'll resolve this for now but do come back if version isn't suitable.

standard/expressions.md
Nigel-Ecma: Changed to A

standard/expressions.md
Nigel-Ecma: Bad copy'n'paste, fixed.

standard/expressions.md
Nigel-Ecma: Done

standard/expressions.md
jskeet: I definitely prefer restriction over subset, yes.

I think I'm mentally distinguishing between "the rule" and "the set of text sequences that match the rule" - the latter is defined by the former, but it's not the same thing to me. At least the way I'm thinking today, which isn't necessarily how I'll be thinking tomorrow :)

Nigel-Ecma: Restriction it is.

jskeet:
Assigned to @MadsTorgersen as his "second item" to focus on.

standard/expressions.md
gafter: *nested_access* ?

standard/expressions.md
```diff
@@ -1280,6 +1282,50 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
...
+  > ((object)P == null) ? (T?)null : P.A
+  > ```
+  Except that `P` is evaluated only once.
+- Otherwise the type of `E` is `T`, and the meaning of `E` is the same as the meaning of:
+  > ```csharp
+  > ((object)P == null) ? null : P.A
+  > ```
+  Except that `P` is evaluated only once.
+
+*Note*: A *null_conditional_member_access*  is left-associative ([§12.4.2](expressions.md#1242-operator-precedence-and-associativity)) so in an expression of the form:
```
gafter: The same would be true if it were right associative, right?  I suggest dropping the comment on associativity and just get to the point:
```suggestion
*Note*: In an expression of the form:
```

standard/expressions.md
```diff
@@ -1464,6 +1510,29 @@ The run-time processing of a delegate invocation of the form `D(A)`, where `D`
 
 See [§20.6](delegates.md#206-delegate-invocation) for details of multiple invocation lists without parameters.
 
+### §null-conditional-invocation-expression Null Conditional Invocation Expression
+
+A *null_conditional_invocation_expression* is syntactically either a *null_conditional_member_access* (§null-conditional-member-access) or *null_conditional_element_access* (§null-conditional-element-access) where the final *captured_access* is an invocation expression ([§12.7.6](expressions.md#1276-invocation-expressions)).
+
+Unlike the syntactically equivalent *null_conditional_member_access* or *null_conditional_element_access*, a *null_conditional_invocation_expression* may be classified as nothing and only occurs as part of a *statement_expression* ([§13.7](statements.md#137-expression-statements)) or *anonymous_function_body* ([§12.16.1](expressions.md#12161-general)).
```
gafter:
```suggestion
Unlike the syntactically equivalent *null_conditional_member_access* or *null_conditional_element_access*, a *null_conditional_invocation_expression* may be classified as nothing when it occurs as a *statement_expression* ([§13.7](statements.md#137-expression-statements)) or *anonymous_function_body* ([§12.16.1](expressions.md#12161-general)).
```

I don't think this is complete.  It is also permitted as the body of an expression-bodied method whose return type is `void`.  Alternatively, perhaps it is overspecified as that is a *statement_expression*, but that is also true for an *anonymous_function_body* when the return type is `void`.

standard/expressions.md
```diff
@@ -1518,6 +1587,38 @@ The binding-time processing of an indexer access of the form `P[A]`, where `P` i
...
+
+A *null_conditional_element_access* consists of a *primary_no_array_creation_expression* followed by the two tokens "`?`" and "`[`", followed by an *argument_list*, followed by a "`]`" token, followed by zero or more *captured_access*es.
+
+```ANTLR
+null_conditional_element_access
+    : primary_no_array_creation_expression '?' '[' argument_list ']' captured_access*
+    ;
+```
+
+A *null_conditional_element_access* is a conditional version of *element_access* ([§12.7.7](expressions.md#1277-element-access)) and it is a binding time error if the result type is `void`. For a null conditional expression where the result type may be `void` see (§null-conditional-invocation-expression).
```
gafter, line 1616: What forbids this?

```csharp
    void M2(Q q) => q?.M()[1];
```

gafter (dismissed):

standard/expressions.md
Nigel-Ecma: OK

standard/expressions.md
Nigel-Ecma: I don't think they’re *nested* – if anything isn't it the *enclosed*/*first* which is impacting the *enclosing*/*following* so nested sounds sort of backwards to me? So while *captured* isn’t great I think its a smidgen better. But is there something better? Words are hard, give me numbers any day ;-)

standard/expressions.md
Nigel-Ecma: You are right, expression-bodied method has been missed.

As currently written the grammar doesn't allow a *null_conditional_invocation_expression* within an expression-bodied method. (However both #7 and this one syntactically accept their respective versions of null conditional access which end in an invocation; the restriction here in #251 that such an access cannot return void is a semantic one.)

I haven't looked at whether this can be accommodated easily in #251's current approach of splitting void handling into different rules vs. the nested bullet semantics (as in #7) – the latter are just messy and this part of #251 was an attempt to make the description easier to follow, but maybe this will turn out to show that cannot be done (at least using seperate rules).

Null conditionals are certainly proving “fun” (as in both #7 & #251 have problems 🙁)

standard/expressions.md
Nigel-Ecma: I think the answer is the type system (both #7 & #251 accept this syntactically) unless I'm misunderstanding the question?

If it was `void M2(Q q) => Console?.WriteLine(q)` then #251’s treatment of void currently rejects it semantically – but see the comments above about expression-bodied functions for that issue.

standard/expressions.md
Nigel-Ecma: @gafter – You wrote “Alternatively, perhaps it is overspecified as that is a *statement_expression*…” For a moment that seemed to provide hope, but then I considered it, doesn't an expression bodied member have an *expression* and not a *statement_expression*? In this context the latter might be nice, but I think it is/should be the former.

Looking at this I realised that expression bodied members may not be allowed for elsewhere, will add an Issue.

standard/expressions.md
jskeet: So just to be clear, a captured/nested access is one that's gated by a null conditional member access - i.e. it looks unconditional in itself, but it's conditional based on the *earlier* null conditional member access, yes?

"Implicitly conditional access"? "Effectively conditional access"? I'm not liking either of those, but if we can agree on what impression we're trying to give, that'll be a start :)

Nigel-Ecma:
Yes, a captured access is one that is “gated” as you say by a preceding one.

So in A?.B.C.D, which groups as ((A?.B).C).D (i.e. same as A.B.C.D does), the .C & .D are captured and if A is null the whole expression is null (but if B or C turn out to be null there is an exception).

“Nested” doesn’t work as C & D are on the outside, it is A?.B that is nested. In an inside-out world it would be it ;-)

Your “implicitly conditional access” is I think the better of the two, but I suspect it implies *too* much – it sounds as though A?.B.C.D is the same as A?.B?.C?.D which isn’t the case, the former will fault if B or C is null while the latter won’t.

“Captured” works I think… partly as it doesn’t try to give the detail and leaves the reader to figure it out from the grammar and prose :-)

standard/expressions.md
Nigel-Ecma: @gafter – your wording suggestion above was resolved differently by truncating the sentence after "as nothing"

jskeet:
> “Captured” works I think… partly as it doesn’t try to give the detail and leaves the reader to figure it out from the grammar and prose :-)

I'm still uncomfortable with captured given that it has a very specific (and entirely different) meaning elsewhere. I think this we should discuss in the meeting.

standard/classes.md
```diff
@@ -1622,12 +1622,20 @@ member_name
...
     | '=>' expression ';'
     | ';'
     ;
 ```
 
-*unsafe_modifier* ([§23.2](unsafe-code.md#232-unsafe-contexts)) is only available in unsafe code ([§23](unsafe-code.md#23-unsafe-code)).
+Grammar notes:
+
+- *unsafe_modifier* ([§23.2](unsafe-code.md#232-unsafe-contexts)) is only available in unsafe code ([§23](unsafe-code.md#23-unsafe-code)).
+
```
jskeet, line 1634: Unnecessary blank line here? The one before the previous bullet may well be unnecessary too.

Nigel-Ecma: Hmmm... I'll admit to thinking that the blank line was optional in this case – sometimes you need a blank line to make a break and sometimes MD will infer the break, but in the latter case I thought you could be explicit and I tend to put them in or not depending on readability of the raw MD.

Just did a quick scan and in §15.3.1 I see lists with and without spaces and at least GitHub is rendering the spacing differently. So this problem, if it is one, could be all over the place...

jskeet:
In current meeting: "dependent" has the most backing. (Somewhat reluctantly from all, but better than alternatives.)

standard/classes.md
```diff
@@ -1650,7 +1658,7 @@ The optional *formal_parameter_list* specifies the parameters of the method ([§
 
 The *return_type* and each of the types referenced in the *formal_parameter_list* of a method shall be at least as accessible as the method itself ([§8.5.5](basic-concepts.md#855-accessibility-constraints)).
 
-The *method_body* is either a semicolon, a ***block body*** or an ***expression body***. A block body consists of a *block*, which specifies the statements to execute when the method is invoked. An expression body consists of `=>` followed by an *expression* and a semicolon, and denotes a single expression to perform when the method is invoked.
+The *method_body* is either a semicolon, a ***block body*** or an ***expression body***. A block body consists of a *block*, which specifies the statements to execute when the method is invoked. An expression body consists of `=>`, followed by an *null_conditional_invocation_expression* or *expression*, followed by a semicolon, and denotes a single expression to perform when the method is invoked.
```
BillWagner: nit:

```suggestion
The *method_body* is either a semicolon, a ***block body*** or an ***expression body***. A block body consists of a *block*, which specifies the statements to execute when the method is invoked. An expression body consists of `=>`, followed by a *null_conditional_invocation_expression* or *expression*, followed by a semicolon, and denotes a single expression to perform when the method is invoked.
```

standard/expressions.md
```diff
@@ -1467,6 +1513,31 @@ The run-time processing of a delegate invocation of the form `D(A)`, where `D`
...
+```ANTLR
+null_conditional_invocation_expression
+    : null_conditional_member_access '(' argument_list? ')'
+    | null_conditional_element_access '(' argument_list? ')'
+    ;
+```
+
+A  *null_conditional_invocation_expression* expression `E` is of the form `P?A`; where `A` is the remainder of the syntactically equivalent *null_conditional_member_access* or *null_conditional_element_access*, `A` will therefore start with `.` or `[`. Let `PA` signify the concatention of `P` and `A`. When `E` occurs as a *statement_expression* the meaning of `E` is the same as the meaning of the *statement*:
+> ```csharp
+> if ((object)P == null) PA
```
BillWagner: Shouldn't this be `!=`?

```suggestion
> if ((object)P != null) PA
```

standard/expressions.md
```diff
@@ -1467,6 +1513,31 @@ The run-time processing of a delegate invocation of the form `D(A)`, where `D`
...
+> ```
+When `E` occurs as a *anonymous_function_body* or *method_body* the meaning of `E` is the same as the meaning of the *block*:
+> ```csharp
+> { if ((object)P == null) PA; }
```
BillWagner: Same as 1533:

```suggestion
> { if ((object)P != null) PA; }
```

BillWagner (approved):
This is great @Nigel-Ecma 

I think I found 2 small nits that I commented on. Once we resolve those, I think we should merge this, in favor of #7

standard/expressions.md
Nigel-Ecma: Whoops, good catch!

standard/expressions.md
Nigel-Ecma: And I copied and pasted it… :-(

standard/classes.md
jskeet: I don't think it's an actual problem, just a personal preference to make it clear that it's a single list. I'll resolve now as I don't want this to get in the way of the PR.

standard/expressions.md
```diff
@@ -1283,6 +1285,50 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
 > ```
 > Within the `A` class, those occurrences of the Color identifier that reference the Color type are delimited by `**`, and those that reference the Color field are not. *end example*
 
+### §null-conditional-member-access Null Conditional Member Access
+
+A *null_conditional_member_access* is a conditional version of *member_access* ([§12.7.5](expressions.md#1275-member-access)) and  it is a binding time error if the result type is `void`. For a null conditional expression where the result type may be `void` see (§null-conditional-invocation-expression).
```
jskeet: Nit: double space in "and it"

standard/expressions.md
```diff
@@ -1283,6 +1285,50 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
...
+
+A *null_conditional_member_access* consists of a *primary_expression* followed by the two tokens "`?`" and "`.`", followed by an *Identifier* with an optional *type_argument_list*, followed by zero or more *dependent_access*es.
+
+```ANTLR
+null_conditional_member_access
+    : primary_expression '?' '.' Identifier type_argument_list? dependent_access*
+    ;
+    
+dependent_access
```
jskeet, line 1299: Just to check, `primary_expression` will cover nested null-conditional member accesses, yes? So if we have:

```csharp
x?.M1()?.M2().M3
```

that ends up with a "top-level" null_conditional_member_access where the primary_expression is `x?.M1()?.M2()`, which is itself a null_conditional_member_access with a primary_expression of `x?.M1()`, which is itself a null_conditional_member_access with a primary expression of `x`... is that right?

standard/expressions.md
```diff
@@ -1283,6 +1285,50 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
 > ```
 > Within the `A` class, those occurrences of the Color identifier that reference the Color type are delimited by `**`, and those that reference the Color field are not. *end example*
 
+### §null-conditional-member-access Null Conditional Member Access
```
jskeet, line 1288: I've just been trying to figure out how this works in a couple of cases, and while I think it's the right approach, we may need a bit more explanation.

For example consider:

```csharp
string x = null;
int? y = x?.GetHashCode();
```

My understanding is that that is a null_conditional_member_access with a dependent_access of `()` - so we get the null value. It's *not* a null_conditional_invocation_expression, because that's only used in places where the result (if any) is discarded. Is that the case? If so, it's a bit confusing because `x.GetHashCode()` is an invocation expresson, isn't it?

I'm not suggesting that we change any of the existing text - just consider adding an example/note.

standard/expressions.md
```diff
@@ -4089,11 +4194,16 @@ implicit_anonymous_function_parameter
...
 
 anonymous_function_body
-    : expression
+    : null_conditional_invocation_expression
+    | expression
     | block
     ;
 ```
 
+When recognising an *anonymous_function_body* if both the *null_conditional_invocation_expression* and *expression* alternatives are applicable then the former shall be chosen.
```
jskeet, line 4214: I'm not sure about this one. To go back to my previous example, consider:

```csharp
Func<string, int?> func = x => x?.GetHashCode();
```

Does that satisfy the requirements of this paragraph? We'd still want to use an *expression* with a *null_conditional_member_access* rather than a *null_conditional_invocation_expression* in order to avoid "losing" the return value, wouldn't we? I may have missed something here.

jskeet (commented):
I'm a bit confused about the difference between null_conditional_invocation_expression and null_conditional_member_access, as noted in comments. I suspect this is a problem of me misunderstanding rather than in the proposal, but I'd like to just check it through during our meeting.

BillWagner:
@jskeet 

> I'm a bit confused about the difference between null_conditional_invocation_expression and null_conditional_member_access, as noted in comments. I suspect this is a problem of me misunderstanding rather than in the proposal, but I'd like to just check it through during our meeting.

I thought the distinction was meant to handle the case where an invocation expression had a `void` result. That makes the discussion of changing the result type to a nullable unnecessary. I too, could be misunderstanding this distinction.

jskeet:
@BillWagner:

> I thought the distinction was meant to handle the case where an invocation expression had a void result. That makes the discussion of changing the result type to a nullable unnecessary. I too, could be misunderstanding this distinction.

I agree with the motivation - but then the part about choosing *null_conditional_invocation_expression* over *expression* confuses me.

jskeet (approved):
Thanks for the change, Nigel. I *think* I now understand and it's all good :)

PR #258, opened 2021-04-09
Fix top-level annex links

jskeet:
The second commit is included as I ran the tooling to check that the Word converter would then work. (Normally we'd just let a new PR be generated.)


BillWagner (approved):
Thanks @jskeet 

Let's :shipit:

Issue #230, opened 2021-03-02
ANTLR Lexer Grammar Notation


RexJaeschke:
## Comments

Line-oriented comments are supported, and we should use them, as appropriate. See below for examples. (Also, my proposed resolution to integrating the unsafe code grammar uses comments.)

## Ranges and Regular-Expression Sets
 
Currently, we have

```ANTLR
Hex_Digit
    : '0' | '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9'
    | 'A' | 'B' | 'C' | 'D' | 'E' | 'F' | 'a' | 'b' | 'c' | 'd' | 'e' | 'f'
	;
```

Using the `..` range notation, this can be rewritten compactly as

```ANTLR
Hex_Digit
    : '0'..'9' | 'A'..'F' | 'a'..'f'
    ;
```

And using regular-expression set notation, this can be rewritten even more compactly as

```ANTLR
Hex_Digit
    : [0-9A-Fa-f]
    ;
```

The regular-expression approach seems attractive, but can compactness interfere with readability? Which of the following do you prefer?

```ANTLR
Input_Character
    : ~('\u000D' | '\u000A' | '\u0085' | '\u2028' | '\u2029')  // anything but New_Line_Character
    ;
```

```ANTLR
Input_Character
    : ~[\u000D\u000A\u0085\u2028\u2029]  // anything but New_Line_Character
    ;
```

In at least one case, the regular-expression form *must* be used; for example:

```ANTLR
Whitespace
    : [\p{Zs}]	// Any character with Unicode class Zs
    ...
    ;
```

The notation `\p{...}` allows Unicode character properties to be specified. (Note this is *not* covered in the ANTLR book, as this feature came later.) It can *only* occur inside a set, even though it's a set of only 1 choice!

**Question: We currently do *not* use ranges or sets. Should we? If so, what are the guiding rules for choosing either of those over the existing approach?**

## Using Escape Sequences

A terminal is written as `'...'` where, typically, characters that are non-printable or difficult/impossible to enter on a keyboard are written as Unicode values. For example, `'A'` and `'\u0041'` both represent "Latin Capital letter A," while `'\t'` and `'\u0009'` both represent "Horizontal tab."

Currently, we have

```ANTLR
New_Line_Character
    : '<Carriage return character (U+000D)>'
    | '<Line feed character (U+000A)>'
    | '<Next line character (U+0085)>'
    | '<Line separator character (U+2028)>'
[...]

PR #233, opened 2021-03-13
Consolidate unsafe-code grammar

RexJaeschke:
This PR resolves https://github.com/ECMA-TC49-TG2/csharpstandard/issues/37, Proposal 6.

PR #261, opened 2021-04-15
Resolve numerous ANTLR issues

RexJaeschke:
This PR addresses the issues raised in Issue #230 as well as Issue #37, Proposal 3. Specifically, it

- Replaces all lexer rule alternatives of the form `'<...>'` with correct content
- Uses semantic predicates to resolve numerous issues, especially w.r.t identifier spelling
- Uses ranges

@BillWagner, to support the semantic predicates, from a validation pov, we'll need to get the following text into the grammar file that we extract/generate, immediately following the current `grammar` directive in tools/validate-grammar.sh:

```ANTLR
echo "grammar CSGrammar;" > $OUTPUT_FILE
```

```ANTLR
@lexer::members {
bool IsLetterCharacter()       { return true; }
bool IsCombiningCharacter()    { return true; }
bool IsDecimalDigitCharacter() { return true; }
bool IsConnectingCharacter()   { return true; }
bool IsFormattingCharacter()   { return true; }
bool IsNotAKeyword()           { return true; }
bool IsNotTrueOrFalse()        { return true; }
}
```

Note that these are simply place-holder methods to allow validation. Clearly, they are *quite incomplete* from a lexer pov.

For now, I'm made this a DRAFT PR pending pushback on the two final uses of semantic predicates I proposed in Issue 230.

### DIFF
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -139,7 +139,7 @@ Line terminators divide the characters of a C# compilation unit into lines.
 ```ANTLR
 New_Line
     : New_Line_Character
-    | '<Carriage return character (U+000D) followed by line feed character (U+000A)>'
+    | '\u000D\u000A'    // carriage return, line feed 
     ;
 ```
 
@@ -197,15 +197,15 @@ Single_Line_Comment
     ;
 
 Input_Character
-    : '<Any Unicode character except a New_Line_Character>'
+    : ~('\u000D' | '\u000A'   | '\u0085' | '\u2028' | '\u2029')   // anything but New_Line_Character
     ;
     
 New_Line_Character
-    : '<Carriage return character (U+000D)>'
-    | '<Line feed character (U+000A)>'
-    | '<Next line character (U+0085)>'
-    | '<Line separator character (U+2028)>'
-    | '<Paragraph separator character (U+2029)>'
+    : '\u000D'  // carriage return
+    | '\u000A'  // line feed
+    | '\u0085'  // next line
+    | '\u2028'  // line separator
+    | '\u2029'  // paragraph separator
     ;
     
 Delimited_Comment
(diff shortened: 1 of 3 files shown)


BillWagner:
@RexJaeschke I think the best way to proceed is for me to make a PR against this branch that adds the changes to the grammar validator. That way, they are in one place.

Does that make sense, or do you need those changes for other work?

RexJaeschke:
@BillWagner. OK; go ahead and make the grammar extractor changes a separate PR.

BillWagner:
@RexJaeschke 

I just created #310 with the updates you mention. It's not working correctly, so we need to figure that out.

RexJaeschke:
@BillWagner, while I don't have a copy of the grammar file you extracted, my guess is these messages are coming from the pseudo-extensions to the grammar from unsafe.md, where an ellipses (...) is used to begin each grammar fragment. If we can get @Nigel-Ecma to review/merge PR #https://github.com/dotnet/csharpstandard/pull/233, all the ... lines (and their parent rules) will go away, and the grammar left in unsafe.md will all be needed and correct as is and no longer be considered as an add-on. This probably will fix the "missing ;" as well, as that is in a rule that is completely removed by that PR.

And you should change this PR (261) from Draft, and merge it as well, as a final test.

BillWagner:
@RexJaeschke #310 had a very obvious shell scripting mistake (obvious, at least, as soon as I saw it).

The results of the grammar validator on that PR is the same as this one now. Is that what you expected, or should we be expecting different diagnostics from ANTLR?

RexJaeschke:
@BillWagner I removed Draft status from this PR and assigned you as a reviewer.

If you want to test further *without* Nigel having reviewed/merged PR 233 to integrate the unsafe grammar changes, you could omit the inclusion of unsafe.md in your script.

BillWagner (approved):
This is a big improvement @RexJaeschke 

Let's merge it.

You can decide if you want to take #311 into this branch, or wait for those kinds of changes for #233

PR #262, opened 2021-04-17
C# 7.x: Allow field-targeted attributes on auto-property declarations

RexJaeschke:
Note that v6 PR #11 modifies much of 15.7.4, and in this new PR, we’re adding new text at the end of that section, which should not overlap with the v6 mods, which had not yet been merged when this new PR was created.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -3458,7 +3458,9 @@ Although the backing field is hidden, that field may have field-targeted attribu
 >     public string MySecret { get; set; }
 > }
 > ```
+>
 > results in the field-targeted attribute `NonSerialized` being applied to the compiler-generated backing field, as if the code had been written as follows:
+>
 > ```csharp
 > [Serializable]
 > public class Foo
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -3452,6 +3452,7 @@ Although the backing field is hidden, that field may have field-targeted attribu
 > *Example*: The following code
 >
 > ```csharp
+> [Serializable]
 > public class Foo
 > {
 >     [field: NonSerialized]
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -3447,6 +3447,34 @@ An auto-property may optionally have a *property_initializer*, which is applied 
 >
 > *end example*
 
+Although the backing field is hidden, that field may have field-targeted attributes applied directly to it via the automatically implemented property's *property_declaration* ([§14.7.1](classes.md#1471-general)).
+
+> *Example*: The following code
+>
+> ```csharp
+> public class Foo
+> {
+>     [field: NonSerialized]
+>     public string MySecret { get; set; }
+> }
+> ```
+> results in the field-targeted attribute `NonSerialized` being applied to the compiler-generated backing field, as if the code had been written as follows:
+> ```csharp
+> [Serializable]
+> public class Foo
+> {
+>     [NonSerialized]
+>     private string _mySecretBackingField;
+>     public string MySecret
+>     {
+>         get { return _mySecretBackingField; }
+>         set { _mySecretBackingField = value; }
+>     }
+> }
+> ```
+>
+> *end example*
+
 ### 14.7.5 Accessibility
 
 If an accessor has an *accessor_modifier*, the accessibility domain ([§7.5.3](basic-concepts.md#753-accessibility-domains)) of the accessor is determined using the declared accessibility of the *accessor_modifier*. If an accessor does not have an *accessor_modifier*, the accessibility domain of the accessor is determined from the declared accessibility of the property or indexer.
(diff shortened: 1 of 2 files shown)


standard/classes.md
```diff
@@ -2852,6 +2852,32 @@ public class ReadOnlyPoint {
...
+> public class Foo
+> {
+>     [field: NonSerialized]
+>     public string MySecret { get; set; }
+> }
+> ```
+> results in the field-targeted attribute `NonSerialized` being applied to the compiler-generated backing field, as if the code had been written as follows:
+> ```csharp
+> [Serializable]
+> public class Foo
```
KalleOlaviNiemitalo, line 3467: The earlier definition of class Foo does not have `[Serializable]`. This inconsistency might give the impression that `[field: NonSerialized]` on a property causes `[Serializable]` to be generated on the containing type.

BillWagner:
converting to draft: C# 7.x feature

jskeet:
I'll fix up the example - possibly replacing the attribute with a different one, to avoid any interpretation that this feature is linked to serialization.

standard/classes.md
jskeet: I've changed it to use `[ThreadStatic]` - does that work for you? (There aren't many common field-targeted attributes available as examples...)

jskeet:
Okay, having changed the attribute in question, I'm personally happy with this - if others are happy, we can merge it tomorrow.

standard/classes.md
KalleOlaviNiemitalo: Well no, ThreadStaticAttribute is not valid on instance fields.

KalleOlaviNiemitalo: It's OK to use `[field: NonSerialized]` in the example, but `[Serializable]` should be on both definitions of `class Foo`, as they are supposed to be equivalent.

jskeet: Whoops, yes, I'd need to make it static as well. And at that point, maybe it's adding more complexity. Okay, will go back to NonSerialized.

jskeet: Done. (Making this a much smaller extra commit, of course.)

KalleOlaviNiemitalo: Now it's OK.

KalleOlaviNiemitalo:
This needs changes in §21.3 Attribute specification as well.

Change to allow `field` on auto properties:
<https://github.com/dotnet/csharpstandard/blob/a14d97c29cf5151e1b3d991ee55293583c6b78a5/standard/attributes.md?plain=1#L278>

Change to list `field` and `property` as valid targets on auto property, and specify that `property` is the default:
<https://github.com/dotnet/csharpstandard/blob/a14d97c29cf5151e1b3d991ee55293583c6b78a5/standard/attributes.md?plain=1#L286>

jskeet:
> This needs changes in §21.3 Attribute specification as well.

Done in another commit. I've currently kept the singular form in the first part of the description, for consistency with the rest, but I'm tempted to change it to "Field-like events and automatically implemented properties can ..."

KalleOlaviNiemitalo (approved):

KalleOlaviNiemitalo:
Shall the compiler issue a warning here?

```csharp
using System;
class C {
    C() { P = 1; }

    [field: Obsolete]
    int P { get; }
}
```

<https://github.com/dotnet/csharpstandard/blob/ab777da743ee4f11d6ba359dbab13908ea7ea3b1/standard/classes.md?plain=1#L3381> <https://github.com/dotnet/csharpstandard/blob/ab777da743ee4f11d6ba359dbab13908ea7ea3b1/standard/attributes.md?plain=1#L770>

KalleOlaviNiemitalo:
Could perhaps add to §14.7.4

> Such an assignment assigns directly to the read-only backing field of the property. <ins>If the backing field is decorated with the `Obsolete` attribute ([§21.5.4](attributes.md#2154-the-obsolete-attribute)), that shall not cause the compiler to issue a warning or an error about the assignment.</ins>

or to §21.5.4

> If a program uses a type or member that is decorated with the `Obsolete` attribute, the compiler shall issue a warning or an error<ins>, unless the member is the backing field of an automatically implemented property ([§14.7.4](classes.md#1474-automatically-implemented-properties))</ins>.

jskeet:
I'm not sure that's necessary, as I'd personally argue that the property (which is the member being used) effectively *isn't* decorated with the `Obsolete` attribute. However, the technical group can discuss that in our meeting in a few hours.

jskeet:
Whoops - I understand your point now, about that specific case referring to the field.
The same is true for field-like events.

Given the latter part, the group has decided that we'll merge this PR, and raise a new issue about it to fix for both. (The reverse case is interesting too, where the property is obsolete but the field isn't.) Intuition from Mads: the code *effectively* refers to the property, and the fact that it refers to a field is "sort of" an implementation detail - so it may be that we tweak 14.7.4 a little. But we'll deal with it in the new issue (which I'll link to this one).

gafter (approved):
🐸

jskeet:
@KalleOlaviNiemitalo: On a separate note, thank you for everything you've been spotting in this PR and others. We really appreciate it.

PR #263, opened 2021-04-18
C# 7.x: Adding rules for improved overload candidates

### LINKED ISSUE #188: 11.8 Method group conversions requires error even if the conversion isn't used
Consider this little test program
```csharp
public class Program1
{
    delegate void MyAction<T>(T x);
 
    void Y(long x) { }
 
    void D(MyAction<int> o, short s) { }
    void D(MyAction<long> o, int i) { }
 
    void T()
    {
        D(Y, 1);
    }
}
```
The first argument in the call `D(Y, 1)` is a method group. During overload resolution, we will seek the existence of a method group conversion to see which methods are applicable.

There is an implicit conversion from argument `Y` to the first parameter of the first `D` method, according to 11.8:

> - A conversion is considered to exist if the algorithm of §12.7.6.2 produces a single best method M having the same number of parameters as `D`.
> - Even if the conversion exists, a compile-time error occurs if the selected method `M` is not compatible (§20.4) with the delegate type `D`.

There certainly is a best method, as there is only one `Y`.  But since `Y` is not compatible with the delegate type `MyAction<int>`, a compile-time error occurs.  This is even though the compiler would not even try to apply that conversion, instead selecting the other overload of `D` as the one to be called!

Actual compilers do not report this compile-time error unless the conversion is attempted to be used. This test program compiles and runs.

### LINKED ISSUE #499: [C# 7.3] Better (but not bestest) betterness fix
I found the following while working on #484 

Consider this code:

```
public class BrokenOverloadResolution
{
    static void Test(Action a)
    {
    }

    static void Test(Func<object> f)
    {
    }

    public void MethodGroupExactMatching()
    {
        // This fails to compile with
        // 
        // error CS0121: The call is ambiguous between the following methods or properties: 'BrokenOverloadResolution.Test(Action)' and 'BrokenOverloadResolution.Test(Func<object>)'
        //
        // Fixed in 7.3
        // Probably in Method group conversion:
        // "For a method group conversion, candidate methods whose return type doesn't match up with the delegate's return type are removed from the set."
        // https://github.com/dotnet/roslyn/issues/8942#issuecomment-299536091
        //Test(Listen);

        //
        // non-void version works
        //
        // Test2 (DontListen);
    }

    void Listen()
    {
    }

    object DontListen()
    {
        return null;
    }
}
```

This compiles without error starting in V7.3, based on this issue: https://github.com/dotnet/roslyn/issues/8942#issuecomment-299536091

RexJaeschke:
Fixes #188
Fixes #499

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -973,9 +973,9 @@ For a function member that includes a parameter array, if the function member is
 
 Additional rules determine whether a method is applicable or not based on the context of the expression:
 
-- An instance method is not applicable when the invocation has no instance receiver and `this` cannot be used explicitly (such as in static methods, field initializers and constructor initializers).
-- A static method is not applicable when the invocation has an instance receiver.
-- When the receiver is ambiguously an instance or type due to a color-color situation, both are applicable.
+- An instance method is not applicable when the invocation has no explicit receiver and `this` cannot be used explicitly (such as in static methods, field initializers and constructor initializers).
+- A static method is not applicable when the invocation has an explicit receiver which is not a type.
+- When the receiver could be either an instance or a type as described in [§11.7.6.2](expressions.md#11762-identical-simple-names-and-type-names), both instance and static methods are applicable.
 - A generic method whose type arguments (explicitly specified or inferred) do not all satisfy their constraints is not applicable.
 - In the context of a method group conversion, there must exist an identity conversion (§10.2.2) between the method return type and the delegate's return type. Otherwise, the candidate method is not applicable.
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -973,9 +973,10 @@ For a function member that includes a parameter array, if the function member is
 
 Additional rules determine whether a method is applicable or not based on the context of the expression:
 
-- An instance method is not applicable when the invocation has no explicit receiver and `this` cannot be used explicitly (such as in static methods, field initializers and constructor initializers).
-- A static method is not applicable when the invocation has an explicit receiver which is not a type.
-- When the receiver could be either an instance or a type as described in [§11.7.6.2](expressions.md#11762-identical-simple-names-and-type-names), both instance and static methods are applicable.
+- A static method is only applicable if the method group results from a *simple_name* or a *member_access* through a type.
+- An instance method is only applicable if the method group results from a *simple_name*, a *member_access* through a variable or value, or a *base_access*.
+  - If the method group results from a *simple_name*, an instance method is only applicable if `this` access is permitted [§11.7.12](expressions.md#11712-this-access).
+- When the method group results from a *member_access* which could be via either an instance or a type as described in [§11.7.6.2](expressions.md#11762-identical-simple-names-and-type-names), both instance and static methods are applicable.
 - A generic method whose type arguments (explicitly specified or inferred) do not all satisfy their constraints is not applicable.
 - In the context of a method group conversion, there must exist an identity conversion (§10.2.2) between the method return type and the delegate's return type. Otherwise, the candidate method is not applicable.
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -972,10 +972,11 @@ For a function member that includes a parameter array, if the function member is
   - for a `ref` or `out` parameter, the type of the argument expression is identical to the type of the corresponding parameter.
 
 In the case of a method group
+
 - If the group contains both instance and static members, instance members invoked without an instance receiver or context are not applicable.
 - If the group contains both instance and static members, static members invoked with an instance receiver are not applicable.
 - When there is no receiver, static members in a static context are applicable; otherwise, both static and instance members are applicable.
-- When the receiver is ambiguously an instance or type due to a color-color situation, both are applicable. 
+- When the receiver is ambiguously an instance or type due to a color-color situation, both are applicable.
 - A static context, where an implicit `this` instance receiver cannot be used, includes the body of members where no `this` is defined, such as static members, as well as places where this cannot be used, such as field initializers and constructor initializers.
 - If the group contains generic methods whose type arguments do not satisfy their constraints, these members are not applicable.
 - For a method group conversion, candidate methods whose return type doesn't match up with the delegate's return type are not applicable.
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -971,6 +971,15 @@ For a function member that includes a parameter array, if the function member is
   - for a fixed value parameter or a value parameter created by the expansion, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from the argument expression to the type of the corresponding parameter, or
   - for a `ref` or `out` parameter, the type of the argument expression is identical to the type of the corresponding parameter.
 
+In the case of a method group
+- If the group contains both instance and static members, instance members invoked without an instance receiver or context are not applicable.
+- If the group contains both instance and static members, static members invoked with an instance receiver are not applicable.
+- When there is no receiver, static members in a static context are applicable; otherwise, both static and instance members are applicable.
+- When the receiver is ambiguously an instance or type due to a color-color situation, both are applicable. 
+- A static context, where an implicit `this` instance receiver cannot be used, includes the body of members where no `this` is defined, such as static members, as well as places where this cannot be used, such as field initializers and constructor initializers.
+- If the group contains generic methods whose type arguments do not satisfy their constraints, these members are not applicable.
+- For a method group conversion, candidate methods whose return type doesn't match up with the delegate's return type are not applicable.
+
 #### 11.6.4.3 Better function member
 
 For the purposes of determining the better function member, a stripped-down argument list `A` is constructed containing just the argument expressions themselves in the order they appear in the original argument list.
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -971,15 +971,13 @@ For a function member that includes a parameter array, if the function member is
   - for a fixed value parameter or a value parameter created by the expansion, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from the argument expression to the type of the corresponding parameter, or
   - for a `ref` or `out` parameter, the type of the argument expression is identical to the type of the corresponding parameter.
 
-In the case of a method group
+Additional rules determine whether a method is applicable or not based on the context of the expression:
 
-- If the group contains both instance and static members, instance members invoked without an instance receiver or context are not applicable.
-- If the group contains both instance and static members, static members invoked with an instance receiver are not applicable.
-- When there is no receiver, static members in a static context are applicable; otherwise, both static and instance members are applicable.
+- An instance method is not applicable when the invocation has no instance receiver and `this` cannot be used explicitly (such as in static methods, field initializers and constructor initializers).
+- A static method is not applicable when the invocation has an instance receiver.
 - When the receiver is ambiguously an instance or type due to a color-color situation, both are applicable.
-- A static context, where an implicit `this` instance receiver cannot be used, includes the body of members where no `this` is defined, such as static members, as well as places where this cannot be used, such as field initializers and constructor initializers.
-- If the group contains generic methods whose type arguments do not satisfy their constraints, these members are not applicable.
-- For a method group conversion, candidate methods whose return type doesn't match up with the delegate's return type are not applicable.
+- A generic method whose type arguments (explicitly specified or inferred) do not all satisfy their constraints is not applicable.
+- In the context of a method group conversion, there must exist an identity conversion (§10.2.2) between the method return type and the delegate's return type. Otherwise, the candidate method is not applicable.
 
 #### 11.6.4.3 Better function member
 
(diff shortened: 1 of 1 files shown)


BillWagner:
converted to draft: C# 7 feature

standard/expressions.md
```diff
@@ -968,6 +968,16 @@ For a function member that includes a parameter array, if the function member is
...
 
+In the case of a method group
+
+- If the group contains both instance and static members, instance members invoked without an instance receiver or context are not applicable.
+- If the group contains both instance and static members, static members invoked with an instance receiver are not applicable.
+- When there is no receiver, static members in a static context are applicable; otherwise, both static and instance members are applicable.
+- When the receiver is ambiguously an instance or type due to a color-color situation, both are applicable.
+- A static context, where an implicit `this` instance receiver cannot be used, includes the body of members where no `this` is defined, such as static members, as well as places where this cannot be used, such as field initializers and constructor initializers.
+- If the group contains generic methods whose type arguments do not satisfy their constraints, these members are not applicable.
+- For a method group conversion, candidate methods whose return type doesn't match up with the delegate's return type are not applicable.
```
gafter: I think "match up with" needs to be tightened up to use spec terminology.

gafter (commented):
Looks good except as noted.

jskeet:
Opened up for broader review given Neal's label.

standard/expressions.md
```diff
@@ -968,6 +968,16 @@ For a function member that includes a parameter array, if the function member is
   - for a fixed value parameter or a value parameter created by the expansion, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from the argument expression to the type of the corresponding parameter, or
   - for a `ref` or `out` parameter, the type of the argument expression is identical to the type of the corresponding parameter.
 
+In the case of a method group
+
+- If the group contains both instance and static members, instance members invoked without an instance receiver or context are not applicable.
+- If the group contains both instance and static members, static members invoked with an instance receiver are not applicable.
+- When there is no receiver, static members in a static context are applicable; otherwise, both static and instance members are applicable.
+- When the receiver is ambiguously an instance or type due to a color-color situation, both are applicable.
+- A static context, where an implicit `this` instance receiver cannot be used, includes the body of members where no `this` is defined, such as static members, as well as places where this cannot be used, such as field initializers and constructor initializers.
```
BillWagner: I think you meant to code-fence the last "this" in this paragraph:

```suggestion
- A static context, where an implicit `this` instance receiver cannot be used, includes the body of members where no `this` is defined, such as static members, as well as places where `this` cannot be used, such as field initializers and constructor initializers.
```

standard/expressions.md
BillWagner: How about this wording:

```suggestion
- For a method group conversion, there must exist an identity conversion (§10.2.2) between the candidate method return type and the delegate's return. Otherwise, the candidate method is not applicable.
```

This wording does mean that the candidate method could return `dynamic` and the delegate type could be `Func<object>`. I tested, and that is the current behavior.

BillWagner:
This PR will address #499

BillWagner (approved):
A had two suggestions. One is a possible typo, the second is suggested wording for Neal's comment.

BillWagner:
We can address it in a second PR, but the text changes here are related to #552 

The sentences noted in that issue in 10.8 are resolved by these changes. I believe the second two quoted sentences can be removed, and a link added to this section.

standard/expressions.md
```diff
@@ -968,6 +968,16 @@ For a function member that includes a parameter array, if the function member is
   - for a fixed value parameter or a value parameter created by the expansion, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from the argument expression to the type of the corresponding parameter, or
   - for a `ref` or `out` parameter, the type of the argument expression is identical to the type of the corresponding parameter.
 
+In the case of a method group
+
+- If the group contains both instance and static members, instance members invoked without an instance receiver or context are not applicable.
+- If the group contains both instance and static members, static members invoked with an instance receiver are not applicable.
```
jskeet: Do these lines mean that if there are *only* instance members or *only* static members, the members that will fail later are still applicable? What's the reasoning behind taking that approach rather than saying they're not applicable anyway?

standard/expressions.md
```diff
@@ -968,6 +968,16 @@ For a function member that includes a parameter array, if the function member is
...
+- When there is no receiver, static members in a static context are applicable; otherwise, both static and instance members are applicable.
+- When the receiver is ambiguously an instance or type due to a color-color situation, both are applicable.
```
jskeet: "due to a color-color situation" doesn't sound like something that should be in the normative text. (Perhaps in an example, or refer to an existing example?)

standard/expressions.md
```diff
@@ -968,6 +968,16 @@ For a function member that includes a parameter array, if the function member is
...
+- A static context, where an implicit `this` instance receiver cannot be used, includes the body of members where no `this` is defined, such as static members, as well as places where this cannot be used, such as field initializers and constructor initializers.
+- If the group contains generic methods whose type arguments do not satisfy their constraints, these members are not applicable.
```
jskeet: Is this about explicit type arguments, or does it include inferred ones? (I can't remember the ordering as to whether this happens before type inference or not...)

standard/expressions.md
jskeet: That works for me.

Nigel-Ecma: "candiate" in last line...

standard/expressions.md
```diff
@@ -968,6 +968,16 @@ For a function member that includes a parameter array, if the function member is
   - for a fixed value parameter or a value parameter created by the expansion, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from the argument expression to the type of the corresponding parameter, or
   - for a `ref` or `out` parameter, the type of the argument expression is identical to the type of the corresponding parameter.
 
+In the case of a method group
```
Nigel-Ecma: We are in a subclause of overload resolution, and surely we're not doing overload resolution unless we have a method group?

Should (some of) these added bullets not be in part of the second bullet list in §11.6.4.1?

standard/expressions.md
Nigel-Ecma: Came to to make a similar comment and found Jon had beaten me to it. A Google of the phrase only seems to turn up references to the use here...

standard/expressions.md
```diff
@@ -968,6 +968,16 @@ For a function member that includes a parameter array, if the function member is
...
+
+- If the group contains both instance and static members, instance members invoked without an instance receiver or context are not applicable.
```
Nigel-Ecma: Should this be about the group containing both *applicable* instance and static members? (I'm wondering whether this algorithm is order-independent and so raising the question)

standard/expressions.md
BillWagner: Thanks. Updated the suggestion.

standard/expressions.md
jskeet: Neal believes this includes inferred ones.

jskeet: (And constraints don't prevent inference - so it could be invalid ones at this stage.)

standard/expressions.md
jskeet: The reasoning is: this way we get a better error message. (This is how the compiler works.) We could potentially simplify this. There might be complexities around extension methods, and we're all somewhat confused.

jskeet: Suggestion: keep it like this for now, but raise an issue to see whether we can convince ourselves that the result is the same (in terms of validity etc) - the compiler can still choose to give a better error message.

jskeet: We now think that it could simplify things a lot if we could:

- Refer to methods rather than members, so that we don't need to say "In the case of a method group"
- Remove references to "If the group contains both instance and static members" so that we can just have the context of "the expression we're considering (e.g. whether it has an instance receiver), the method we're considering, and the argument list" - not the rest of the method group.

jskeet: (But we're not sure if this changes extension method invocation. Neal is checking.)

jskeet: We *should* remove the condition here. Demo code from Neal:

```csharp
using System;

static class Extensions {
    public static void M(this A a) {
        Console.WriteLine("extension A::M");
    }
}

class A {

    static void M() {
        Console.WriteLine("A::M");
    }
    static void Main() {
        A a = new A();
        a.M();
    }
}
```

This picks the extension method, *so must have determined that the static method is not applicable*.

jskeet: We should also remove the final validation bullet from 11.7.8.2.

standard/expressions.md
jskeet: And let's make this (and the one below) refer to a single method.

jskeet:
Jon to apply selected tweaks and send out for review again.

standard/expressions.md
jskeet: Done (locally).

standard/expressions.md
```diff
@@ -968,6 +968,16 @@ For a function member that includes a parameter array, if the function member is
...
+- If the group contains both instance and static members, static members invoked with an instance receiver are not applicable.
+- When there is no receiver, static members in a static context are applicable; otherwise, both static and instance members are applicable.
```
jskeet: I believe we can remove this entirely, as it's covered by the conditions above.

standard/expressions.md
jskeet: With the above bullet removed, we don't need to define static context here - although we probably *do* need to define instance context. The term "static context" is only used elsewhere in the definition of the `nameof` operator. It probably makes sense to define "instance context" and "static context" somewhere, but it's not clear to me where that should be.

jskeet: Hmm... I've tried something else in the end, i.e. clarifying where an instance method with no receiver is valid.

jskeet (commented):
I've made some progress and pushed a new commit, but I think we need more clarity around what's meant by an "instance receiver". (I'm sure we all know what it's intended to mean, but I don't think we define it clearly enough.)

standard/expressions.md
gafter: Change "due to a color-color situation" to a reference to https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#11762-identical-simple-names-and-type-names .  Ambiguous is perhaps not the correct word to describe this situation.

jskeet: Yup, there's still work to do on this bit :)

Nigel-Ecma: Go with suitable wording around @gafter's reference to §11.7.6.2. Not sure I agree that _ambiguous_ is perhaps not the correct word, but if it is changed here it also needs to be changed in §11.7.6.2

standard/expressions.md
```diff
@@ -968,15 +968,13 @@ For a function member that includes a parameter array, if the function member is
   - for a fixed value parameter or a value parameter created by the expansion, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from the argument expression to the type of the corresponding parameter, or
   - for a `ref` or `out` parameter, the type of the argument expression is identical to the type of the corresponding parameter.
 
-In the case of a method group
+Additional rules determine whether a method is applicable or not based on the context of the expression:
 
-- If the group contains both instance and static members, instance members invoked without an instance receiver or context are not applicable.
-- If the group contains both instance and static members, static members invoked with an instance receiver are not applicable.
-- When there is no receiver, static members in a static context are applicable; otherwise, both static and instance members are applicable.
+- An instance method is not applicable when the invocation has no instance receiver and `this` cannot be used explicitly (such as in static methods, field initializers and constructor initializers).
```
Nigel-Ecma: I don't know "instance receiver" is used elsewhere. I think what you're looking for is "when the receiver is not a variable or value", and on the next line "when the receiver is not a type" – this phraseology comes from §11.6.5.

standard/expressions.md
gafter: 11.7.6.2 specifically says that it is *not* ambiguous.

Maybe something like

> When the receiver could be either an instance or a type as described in §11.7.6.2, both instance and static methods are applicable.

standard/expressions.md
jskeet: I'll go with that, but it needs to also take into account the possibility of there not being a receiver.

jskeet: Not sure on this, actually - no explicit receiver? I have yet to find a good canonical definition of "receiver".

standard/expressions.md
jskeet: Done.

jskeet (commented):
Okay, a bit more progress, but we'll definitely want to discuss this in more detail.

standard/expressions.md
```diff
@@ -971,6 +971,14 @@ For a function member that includes a parameter array, if the function member is
   - for a fixed value parameter or a value parameter created by the expansion, an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from the argument expression to the type of the corresponding parameter, or
   - for a `ref` or `out` parameter, the type of the argument expression is identical to the type of the corresponding parameter.
 
+Additional rules determine whether a method is applicable or not based on the context of the expression:
+
+- An instance method is not applicable when the invocation has no explicit receiver and `this` cannot be used explicitly (such as in static methods, field initializers and constructor initializers).
+- A static method is not applicable when the invocation has an explicit receiver which is not a type.
+- When the receiver could be either an instance or a type as described in [§11.7.6.2](expressions.md#11762-identical-simple-names-and-type-names), both instance and static methods are applicable.
```
jskeet: Let's try to go back to the style of text we deleted - but the "this" aspect was a problem. (We could come up with an example of static+instance overloads; the instance one should be removed earlier.)

jskeet: Can refer to "where a this_access is permitted" and refer to https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#11712-this-access

standard/expressions.md
```diff
@@ -971,6 +971,14 @@ For a function member that includes a parameter array, if the function member is
...
+- A generic method whose type arguments (explicitly specified or inferred) do not all satisfy their constraints is not applicable.
+- In the context of a method group conversion, there must exist an identity conversion (§10.2.2) between the method return type and the delegate's return type. Otherwise, the candidate method is not applicable.
```
BillWagner, line 981: This is the change that (I believe) fixes #188

jskeet:
Okay, I've had another pass at this, to remove the undefined term "receiver". PTAL before the next meeting (where we'll discuss it).

PR #271, opened 2021-04-27
Remove unnecessary and inaccurate sentence from simple name resolution

### LINKED ISSUE #187: 12.7.3 Missing spec rule (error for instance method of enclosing class called in nested class)
I cannot find the C# spec rule that makes this an error. Where would I find it?

``` c#
class C1
{
    void M() { }
    class C2
    {
        void Entry()
        {
            M(); // error: object reference required for M
        }
    }
}
```

The closest thing I find is is in 12.7.3 Simple names…

•	Otherwise, the result is the same as a member access (Member access) of the form T.I or T.I<A1,...,Ak>. In this case, it is a binding-time error for the simple_name to refer to an instance member

However, it is a bit misleading. M doesn’t “refer to an instance member”. It refers to a method group. The method group might contain both instance and static methods.

When I look at the section on binding an invocation, I don’t see what error is supposed to be given for this case.

jskeet:
The sentence attempts to simply state that a simple name from a
nested type that resolves to a name in an enclosing type must refer
to a static member - but it gets more complex than that when the
name refers to a method group.

Fortunately, this is all handled by effectively delegating to a
member access using the type name - that member access already
handles attempts to access instance members as if they were static.

Fixes #187


Nigel-Ecma:
Modulo the concerns expressed over the wording (see #187)I think the statement was helpful, however I concur with @jskeet that member access covers it so the statement isn't essential and therefore removing it is the easiest fix. I will leave the approval to @gafter so he can confirm it resolves his concerns.

BillWagner (approved):
This LGTM

PR #309, opened 2021-05-04
Remove the guarantee that an implementation runs finalizers on exit.

### LINKED ISSUE #176: Finalizers are not run on shutdown in .NET Core
See dotnet/corefx#5205

We had discussed how strict to make the statements around finalizers running on program exit when updating the spec for C# 5. We should revisit for C# 6 in light of the referenced issue above.

### LINKED ISSUE #291: Spec shouldn't promise that finalizers run on application termination
In .Net Framework, when an application exits, finalizers for finalizable objects that were not yet garbage collected are called. On the other hand, on .Net Core, this does not happen, see https://github.com/dotnet/corefx/issues/5205.

The .Net Framework behavior is consistent with [the draft C# 6.0 specification from MS](https://github.com/dotnet/csharplang/blob/6027ad5a4ab013f4fb42f5edd2d667d649fe1bd8/spec/basic-concepts.md#application-termination), but the .Net Core behavior isn't:

> Prior to an application's termination, destructors for all of its objects that have not yet been garbage collected are called, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example).

[The latest draft of the C# 5.0 specification from ECMA](https://www.ecma-international.org/activities/Languages/C%20sharp%20Standard%20WD3.16.pdf#%5B%7B%22num%22%3A1738%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C40%2C534%2C0%5D) weakens this requirement:

> Prior to an application’s termination, <ins>an implementation should make every reasonable effort to call</ins> finalizers (§16.13) for all of its objects that have not yet been garbage collected<s> are called</s>, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example).<ins> The implementation should document any conditions under which this behavior cannot be guaranteed.</ins>

This means that .Net Core follows the letter of this version of the spec (as long as it documents that this behavior is *never* guaranteed). But I think it does not follow the spirit of the spec: .Net Core does not make *any* effort to call finalizers and so I think that the spec should be weakened even further.

Is the new wording in the ECMA draft good enough? Could someone from the ECMA TC49-TG2, which is currently updating the ECMA version of the spec, comment on this?

This was inspired by [an SO question](https://stackoverflow.com/q/44732234/41071).

---

To show this behavior:

1. Create an application consisting of Program.cs:

    ```c#
    using System;

    class C
    {
        static void Main()
        {
            new C();

            Console.WriteLine("Exiting.");
        }

        ~C() => Console.WriteLine("Finalizing.");
    }
    ```

   and app.csproj:

    ```xml
    <Project Sdk="Microsoft.NET.Sdk">
      <PropertyGroup>
        <OutputType>Exe</OutputType>
        <TargetFrameworks>net47;netcoreapp1.0</TargetFrameworks>
      </PropertyGroup>
    </Project>
    ```
2. Run the app on .Net Framework and .Net Core to observe the difference:

    ```powershell
    > dotnet run -f netcoreapp1.0 # finalizer not called
    Exiting.
    > dotnet run -f net47         # finalizer called
    Exiting.
    Finalizing.
    ```

BillWagner:
Fixes #176
Fixes #291

The normative text now declares that running finalizers on exit is implementation specific.

I added an additional *note* that the .NET Framework follows the earlier standard text. (I looked for a definitive statement on the Mono environment, and I found resources that implied both that it did and it didn't follow this guarantee.

I'll propose three resolutions in comments for committee members to consider.

### DIFF
--- a/standard/basic-concepts.md
+++ b/standard/basic-concepts.md
@@ -40,7 +40,7 @@ If the return type of the entry point method is `void`, reaching the right brace
 
 Whether or not finalizers ([§15.13](classes.md#1513-finalizers)) are run as part of application termination is implementation-specific.
 
-> *Note*: The .NET Framework implementation makes every reasonable effort to call finalizers ([§15.13](classes.md#1513-finalizers)) for all of its objects that have not yet been garbage collected, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example). No other implementation makes this guarantee. *end note*
+> *Note*: The .NET Framework implementation makes every reasonable effort to call finalizers ([§15.13](classes.md#1513-finalizers)) for all of its objects that have not yet been garbage collected, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example). *end note*
 
 ## 8.3 Declarations
 
--- a/standard/basic-concepts.md
+++ b/standard/basic-concepts.md
@@ -38,7 +38,9 @@ If the return type of the application's entry point method is `int`, the value r
 
 If the return type of the entry point method is `void`, reaching the right brace (`}`) that terminates that method, or executing a `return` statement that has no expression, results in a termination status code of `0`. If the entry point method terminates due to an exception ([§21.4](exceptions.md#214-how-exceptions-are-handled)), the exit code is implementation-specific. Additionally, the implementation may provide alternative APIs for specifying the exit code.
 
-Prior to an application's termination, an implementation should make every reasonable effort to call finalizers ([§15.13](classes.md#1513-finalizers)) for all of its objects that have not yet been garbage collected, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example). The implementation should document any conditions under which this behavior cannot be guaranteed.
+Whether or not finalizers ([§15.13](classes.md#1513-finalizers)) are run as part of application termination is implementation-specific.
+
+> *Note*: The .NET Framework implementation makes every reasonable effort to call finalizers ([§15.13](classes.md#1513-finalizers)) for all of its objects that have not yet been garbage collected, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example). No other implementation makes this guarantee. *end note*
 
 ## 8.3 Declarations
 


BillWagner:
***Proposal 1:***

Accept these changes as is.

BillWagner:
***Proposal 2:***

Accept the normative text change, but do not include the additional note.

BillWagner:
***Proposal 3:***

Delete the original paragraph, but do not add any new text.

standard/basic-concepts.md
```diff
@@ -38,7 +38,9 @@ If the return type of the application's entry point method is `int`, the value r
...
 
-Prior to an application's termination, an implementation should make every reasonable effort to call finalizers ([§15.13](classes.md#1513-finalizers)) for all of its objects that have not yet been garbage collected, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example). The implementation should document any conditions under which this behavior cannot be guaranteed.
+Whether or not finalizers ([§15.13](classes.md#1513-finalizers)) are run as part of application termination is implementation-specific.
+
+> *Note*: The .NET Framework implementation makes every reasonable effort to call finalizers ([§15.13](classes.md#1513-finalizers)) for all of its objects that have not yet been garbage collected, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example). No other implementation makes this guarantee. *end note*
```
jskeet:
```suggestion
> *Note*: The .NET Framework implementation makes every reasonable effort to call finalizers ([§15.13](classes.md#1513-finalizers)) for all of its objects that have not yet been garbage collected, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example). *end note*
```

(As suggested during the meeting.)

gafter:
```suggestion
> *Note*: The .NET Framework implementation makes every reasonable effort to call finalizers ([§15.13](classes.md#1513-finalizers)) for all of its objects that have not yet been garbage collected, unless such cleanup has been suppressed (by a call to the library method `GC.SuppressFinalize`, for example).*end note*
```

gafter (approved):

jskeet (approved):

PR #310, opened 2021-05-04
Add lexer rules to grammar validation

BillWagner:
Include the stub definitions for lexer members into the grammar file.

### DIFF
--- a/tools/GetGrammar/grammar-lexer-members.txt
+++ b/tools/GetGrammar/grammar-lexer-members.txt
@@ -1,0 +1,10 @@
+
+@lexer::members {
+bool IsLetterCharacter()       { return true; }
+bool IsCombiningCharacter()    { return true; }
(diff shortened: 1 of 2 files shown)


tools/GetGrammar/grammar-lexer-members.txt
```diff
@@ -0,0 +1,10 @@
+
+@lexer::members {
```
BillWagner, line 2: @RexJaeschke 

This appears to be incorrect syntax. Where should we look for this?

BillWagner: OK, it was completely my fault for using `>` instead of `>>`.

This is fixed now.

BillWagner:
@RexJaeschke If you merge this PR, it will go into your ANTLR branch, and it will update #261

PR #325, opened 2021-05-22
Fix editorial, formatting, example nits

RexJaeschke:
Corrections for various errors I spotted while reading the spec.

The `stackalloc` example did not correctly handle the smallest negative value, which has no positive counterpart.


BillWagner (approved):
These all LGTM @RexJaeschke 

I'll :shipit: now.

PR #327, opened 2021-05-27
Add note (with examples) around delegate combination and removal issues

### LINKED ISSUE #122: 12.9.5 Document (as a vague note) exceptions when combining different delegate types
We also need to check whether it can happen on subtraction (13.9.6)

jskeet:
Fixes #122

### DIFF
--- a/standard/delegates.md
+++ b/standard/delegates.md
@@ -207,6 +207,43 @@ Once instantiated, a delegate instance always refers to the same invocation list
 
 > *Note*: Remember, when two delegates are combined, or one is removed from another, a new delegate results with its own invocation list; the invocation lists of the delegates combined or removed remain unchanged. *end note*
 
+> *Note*: Delegate combination and removal is known to fail with an exception in some implementations when the delegate types involved at runtime are different due to variant conversion ([§18.2.3.3](interfaces.md#18233-variance-conversion). The following example demonstrates a situation which may fail.
+> ```csharp
+> delegate void Action<in T>(T obj);
+>
+> class Test
+> {
+>    static void Main() {
+>       Action<string> stringAction = str => {};
+>       Action<object> objectAction = obj => {};
+>        
+>       // Valid due to implicit reference conversion of
+>       // objectAction to Action<string>, but may fail
+>       // at runtime.
+>       Action<string> combination = stringAction + objectAction;
+>    }
+> }
+> ```
+>
+> If combination or removal is required between  different but compatible delegate types, a *delegate_creation_expression* ([§12.7.11.6](expressions.md#127116-delegate-creation-expressions)) can be used to create a delegate with the correct runtime type. The following example demonstrates how this workaround may be applied to the preceding example.
+> ```csharp
+> delegate void Action<in T>(T obj);
+>
+> class Test
+> {
+>    static void Main() {
+>       Action<string> stringAction = str => {};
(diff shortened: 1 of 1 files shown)


standard/delegates.md
```diff
@@ -207,6 +207,43 @@ Once instantiated, a delegate instance always refers to the same invocation list
...
+>        
+>       // Valid due to implicit reference conversion of
+>       // objectAction to Action<string>, but may fail
+>       // at runtime.
+>       Action<string> combination = stringAction + objectAction;
+>    }
+> }
+> ```
+>
+> If combination or removal is required between  different but compatible delegate types, a *delegate_creation_expression* ([§12.7.11.6](expressions.md#127116-delegate-creation-expressions)) can be used to create a delegate with the correct runtime type. The following example demonstrates how this workaround may be applied to the preceding example.
```
jskeet, line 228: Note: I think "delegate creation expression" (rather than "*delegate_creation_expression*") would be better here, as we're referring to the semantic construct more than the syntax, but the rest of the section seems to use the latter. Thoughts welcome on this - and indeed whether it's even worth mentioning this workaround.

standard/delegates.md
```diff
@@ -207,6 +207,43 @@ Once instantiated, a delegate instance always refers to the same invocation list
 
 > *Note*: Remember, when two delegates are combined, or one is removed from another, a new delegate results with its own invocation list; the invocation lists of the delegates combined or removed remain unchanged. *end note*
 
+> *Note*: Delegate combination and removal is known to fail with an exception in some implementations when the delegate types involved at runtime are different due to variant conversion ([§18.2.3.3](interfaces.md#18233-variance-conversion). The following example demonstrates a situation which may fail.
```
BillWagner, line 210: Note: to save time, you can just place "§18.2.3.3" in the text, and the automatic section renumbering tool will format the link to the correct anchor.  These are correct, so no need to make any edits.

BillWagner (approved):
This LGTM @jskeet 

You can :shipit: when ready.

standard/delegates.md
jskeet: Ah right. :)

jskeet:
Will wait until Wednesday and we can just get this checked out by the others too - I wanted to get an early first pass from you though :)

standard/delegates.md
svick, line 210: Tiny nit: missing closing paren:

```suggestion
> *Note*: Delegate combination and removal is known to fail with an exception in some implementations when the delegate types involved at runtime are different due to variant conversion ([§18.2.3.3](interfaces.md#18233-variance-conversion)). The following example demonstrates a situation which may fail.
```

Nigel-Ecma, line 210: Maybe something along the lines of:

> *Note*: At the time of writing delegate combination...

Rationale: Knowing when an implementation fails to meet the Standard is clearly useful to the (potential) user of that implementation. However whether it is the role of the Standard to do that is less obvious; the implementation’s own documentation should, an Annotated Standard might, but the Standard itself? If the Standard is to do so then indicating the *Note* refers to a point in time reduces the rather definitive tone – after all this anonymous implementation could become compliant (in a ideal world) surely?

jskeet:
Closing after discussion as we've decided not to include it in the Standard. @BillWagner will make a similar change in the language reference though.

PR #332, opened 2021-06-11
Fix Word converter

jskeet:


BillWagner (approved):
This is awesome @jskeet 

One more check that passes.

I'll approve this. You can :shipit: when ready.

PR #333, opened 2021-06-15
C# 7.x: Add support for readonly structs

RexJaeschke:
This PR adds one new modifier to structs, `readonly`, in *struct_modifier*.

### DIFF
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -56,7 +56,7 @@ A readonly struct has the following constraints:
 
 - Each of its instance fields shall also be declared `readonly`.
 - None of its instance properties shall have a *set_accessor_declaration* ([§14.7.3](classes.md#1473-accessors)).
-- It shall not declare any field-like events ([§14.8.2](classes.md#1482-field-like events)).
+- It shall not declare any field-like events ([§14.8.2](classes.md#1482-field-like-events)).
 
 When an instance of a readonly struct is passed to a method, its `this` is treated like an `in` argument/parameter, which disallows write access to any instance fields (except by constructors).
 
@@ -65,6 +65,7 @@ When an instance of a readonly struct is passed to a method, its `this` is treat
 The `ref` modifier indicates that the *struct_declaration* declares a type whose instances are allocated on the execution stack.
 
 It is a compile-time error if a ref struct type is used in any of the following contexts:
+
 - As the element type of an array.
 - As the declared type of a field of a class or a non-ref struct.
 - To implement an interface.
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -16,13 +16,13 @@ A *struct_declaration* is a *type_declaration* ([§13.7](namespaces.md#137-type-
 
 ```ANTLR
 struct_declaration
-    : attributes? struct_modifier* 'partial'? 'struct' identifier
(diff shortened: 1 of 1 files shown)


KalleOlaviNiemitalo:
Should an implicit conversion from `dynamic` to a ref struct type be rejected at compile time? Microsoft's implementation of `dynamic` uses the ref struct as a type argument, which is then not supported by the CLR. However, as the C# standard does not specify the details of `dynamic`, I imagine a different compiler with a different library could find a way to make the conversion work. <https://github.com/dotnet/csharplang/discussions/4713>, <https://github.com/dotnet/csharpstandard/issues/282>

standard/structs.md
```diff
@@ -38,12 +38,38 @@ struct_modifier
...
+The `ref` modifier indicates that the *struct_declaration* declares a type whose instances are allocated on the execution stack.
+
+It is a compile-time error if a ref struct type is used in any of the following contexts:
+- As the element type of an array.
+- As the declared type of a field of a class or a non-ref struct.
+- To implement an interface.
+- Being boxed to `System.ValueType` or `System.Object`.
+- As a type argument.
+- An async method.
+- An iterator.
```
KalleOlaviNiemitalo: Does "a ref struct type is used" include this case:

```C#
using System;
using System.Threading.Tasks;

class Widget
{
}

static class Class1
{
    static async Task M(Task t)
    {
        Widget w = stackalloc int[] { 42 };
        await t;
    }
}
```

.NET SDK 6.0.101 compiles it without errors. The Span\<int\> becomes a local variable in the MoveNext method of the state machine.

gafter:
This draft does not reflect the draft specification at https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.2/span-safety.md

RexJaeschke:
Reminder to look at https://github.com/dotnet/csharpstandard/issues/334 for other open issues w.r.t this PR (including Neal's concerns re missing text re span safety).

RexJaeschke:
Based on Neal’s suggestion during yesterday's TG2 meeting, we agreed to merge as is, but only call `readonly structs` completed. Then, create another PR for all the additional work for `ref structs`.

BillWagner:
I'll rebase this branch to resolve conflicts, and mark it ready for review.

jskeet:
Assigned to @gafter for the splitting off - we should have assigned this before...

gafter:
I've removed ref struct from this PR and moved it to https://github.com/dotnet/csharpstandard/pull/601.  This should be fine for `readonly struct`.

standard/structs.md
BillWagner: Resolving this conversation in this PR, as this concern should be covered in #601

KalleOlaviNiemitalo: Reposted as <https://github.com/dotnet/csharpstandard/pull/601#discussion_r925709489>.

BillWagner:
@jskeet 

We should be able to do the final approval for this and merge it during this week's meeting. The `ref` struct portion has all been moved to #601

BillWagner (approved):
This LGTM. Let's :shipit:

jskeet (approved):

jskeet:
Yup, just done a final review. Will merge now.

PR #335, opened 2021-06-15
Fix bad link references

RexJaeschke:
In 4 places in 3 files, I'd written $ instead of § to introduce a link's destination for a link that did not yet exist.

I also fixed a code-fencing error.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -2173,7 +2173,7 @@ The `typeof` operator can be used on a type parameter. The result is the `System
 
 ### 12.7.13 The sizeof operator
 
-The `sizeof` operator returns the number of 8-bit bytes occupied by a variable of a given type. The type specified as an operand to sizeof shall be an *unmanaged_type* ($unmanaged-types-new-clause).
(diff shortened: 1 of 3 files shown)


BillWagner (approved):
This LGTM @RexJaeschke 

I'll :shipit: now.

BillWagner:
@RexJaeschke I merged this, and expected to see a related PR that updates these links. That didn't happen. I think it's because the previous action had replaced the key `unmanaged-types-new-clause` with its section number.  

I'll make an issue for the section renumbering tool so that it flags an error in that case. In the meantime, what is the reference that should go in its place? Once I test my fixes in the section renumbering tool, I'll update the references.

RexJaeschke:
@BillWagner (Hopefully) I just fixed this with PR #337. Clearly, I didn't thing this through before. By changing the $ to §, I naively figured the number-fixup tool would kick in and update that. However, when that was run on the original merge some time ago, it assigned that new section a number, so no new reconciliation was possible this time. It sure is hard to get good help nowadays!

BillWagner:
@RexJaeschke Your naive plan helped me find a few errors in the number-fixup tool.  See #336, which I think makes #337 unnecessary (but also prevents those kinds of errors coming back).

PR #339, opened 2021-06-19
The ANTLR-isation of the lexical grammar. The C# "lexical" grammar bo…

Nigel-Ecma:
…th tokenises and parses the input and this

requires using both lexical and parser ANTLR rules while producing a lexer which produces a token stream and
not some AST of the input!

In this stage of ANTLR-isation the pre-processor is not tackled; the pre-processor definition is a context-free
grammar which is atypical for a lexical grammar and defining it with lexical rules does not produce the expected
result... That said the semantics of the pre-processor can be acheived in an ANTLR lexical grammar which includes
predicates/actions and that will be explored in a subsequent stage. The grammar in this stage can be used along
with a factored syntactical grammar to lex & parse C# source which does not include pre-processor constructs.

The majority of changes are to lexical-structure.md, many of the other files are only altered as the lexical
rule *Identifier* has been replaced by the parser rule *identifier*.

I will add some comments within the PR to explain why it is the way it is...

PR #342, opened 2021-06-24
ANTLR-isation of the lexical and pre-processor grammars

Nigel-Ecma:
This is a follow on from PR#339 which only ANTLR-ised the lexical grammar:

- All the non-pre-processor changes in PR#339 are included.
- The pre-processor changes in PR#339 made only so the grammar would pass
  ANTLR are *replaced* by different changes in this PR.
- The grammar in PR#339 while passing ANTLR does not work unless the
  pre-processor productions are omitted.
- The original grammar also passed ANTLR but did not work for more reasons

The lexical and pre-processor grammars in this PR both pass ANTLR and should
work 99.9999% (approximately ;-)) of the time:

- To achieve this the way the pre-processor is described and it's grammar
  have been modified.
- In this grammar each pre-processor directive is lexer as a single instance
  of a `PP_Directive` token.
- To build a working ANTLR lexer these tokens can either (a) be `skip`'ed so
  the parser does not see them or (b) a semantic action can be defined which
  implements the pre-processing procces.
- Both options have been tested and a fully working pre-processor implemented
  as a semantic action. This is (currently) written in Java and, even if a C#
  version is written, is not intended for anything other than interal use in
  testing.
- So what's that 99.9999% approximation? Well the Standard (currently) allows
  input skipped over by `#if`/.../`endif` processing to be lexically invalid.
  While the Standard describes this the grammar when run through ANTLR does
  not. So if the feature has been taken advantage of ANTLR will report lexical
  errors.

Note: the combined lexical, pre-processor & syntatic grammar remains
unacceptable to ANTLR due to the mutual left recursion. This still needs
to be removed by hand as shown in the documents distributed on the mailing
list. At least some of the changes required will be submitted in PRs.

### DIFF
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -300,7 +300,7 @@ token
 
 ### 7.4.2 Unicode character escape sequences
 
-A Unicode escape sequence represents a Unicode code point. Unicode escape sequences are processed in identifiers ([§7.4.3](lexical-structure.md#743-identifiers)), character literals ([§7.4.5.5](lexical-structure.md#7455-character-literals)), regular string literals ([§7.4.5.6](lexical-structure.md#7456-string-literals)), and interpolated regular string literals (§interpolated-string-literals-new-clause). A Unicode escape sequence is not processed in any other location (for example, to form an operator, punctuator, or keyword).
+A Unicode escape sequence represents a Unicode code point. Unicode escape sequences are processed in identifiers ([§7.4.3](lexical-structure.md#743-identifiers)), character literals ([§7.4.5.5](lexical-structure.md#7455-character-literals)), and regular string literals ([§7.4.5.6](lexical-structure.md#7456-string-literals)). A Unicode escape sequence is not processed in any other location (for example, to form an operator, punctuator, or keyword).
 
 ```ANTLR
 fragment Unicode_Escape_Sequence
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -26,23 +26,29 @@ All terminal characters are to be understood as the appropriate Unicode characte
 
 ### 7.2.2 Grammar notation
 
-The lexical and syntactic grammars are presented in Backus-Naur form using the notation of the ANTLR grammar tool.
+The lexical and syntactic grammars are presented in the ANTLR grammar tool’s Extended Backus-Naur form.
+
+While the ANTLR notation is used this Standard does not present a complete ANTLR-ready "reference grammar" for C#; writing a lexer and parser, either by hand or using a tool such as ANTLR, is outside the scope of a language specification. With that qualification, this Standard attempts to minmise the gap between the specified grammar and that required to build a lexer and parser in ANTLR with the notable exception of the preprocessor ([§7.5](lexical-structure.md#75-pre-processing-directives) which requires more substantial work to fit into the ANTLR model.
+
+ANTLR distinguishes between lexical and syntactic, termed parser by ANTLR, grammars in its notation by starting lexical rules with an initial uppercase letter and parser rules with an initial lowercase letter.
+
+> *Note*: The C# lexical grammar ([§7.2.3](lexical-structure.md#723-lexical-grammar)) and syntactic grammar ([§7.2.4](lexical-structure.md#724-syntactic-grammar)) are not in exact correspondence with the ANTLR division into lexical and parser grammers. This small mismatch means that some ANTLR parser rules are used when specifying the C# lexical grammar. *end note* 
 
 ### 7.2.3 Lexical grammar
 
 The lexical grammar of C# is presented in [§7.3](lexical-structure.md#73-lexical-analysis), [§7.4](lexical-structure.md#74-tokens), and [§7.5](lexical-structure.md#75-pre-processing-directives). The terminal symbols of the lexical grammar are the characters of the Unicode character set, and the lexical grammar specifies how characters are combined to form tokens ([§7.4](lexical-structure.md#74-tokens)), white space ([§7.3.4](lexical-structure.md#734-white-space)), comments ([§7.3.3](lexical-structure.md#733-comments)), and pre-processing directives ([§7.5](lexical-structure.md#75-pre-processing-directives)).
 
-Every compilation unit in a C# program shall conform to the *Input* production of the lexical grammar ([§7.3.1](lexical-structure.md#731-general)).
+Many of the terminal symbols of the syntactic grammar are not defined explicitly as tokens in the lexical grammar. Rather advantage is taken of the ANTLR behaviour that literal strings in the grammar are extracted as implicit lexical tokens; this allows keywords, operators, etc. to be represented in the grammar by their literal representation rather than a token name.
 
-Using the ANTLR convention, lexer rule names are spelled with an initial uppercase letter.
+The same behaviour is also used to simplify the lexical grammar, see [§7.3.1](lexical-structure.md#731-general).
+
+Every compilation unit in a C# program shall conform to the *input* production of the lexical grammar ([§7.3.1](lexical-structure.md#731-general)).
 
 ### 7.2.4 Syntactic grammar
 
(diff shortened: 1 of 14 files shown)


jskeet:
I'm hoping to fix the Word converter to allow fragments today.

Nigel-Ecma:
So you can overview the grammar easily I've manually updated the grammar.md file so it should now reflect the grammar in the other sections.

jskeet:
I'm going to wait until we've talked about #339 before I review this - my brain can't take any more ANTLR right now...

standard/lexical-structure.md
```diff
@@ -26,23 +26,29 @@ All terminal characters are to be understood as the appropriate Unicode characte
 
 ### 7.2.2 Grammar notation
```
MadsTorgersen, line 27: Great clarifying text here; it helps a lot with setting the context for how to think of the ANTLR grammars throughout the spec.

standard/lexical-structure.md
```diff
@@ -101,37 +107,49 @@ SLASH    : '/' ;
...
 
 The lexical processing of a C# compilation unit consists of reducing the file into a sequence of tokens that becomes the input to the syntactic analysis. Line terminators, white space, and comments can serve to separate tokens, and pre-processing directives can cause sections of the compilation unit to be skipped, but otherwise these lexical elements have no impact on the syntactic structure of a C# program.
 
 When several lexical grammar productions match a sequence of characters in a compilation unit, the lexical processing always forms the longest possible lexical element.
 
 > *Example*: The character sequence `//` is processed as the beginning of a single-line comment because that lexical element is longer than a single `/` token. *end example*
 
+The rules of the lexical grammar are ordered top-to-bottom, with the first matching rule determining the recognised token. Implicit rules ([§7.2.3](lexical-structure.md#723-lexical-grammar)) derived from literal strings in the grammar are ordered before the explicit rules. This ordering simplifies the grammar.
```
MadsTorgersen: I am concerned about the reliance on ordering. The old spec didn't have that. It introduces an implicit context for grammar rules that prevents them from standing alone, and for instance makes it hard to reliably quote them. I'd like to suggest that we minimize the reliance on ordering, and where necessary, make a comment as to the exceptions incurred by previous rules.

standard/lexical-structure.md
```diff
@@ -846,10 +934,14 @@ Pre-processing directives are not tokens and are not part of the syntactic gramm
 The conditional compilation functionality provided by the `#if`, `#elif`, `#else`, and `#endif` directives is controlled through pre-processing expressions ([§7.5.3](lexical-structure.md#753-pre-processing-expressions)) and conditional compilation symbols.
 
 ```ANTLR
-Conditional_Symbol
-    : Identifier_Or_Keyword { IsNotTrueOrFalse() }?
+fragment PP_Conditional_Symbol
```
MadsTorgersen, line 927: I wonder whether the ANTLR semantic predicates are a good idea within the text of the standard. I'd almost rather have an end-of-line comment to the same effect. I think this is a good topic for discussion.

standard/lexical-structure.md
```diff
@@ -974,56 +1058,63 @@ A `#undef` may "undefine" a conditional compilation symbol that is not defined.
 The conditional compilation directives are used to conditionally include or exclude portions of a compilation unit.
 
 ```ANTLR
-Pp_Conditional
-    : Pp_If_Section Pp_Elif_Section* Pp_Else_Section? Pp_Endif
+fragment PP_Conditional
```
MadsTorgersen, line 1050: This is the bread and butter of this approach. This is where the change is made to have conditional directives include only the directive itself, not the source text guarded by it, thus forcing tokenization of the intermediate text.

I have to say that I think it is elegant. I wouldn't want to complicate the grammar to achieve that last 0.001% expressiveness; instead we can wave it away in prose.

standard/lexical-structure.md
```diff
@@ -974,56 +1058,63 @@ A `#undef` may "undefine" a conditional compilation symbol that is not defined.
...
 
-Skipped_Section_Part
-    : Skipped_Characters? New_Line
-    | Pp_Directive
-    ;
-    
-Skipped_Characters
-    : Whitespace? Not_Number_Sign Input_Character*
-    ;
+Conditional compilation directives shall be written as sets consisting of, in order, a `#if` directive, zero or more `#elif` directives, zero or one `#else` directive, and a `#endif` directive. Between the directives are ***conditional sections*** of source code. Each section is controlled by the immediately preceding directive. A conditional section may itself contain nested conditional compilation directives provided these directives form complete sets.
```
MadsTorgersen: This is where we pay the price of not being able to sequence the conditional directives in grammar. That's fine - the term "set" may be confusing here, because it does not refer to a mathematical set (unordered by definition). Maybe another term, like "group", would be better? (Yes, groups are also a mathematical notion, but less likely to be understood as such in this context!)

MadsTorgersen (approved):
This is great. I've brought up a few issues for discussion in comments, and would like to address those in the next meeting, but they should not hold up the PR.

standard/lexical-structure.md
Nigel-Ecma: The old spec more glossed over the ordering than not have it, so the *text* did not mention it. The adding of `fragment` removes most of the ordering issues leaving the overlap between the (implicit) keyword and *available_identifier* rules. The old spec has:

```
available-identifier::
   An identifier-or-keyword that is not a keyword
```

while this one (line 357) has:

```
fragment Available_Identifier
     : Basic_Identifier     // does not include keywords or contextual keywords,
                            // see note below
     ;
```

That comment could change from “does not include” to say ”excluding”. The referenced note (line 427), which is of course informative, could maybe be reworded to be clearer that *in ANTLR* the *Available_Identifier* rule automagically (a word we should have in the Standard somewhere ;-)) enforces the comment on it. The example below (line 147), and the para (line 145) this comment is attached to, would also need similar treatment.

I put the ordering explanation in as the grammar is written in ANTLR which, along with most lexer generators, does have such ordering. (Just as most follow "maximum munch", which the Standard does mention (line 141)). 

Adding `fragment` resolves most of the ordering issues, leaving just (I think) *Available_Identifier*, so making changes to the text as suggested above to move all mention of it to informative text is probably doable.

standard/lexical-structure.md
Nigel-Ecma: I wonder as well...

I'm currently tending towards the idea of having a section in Appendix A (which is all informative) which suggests where they might be used etc. as part of helping people actually use the grammar but without making any of it normative or including implementations (both of which are satisfied already, just distributed within the text).

On the other hand the inline predicate in `PP_Start` is rather key from an ANTLR perspective... dither, dither ;-)

standard/lexical-structure.md
Nigel-Ecma: I think "group" would fit.

Nigel-Ecma:
Thanks for the review Mads. I’ve added a few comments to the comments to help the discussion along.

standard/lexical-structure.md
jskeet: Nigel will rephrase to make it less intimidating, and make it informative instead of normative. But it's still important "as a thing".

standard/lexical-structure.md
jskeet: Resolved in meeting to remove them all and potentially add a section in the appendix later.

standard/lexical-structure.md
jskeet: The group agreed with group.

jskeet: (Group-think in action.)

standard/lexical-structure.md
Nigel-Ecma: All mention of ordering has been relegated to an ANTLR specific note.

standard/lexical-structure.md
Nigel-Ecma: OK, all semantic predicates except the one on `PP_START` have been removed. The `PP_Start` one is in-line and fundamental – without it or something else tokenising might be incorrect. However we could move it to the note if that is the consensus.

PR #343, opened 2021-06-24
Parse (and format) ANTLR fragment productions in the Word converter

jskeet:
This should enable #339 and #342 in terms of Word conversion.

### DIFF
--- a/tools/MarkdownConverter/Grammar/Production.cs
+++ b/tools/MarkdownConverter/Grammar/Production.cs
@@ -8,6 +8,11 @@
         public EbnfNode Ebnf { get; set; }
 
         /// <summary>
+        /// Whether or not this production is a fragment
+        /// </summary>
+        public bool Fragment { get; set; }
+
+        /// <summary>
         /// Optional. Contains no whitespace and is not delimited by '
         /// </summary>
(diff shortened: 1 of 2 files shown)


tools/MarkdownConverter/Grammar/Production.cs
```diff
@@ -7,6 +7,11 @@ internal class Production
         /// </summary>
         public EbnfNode Ebnf { get; set; }
 
+        /// <summary>
+        /// Whether or not this production is a fragment
+        /// </summary>
+        public bool Fragment { get; set; }
```
BillWagner, line 13: Not necessary, but it looks like this could be an init only property once we move to C# 9. I only see it set in the property initializer.

```suggestion
        public bool Fragment { get; init; }
```

BillWagner (approved):
This LGTM @jskeet 

Let's see if this gets us further.

tools/MarkdownConverter/Grammar/Production.cs
jskeet: Yes, I suspect that may well be fine. I was mostly following the rest of the properties. I'd prefer to leave it as it is for now, and then revisit holistically.

PR #344, opened 2021-06-28
Change in-line Note(s) to stand-alone Notes

RexJaeschke:
@Nigel-Ecma this fixes the instances you reported to me.


Nigel-Ecma (approved):
LGTM

PR #349, opened 2021-07-01
Remove the language option

BillWagner:
The current ANTLR tools don't generate C# well.

Here's the language fix. I ran this against the latest draft v6 branch, and the only error was the final remaining mutual left recursion error.


Nigel-Ecma (approved):
LGTM

PR #352, opened 2021-07-07
generate the TOC in the standard folder

### LINKED ISSUE #329: Add a standalone TOC file
I'd like to update the section numbering tool to create a standalone toc.md file in the spec folder (in addition to updating the readme.md file in this repo).

As we complete the features for C# 6.0, I'd like to publish the draft standard text on docs.microsoft.com instead of the Microsoft spec from the [csharplang](https://github.com/dotnet/csharplang/tree/main/spec) repo. The format for the updated TOC is much nicer, so would be a great detailed table of contents.

BillWagner:
Update the logic so the TOC is added to the end of the README.md file in the standard folder.

Also, fix one regression where if a new link was added at the end of a paragraph, the logic walked off the end of the line.

Fixes #329


tools/StandardAnchorTags/ReferenceUpdateProcessor.cs
```diff
@@ -106,14 +106,14 @@ private string ProcessSectionLinks(string line, int lineNumber, string file)
             // .
             // - 
             // indicates the end of the link reference.
-            while (line[endIndex] switch
+            while ((endIndex < line.Length) && (line[endIndex] switch
```
BillWagner, line 109: This was the one regression. The test for the end of the line was added.

tools/StandardAnchorTags/Program.cs
```diff
@@ -141,7 +141,7 @@ private static async Task<string> ReadExistingReadMe()
             // This is the first node in the TOC, so truncate here:
             var index = contents.IndexOf(TOCHeader);
 
-            return contents.Substring(0, index);
+            return (index == -1) ? contents : contents.Substring(0, index);
```
BillWagner: This strictly isn't necessary, but it means that we can remove the TOC nested list by hand, and the next run will add it. It made these changes easier to test, and seems like a nice to have.

standard/README.md
```diff
@@ -0,0 +1,728 @@
+# Detailed table of contents
+
+<!-- The remaining text is generated by a tool. Do not hand edit -->
```
jskeet, line 3: Do we actually need the marker any more? I'd expect we could simplify the code by just making it generate the complete file, including the heading.

tools/StandardAnchorTags/Program.cs
jskeet: Again, I think this can go away entirely :)

jskeet (approved):
One suggestion to potentially simplify things, but I'm fine as-is as well.

standard/README.md
BillWagner: I left this in for two reasons:

1. It's a note to any contributors not to edit the TOC.
1. If we add more text in the file, it will just work (as it was already tested in the parent folder).

jskeet: Okay - I'm easy either way.

PR #354, opened 2021-07-09
Ignore README.md in the input to the Word converter

jskeet:
Previously there was a whole pile of code which would expect the
README to specify the order - whereas now we generate the README
separately.


BillWagner (approved):
This is quite the simplification @jskeet 

Let's :shipit:

PR #356, opened 2021-07-28
Update section 11.7.3 to be more accurate and specific about expression tree conversions

### LINKED ISSUE #341: Strange reference to Section 12.7.3 in §C.3 Expression<TDelegate>.Compile()
In draft-v6, Section C.3 "Standard Library Types not defined in ISO/IEC 23271" includes this reference to Section 12.7.3: <https://github.com/dotnet/csharpstandard/blob/331aae414c0955eb460bd1cbdbcb94792ba83845/standard/standard-library.md#L568-L577>

However, [12.7.3 Simple names](https://github.com/dotnet/csharpstandard/blob/331aae414c0955eb460bd1cbdbcb94792ba83845/standard/expressions.md#1273-simple-names) does not mention "delegate". Perhaps the reference should be to some other section.

jskeet:
Expression tree conversions are implementation-defined, so it
doesn't make sense to refer to 11.7.3 for more details (or to fail
to do so properly due to using the wrong number, as is the case in
the standard library).

Fixes #341

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -643,7 +643,7 @@ Given a user-defined conversion operator that converts from a non-nullable value
 
 ### 11.7.1 General
 
-An *anonymous_method_expression* or *lambda_expression* is classified as an anonymous function ([§12.16](expressions.md#1216-anonymous-function-expressions)). The expression does not have a type, but can be implicitly converted to a compatible delegate type. Some lambda expressions may also be implicitly converted to a compatible expression-tree type.
+An *anonymous_method_expression* or *lambda_expression* is classified as an anonymous function ([§12.16](expressions.md#1216-anonymous-function-expressions)). The expression does not have a type, but can be implicitly converted to a compatible delegate type. Some lambda expressions may also be implicitly converted to a compatible expression tree type.
 
 For the purpose of brevity, this subclause uses the short form for the task types `Task` and `Task<T>` ([§15.15.1](classes.md#15151-general)).
 
@@ -717,13 +717,6 @@ Specifically, an anonymous function `F` is compatible with a delegate type `D`
 
 A lambda expression `F` is compatible with an expression tree type `Expression<D>` if `F` is compatible with the delegate type `D`. This does not apply to anonymous methods, only lambda expressions.
 
-Certain lambda expressions cannot be converted to expression tree types: Even though the conversion *exists*, it fails at compile-time. This is the case if the lambda expression:
-
-- Has a *block* body
-- Contains simple or compound assignment operators
-- Contains a dynamically bound expression
-- Is async
-
 Anonymous functions may influence overload resolution, and participate in type inference. See [§12.6](expressions.md#126-function-members) for further details.
 
 ### 11.7.2 Evaluation of anonymous function conversions to delegate types
@@ -753,9 +746,27 @@ static void F(double[] a, double[] b) {
 ```
 Since the two anonymous function delegates have the same (empty) set of captured outer variables, and since the anonymous functions are semantically identical, the compiler is permitted to have the delegates refer to the same target method. Indeed, the compiler is permitted to return the very same delegate instance from both anonymous function expressions.
 
-### 11.7.3 Evaluation of anonymous function conversions to expression tree types
+### 11.7.3 Evaluation of lambda expression conversions to expression tree types
 
-Conversion of an anonymous function to an expression-tree type produces an expression tree ([§9.6](types.md#96-expression-tree-types)). More precisely, evaluation of the anonymous-function conversion produces an object structure that represents the structure of the anonymous function itself. The precise structure of the expression tree, as well as the exact process for creating it, are implementation-defined.
+Conversion of a lambda expression to an expression tree type produces an expression tree ([§9.6](types.md#96-expression-tree-types)). More precisely, evaluation of the lambda expression conversion produces an object structure that represents the structure of the lambda expression itself.
+
+Not every lambda expression can be converted to expression tree types. The conversion to a compatible delegate type always *exists*, but it may fail at compile-time for implementation-specific reasons.
+
+> *Note*: Common reasons for a lambda expression to fail to convert to an expression tree type include:
+>
+> - It has a block body
+> - It has the `async` modifier
+> - It contains an assignment operator
+> - It contains an `out` or `ref` parameter
+> - It contains a dynamically bound expression
+>
+> *end note*
+
+The following aspects of conversions from lambda expressions to expression trees are implementation-specific:
+
+- The precise structure of the expression tree
+- The exact process for creating it
+- The rules for which lambda expressions can be successfully converted into expression trees, and for which expression tree types.
 
 ## 11.8 Method group conversions
 
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -762,12 +762,6 @@ Not every lambda expression can be converted to expression tree types. The conve
 >
 > *end note*
 
-The following aspects of conversions from lambda expressions to expression trees are implementation-specific:
-
-- The precise structure of the expression tree
-- The exact process for creating it
-- The rules for which lambda expressions can be successfully converted into expression trees, and for which expression tree types.
-
 ## 11.8 Method group conversions
 
 An implicit conversion exists from a method group ([§12.2](expressions.md#122-expression-classifications)) to a compatible delegate type ([§20.4](delegates.md#204-delegate-compatibility)). If `D` is a delegate type, and `E` is an expression that is classified as a method group, then `D` is compatible with `E` if and only if `E` contains at least one method that is applicable in its normal form ([§12.6.4.2](expressions.md#12642-applicable-function-member)) to any argument list ([§12.6.2](expressions.md#1262-argument-lists)) having types and modifiers matching the parameter types and modifiers of `D`, as described in the following.
--- a/standard/types.md
+++ b/standard/types.md
@@ -551,11 +551,9 @@ As a type, type parameters are purely a compile-time construct. At run-time, eac
 
 ***Expression trees*** permit lambda expressions to be represented as data structures instead of executable code. Expression trees are values of ***expression tree types*** of the form `System.Linq.Expressions.Expression<TDelegate>`, where `TDelegate` is any delegate type. For the remainder of this specification we will refer to these types using the shorthand `Expression<TDelegate>`.
 
-If a conversion exists from a lambda expression to a delegate type `D`, a conversion also exists to the expression tree type `Expression<TDelegate>`. Whereas the conversion of a lambda expression to a delegate type generates a delegate that references executable code for the lambda expression, conversion to an expression tree type creates an expression tree representation of the lambda expression.
+If a conversion exists from a lambda expression to a delegate type `D`, a conversion also exists to the expression tree type `Expression<TDelegate>`. Whereas the conversion of a lambda expression to a delegate type generates a delegate that references executable code for the lambda expression, conversion to an expression tree type creates an expression tree representation of the lambda expression. More details of this conversion are provided in [§11.7.3](conversions.md#1173-evaluation-of-lambda-expression-conversions-to-expression-tree-types)).
 
-Expression trees are efficient in-memory data representations of lambda expressions and make the structure of the lambda expression transparent and explicit.
-
-Just like a delegate type `D`, `Expression<TDelegate>` is said to have parameter and return types, which are the same as those of `D`.
+Expression trees are in-memory data representations of lambda expressions and make the structure of the lambda expression transparent and explicit.
 
 > *Example*: The following program represents a lambda expression both as executable code and as an expression tree. Because a conversion exists to `Func<int,int>`, a conversion also exists to `Expression<Func<int,int>>`:
 > ```csharp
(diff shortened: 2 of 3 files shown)


jskeet:
Meeting feedback: restore the references, but state in the text around the reference that it's implementation-specific, so that people don't *expect* more details.

jskeet:
We discussed whether we could fix the ugliness of "the conversion exists but may fail", but decided not to.

jskeet:
This is ready for another look in the meeting later.

jskeet:
This is now *not* ready, as I've messed up with encodings. Jon to reapply the desired changes in a suitably UTF-8 aware editor.

jskeet:
Looks like it was just a single character that was wrong - but that confused GitHub into thinking it wasn't UTF-8 after all.

Fixed in a third commit.

I can't remember whether we actually approved this or not, so I won't merge yet, but let's try to get it done next time.

Nigel-Ecma (changes requested):
The motivation for this PR is good but unfortunately I don't think it's a good idea as the issue is more involved.

§11.7.1 starts:

> An *anonymous_method_expression* or *lambda_expression* is classified as an anonymous function (§12.16). The expression does not have a type, but can be implicitly converted to a compatible delegate type. Some lambda expressions may also be implicitly converted to a compatible expression-tree type.

Could be read as limiting expression-tree conversions to a (subset of) *lambda* expressions, what about anonymous method expressions?

Later in the same clause we get:

> A lambda expression `F` is compatible with an expression tree type` Expression<D>` if `F` is compatible with the delegate type `D`. This does not apply to anonymous methods, only lambda expressions.

> Certain lambda expressions cannot be converted to expression tree types: Even though the conversion *exists*, it fails at compile-time.  This is the case if the lambda expression:
> - Has a block body
> - Contains simple or compound assignment operators
> - Contains a dynamically bound expression
> - Is async

Here it is still talking only about lambda expressions and gives an explicit list of those that cannot be converted to expression tree types. Is the list exhaustive? Could possibly be read either way.

Which gets us to:

> ### 11.7.3 Evaluation of anonymous function conversions to expression tree types

> Conversion of an anonymous function to an expression-tree type produces an expression tree (§9.6). More precisely, evaluation of the anonymous-function conversion produces an object structure that represents the structure of the anonymous function itself. The precise structure of the expression tree, as well as the exact process for creating it, are implementation-defined.

while the PR replaces the last sentence with:

> The following aspects are implementation-defined:
> - The precise structure of the expression tree
> - The exact process for creating it
> - The rules for which lambda expressions can be successfully converted into expression trees, and for which expression tree types.

The original talks about anonymous functions, says the process/result is implementation defined and does not repeat the list above of those lambda expressions which are not convertible.

The PR version contradicts the previous list by saying that those that are not convertible is implementation defined. 

Overall in the original §11.7 there is a lack of clarity of what the rules are for anonymous functions (which the Standard is using to cover both *anonymous_method_expression* and *lambda_expression*) and what rules, if any, apply only to lambda expressions. The original provides a list of lambda expressions which cannot be converted, but it is unclear whether this is exhaustive; the PR contradicts the list by defining it as implementation defined.

Moving now to §9.6, para 2 starts:

> If a conversion exists from a lambda expression to a delegate type `D`, a conversion also exists to the expression tree type `Expression<TDelegate>`.

Seems we're back to lambda expressions, not §11.7.3’s anonymous functions. An x-ref to §11.7(*.x*?) might be helpful.

> Expression trees are efficient in-memory data representations of lambda expressions and make the structure of the lambda expression transparent and explicit.

Definitely *lambda expressions*, and yet §11.7.3 is about conversion of anonymous functions. (BTW “efficient” seems a bit odd as well, unless “implementation defined” is *required* to be efficient.)

And finally we come to:

> - Not all lambda expressions can be converted to expression trees. For instance, lambda expressions with statement bodies, and lambda expressions containing assignment expressions cannot be represented. In these cases, a conversion still exists, but it will fail at compile-time. These exceptions are detailed in §11.7.3).

(The unbalanced “)” is in the original.)

Here it does seem to suggest that the subset of lambda’s which cannot be convert is exhaustive (”are detailed”), but that list is in §11.7.1 not §11.7.3.

The PR changes this bullet to:

> - Not all lambda expressions can be converted to expression trees. For instance, lambda expressions with statement bodies, and lambda expressions containing assignment expressions cannot be represented. In these cases, a conversion still exists, but it will fail at compile-time. The conversion's existence is specified in §11.7.3), but this does not provide details due to the implementation-dependent nature of the conversion.

(The lone “)” survived).

Which moves us back to implementation-defined...

The stated aim of the PR is to remove an x-ref to a list which doesn’t exist, but a list does exist – just in §11.7.1 not §11.7.3.

There is some lack of clarity in the original between:

- what applies to anonymous function and what only applies to lambda expressions; and
- what is implementation-defined and what is not

The issues are larger than the PR tackles, so though the motivation is good I wouldn't recommend adopting it but rather aiming for a PR which tackles all the issues.

jskeet:
(The quotes are from the above comment... hopefully it's intelligible that way.)

> Could be read as limiting expression-tree conversions to a (subset of) lambda expressions, what about anonymous method expressions?

This is deliberate. There is no conversion from anonymous method expressions to expression trees. It really is only lambda expressions, despite the heading of "anonymous function conversions". We should discuss (in a meeting) whether it's worth changing the headings here to be more explicit about only supporting lambda expressions.

> Here it is still talking only about lambda expressions and gives an explicit list of those that cannot be converted to expression tree types. Is the list exhaustive? Could possibly be read either way.

No, the list isn't exhaustive. We should fix it to be explicit that this is *not* exhaustive - or omit it entirely. It's not obvious to me whether *everything* should be implementation-defined, or whether it's useful for the standard to say "The following restrictions definitely apply, and implementations can impose additional restrictions".

I'll fix the closing parenthesis now - thanks for spotting that - but leave the rest of the PR right now; we can discuss the approach we want to take in the next meeting, and I'll update the PR afterwards.

jskeet:
In our last meeting we wanted to know whether a conversion from "anonymous function" to expression tree existed, or only a conversion from "lambda expression". Experimentally, it looks like it's only from lambda expression:

```csharp

int a = 0;
var derived = new DerivedClass();
derived.Method(delegate { return 5; }); // Fine
derived.Method(() => 5); // Fine
derived.Method(() => a = 5); // Fails
derived.Method(() => { return 5; }); // Fails

public class BaseClass
{
    public void Method(Func<int> func) { }
}

public class DerivedClass : BaseClass
{
    public void Method(Expression<Func<int>> expression) { }
}
```

The last two calls fail because the conversion exists but fails; the anonymous method call succeeds because the method in `DerivedClass` isn't applicable.

Next step is to make that clear in the text :)

jskeet:
This is now updated - PTAL, particularly @Nigel-Ecma who had lots of comments before. I'm sure I've missed something.

standard/conversions.md
```diff
@@ -753,9 +746,27 @@ static void F(double[] a, double[] b) {
...
+>
+> - It has a block body
+> - It has the `async` modifier
+> - It contains an assignment operator
+> - It contains an `out` or `ref` parameter
+> - It contains a dynamically bound expression
+>
+> *end note*
+
+The following aspects of conversions from lambda expressions to expression trees are implementation-specific:
```
Nigel-Ecma: This reads to be as out-of-order – the examples in lines 757-761 come _before_ the list of things in lines 767-769 of things which are implantation specific. Surely it should be the other way around?

standard/types.md
```diff
@@ -568,7 +568,7 @@ The exact definition of the generic type `Expression<TDelegate>` as well as the
 
 Two things are important to make explicit:
 
-- Not all lambda expressions can be converted to expression trees. For instance, lambda expressions with statement bodies, and lambda expressions containing assignment expressions cannot be represented. In these cases, a conversion still exists, but it will fail at compile-time. These exceptions are detailed in [§11.7.3](conversions.md#1173-evaluation-of-anonymous-function-conversions-to-expression-tree-types)).
+- Not all lambda expressions can be converted to expression trees. For instance, lambda expressions with statement bodies, and lambda expressions containing assignment expressions cannot be represented. In these cases, a conversion still exists, but it will fail at compile-time.
```
Nigel-Ecma: Starting back at line 556 (I can't attach a comment there) it states that expression trees, an implementation defined type, "are efficient" – maybe "shall be", "must be", "are required to be", but surely not "are"? It carries on to state that they make "make the structure of the lambda expression transparent and explicit" – again this is an implementation defined type, so maybe "shall make…", "must make…" etc but not simply "and make".

In this section we also see repetition of material from §11.7 without at least x-refs, and maybe less repetition is possible.

I think this is proving a hard feature to Standardise because it is essentially implementation defined and the Standard is trying to be implementation neutral but hasn't much to work with, while also implementation specific material (e.g. it being efficient above) has crept in.

Nigel-Ecma (changes requested):
I don't know *what* changes, sorry, but this still doesn't feel right – see comments.

standard/conversions.md
jskeet: I went back and forth on this. I don't mind whether we go from the detail to the summary or vice versa. Let's talk about it in the meeting.

standard/types.md
jskeet: Agree that we shouldn't have efficient in there - removed.

I've reworked this section a bit, restoring the cross-reference (but moving it) and removing duplication. I've retained the part about the `Compile` method, but it's no longer in a list.

jskeet: Also removed the bit about return types and parameters which seemed to be unnecessary.

jskeet (commented):
Updated - PTAL. I've left the changes in a second commit; I'll rebase them down to one when we're happy.

standard/conversions.md
```diff
@@ -753,9 +746,27 @@ static void F(double[] a, double[] b) {
...
 
-Conversion of an anonymous function to an expression-tree type produces an expression tree ([§9.6](types.md#96-expression-tree-types)). More precisely, evaluation of the anonymous-function conversion produces an object structure that represents the structure of the anonymous function itself. The precise structure of the expression tree, as well as the exact process for creating it, are implementation-defined.
+- The precise structure of the expression tree
```
Nigel-Ecma: Not even the imprecise structure of the tree is defined, it may not even be a tree surely? `Expression<>` is an implementation defined type, the implementation provides a compile time conversion (so not a conversion operator? Or a constructor? Just a compilation strategy so not really a “conversion” *per se*?) which results in an opaque value of type `Expression<>`. And there is just one required method on this type `Compile`.

standard/types.md
```diff
@@ -564,25 +562,20 @@ Just like a delegate type `D`, `Expression<TDelegate>` is said to have parameter
 > ```
 > Following these assignments, the delegate `del` references a method that returns `x + 1`, and the expression tree exp references a data structure that describes the expression `x => x + 1`. *end example*
 
-The exact definition of the generic type `Expression<TDelegate>` as well as the precise rules for constructing an expression tree when a lambda expression is converted to an expression tree type, are implementation dependent.
```
Nigel-Ecma, line 567: I think this particular deletion removes all reference to `Expression<>` being (largely) implementation-defined for the types.md, it is mentioned in (the later) conversions.md (see comment there).

standard/types.md
```diff
@@ -564,25 +562,20 @@ Just like a delegate type `D`, `Expression<TDelegate>` is said to have parameter
...
-
-Two things are important to make explicit:
-
-- Not all lambda expressions can be converted to expression trees. For instance, lambda expressions with statement bodies, and lambda expressions containing assignment expressions cannot be represented. In these cases, a conversion still exists, but it will fail at compile-time. These exceptions are detailed in [§11.7.3](conversions.md#1173-evaluation-of-anonymous-function-conversions-to-expression-tree-types)).
-- `Expression<TDelegate>` offers an instance method Compile which produces a delegate of type `TDelegate`:
+`Expression<TDelegate>` offers an instance method `Compile` which produces a delegate of type `TDelegate`:
```
Nigel-Ecma: I’m not sure “offers an” is right here, `Compile` is the only method the otherwise implementation-defined `Expression<>` type is *required* to provide.

standard/types.md
```diff
@@ -551,11 +551,9 @@ As a type, type parameters are purely a compile-time construct. At run-time, eac
...
+If a conversion exists from a lambda expression to a delegate type `D`, a conversion also exists to the expression tree type `Expression<TDelegate>`. Whereas the conversion of a lambda expression to a delegate type generates a delegate that references executable code for the lambda expression, conversion to an expression tree type creates an expression tree representation of the lambda expression. More details of this conversion are provided in [§11.7.3](conversions.md#1173-evaluation-of-lambda-expression-conversions-to-expression-tree-types)).
 
-Expression trees are efficient in-memory data representations of lambda expressions and make the structure of the lambda expression transparent and explicit.
-
-Just like a delegate type `D`, `Expression<TDelegate>` is said to have parameter and return types, which are the same as those of `D`.
+Expression trees are in-memory data representations of lambda expressions and make the structure of the lambda expression transparent and explicit.
```
Nigel-Ecma: `Expression<TDelegate>`'s internals are implementation defined, as far the Standard goes they are *opaque* and not “transparent and explicit”.

This doesn’t mean a particular implementation may not choose to surface the internals, that is allowed by the Standard.

Nigel-Ecma (changes requested):
Apologies but I think we’re still not there on this one.

As far as I can see we are standardising *only*:

- There is a type `Expression<TDelegate>`
- Instances of it are constructed/converted/compiled from a lambda expression at compile time in an implementation defined way.
   - This construction/conversion/compilation may fail at compile time for an implementation defined reason  
- It provides a single method `Compile()` which returns a `TDelegate`
   - This method may fail at run time for an implementation defined reason  

And nothing else.

Unless of course the intention is to Standardise more?

standard/conversions.md
jskeet: I've now removed this list anyway, due to other changes.

standard/conversions.md
jskeet: This is now gone, but there's new wording in types.md.

standard/types.md
jskeet: Changed to just "provides". (I think it's reasonably clear that this means "the implementation is required to provide".)

standard/types.md
jskeet: Right - I've moved some aspects back to types.md.

standard/types.md
jskeet: I've rewritten all of this as a note. I don't even know what "transparent" was meant to be saying before.

jskeet (commented):
I've taken yet another pass at this.

Note that I *don't* want to be in the business of specifying anything extra - if anyone thinks we should be requiring more aspects to be specified, we should discuss this. I'm just trying to make sure it's clearer what the limits of the specification are.

Nigel-Ecma (commented):
I think all vestiges of implementation specific stuff are gone and just the bare bones that are being standardised remain – the process was long but the result is short and sweet :-)

I don't know if the notes at the ends of §9.6 & §11.7.3 are what we wish to say or not, I'll defer to the collective wisdom on that.

So this is an approval on the normative text, leaving the collective wisdom at the meeting to approve the notes :-)

Nigel-Ecma (approved):

PR #362, opened 2021-08-04
Fix an ANTLR syntax issue

BillWagner:
The filename and the grammar name must match for ANTLR to validate the grammar.


Nigel-Ecma (approved):
Fixes what ANTLR's error message said the issue was, I guess we'll find out if ANTLR is keeping anything back as it did with the last config error it reported....

PR #376, opened 2021-09-19
Remove duplicate text and fix table format

RexJaeschke:
The definition of an unmanaged_type was moved to §9.8 by PR https://github.com/dotnet/csharpstandard/pull/233. However, the old, now duplicate, definition still resides in 23.3, “Pointer types.” That duplicate is being deleted.

Also, an Example table does not render correctly; the table header is treated as run-on text to the previous line:


Nigel-Ecma (approved):
LGTM

jskeet (approved):

PR #377, opened 2021-09-22
Empty params arrays are implementation-defined; see issue #200

### LINKED ISSUE #200: Need spec text for new V6 feature "empty array for params parameter"
See https://github.com/dotnet/csharplang/blob/master/proposals/csharp-6.0/empty-params-array.md

Can someone write up this new spec text?

ericlippert:
We decided to make whether empty params arrays are always allocated new or re-used an implementation detail; we also resolved to not mention that some implementations use Array.Empty<T>() -- which caches -- as their implementation choice.

Fixes https://github.com/dotnet/csharpstandard/issues/200


jskeet (approved):

jskeet:
(Good to see other team members taking the same ECMA work scheduling approach as I do...)

jskeet:
Is the first commit in this PR meant to be present? (I only looked at the final one.)
(Ah, I suspect it may go away when we rebase.)

KalleOlaviNiemitalo:
Did you deliberately make this implementation-defined rather than unspecified?

ericlippert:
> Did you deliberately make this implementation-defined rather than unspecified?

(The difference being that an implementation-defined behaviour should be documented, whereas an unspecified behaviour merely must do something sensible.)

I did not give it a lot of thought, no. If anyone wishes to make the argument that it is burdensome on compiler developers to document their choice here, we could change the text. Thanks for the comment!

BillWagner:
> The difference being that an implementation-defined behaviour should be documented, whereas an unspecified behaviour merely must do something sensible.

I looked at our current docs. We don't specify which the implementation does. I have a mild preference for "unspecified", to avoid stating definitively what the implementation does. That way, the product team can make changes based on  any performance or other research.  Adding @jaredpar and @stephentoub for their opinions.

stephentoub:
> an implementation-defined behaviour should be documented

Does documented imply the implementation is guaranteeing that behavior?  While I don't see us changing Roslyn to start allocating new empty arrays here, I hesitate to make documented guarantees about exactly what code it's generating.  In which case, I'd also prefer "unspecified", given the choice.

What's the downside to using unspecified in situations like this?

jaredpar:
I'd prefer unspecified as well. I'm less interested in this particular case (pretty easy to doc what we do now and I don't see a reasonable avenue for it changing). Looking outwards towards `params ReadOnlySpan<T>` though how we allocate the storage will be more complex and I'd prefer unspecified there. Hence use unspecified in both places for consistency.

standard/expressions.md
```diff
@@ -635,25 +635,23 @@ The array co-variance rules ([§17.6](arrays.md#176-array-covariance)) permit a
 > ```
 > the second invocation of `F` causes a `System.ArrayTypeMismatchException` to be thrown because the actual element type of `b` is `string` and not `object`. *end example*
 
-When a function member with a parameter array is invoked in its expanded form, the invocation is processed exactly as if an array creation expression with an array initializer ([§12.7.11.5](expressions.md#127115-array-creation-expressions)) was inserted around the expanded parameters.
+When a function member with a parameter array is invoked in its expanded form, the invocation is processed as if an array creation expression with an array initializer ([§12.7.11.5](expressions.md#127115-array-creation-expressions)) was inserted around the expanded arguments. An empty array is passed when there are no arguments for the parameter array; it is implementation-defined whether the reference passed is to a newly allocated or existing empty array.
```
```suggestion
When a function member with a parameter array is invoked in its expanded form with at least one expanded argument, the invocation is processed as if an array creation expression with an array initializer ([§12.7.11.5](expressions.md#127115-array-creation-expressions)) was inserted around the expanded arguments. An empty array is passed when there are no arguments for the parameter array; it is unspecified whether the reference passed is to a newly allocated or existing empty array.
```

standard/expressions.md
```diff
@@ -635,25 +635,23 @@ The array co-variance rules ([§17.6](arrays.md#176-array-covariance)) permit a
 > ```
 > the second invocation of `F` causes a `System.ArrayTypeMismatchException` to be thrown because the actual element type of `b` is `string` and not `object`. *end example*
 
-When a function member with a parameter array is invoked in its expanded form, the invocation is processed exactly as if an array creation expression with an array initializer ([§12.7.14.5](expressions.md#127145-array-creation-expressions)) was inserted around the expanded parameters.
+When a function member with a parameter array is invoked in its expanded form with at least one expanded argument, the invocation is processed as if an array creation expression with an array initializer ([§12.7.14.5](expressions.md#127115-array-creation-expressions)) was inserted around the expanded arguments. An empty array is passed when there are no arguments for the parameter array; it is unspecified whether the reference passed is to a newly allocated or existing empty array.
```
```suggestion
When a function member with a parameter array is invoked in its expanded form with at least one expanded argument, the invocation is processed as if an array creation expression with an array initializer ([§12.7.14.5](expressions.md#127145-array-creation-expressions)) was inserted around the expanded arguments. An empty array is passed when there are no arguments for the parameter array; it is unspecified whether the reference passed is to a newly allocated or existing empty array.
```

jskeet (approved):
Approved in committee

PR #381, opened 2021-10-05
Provide guidance about creating issues

jskeet:
In particular, request that contributors hold off mentioning
features that aren't in the version we're currently standardising.

We don't want to discourage contributors in general, and if we miss
something we definitely want to know - but having huge numbers of
issues for features we aren't ready to standardize ends up making it
harder to find issues we *do* want to address.

Example issue: #296 - we'll close that once this change is in.


CONTRIBUTING.md
```diff
@@ -18,6 +18,16 @@ This article classifies three types of changes to the standard:
...
 
+## Filing issues
+
+If you have found a problem in the standard, please let us know by filing an issue. If possible, please let us know what kind of problem you've found:
+
+- Does an existing implementation (e.g. Roslyn / .NET) behave contrary to the standard? If you can provide a complete example program and explain how it actually behaves and how you think it should behave according to the standard (ideally referencing all the relevant sections of the standard) that's enormously helpful
+- Is there a relatively simple typo in the standard? Feel free to create a PR to address this directly should you wish, but it's equally fine just to create an issue. Please be as specific as possible about where the typo is and what you'd expect instead.
+- Is there a C# feature missing in the standard?
+  - If this is with respect to a feature introduced an older version, or the version that is currently the draft specified in the default branch, we definitely want to capture anything missing. Please check for existing issues and file a new one if the feature isn't already covered.
```
BillWagner: I like this. Should we point people to the admin folder's feature tracker:

```suggestion
  - If this is with respect to a feature introduced an older version, or the version that is currently the draft specified in the default branch, we definitely want to capture anything missing. The [feature tracker](https://github.com/dotnet/csharpstandard/blob/draft-v6/admin/v6-feature-tracker.md) lists features for the current version. Please check that document and existing issues then file a new one if the feature isn't already covered.
```

BillWagner (approved):
This LGTM @jskeet 

I did have one suggestion, if you agree. (Totally optional, what you have is fine)

CONTRIBUTING.md
jskeet: Yes, absolutely :)

PR #387, opened 2021-11-08
fix join sample (sample ignored rule "A query expression with a join …

bernd5:
fix join sample (sample ignored rule "A query expression with a join … clause without an into followed by a select clause...")


jskeet:
Bill, do you have time to look at this one? I've got a comment from Nigel that I've been putting off for a week but which really deserves attention...

jskeet:
(Closing was accidental.)

BillWagner (approved):
Thanks @bernd5 

I agree with your changes.

I'll approve this PR, and I'll :shipit: now.

PR #390, opened 2021-11-09
Interpolated strings

Nigel-Ecma:
…e a first draft.

Once the decision on how much of the ANLTR language we would use was made the grammar
was fairly straghtforward.

The prose describing the feature is more of a challenge as it is a language-level
surfacing of features of the Standard Library, and those features are either defined
in another Standard or not defined in any Standard but which we pretend are…

So the challenge is how much of the semantics should be included somewhere in this
Standard: what is a `FormattableString` vs. a `string` for the result type of an
*interpolated_string_expression*; what is the field width, what is the format, etc.?

This draft punts on some of these (e.g. the first one), includes details (e.g. for the the second
one), and even a new *informative* appendix, §C.4 (for the third one).

Is this balance right?

Well that depends on how we view our aim here…

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1153,11 +1153,11 @@ A *primary_expression* that consists of a *literal* ([§7.4.5](lexical-structure
 
 ### §interpolated-string-expressions Interpolated string expressions
 
-An *interpolated_string_expression* consists of a `$` character immediately followed by text within `"` characters. Within the quoted text there are zero or more **_interpolations_** delimited by `{` and `}` characters, each of which encloses an *expression* and optional formatting specifications.
+An *interpolated_string_expression* consists of a `$` character immediately followed by text within `"` characters. Within the quoted text there are zero or more ***interpolations*** delimited by `{` and `}` characters, each of which encloses an *expression* and optional formatting specifications.
 
 Interpolated string expressions have two forms; regular (*interpolated_regular_string_expression*)
 and verbatim (*interpolated_verbatim_string_expression*); which are lexically similar to, but differ semantically from, the two forms of string
-literals ([§7.4.5.6](lexical-structure.md#7456-string-literals)).
+literals (§7.4.5.6).
 
 ```ANTLR
 interpolated_string_expression
@@ -1185,6 +1185,7 @@ Interpolated_Regular_String_Start
     ;
 
 // the following three lexical rules are context sensitive, see details below
+
 Interpolated_Regular_String_Mid
     : Interpolated_Regular_String_Element+
     ;
@@ -1207,8 +1208,8 @@ fragment Interpolated_Regular_String_Element
     ;
 
 fragment Interpolated_Regular_String_Character
-    // Any character except " (U+0022), \\ (U+005C), { (U+007B), } (U+007D), and New_Line_Character
-    : ~["\\{}\u000D\u000A\u0085\u2028\u2029]
+    : ~["\\{}\u000D\u000A\u0085\u2028\u2029]    // Any character except " (U+0022), \\ (U+005C),
+                                                //  { (U+007B), } (U+007D), and New_Line_Character
     ;
 
 // interpolated verbatim string expressions
@@ -1227,6 +1228,7 @@ Interpolated_Verbatim_String_Start
     ;
 
 // the following three lexical rules are context sensitive, see details below
+
 Interpolated_Verbatim_String_Mid
     : Interpolated_Verbatim_String_Element+
     ;
@@ -1247,8 +1249,7 @@ fragment Interpolated_Verbatim_String_Element
     ;
 
 fragment Interpolated_Verbatim_String_Character
-    // Any character except " (U+0022), { (U+007B) and } (U+007D)
-    : ~["{}]
+    : ~["{}]    // Any character except " (U+0022), { (U+007B) and } (U+007D)
     ;
 
 // lexical fragments used by both regular and verbatim interpolated strings
@@ -1267,26 +1268,26 @@ Six of the lexical rules defined above are *context sensitive* as follows:
 | Rule | Contextual Requirements |
 | :--- | :---------------------- |
 | *Interpolated_Regular_String_Mid* | Only recognised after an *Interpolated_Regular_String_Start*, between any following interpolations, and before the corresponding *Interpolated_Regular_String_End*. |
-| *Regular_Interpolation_Format* | Only recognised immediately before the closing delimiter (`}`) of a *regular_interpolation*. |
+| *Regular_Interpolation_Format* | Only recognised within a *regular_interpolation* and when the starting colon (:) is not nested within any kind of bracket (parentheses/braces/square). |
 | *Interpolated_Regular_String_End* | Only recognised after an *Interpolated_Regular_String_Start* and only if any intervening tokens are either *Interpolated_Regular_String_Mid*s or tokens that can be part of *regular_interpolation*s, including tokens for any *interpolated_regular_string_expression*s contained within such interpolations. |
 | *Interpolated_Verbatim_String_Mid* *Verbatim_Interpolation_Format* *Interpolated_Verbatim_String_End* | Recognition of these three rules follows that of the corresponding rules above with each mentioned *regular* grammar rule replaced by the corresponding *verbatim* one. |
 
-*Note:* The above rules are context sensitive as their definitions overlap with those of
+> *Note:* The above rules are context sensitive as their definitions overlap with those of
 other tokens in the language. *end note*
 
-*Note:* The above grammar is not ANTLR-ready due to the context sensitive lexical rules. As with
-other lexer generators ANTLR supports context sensitive lexical rule, for example using its *lexical modes*,
+> *Note:* The above grammar is not ANTLR-ready due to the context sensitive lexical rules. As with
+other lexer generators ANTLR supports context sensitive lexical rules, for example using its *lexical modes*,
 but this is an implementation detail and therefore not part of this Standard. *end note*
 
 An *interpolated_string_expression* is classified as a value. If it is immediately converted to `System.IFormattable` or `System.FormattableString` with an implicit interpolated string conversion (§implicit-interpolated-string-conversions), the interpolated string expression has that type. Otherwise, it has the type `string`.
 
-*Note:* The differences between the possible types an *interpolated_string_expression* may be determined from the documentation for `System.String` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) and `System.FormattableString` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)). *end note*
+> *Note:* The differences between the possible types an *interpolated_string_expression* may be determined from the documentation for `System.String` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) and `System.FormattableString` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)). *end note*
 
-The meaning on an interpolation, both *regular_interpolation* and *verbatim_interpolation*, is to format the value of the *expression* as a `string` either according to the format specified by the *Regular_Interpolation_Format* or *Verbatim_Interpolation_Format*, or according to a default format for the type of *expression*. The formatted string is then modified by the *interpolation_minimum_width*, if any, to produce the final `string` to be interpolated into the *interpolated_string_expression*.
+The meaning of an interpolation, both *regular_interpolation* and *verbatim_interpolation*, is to format the value of the *expression* as a `string` either according to the format specified by the *Regular_Interpolation_Format* or *Verbatim_Interpolation_Format*, or according to a default format for the type of *expression*. The formatted string is then modified by the *interpolation_minimum_width*, if any, to produce the final `string` to be interpolated into the *interpolated_string_expression*.
 
-*Note:* How the default format for a type is determined is detailed in the documentation for `System.String` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) and `System.FormattableString` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)). Descriptions of standard formats, which are identical for *Regular_Interpolation_Format* and *Verbatim_Interpolation_Format*, may be found in the documentation for `System.IFormattable` ([§C.4](standard-library.md#c4-format-specifications)) and in other types in the standard library ([§C](standard-library.md#annex-c-standard-library)). *end note*
+> *Note:* How the default format for a type is determined is detailed in the documentation for `System.String` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) and `System.FormattableString` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)). Descriptions of standard formats, which are identical for *Regular_Interpolation_Format* and *Verbatim_Interpolation_Format*, may be found in the documentation for `System.IFormattable` ([§C.4](standard-library.md#c4-format-specifications)) and in other types in the standard library ([§C](standard-library.md#annex-c-standard-library)). *end note*
 
-In an *interpolation_minimum_width* the *constant_expression* shall have an implicit conversion to `int` and it termed the *alignment*, the *field width* is the absolute value of the alignment:
+In an *interpolation_minimum_width* the *constant_expression* shall have an implicit conversion to `int`. Let the *field width* be the absolute value of this *constant_expression* and the *alignment* be the sign (positive or negative) of the value of this *constant_expression*:
 
 - If the value of field width is less than or equal to the length of the formatted string the formatted string is not modified.
 - Otherwise the formatted string is padded with white space characters so that its length is equal to field width:
@@ -1321,7 +1322,9 @@ This example uses the following format specification features:
 
  - the `X` format specification which formats integers as uppercase hexadecimal,
  - the default format for a `string` value is the value itself,
- - positive alignment values that right-justify within the specified field width, and
+ - positive alignment values that right-justify within the specified minimum field width,
+ - negative alignment values that left-justify within the specified minimum field width,
+ - defined constants for the *interpolation_minimum_width*, and
  - that `{{` and `}}` are formatted as `{` and `}` respectively.
 
 Given:
@@ -1329,6 +1332,7 @@ Given:
 ```csharp
 string text = "red";
 int number = 14;
+const int width = -4;
 ```
 
 Then:
@@ -1338,6 +1342,7 @@ Then:
 | `$"{text}"`                          | `string.Format("{0}", text)`                                  | `"red"`      |
 | `$"{{text}}"`                        | `string.Format("{{text}})`                                    | `"{text}"`   |
 | `$"{ text , 4 }"`                    | `string.Format("{0,4}", text)`                                | `" red"`     |
+| `$"{ text , width }"`                | `string.Format("{0,-4}", text)`                               | `"red "`     |
 | `$"{number:X}"`                      | `string.Format("{0:X}", number)`                              | `"E"`        |
 | `$"{text + '?'} {number % 3}"`       | `string.Format("{0} {1}", text + '?', number % 3)`            | `"red? 2"`   |
 | `$"{text + $"[{number}]"}"`          | `string.Format("{0}", text + string.Format("[{0}]", number))` | `"red[14]"`  |
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1106,6 +1106,7 @@ primary_expression
 
 primary_no_array_creation_expression
     : literal
+    | interpolated_string_expression
     | simple_name
     | parenthesized_expression
     | member_access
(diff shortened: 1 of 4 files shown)


standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+// the following three lexical rules are context sensitive, see details below
+Interpolated_Verbatim_String_Mid
+    : Interpolated_Verbatim_String_Element+
+    ;
+
+Verbatim_Interpolation_Format
+    : ':' Interpolated_Verbatim_String_Element+
+    ;
+
+Interpolated_Verbatim_String_End
```
jskeet, line 1240: Do we particularly need this to be separate from Interpolated_Regular_String_End? Could we just have Interpolated_String_End and use it in both rules? Or is that harder due to the context-sensitive nature?

Same applies for some other rules as well, I believe - it's just that this was the simplest one to validate.

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+
+regular_interpolation
+    : expression (',' interpolation_minimum_width)? Regular_Interpolation_Format?
+    ;
+
+interpolation_minimum_width
+    : constant_expression
+    ;
+
+Interpolated_Regular_String_Start
```
jskeet, line 1183: I can't remember the different casing rules we've got, and I probably couldn't check whether these are appropriate anyway. I strongly suspect they're all fine, but I bring it up so that someone with more of a clue can review that aspect specifically.

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+| *Regular_Interpolation_Format* | Only recognised immediately before the closing delimiter (`}`) of a *regular_interpolation*. |
+| *Interpolated_Regular_String_End* | Only recognised after an *Interpolated_Regular_String_Start* and only if any intervening tokens are either *Interpolated_Regular_String_Mid*s or tokens that can be part of *regular_interpolation*s, including tokens for any *interpolated_regular_string_expression*s contained within such interpolations. |
+| *Interpolated_Verbatim_String_Mid* *Verbatim_Interpolation_Format* *Interpolated_Verbatim_String_End* | Recognition of these three rules follows that of the corresponding rules above with each mentioned *regular* grammar rule replaced by the corresponding *verbatim* one. |
+
+*Note:* The above rules are context sensitive as their definitions overlap with those of
+other tokens in the language. *end note*
+
+*Note:* The above grammar is not ANTLR-ready due to the context sensitive lexical rules. As with
+other lexer generators ANTLR supports context sensitive lexical rule, for example using its *lexical modes*,
```
jskeet: lexical rule => lexical rules

(unless I'm misreading)

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+
+An *interpolated_string_expression* is classified as a value. If it is immediately converted to `System.IFormattable` or `System.FormattableString` with an implicit interpolated string conversion (§implicit-interpolated-string-conversions), the interpolated string expression has that type. Otherwise, it has the type `string`.
+
+*Note:* The differences between the possible types an *interpolated_string_expression* may be determined from the documentation for `System.String` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) and `System.FormattableString` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)). *end note*
+
+The meaning on an interpolation, both *regular_interpolation* and *verbatim_interpolation*, is to format the value of the *expression* as a `string` either according to the format specified by the *Regular_Interpolation_Format* or *Verbatim_Interpolation_Format*, or according to a default format for the type of *expression*. The formatted string is then modified by the *interpolation_minimum_width*, if any, to produce the final `string` to be interpolated into the *interpolated_string_expression*.
```
jskeet: meaning on => meaning of

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+
+*Note:* How the default format for a type is determined is detailed in the documentation for `System.String` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) and `System.FormattableString` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)). Descriptions of standard formats, which are identical for *Regular_Interpolation_Format* and *Verbatim_Interpolation_Format*, may be found in the documentation for `System.IFormattable` ([§C.4](standard-library.md#c4-format-specifications)) and in other types in the standard library ([§C](standard-library.md#annex-c-standard-library)). *end note*
+
+In an *interpolation_minimum_width* the *constant_expression* shall have an implicit conversion to `int` and it termed the *alignment*, the *field width* is the absolute value of the alignment:
```
jskeet: it termed => is termed

I'd suggest a semi-colon before "the *field width*" rather than a comma... or possibly even a period, with a new sentence at the end of the paragraph of something like: "The alignment and field width are used to determine the padding, if any, to be used when formatting the string:"

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+int number = 14;
+```
+
+Then:
+
+| Interpolated String Expression       | Equivalent Meaning As `string`                                | Value        |
+| :----------------------------------- | :------------------------------------------------------------ | :----------- |
+| `$"{text}"`                          | `string.Format("{0}", text)`                                  | `"red"`      |
+| `$"{{text}}"`                        | `string.Format("{{text}})`                                    | `"{text}"`   |
+| `$"{ text , 4 }"`                    | `string.Format("{0,4}", text)`                                | `" red"`     |
```
jskeet, line 1344: Is it worth having an example that shows the interpolation_minimum_width being an expression using a named constant rather than just digits?

standard/standard-library.md
```diff
@@ -658,3 +652,400 @@ namespace System.Runtime.CompilerServices
...
+
+The meaning of the formats, as used in interpolated string expressions (§interpolated-string-expressions), are defined in ISO/IEC 23271:2012. For convenience the following text is copied from the description of `System.IFormatable`.
+
+**This text is informative.**
+
+A *format* is a string that describes the appearance of an object when
+it is converted to a string. Either standard or custom formats can be used. A
+standard format takes the form *Axx*, where *A* is a single
+alphabetic character called the *format specifier*
+, and *xx* is an integer between zero and 99 inclusive, called the *precision specifier*. The format specifier controls the type
```
jskeet: Move the comma to the previous line, here and below? (In general we might want to rewrap this paragraph.)

standard/standard-library.md
```diff
@@ -658,3 +652,400 @@ namespace System.Runtime.CompilerServices
...
+of formatting applied to the value being represented as a string. The
+*precision specifier* controls the number 
+of significant digits or decimal places in the string, if applicable. [*Note:* For the list of standard format specifiers, see the
+table below. Note that a given data type, such as `System.Int32`
+, might not support one
+or more of the standard format specifiers.]
+
+[*Note:* When a format includes symbols that vary by culture, such as the currency
+symbol included by the ‘C’ and ‘c’ formats, a formatting object supplies the
```
jskeet: Do we want to use curly quotes here?

standard/standard-library.md
```diff
@@ -658,3 +652,400 @@ namespace System.Runtime.CompilerServices
...
+
+**This text is informative.**
+
+A *format* is a string that describes the appearance of an object when
+it is converted to a string. Either standard or custom formats can be used. A
+standard format takes the form *Axx*, where *A* is a single
+alphabetic character called the *format specifier*
+, and *xx* is an integer between zero and 99 inclusive, called the *precision specifier*. The format specifier controls the type
+of formatting applied to the value being represented as a string. The
+*precision specifier* controls the number
```
jskeet, line 667: Precision specifiers are only relevant for numeric format strings, of course... there are standard formats for date/time values too.

jskeet: (I appreciate this is a copy/paste job, and we may well not want to bother changing it at all.)

standard/standard-library.md
```diff
@@ -658,3 +652,400 @@ namespace System.Runtime.CompilerServices
...
+Class Library, the formatting object for system-supplied numeric types is a
+`System.Globalization.NumberFormatInfo` instance. For `System.DateTime` instances, a 
+`System.Globalization.DateTimeFormatInfo` is
+used.]
+
+The following table describes the standard format specifiers and associated formatting
+object members that are used with numeric data types in the Base Class
+Library.
+
+<table>
```
jskeet, line 678: I seem to remember that I make very specific assumptions about the formats of tables in the Word converter. This may very well not work at the moment.

standard/standard-library.md
```diff
@@ -658,3 +652,400 @@ namespace System.Runtime.CompilerServices
...
+Note that for fixed-point format strings (strings not containing an ‘E0’,
+‘E+0’, ‘E-0’, ‘e0’, ‘e+0’, or ‘e-0’), numbers are rounded to as many decimal
+places as there are Zero or Digit placeholders to the right of the decimal
+separator. If the custom format does not contain a decimal separator, the number is
+rounded to the nearest integer. If the number has more digits than there are
+Zero or Digit placeholders to the left of the decimal separator, the extra
+digits are copied to the output string immediately before the first Zero or
+Digit placeholder.
+
+A custom format can contain
```
jskeet, line 1002: Again, this is all about numeric format strings.

jskeet (commented):
Various comments - but above all, thanks for your work on this.

I'm not at all sure about C.4 at the moment. It doesn't sit quite right, but I don't know the best solution to that. I need to ponder, then look forward to a discussion in the meeting next week.

standard/expressions.md
Nigel-Ecma: Correct, it is the context-sensitivity.

In the working ANTLR grammar there are two (in addition to the standard one) modes, one for each of regular & verbatim interpolated strings. In these two modes some rules are different while others are the same but with different names (as there is a single scope for names). This is a common situation when using modes, nothing special for the use case here.

standard/expressions.md
Nigel-Ecma: I suspect it looks a bit off as this is the first place where we really intermingle lexer & parser rules a lot. The casing is correct or ANTLR would have barfed.

standard/expressions.md
Nigel-Ecma: Thanks, will be fixed.

standard/expressions.md
Nigel-Ecma: Thanks, will be fixed.

standard/expressions.md
Nigel-Ecma: Will fix the typo.

The prose here is not good. The source material often uses "alignment" when they mean "field width" and generally mixes the two up. I've partially fixed this but it is still wrong – the alignment is the *sign* of the *constant_expression* (not its value), the field width is its absolute value. More wordsmithing required...

standard/standard-library.md
Nigel-Ecma: See below…

standard/standard-library.md
Nigel-Ecma: Keep going…

standard/standard-library.md
Nigel-Ecma: I too pondered over how much to change this. The original style and formatting doesn't even fit the rest of the CLI Standard, but then it is distributed as an XML file that was mechanically produced from CLR documentation so it wasn't really noticable. I mainly did a mechanical translation into markdown and then a few hand tweaks so it's still recognisable as the source material.

This is the largest single block of format specs I spotted but it is about numeric types, the closing para directs the reader to `Enum` and `DateTime`. How much to include?

I think TG2 needs to ask whether it should foster (not adopt) the whole of the standard library docs while TG3 is in hibernation, to be returned when TG3 wakes up. That would of course address the "how much" issue, and would allow TG2 to remove §C.3 as it would be empty and TG2 would no longer have to fudge and pretend these the types in §C.3 have been Standardised…

standard/standard-library.md
Nigel-Ecma: I think it looks like the other tables, and the Word Convertor didn't barf but I've not seen its output…

standard/standard-library.md
Nigel-Ecma: Look two comments back

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+*Note:* The above rules are context sensitive as their definitions overlap with those of
+other tokens in the language. *end note*
+
+*Note:* The above grammar is not ANTLR-ready due to the context sensitive lexical rules. As with
+other lexer generators ANTLR supports context sensitive lexical rule, for example using its *lexical modes*,
+but this is an implementation detail and therefore not part of this Standard. *end note*
+
+An *interpolated_string_expression* is classified as a value. If it is immediately converted to `System.IFormattable` or `System.FormattableString` with an implicit interpolated string conversion (§implicit-interpolated-string-conversions), the interpolated string expression has that type. Otherwise, it has the type `string`.
+
+*Note:* The differences between the possible types an *interpolated_string_expression* may be determined from the documentation for `System.String` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) and `System.FormattableString` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)). *end note*
```
BillWagner: More a note for Nigel:

If you just write "§C.2" and "§C.3", the section numbering tool will convert that text to the correct link.

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+
+| Rule | Contextual Requirements |
+| :--- | :---------------------- |
+| *Interpolated_Regular_String_Mid* | Only recognised after an *Interpolated_Regular_String_Start*, between any following interpolations, and before the corresponding *Interpolated_Regular_String_End*. |
+| *Regular_Interpolation_Format* | Only recognised immediately before the closing delimiter (`}`) of a *regular_interpolation*. |
+| *Interpolated_Regular_String_End* | Only recognised after an *Interpolated_Regular_String_Start* and only if any intervening tokens are either *Interpolated_Regular_String_Mid*s or tokens that can be part of *regular_interpolation*s, including tokens for any *interpolated_regular_string_expression*s contained within such interpolations. |
+| *Interpolated_Verbatim_String_Mid* *Verbatim_Interpolation_Format* *Interpolated_Verbatim_String_End* | Recognition of these three rules follows that of the corresponding rules above with each mentioned *regular* grammar rule replaced by the corresponding *verbatim* one. |
+
+*Note:* The above rules are context sensitive as their definitions overlap with those of
+other tokens in the language. *end note*
```
BillWagner: All our other "Note"s have been indented:

```suggestion
> *Note:* The above rules are context sensitive as their definitions overlap with those of other tokens in the language. *end note*
```

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+
+*Note:* The above grammar is not ANTLR-ready due to the context sensitive lexical rules. As with
+other lexer generators ANTLR supports context sensitive lexical rule, for example using its *lexical modes*,
+but this is an implementation detail and therefore not part of this Standard. *end note*
```
BillWagner: Same comment (and made the change Jon suggested):

```suggestion
> *Note:* The above grammar is not ANTLR-ready due to the context sensitive lexical rules. As with other lexer generators ANTLR supports context sensitive lexical rules, for example using its *lexical modes*, but this is an implementation detail and therefore not part of this Standard. *end note*
```

standard/expressions.md
BillWagner: Note formatting:

```suggestion
> *Note:* The differences between the possible types an *interpolated_string_expression* may be determined from the documentation for `System.String` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) and `System.FormattableString` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)). *end note*
```

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+
+The meaning on an interpolation, both *regular_interpolation* and *verbatim_interpolation*, is to format the value of the *expression* as a `string` either according to the format specified by the *Regular_Interpolation_Format* or *Verbatim_Interpolation_Format*, or according to a default format for the type of *expression*. The formatted string is then modified by the *interpolation_minimum_width*, if any, to produce the final `string` to be interpolated into the *interpolated_string_expression*.
+
+*Note:* How the default format for a type is determined is detailed in the documentation for `System.String` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) and `System.FormattableString` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)). Descriptions of standard formats, which are identical for *Regular_Interpolation_Format* and *Verbatim_Interpolation_Format*, may be found in the documentation for `System.IFormattable` ([§C.4](standard-library.md#c4-format-specifications)) and in other types in the standard library ([§C](standard-library.md#annex-c-standard-library)). *end note*
```
BillWagner: same:

```suggestion
> *Note:* How the default format for a type is determined is detailed in the documentation for `System.String` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) and `System.FormattableString` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)). Descriptions of standard formats, which are identical for *Regular_Interpolation_Format* and *Verbatim_Interpolation_Format*, may be found in the documentation for `System.IFormattable` ([§C.4](standard-library.md#c4-format-specifications)) and in other types in the standard library ([§C](standard-library.md#annex-c-standard-library)). *end note*
```

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
 
 A *primary_expression* that consists of a *literal* ([§7.4.5](lexical-structure.md#745-literals)) is classified as a value.
 
+### §interpolated-string-expressions Interpolated string expressions
+
+An *interpolated_string_expression* consists of a `$` character immediately followed by text within `"` characters. Within the quoted text there are zero or more **_interpolations_** delimited by `{` and `}` characters, each of which encloses an *expression* and optional formatting specifications.
```
BillWagner: Q for @jskeet 

Will the markdown converter parse the `**_` for bold italic, or should we use `***`?

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+
+Interpolated string expressions have two forms; regular (*interpolated_regular_string_expression*)
+and verbatim (*interpolated_verbatim_string_expression*); which are lexically similar to, but differ semantically from, the two forms of string
+literals ([§7.4.5.6](lexical-structure.md#7456-string-literals)).
```
BillWagner: Note for @Nigel-Ecma 

You can just write the section number for the link, and the tool will fill in the correct section:

```suggestion
literals (§7.4.5.6).
```

standard/expressions.md
```diff
@@ -1150,6 +1151,200 @@ object o = (new int[3])[1];
...
+    : '}}'
+    ;
+```
+
+Six of the lexical rules defined above are *context sensitive* as follows:
+
+| Rule | Contextual Requirements |
+| :--- | :---------------------- |
+| *Interpolated_Regular_String_Mid* | Only recognised after an *Interpolated_Regular_String_Start*, between any following interpolations, and before the corresponding *Interpolated_Regular_String_End*. |
+| *Regular_Interpolation_Format* | Only recognised immediately before the closing delimiter (`}`) of a *regular_interpolation*. |
```
BillWagner: Should this key on the `:` in the interpolation rather than "Only recognized immediately before the closing `}` delimiter?  Would this rule mean that a conditional (`? :`) expression should parse correctly without parentheses if there's an additional format specifier after the expression?

standard/expressions.md
BillWagner: Proposal: If the only remaining issue is with some of the prose, let's merge this and open a new issue to update the prose in this section.

standard/standard-library.md
BillWagner: Hmm. I'd be concerned unless we can expand the membership in TG2, including some people who have more expertise in this area.

BillWagner (approved):
Overall, I like this. I had a few smaller comments, but none that are required.

I'm ready to approve it.

standard/expressions.md
Nigel-Ecma: Fixed

standard/expressions.md
Nigel-Ecma: Fixed

standard/expressions.md
Nigel-Ecma: Fixed

standard/expressions.md
Nigel-Ecma: Fixed

standard/expressions.md
Nigel-Ecma: OK, changed. Does it make any difference as the long hand version will be there next time, or is it just to help reduce errors?

BillWagner: It won't make any difference later. It's to save you time when you add links.

standard/expressions.md
Nigel-Ecma: The tricky part here is the format can contain just about anything, including parentheses etc. So the rule matches if there is a colon, some text, and then a following '}' (which is not part of the token itself) *and* the whole token is at the outmost level.

What we're not making clear here, because its an implementation thing, is that in order to implement the context sensitivity you need to track *all* bracket kinds.

The question is whether the text here is a correct/sufficient description so that implementors can produce a correct implementation. I'll elaborate on the TG2 list…

BillWagner: I like the alternative you suggested to the committee.  Adding it here for discussion:

```suggestion
| *Regular_Interpolation_Format* | Only recognised within a *regular_interpolation*, when the starting colon (`:`) is not nested within any kind of bracket (parentheses/braces/square), and when the following brace (`}`) is the terminating delimiter of a *regular_interpolation*. |
```

standard/expressions.md
jskeet: SGTM.

standard/expressions.md
jskeet: Nigel will tidy it up a little, but not worry too much about achieving perfection: we can adjust slightly later.

standard/expressions.md
jskeet: Good question - no idea; I believe `***` would work though and is what we've done elsewhere.

standard/standard-library.md
jskeet: Resolving not to change.

standard/expressions.md
jskeet: Agreed with the suggestion without `, and when the following brace (`}`) is the terminating delimiter of a *regular_interpolation*`.

standard/expressions.md
jskeet: Example of valid code that surprised me:

```csharp
const int Width = 4;
string text = "abc";
string formatted = $"{text, Width}";
```

standard/standard-library.md
jskeet: Resolving as we are going to keep just the XML as-is.

jskeet: Unresolving due to an extra space before the comma being incorrect.

standard/standard-library.md
jskeet: Jon to attach a screenshot after fetching and running the Word converter.

standard/standard-library.md
```diff
@@ -658,3 +652,400 @@ namespace System.Runtime.CompilerServices
...
+`NaN`,
+`PositiveInfinity`, or `NegativeInfinity`, the format 
+specifier is ignored, and one of the following is returned: `System.Globalization.NumberFormatInfo.NaNSymbol`, `System.Globalization.NumberFormatInfo.PositiveInfinitySymbol`, or `System.Globalization.NumberFormatInfo.NegativeInfinitySymbol`.
+
+A custom format is any string specified as a format that
+is not in the form of a standard format string (Axx) described above. The
+following table describes the characters that are used in constructing custom
+formats.
+
+<table>
```
jskeet, line 868: Another table for Jon to screenshot

jskeet:
Resolution from meeting of 2021-11-17:

- Nigel to check in new version after applying suggestions, typo fixes etc as per comments
- Nigel to change from draft to "ready to review"
- Jon to check all changes decided in meeting have been applied
- Jon to run Word converter and take screenshots of long comment in grammar, and tables in the appendix C.4
- Mads to review and confirm he's happy (as he wasn't at the meeting)
- Nigel to press the button and merge :)

standard/expressions.md
```diff
@@ -1150,6 +1151,205 @@ object o = (new int[3])[1];
...
+fragment Interpolated_Regular_String_Element
+    : Interpolated_Regular_String_Character
+    | Simple_Escape_Sequence
+    | Hexadecimal_Escape_Sequence
+    | Unicode_Escape_Sequence
+    | Open_Brace_Escape_Sequence
+    | Close_Brace_Escape_Sequence
+    ;
+
+fragment Interpolated_Regular_String_Character
```
jskeet, line 1210: This didn't come out well in the Word converter :(

Screenshot (15)

I'll think about options here. Given that it's only a single rule, how would you feel about putting the comments before the production?

standard/standard-library.md
```diff
@@ -658,3 +652,393 @@ namespace System.Runtime.CompilerServices
...
+
+**This text is informative.**
+
+A *format* is a string that describes the appearance of an object when
+it is converted to a string. Either standard or custom formats can be used. A
+standard format takes the form *Axx*, where *A* is a single
+alphabetic character called the *format specifier*, and *xx* is an integer between zero and 99 inclusive, called the *precision specifier*. The format specifier controls the type
+of formatting applied to the value being represented as a string. The
+*precision specifier* controls the number 
+of significant digits or decimal places in the string, if applicable. [*Note:* For the list of standard format specifiers, see the
```
jskeet: This note needs to be changed to use the format we use elsewhere. (Currently the converter thinks it's a link.) I'd suggest making it a new paragraph - I'm not sure how "inline note which isn't in a list" works at the moment.

standard/standard-library.md
```diff
@@ -658,3 +652,393 @@ namespace System.Runtime.CompilerServices
...
+table below. Note that a given data type, such as `System.Int32`, might not support one
+or more of the standard format specifiers.]
+
+[*Note:* When a format includes symbols that vary by culture, such as the currency
```
jskeet: Ditto (note format)

standard/standard-library.md
jskeet: Nope, the Word converter fails entirely for this. This will take a fair amount of work... I'll see if the same table converter that manages to handle the table in the unsafe types section would work here - but it does mean a rather larger tweak of the XML we're "copying" the content from.

jskeet (approved):

standard/standard-library.md
```diff
@@ -658,3 +652,393 @@ namespace System.Runtime.CompilerServices
...
+
+A *format* is a string that describes the appearance of an object when
+it is converted to a string. Either standard or custom formats can be used. A
+standard format takes the form *Axx*, where *A* is a single
+alphabetic character called the *format specifier*, and *xx* is an integer between zero and 99 inclusive, called the *precision specifier*. The format specifier controls the type
+of formatting applied to the value being represented as a string. The
+*precision specifier* controls the number 
+of significant digits or decimal places in the string, if applicable. [*Note:* For the list of standard format specifiers, see the
+table below. Note that a given data type, such as `System.Int32`, might not support one
+or more of the standard format specifiers.]
```
BillWagner: Suggestion for formatting:

```suggestion
of significant digits or decimal places in the string, if applicable. 

> *Note:* For the list of standard format specifiers, see the table below. Note that a given data type, such as `System.Int32`, might not support one or more of the standard format specifiers. *end note*
```

standard/standard-library.md
```diff
@@ -658,3 +652,393 @@ namespace System.Runtime.CompilerServices
...
+symbol included by the ‘C’ and ‘c’ formats, a formatting object supplies the
+actual characters used in the string representation. A method might include a
+parameter to pass a `System.IFormatProvider` object that supplies a
+formatting object, or the method might use the default formatting object, which
+contains the symbol definitions for the current culture. The current culture
+typically uses the same set of symbols used system-wide by default. In the Base
+Class Library, the formatting object for system-supplied numeric types is a
+`System.Globalization.NumberFormatInfo` instance. For `System.DateTime` instances, a 
+`System.Globalization.DateTimeFormatInfo` is
+used.]
```
BillWagner: Suggestion to use the standard *note* format:

```suggestion
> *Note:* When a format includes symbols that vary by culture, such as the currencysymbol included by the ‘C’ and ‘c’ formats, a formatting object supplies the actual characters used in the string representation. A method might include a parameter to pass a `System.IFormatProvider` object that supplies a formatting object, or the method might use the default formatting object, which contains the symbol definitions for the current culture. The current culture typically uses the same set of symbols used system-wide by default. In the Base Class Library, the formatting object for system-supplied numeric types is a `System.Globalization.NumberFormatInfo` instance. For `System.DateTime` instances, a `System.Globalization.DateTimeFormatInfo` is used. *end note*
```

BillWagner (approved):
Once we update the note formats, this is ready.

PR #396, opened 2021-11-16
Various Word converter improvements

jskeet:
I'd suggest reviewing one commit at a time.

The result is *much* nicer than it was before.


BillWagner (approved):
I like the improvements you made here Jon.

You can :shipit: when ready.

PR #402, opened 2021-11-18
Replace nested bullets with prose

### LINKED ISSUE #383: Try statement: Replace nested bulleted list with prose
This is additional work based on #2 

See https://github.com/dotnet/csharpstandard/pull/2#issuecomment-744665141

The existing nested bulleted lists are are to parse and follow. We will replace them with a series of short prose paragraphs instead.

BillWagner:
Fixes #383

The nested bullets that specify the control flow for the `try` statement are complicated. This PR replaces them with paragraphs of prose.


standard/statements.md
```diff
@@ -1173,26 +1173,13 @@ It is a compile-time error for a `break`, `continue`, or `goto` statement to tra
...
+
+If an exception has been propagated and a matching `catch` clause is found, control is transferred to the first matching `catch` block. If control reaches the end point of the `catch` block without an exception being propagated, control is transferred to the `finally` block if one exists. If no `finally` block exists, control is transferred to the end point of the `try` statement. If an exception has been propagated from the `catch` block, control transfers to the `finally` block if one exists. The exception is propagated to the next enclosing `try` statement.
+
+If an exception has been propagated, and no matching `catch` clause is found, control transfers to the `finally` block, if it exists. The exception is propagated to hte next enclosing `try` statement.
```
```suggestion
If an exception has been propagated, and no matching `catch` clause is found, control transfers to the `finally` block, if it exists. The exception is propagated to the next enclosing `try` statement.
```

jskeet (commented):
Unless I'm mistaken - which is very possible - I don't believe this accurately captures the "two pass" exception handling that's implemented, where in the first phase a matching handler is found (potentially going up several enclosing try blocks) and only then the finally block is executed.

Example:

```csharp
using System;

public class Test
{
    static void Main()
    {
        try
        {
            try
            {
                throw new ArgumentException();
            }
            finally
            {
                Console.WriteLine("Finally");
            }
        }
        catch (Exception ex) when (ExceptionFilter(ex))
        {
            Console.WriteLine("Catch");
        }

        bool ExceptionFilter(Exception ex)
        {
            Console.WriteLine("Filter");
            return true;
        }
    }
}
```

I believe that according to this PR, the output would be Finally, Filter, Catch. It's actually Filter, Finally, Catch.

standard/statements.md
```diff
@@ -1179,26 +1179,13 @@ It is a compile-time error for a `break`, `continue`, or `goto` statement to tra
...
-    - When and if control reaches the end point of the `catch` block:
-      - If the `try` statement has a `finally` block, the `finally` block is executed.
-      - Control is transferred to the end point of the `try` statement.
-    - If an exception is propagated to the `try` statement during execution of the `catch` block:
-      - If the `try` statement has a `finally` block, the `finally` block is executed.
-      - The exception is propagated to the next enclosing `try` statement.
-- If the `try` statement has no `catch` clauses or if no `catch` clause matches the exception:
-  - If the `try` statement has a `finally` block, the `finally` block is executed.
-  - The exception is propagated to the next enclosing `try` statement.
+When execution reaches a`try` statement, control is transferred to the `try` block. If control reaches the end point of the `try` block without an exception being propagated, control is transferred to the `finally` block if one exists. If no `finally` block exists, control is transferred to the end point of the `try` statement.
```
Nigel-Ecma, line 1182: Original:

> - When and if control reaches the end point of the `try` block:
>   - If the `try` statement has a `finally` block, the `finally` block is executed.
>   - Control is transferred to the end point of the `try` statement.

This states the finally block is conditionally executed and then control is transferred to the end of the try statement.

Replacement:

> If control reaches the end point of the `try` block without an exception being propagated, control is transferred to the `finally` block if one exists. If no `finally` block exists, control is transferred to the end point of the `try` statement.

This says control is transferred to the finally block if it exists *otherwise* to the end of the try statement.

Hasn't this lost what happens when a finally block finishes executing?

BillWagner: Thanks @Nigel-Ecma 

I addressed this in the paragraph on 1190. That defines how the statements in a `finally` block execute. I think the prose worked better.

Nigel-Ecma: The original talks about executing the `finally` and then transferring control to the endpoint of the `try` statement. Doesn’t 1190 cover the executing part and not the transfer to the endpoint bit?

BillWagner: @Nigel-Ecma I added a new sentence at the end of that paragraph that declares that control transfers to the end of the `try` statement.

Nigel-Ecma: @BillWagner – So you have, my bad I was looking at the wrong commit, apologies.

BillWagner:
Regarding @jskeet 's comment: The original bullets didn't specify that order either. (I checked the V5 text, and it's not specified in that version either). ~~I'll add a new commit to declare that.~~

After review, I didn't add another commit [§13.10.6](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/statements.md#13106-the-throw-statement). See the third text paragraph that starts with "Propagation of an exception consists of repeatedly evaluating the following steps...". I can add a reference here, but I don't think it's necessary.

BillWagner:
@jskeet Marking this ready for discussion / review.

jskeet:
I'm happy with the change in terms of bullet points - but I still think it's inaccurate in terms of the two pass exception filter model. We could potentially merge this but raise another issue for that aspect. Let's discuss in the meeting.

BillWagner:
@jskeet I did make one more commit. While reading it again carefully, the `throw` statement clause did not mention exception filters, but did specify how the search for a matching `catch` propagates before control enters any `catch` or `finally` clauses. 

The section on `catch` now references that section.

jskeet (approved):
Okay... I *think* this now specifies the behavior if you read it really carefully, but given that it's surprising (an "outer" catch block filter being evaluated before a finally block) I suspect an example would be useful.

Do you want to get this PR merged and then we can add an example later?

BillWagner:
> Do you want to get this PR merged and then we can add an example later?

@jskeet If you're happy with the example in your earlier comment, I'll propose adding it to this PR, and then merging. I can make those edits tomorrow after our meeting.

jskeet:
> > Do you want to get this PR merged and then we can add an example later?
> 
> @jskeet If you're happy with the example in your earlier comment, I'll propose adding it to this PR, and then merging. I can make those edits tomorrow after our meeting.

I'd suggest a slight change for clarification:

```csharp
using System;

public class Test
{
    static void Main()
    {
        try
        {
            Method();
        }
        catch (Exception ex) when (ExceptionFilter(ex))
        {
            Console.WriteLine("Catch");
        }

        bool ExceptionFilter(Exception ex)
        {
            Console.WriteLine("Filter");
            return true;
        }
    }

    static void Method()
    {
        try
        {
            throw new ArgumentException();
        }
        finally
        {
            Console.WriteLine("Finally");
        }
    }
}
```

That shows that it's not just a matter of looking for all the blocks in a method.

PR #409, opened 2021-12-01
Add create permissions for this action

BillWagner:
This action creates a new branch and opens a PR to merge that branch into the current default.

The default `permissions` for GitHub actions are `read` for the repo contents. This adds `write` for contents, and `write` for pull requests to push the branch and create a new PR.

### DIFF
--- a/.github/workflows/update-on-merge.yaml
+++ b/.github/workflows/update-on-merge.yaml
@@ -16,6 +16,9 @@ on:
 jobs:
   create-pull-request-for-sections-and-grammar:
     runs-on: ubuntu-18.04
+    permissions:
+      contents: write
+      pull-requests: write    
     env:
       DOTNET_NOLOGO: true
 


Nigel-Ecma:
I'm guessing "write for contents" doesn't mean what it sounds like? Or does this make the C# Standard world writable?

jskeet:
> I'm guessing "write for contents" doesn't mean what it sounds like?

It means what it says *for the GitHub action itself*. That's fine so long as we're confident that the action only does what we want.

@BillWagner If someone creates a PR that changes the renumbering tool to do something harmful, would that still execute (and do so within the writable context)? Perhaps we should make this action run manually rather than automatically?

Nigel-Ecma:
OK, I see. Are we taking regular off-GitHub backups?

KalleOlaviNiemitalo:
> If someone creates a PR that changes the renumbering tool to do something harmful, would that still execute

If I understand correctly, `update-on-merge.yaml` runs after a `push` event to the draft-v6 branch, or when manually triggered by someone with write access ([Manually running a workflow](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow)). The `pull_request` event type is not listed, and even if it were, GitHub could still be configured to require approval per [Approving workflow runs from public forks](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks).

jskeet:
> > If someone creates a PR that changes the renumbering tool to do something harmful, would that still execute
> 
> If I understand correctly, `update-on-merge.yaml` runs after a `push` event to the draft-v6 branch, or when manually triggered by someone with write access ([Manually running a workflow](https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow)). The `pull_request` event type is not listed, and even if it were, GitHub could still be configured to require approval per [Approving workflow runs from public forks](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks).

Yup, you're right. In which case I think this should be okay.

jskeet (approved):

BillWagner:
@jskeet @Nigel-Ecma 

> If someone creates a PR that changes the renumbering tool to do something harmful, would that still execute (and do so within the writable context)? Perhaps we should make this action run manually rather than automatically?

There are three different protections for us here. First, this change does elevate privileges on the default token because we use the default perms of read access. Second, we've employed branch protection rules on the "trunk" branches for each version (`v5`, `draft-v6` and `draft-v7`). No account (even ours) can push directly to those branches except by merging a PR that has at least one approval. Third, any PR whose author hasn't contributed before must have someone with write access approve before actions run. 

I've been reading and reaching out to GitHub folks to follow the recommended security practices. Thanks to all of you for checking my work, and helping us be safe.

Finally, to Nigel's point, I do fetch on a regular basis in case I've missed anything.

Issue #399, opened 2021-11-17
19.2 Enum base type


gafter:
An enum base type may be an integral type's keyword or any type name that *resolves* to one of the integral types.

The current text of the spec requires that it be the fully qualified name, but that's not actually what is implemented in C# 6. It could use an alias or an imported type name.

RexJaeschke:
During the Nov 17 TG2 call, @jskeet noticed that [admin/v6-feature-tracker.md ](https://github.com/dotnet/csharpstandard/blob/draft-v6/admin/v6-feature-tracker.md) did not seem to reflect the current status of that topic, and he asked me to investigate. Here are my findings:

## 15.5.4 Volatile fields

This currently contains the following:
> These restrictions ensure that all threads will observe volatile writes performed by any other thread in the order in which they were performed. A conforming implementation is not required to provide a single total ordering of volatile writes as seen from all threads of execution. The type of a volatile field shall be one of the following:
> -  …
> -  An *enum_type* having an *enum_base* type of `byte`, `sbyte`, `short`, `ushort`, `int`, or `uint`.

Unless I'm reading it too literally, this constraint does not allow the `System.*` alias names for these keywords. Nor does it disallow the `System.*` alias names for `long` and `ulong`. I propose changing this constraint to the following:

> -  An *enum_type* having an *enum_base* type other than `long`, `ulong`, and their corresponding `System.*` aliases.

## 19.2 Enum declarations

As a result of PR https://github.com/dotnet/csharpstandard/pull/351, we have the following:

> ```ANTLR
> enum_base
>     : ':' integral_type
>     | ':' integral_type_name
>     ;
> integral_type_name
>     : type_name // restricted to one of System.{SByte,Byte,Int16,UInt16,Int32,UInt32,Int64,UInt64}
>     ;
> ```
> Each enum type has a corresponding integral type called the ***underlying type*** of the enum type. This underlying type shall be able to represent all the enumerator values defined in the enumeration. If the *enum_base* is present, it explicitly declares the underlying type. The underlying type shall be one of the *integral types* ([§9.3.6](types.md#936-integral-types)) other than `char`; specified either by keyword (*integral_type*), or by one of the full type names that the integral types alias ([§9.3.5](types.md#935-simple-types)) (*integral_type_name*).

> *Note*: Neither `char` nor `System.Char` can be used as an underlying type. *end note*

[...]

PR #424, opened 2022-01-13
Clarify enum_base specification

### LINKED ISSUE #412: Sort out what's happening about base enums
- Possibly an issue in terms of which types are referred to
- Possibly an issue with some change in the old repo not being transferred

jskeet:
The `integral_type_name` approach doesn't work, as it's fine to use
namespace aliases etc. What's important is what type the type name
resolves to after taking into account using directives etc.

This may not be quite the wording we want, but I'm hopeful that it's
on the right track.

Fixes #412

### DIFF
--- a/standard/enums.md
+++ b/standard/enums.md
@@ -31,7 +31,7 @@ enum_base
     ;
 
 integral_type_name
-    : type_name // restricted to one of System.{SByte,Byte,Int16,UInt16,Int32,UInt32,Int64,UInt64}
+    : type_name // Shall resolve to an integral type other than char
     ;
 
 enum_body
@@ -40,9 +40,9 @@ enum_body
     ;
 ```
 
-Each enum type has a corresponding integral type called the ***underlying type*** of the enum type. This underlying type shall be able to represent all the enumerator values defined in the enumeration. If the *enum_base* is present, it explicitly declares the underlying type. The underlying type shall be one of the *integral types* ([§8.3.6](types.md#836-integral-types)) other than `char`; specified either by keyword (*integral_type*), or by one of the full type names that the integral types alias ([§8.3.5](types.md#835-simple-types)) (*integral_type_name*).
+Each enum type has a corresponding integral type called the ***underlying type*** of the enum type. This underlying type shall be able to represent all the enumerator values defined in the enumeration. If the *enum_base* is present, it explicitly declares the underlying type. The underlying type shall be one of the *integral types* ([§8.3.6](types.md#836-integral-types)) other than `char`. The underlying type may be specified either by an `integral_type` ([§8.3.5](types.md#835-simple-types)), or an `integral_type_name`. The `integral_type_name` is resolved in the same way as `type_name` ([§7.8.1](basic-concepts.md#781-general)), including taking any using directives ([§13.5](namespaces.md#135-using-directives)) into account.
 
-> *Note*: Neither `char` nor `System.Char` can be used as an underlying type. *end note*
+> *Note*: The `char` type cannot be used as an underlying type, either by keyword or via an `integral_type_name`. *end note*
 
 An enum declaration that does not explicitly declare an underlying type has an underlying type of `int`.
 
--- a/standard/grammar.md
+++ b/standard/grammar.md
@@ -2222,7 +2222,7 @@ enum_base
     ;
 
 integral_type_name
-    : type_name // restricted to one of System.{SByte,Byte,Int16,UInt16,Int32,UInt32,Int64,UInt64}
+    : type_name // Shall resolve to an integral type other than char
     ;
 
 enum_body


jskeet:
Also fixes #399.

BillWagner:
The wording in the prose is fine to me. I'd like to hear @Nigel-Ecma 's opinion on the grammar, and if there is a way to express that the `base_type` must resolve to one of the allowed types in ANTLR.

jskeet:
> and if there is a way to express that the base_type must resolve to one of the allowed types in ANTLR.

I find it hard to see how we could express that, given that we'd have to make things like this work:

```csharp
using Xyz = System;

public enum Colors : Xyz.Int32
{
}
```

This feels like it's something we can't really do in the grammar - it's got to be validated later.

Nigel-Ecma:
Just took a quick look:

The grammar does not need to change – the revised grammar might be OK (I haven’t confirmed) but the current grammar was constructed to resolve mutual recursion issues let’s not mess with it if we don’t have to!
The comment in the grammar and the descriptive text might need to change.

The comment says “restricted to one of…”, it doesn’t define how this restriction is performed.

To address Jon’s concerns that it doesn’t allow aliases the wording could instead be “must resolve to one of…” and the descriptive text could be:

The underlying type shall be one of the integral types (§9.3.6) other than char; specified either by keyword (integral_type), or by a type_name which resolves to one of the full type names that the integral types alias (§9.3.5) (integral_type_name).

or something like that (it’s a quick look, no wordsmithing ;-))

Personal opinion:

Whether we should standardise this is another matter. I think the original language design with keywords was clean and logical; that some compiler writers choose to allow full type names, aliases, etc. doesn’t mean it is good practice and should be enshrined in a Standard. Those keywords exist for a reason.

But I’m going to lose that one… ;-)

jskeet:
I don't think I've introduced anything with mutual recursion. I've just avoided introducing a new-and-unnecessary rule. Let's discuss that in the next meeting.

In terms of standardising the change: personally I'm in favour of this change, as it's always felt odd to me that you can refer to the types via either keywords or full type names *everywhere* apart from enum base types.

Nigel-Ecma:
> I don't think I've introduced anything with mutual recursion.  I've just avoided introducing a new-and-unnecessary rule. Let's discuss that in the next meeting.

There are at least 15 (I stopped counting) other such rules (of the form `non_terminal_A : non_terminal_B ;`). The grammar uses them to convey meaning/restrictions. e.g. `boolean_expression : expression ;` to indicate the expression must have a boolean type.

Here the rule name conveys the restriction to integral types, leave it as is.

Nigel-Ecma (changes requested):
Leave the grammar *rule* as is, no change is required and its style fits the rest of the grammar.

Change the grammar *comment* and the associated text replacing “restricted” with “must resolve to” and any associated wordsmithing required to produce valid and understandable grammar – see previous comment for a quick suggestion.

Nigel-Ecma:
> In terms of standardising the change: personally I'm in favour of this change, as it's always felt odd to me that you can refer to the types via either keywords or full type names _everywhere_ apart from enum base types.

As I said, I’ve already lost this one ;-(

<subjective>
The odd thing is you can specify these types other than by keyword.

Coincidentally I've just read an article on the 27 (OK, slight exaggeration ;-)) different ways to write an iterative loop in Swift which included some suggestions on how to pick best one based on the context... AARGH!

In prose one may waffle, in language design and code one aims for clarity, directness and elegance; not aliasing `System` or avoiding specifying enum base types by keyword.
</subjective>

jskeet:
I've pushed a new version which restores `integral_type_name`. Please review, folks.

standard/enums.md
```diff
@@ -31,7 +31,7 @@ enum_base
     ;
 
 integral_type_name
-    : type_name // restricted to one of System.{SByte,Byte,Int16,UInt16,Int32,UInt32,Int64,UInt64}
+    : type_name // Must resolve to an integral type other than char
```
KalleOlaviNiemitalo: Would "shall" be better here?

jskeet: Probably - it always seems weird to me, but that's the style we have...

standard/enums.md
```diff
@@ -40,9 +40,9 @@ enum_body
...
+Each enum type has a corresponding integral type called the ***underlying type*** of the enum type. This underlying type shall be able to represent all the enumerator values defined in the enumeration. If the *enum_base* is present, it explicitly declares the underlying type. The underlying type shall be one of the *integral types* ([§9.3.6](types.md#936-integral-types)) other than `char`. The underlying type may be specified either by a `simple_type` ([§9.3.5](types.md#935-simple-types)), or an `integral_type_name`. The `integral_type_name` is resolved in the same way as `type_name`, including taking any using directives ([§14.5](namespaces.md#145-using-directives)) into account.
 
-> *Note*: Neither `char` nor `System.Char` can be used as an underlying type. *end note*
+> *Note*: The `char` type cannot be used as an underlying type, either by keyword or via a `integral_type_name`. *end note*
```
Nigel-Ecma: Grammar: via *an* `integral_type_name`

standard/enums.md
```diff
@@ -40,9 +40,9 @@ enum_body
...
+Each enum type has a corresponding integral type called the ***underlying type*** of the enum type. This underlying type shall be able to represent all the enumerator values defined in the enumeration. If the *enum_base* is present, it explicitly declares the underlying type. The underlying type shall be one of the *integral types* ([§9.3.6](types.md#936-integral-types)) other than `char`. The underlying type may be specified either by a `simple_type` ([§9.3.5](types.md#935-simple-types)), or an `integral_type_name`. The `integral_type_name` is resolved in the same way as `type_name`, including taking any using directives ([§14.5](namespaces.md#145-using-directives)) into account.
```
Nigel-Ecma: Typo, replace: “specified either by a `simple_type`”
With: “specified either by an `integral_type`”

Nigel-Ecma: If the last sentence is to include the text after the comma then a reference to §8.8.1 should probably be added before the comma as that describes the resolution process, e.g.:

> The `integral_type_name` is resolved in the same way as `type_name` (§8.8.1), including taking any using directives ([§14.5](namespaces.md#145-using-directives)) into account.

Nigel-Ecma (changes requested):
This is "request changes" as the is one error – `simple_type` instead of `integral_type` – which must be fixed. The other comments you can take or leave!

standard/enums.md
jskeet: Doh. Fixed, thanks.

standard/enums.md
jskeet: Indeed - I'm not sure how I ended up with that :(

jskeet: Good plan - done.

Nigel-Ecma (approved):

jskeet:
This needed a bit of rebasing, but I believe it's okay - it looks like we were actually ready to merge it before, so next week we can just give it one final look and merge if we're happy.

BillWagner (approved):
LGTM

PR #430, opened 2022-01-19
Clarify that object initializer can have multiple indexer targets

### LINKED ISSUE #411: Clarify that indexers can be used multiple times in object initializers
Follow-up action from #8

jskeet:
Fixes #411.

(This is a bit of a straw man to start with, but is hopefully enough to work out how it needs to change before being merged.)


BillWagner (approved):
This LGTM @jskeet 

Let's :shipit:

Issue #425, opened 2022-01-15
Revising "4 Acronyms and abbreviations"


RexJaeschke:
When we submit V6 to Ecma in April, this will be our first time starting with an md base and generating from that the Word rendering we submit. As such, I'm working with Jon to automate as much of that process as is reasonably possible. (That said, it is almost certain I will have to make a number of manual changes to the resulting Word document.) At the same time, I'm looking at our spec from the point of view of eventually submitting it to ISO (but not for V6), which has *many* requirements, more than a few of which we do *not* follow.

While reviewing "4 Acronyms and abbreviations," I came to the following conclusions. Nothing is broken, per se; this is just some editorial housekeeping.

## **Recommendation 1** 

We strike the following:

> The following acronyms and abbreviations are used throughout this specification:
> BCL --- Base Class Library, which provides types to represent the built-in data types of the CLI, simple file access, custom attributes, security attributes, string manipulation, formatting, streams, and collections.
> CLI --- Common Language Infrastructure
> CLS --- Common Language Specification
> IEC --- the International Electrotechnical Commission
> IEEE --- the Institute of Electrical and Electronics Engineers
> ISO --- the International Organization for Standardization

BCL and CLI are explained in "Normative references;" CLS is never used; and IEC, IEEE, and ISO are well-known in the IT world, so no need to mention them here.

## **Recommendation 2**

We move the following:

> The name C# is pronounced "C Sharp".
> The name C# is written as the LATIN CAPITAL LETTER C (U+0043) followed by the NUMBER SIGN # (U+0023).

to the end of "Introduction."

## **Recommendation 3** 

We change

> ... Throughout this specification, these types will appear as the fully qualified name, omitting the global namespace qualifier, or as a simple unqualified type name, omitting the namespace. ...

to

> ... Throughout this specification, these types appear as either the fully qualified name; with the global namespace qualifier omitted; or as a simple unqualified type name, with the namespace omitted as well. ...

## **Recommendation 4**

We change the clause title to “Library type name abbreviations.” We can leave the file name as is, so we don’t have to change any tooling.

I even thought about striking this clause and moving its content to a new section in Type.md following 9.1. As "Types" is the first substantive clause, this topic would not be out of place there.

## **Recommendation 5** 

We strike the Clause delimiter lines **This clause is informative.** and **End of informative text.**, as all the remaining contents are normative.

jskeet:
[...]

PR #444, opened 2022-01-20
Ensure section renumbering tool handles removing clauses.

BillWagner:
In preparation for **Recommendation 5** in #425

This PR updates the section renumbering tool to handle situations where we remove a clause. (I think it will handle adding a clause as well, but that wasn't tested.)

The focus of the change is to remove the hardcoded list of files from the `Program.cs` file. Instead, read the list of frontmatter, main body clauses, and annexes from a .json config file. The config file is stored in the same folder as the standard text.

- If you remove a file from the standard, and don't remove it from the config file, the dry run fails. That should help us catch any instances when the PR is opened.
- The dry run will catch any links in the text that resolved to a clause in the removed section. 
- The full run renumbers headers and all links.


tools/StandardAnchorTags/TocSectionNumberBuilder.cs
```diff
@@ -33,13 +33,13 @@ private struct SectionHeader
...
-        private StringBuilder tocContent = new StringBuilder();
-        private readonly Dictionary<string, SectionLink> sectionLinkMap = new Dictionary<string, SectionLink>();
+        private readonly StringBuilder tocContent = new();
+        private readonly Dictionary<string, SectionLink> sectionLinkMap = new();
         private bool isAnnexes;
 
         // Running array of entries for the current headings. 
         // Starting with H1 is headings[0], H2 is headings[1] etc.
-        int[] headings = new int[8];
+        readonly int[] headings = new int[8];
```
jskeet: Add an explicit `private` as well for consistent?

tools/StandardAnchorTags/TocSectionNumberBuilder.cs
```diff
@@ -53,12 +53,24 @@ public TocSectionNumberBuilder(string pathFromToolToStandard, bool dryRun)
...
+        /// <param name="fileName">The front matter file name.</param>
+        /// <remarks>
+        /// The front matter entries don't add section numbers in the headers.
+        /// Otherwise, this is the same logic for the main file.
+        /// </remarks>
+        public async Task AddFrontMatterTocEntries(string fileName)
         {
-            foreach (var entry in entries)
+            string? line;
+            using var stream = new StreamReader($"{PathToStandardFiles}/{fileName}");
```
jskeet: Maybe use Path.Combine here? (I'm sure it's fine either way... just my habit.)
(I also generally use `File.OpenText` but it's a personal preference thing...)

tools/StandardAnchorTags/TocSectionNumberBuilder.cs
```diff
@@ -53,12 +53,24 @@ public TocSectionNumberBuilder(string pathFromToolToStandard, bool dryRun)
...
+            while ((line = await stream.ReadLineAsync()) != null)
             {
-                tocContent.AppendLine($"- [{entry.linkText}]({entry.fileName})");
+                if (line.StartsWith("# "))
+                {
+                    var linkText = line[2..];
+                    tocContent.AppendLine($"- [{linkText}]({fileName})");
+                    // Done: return.
+                    return;
+                }
```
jskeet, line 72: Do we want to throw an exception if we get to the end of this method? It means we haven't found a heading, which is presumably a problem.

Also, can we not just read the first line of the file and assert that it starts with `# `? (Again, if it doesn't that's presumably something wrong...)

tools/Utilities/Clauses.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Utilities
+{
+    public class Clauses
+    {
+        public string[] FrontMatter { get; set; } = Array.Empty<string>();
+        public string[] MainBody { get; set; } = Array.Empty <string>();
```
jskeet: Nit: remove space after `Empty`.
(I'm interested in your preference for arrays over `List<string>` here, but that can be something we discuss at leisure...)

jskeet (approved):
Fine to merge as-is, but with a few optional suggestions.

tools/StandardAnchorTags/TocSectionNumberBuilder.cs
BillWagner: Good suggestion. I applied both.

tools/StandardAnchorTags/TocSectionNumberBuilder.cs
BillWagner: Both good suggestions. (I thought about the 2nd, but then wondered if we might ever add comments above the H1. I tend to agree that we won't do that).

tools/Utilities/Clauses.cs
BillWagner: I suppose I could say something brilliant about these never being resized, only set during serialization, etc.

But, the truth is I used Visual Studio's "paste JSON text as class" and used that definition. 🔨

PR #452, opened 2022-02-03
Updated typography

Nigel-Ecma:
**This is not a PR as such, but read on…**

A program to update the typography of the Standard; restoring the ‘single’, “double”, apostrophe’s and — long dashes that got lost in the move to Markdown; has been developed based on the *Markdig* Markdown processor.

The recommendation is that this program is used before V6 is published, so it doesn’t look like a typewriter manuscript from the 70’s ;-)

This PR contains the results of running the Standard through this program. This is currently a manual process using a GUI frontend and is therefore one of the last steps before publishing.

The longer term plan is to package this as a check-in action and so the typography will be fixed automatically. There is a problem with doing this at the moment — *Markdig* turned out to be buggy and has been patched in a Heath Robinson manner to get it to work.

*After applying it to the Standard each file has been checked manually in a diff program to make sure the changes are valid — so every change is good. However some typewriter-typography may been missed.*

So please take a look and weigh in on whether we run this before publishing V6.

*If you spot errors in the conversion just add comments, don’t edit the files as they won’t ever be merged!*


Nigel-Ecma:
> *However some typewriter-typography may been missed.*

This has now been checked and there are eight cases in `standard-library.md` which typography-wise should be converted to proper open & close quotes but are not.

They are “missed” by this app as they are inside a HTML table the contents of which is treated as HTML, not Markdown, content (according to Markdown’s rules) and only Markdown text is converted.

So it’s not a bug *per se*, the code is behaving to the Markdown/Markdig spec even if it gives the wrong results for us in this case. With only 8 occurrences and our minimal use of HTML this can be dealt with by a hand-edit.

jskeet (approved):

jskeet:
Decision: Bill to attempt to include the Smarten tool into our CI.

jskeet:
I can confirm this looks fine in the converted Word doc:

image

Nigel-Ecma:
Following up on our last meeting I’ve added PR #489 for the manual changes required to standard-library.md. After that the smarten tool can be run on the final v6 or integrated into the GitHub workflow.

jskeet:
Closing now that Smarten is implemented.

PR #454, opened 2022-02-09
Remove unnecessary heading

BillWagner:
***Simple Names*** has only one child heading: General. 

I think that child heading can be removed.


RexJaeschke (approved):
Yes, we deleted the 2nd subsection from here in V6.

The only link reference I see to the current section name is in README, but I presume that will be automatically corrected/regenerated.

Issue #369, opened 2021-09-07
Expression-bodied members vs. expressions that return nothing


Nigel-Ecma:
v6 introduces expression-bodied members, v7 will introduce expression-bodied finalisers. In the former case result of the *expression* of the body may be classified as nothing (§12.2.1) if the return type of the method is `void`; in the latter case it may be the case that the result type *must* be classified as nothing.

The introduction of at least the above two kinds of expression bodied members extends the places expressions whose result is classified as nothing may occur: §12.2.1 requires updating for v6 (and again for v7); it *may* impact how null conditionals are specified (PR #7, #251); and there could well be other places…

Let the hunt begin!

jskeet:
Jon will poke around to try to find other examples.

jskeet:
Research notes as I go...

There are 13 matches for "nothing" (case-insensitive) in the standard right now. Three of these are irrelevant - for the record, they are:

- In standard-library.md: "otherwise, nothing is stored in that position in the output string"
- In standard-library.md: "(nothing appears between the semicolons)"
- In statements.md: "An *empty_statement* does nothing."

The remaining matches are all in expressions.md (somewhat predictably). They are:

- 11.2.1 (Expression classifications / general): "Nothing. This occurs when the expression is an invocation of a method with a return type of void. An expression classified as nothing is only valid in the context of a statement_expression (§12.7) or as the body of a lambda_expression (§11.16)."
- 11.2.2 (Values of expressions): "Most of the constructs that involve an expression ultimately require the expression to denote a value. In such cases, if the actual expression denotes a namespace, a type, a method group, or nothing, a compile-time error occurs."
- 11.6.3.13 (Type inference / Inferred return type): "If F is async and the body of F is either an expression classified as nothing (§11.2), or a statement block where no return statements have expressions, the inferred return type is System.Threading.Tasks.Task."
- 11.7.8.1 (Invocation expressions / general): "If the invocation_expression invokes a method or delegate that returns void, the result is nothing. An expression that is classified as nothing is permitted only in the context of a statement_expression (§12.7) or as the body of a lambda_expression (§11.16). Otherwise a binding-time error occurs."
- 11.7.9 (Null Conditional Invocation Expression): "Unlike the syntactically equivalent null_conditional_member_access or null_conditional_element_access, a null_conditional_invocation_expression may be classified as nothing."
[...]

PR #458, opened 2022-02-18
Fix inconsistency around the effective return type of a method

jskeet:
Found while investigating #369


BillWagner (approved):
This wording looks good to me @jskeet 

I'll approve.

PR #461, opened 2022-02-24
Improve text around extern aliases

### LINKED ISSUE #193: 14.4 Extern alias directives
As well as looking at the diffs between the md from both, it will be useful to also look at the (few) tracked changes TG2 made in this subclause (§14.4) of the WD3.17 Word document.

**V5std has**

“Within C# source code, a type is declared a member of a single namespace. However, a namespace hierarchy referenced by an extern alias may contain types that are also members of other namespaces. For example, if `A` and `B` are extern aliases, the names `A::X`,`B::C.Y` and `global::D.Z` may, depending on the external specification supported by the particular compiler, all refer to the same type.

The alias introduced by an *extern_alias_directive* is very similar to the alias introduced by a *using_alias_directive*. See [§14.5.2](#using-alias-directives) for more detailed discussion of *extern_alias_directive*s and *using_alias_directive*s.

`alias` is a contextual keyword ([§7.4.4](lexical-structure.md#keywords)) and only has special meaning when it immediately follows the `extern` keyword in an *extern_alias_directive*.

> [*Example*: In fact an extern alias could use the identifier `alias` as its name:
> ```csharp
> extern alias alias;
> ```
> *end example*]”

**While v6spec has**

“An *extern_alias_directive* makes an alias available within a particular compilation unit or namespace body, but it does not contribute any new members to the underlying declaration space. In other words, an *extern_alias_directive* is not transitive, but, rather, affects only the compilation unit or namespace body in which it occurs.

The following program declares and uses two extern aliases, `X` and `Y`, each of which represent the root of a distinct namespace hierarchy:
```csharp
extern alias X;
extern alias Y;
class Test
{
    X::N.A a;
    X::N.B b1;
    Y::N.B b2;
    Y::N.C c;
}
```

The program declares the existence of the extern aliases `X` and `Y`, but the actual definitions of the aliases are external to the program. The identically named `N.B` classes can now be referenced as `X.N.B` and `Y.N.B`, or, using the namespace alias qualifier, `X::N.B` and `Y::N.B`. An error occurs if a program declares an extern alias for which no external definition is provided.“

Which is correct?

jskeet:
- Use a more sensible example (sourced from the draft C# 6 Microsoft spec)
- Remove a paragraph around types being accessible via different names, as we can't see how the example given would actually work

Fixes #193

### DIFF
--- a/standard/namespaces.md
+++ b/standard/namespaces.md
@@ -112,16 +112,28 @@ The scope of an *extern_alias_directive* extends over the *using_directive*s, *g
 
 Within a compilation unit or namespace body that contains an *extern_alias_directive*, the identifier introduced by the *extern_alias_directive* can be used to reference the aliased namespace. It is a compile-time error for the *identifier* to be the word `global`.
 
-Within C# source code, a type is declared a member of a single namespace. However, a namespace hierarchy referenced by an extern alias may contain types that are also members of other namespaces. For example, if `A` and `B` are extern aliases, the names `A::X`,`B::C.Y` and `global::D.Z` may, depending on the external specification supported by the particular compiler, all refer to the same type.
-
 The alias introduced by an *extern_alias_directive* is very similar to the alias introduced by a *using_alias_directive*. See [§13.5.2](namespaces.md#1352-using-alias-directives) for more detailed discussion of *extern_alias_directive*s and *using_alias_directive*s.
 
 `alias` is a contextual keyword ([§6.4.4](lexical-structure.md#644-keywords)) and only has special meaning when it immediately follows the `extern` keyword in an *extern_alias_directive*.
 
-> *Example*: In fact an extern alias could use the identifier `alias` as its name:
+An error occurs if a program declares an extern alias for which no external definition is provided.
+
+> *Example*: The following program declares and uses two extern aliases, `X` and `Y`, each of which represent the root of a distinct namespace hierarchy:
+>
 > ```csharp
-> extern alias alias;
+> extern alias X;
+> extern alias Y;
+> 
+> class Test
+> {
+>     X::N.A a;
+>     X::N.B b1;
+>     Y::N.B b2;
+>     Y::N.C c;
+> }
 > ```
+>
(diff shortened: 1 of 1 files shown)


standard/namespaces.md
```diff
@@ -112,16 +112,28 @@ The scope of an *extern_alias_directive* extends over the *using_directive*s, *g
...
-
 The alias introduced by an *extern_alias_directive* is very similar to the alias introduced by a *using_alias_directive*. See [§13.5.2](namespaces.md#1352-using-alias-directives) for more detailed discussion of *extern_alias_directive*s and *using_alias_directive*s.
 
 `alias` is a contextual keyword ([§6.4.4](lexical-structure.md#644-keywords)) and only has special meaning when it immediately follows the `extern` keyword in an *extern_alias_directive*.
 
-> *Example*: In fact an extern alias could use the identifier `alias` as its name:
+An error occurs if a program declares an extern alias for which no external definition is provided.
+
+> *Example*: The following program declares and uses two extern aliases, `X` and `Y`, each of which represent the root of a distinct namespace hierarchy:
```
BillWagner, line 121: Would this help clarify the example? I wonder if we'll remember why this exists when we look at it later. (I know I keep forgetting).

```suggestion
> *Example*: The following program declares and uses two extern aliases, `X` and `Y`, each of which represent the root of a distinct namespace hierarchy. The external definition maps the `X` and `Y` namespaces to the assemblies that contain the distinct types with matching fully qualified names:
```

BillWagner (approved):
This does help clarify things @jskeet I like it.

I added a sentence in the example (informative) text that provides some context for the extern definitions. What are your thoughts?

standard/namespaces.md
jskeet: I'm not sure... that feels like it's moving away from the generalized view of extern aliases to something specific. On the other hand, it *is* only an example. @Nigel-Ecma, thoughts?

Nigel-Ecma: I think I agree with @jskeet here. @BillWagner's suggestion makes sense if we're standardising the Roslyn copier's external mechanism that presents an assembly as a namespace that encloses all the namespaces in the assembly; but if we're doing that it has to be specified normatively somewhere.

Maybe we should standardise the external mechanism, or at least as much as what is allowed to be aliased (e.g. assemblies and not individual namespaces within them), but that is probably too big a decision to make as this stage for v6 and is a (theoretical) breaking change.

TL;DR let's fix as per this PR and revisit in v7

Nigel-Ecma (approved):
LGTM for what we should do in v6

Issue #426, opened 2022-01-15
Reviewing "2. Normative references"


RexJaeschke:
Each time a committee produces a new edition of a spec, it should review its list of normative references to see if it is complete and current. 

1. I'm not aware that we've added any new requirements needing new entries. Is anyone else?
2. ISO/IEC/IEEE 60559:2011, *Information technology -- Microprocessor Systems -- Floating-Point arithmetic* contains a dated reference to the 2011 edition. (Note that Political Correctness has us referring to the IEC version rather than the IEEE one, which is equivalent and much better known.) The ISO website tells me that the 2011 edition has been superseded (and withdrawn) by one published in 2020 (but IEEE has been dropped from the title). Are we happy to change our reference to the the 2020 edition? **I will take an action item to ping Mike Cowlishaw (former IBM Fellow and TG2 and TG3 participant, and long-time IEEE FP spec editor) to see of he can think of any additions/changes that might impact us.**
3. For Unicode and the related ISO 10646 stuff, we refer to undated editions, which means that we are prepared to deal with **anything** they throw at us in a new edition. The only thing I can think of in this area is the addition of new code points as letters, digits, and such in identifiers. Our new ANTLR lexical grammar punts to Unicode character classifications by name, and does *not* spell out individual code points, so I think we're safe in that regard. @Nigel-Ecma does that sound right?
4. BTW, we do *not* reference ISO 10646, so I propose we strike this entry.

RexJaeschke:
On today's call, after some discussion, we agreed to keep using undated references in Point 3. And we agreed to Point 4.

Mike Cowlishaw (of the IEEE FP committee) acknowledged receipt of my question re this, and will try to get back to me soon.

RexJaeschke:
Feedback from Mike Cowlishaw:

> David Hough has collated a list of changes at:  https://754r.ucbtest.org/background.  I discussed your question on the CFP telecom the other day with David, Jim Thomas, Fred Tydeman, etc., and we all agreed that it is extremely unlikely that any changes would affect your document, especially as any major changes are only recommendations.  Probably worth you checking the changes list, however.
> In short:  yes go for the 2020 ISO edition.

BTW, the 3 others Mike mentioned all worked with me on the C language standard.

jskeet:
Meeting result: agreed; Rex to finish up.

PR #463, opened 2022-02-24
update FP ref, drop 10646

RexJaeschke:
Closes Issue #426.

Note that "IEEE" is no longer part of the official title. (IEEE has its own equivalent standard.)

### DIFF
--- a/standard/normative-references.md
+++ b/standard/normative-references.md
@@ -8,8 +8,6 @@ ISO 80000-2, *Quantities and units --- Part 2: Mathematical signs and symbols to
 
 ISO/IEC 2382, *Information technology --- Vocabulary*.
 
-ISO/IEC 10646 (all parts), *Information technology --- Universal Multiple-Octet Coded Character Set (UCS)*.
-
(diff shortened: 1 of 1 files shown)


jskeet:
> (IEEE has its own equivalent standard.)

Should that be noted somewhere? I don't know about anyone else, but I typically talk about "IEEE-754" when it comes to FP stuff - it might be alarming to see a different standard referenced, without any acknowledgement of IEEE. Completely understand if this is undesirable for various reasons, however.

RexJaeschke:
Yes, the whole world knows and uses the term "IEEE FP", while very few know about the ISO/IEC equivalent spec. However, the politically correct thing for an internationally distributed spec is to reference the ISO/IEC spec, and that's what all language specs I've worked on do.

We have a similar situation with Unicode and ISO 10646. The 10646 folks always want acknowledgement for their contributions when most people have no idea what 10646 is, but they all know Unicode! (And we are in the processing of removing the 10646 reference.)

jskeet (approved):

Issue #464, opened 2022-02-24
Inconsistent use of formatting for kinds of accessors


RexJaeschke:
The text "get accessor" is sometimes set as "`get` accessor," although the two appear to represent the same thing. Here's a summary of what I found in the md files:

Text | Uses a code fence | No code fence
---- | ---------- | ---------------
get accessor    | 28 | 10
set accessor    | 39 | 18
add accessor   | 2  | 0
remove accessor | 4 | 3

BTW, separately, the term "getter" (sometimes with the suffix "-only")---without a code fence---occurs 5 times. "setter" is never used.

The solution is not quite a "global" pattern edit, as we have "`get` and `set` accessors" (vs. `get` and `set` tokens). Of course, many occurrences of `get` and `set` need to stay fenced.

BTW, V9 will add an `init` accessor, so it will be good to clean up this inconsistent use.

@jskeet, @BillWagner, as this is a relatively simple fix, I can do it for V6. I propose removing the fence from *all* instances that are *not* referring to the spelling of the tokens. (Note that some unfenced instances are in code comments, and we certainly don't want to add fences to them!)

jskeet:
Yup, I'd be fine with that.

BillWagner:
Thanks @RexJaeschke 

I agree, this is the right plan.

PR #465, opened 2022-02-24
Format accessor names

RexJaeschke:
This addresses Issue #464.

Almost all edits involve removing the md code fence from around a single word. However, I also removed a couple of parenthetic remarks of the form "accessors (as in, \`get\` and \`set\` blocks)," as they were redundant; just saying "accessor" implied those, so no need to spell it out! (In any event, we don't define the terms "get"/"set" blocks.)


jskeet (approved):
All changes look good to me. I can't confirm that there aren't others that have been missed, of course.

BillWagner (approved):
This also LGTM @RexJaeschke 

Let's :shipit:

PR #466, opened 2022-02-24
accessor term usage cleanup

RexJaeschke:
Bill, this might be the first time I've created a PR *without* having an issue first. As it seems like a "no-brainer," I skipped that step. Is that OK?

For properties, we define the adjectival (noun not included) terms ***read-only***, ***read-write***, and ***write-only***. However, we have no analogous terms defined for indexers, although such critters exist. So, I extended the 3 existing property-related terms to include the suffix "property", and cloned the whole set of definitions for indexers.

Separately, I replaced the 5 instances of "getter" (an undefined term used in the context of properties and indexers) with "read-only" or "get accessor," as appropriate. We do not use "setter," so no changes were needed for that.


BillWagner (approved):
This all LGTM @RexJaeschke 

To answer your question: yes, it's fine to create PRs without associated issues when the task is obvious.

PR #471, opened 2022-03-03
Reformatting of examples in section 8, 9 and 10

jskeet:
I believe this also fixes a mistake in 9.4.4.16 where there's a
spurious semicolon after `LABEL:`. (It's not invalid, but I don't
believe it's intentional.)


BillWagner (approved):
This is all set to go @jskeet 

Let's :shipit:

jskeet:
Thanks! Will keep going tomorrow...

PR #474, opened 2022-03-04
format code

BillWagner:
This PR formats the sample code in the two clauses for:

- Attributes
- Unsafe code

There were no changes needed in the other annexes (like the standard library annex)

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -116,20 +116,19 @@ Attribute classes can have ***positional parameters*** and ***named parameters**
 > [AttributeUsage(AttributeTargets.Class)]
 > public class HelpAttribute : Attribute
 > {
->     public HelpAttribute(string url)
+>     public HelpAttribute(string url) // url is a positional parameter
 >     { 
->         // url is a positional parameter
 >         ...
 >     }
+>
+>     // Topic is a named parameter
 >     public string Topic
->     { // Topic is a named parameter
+>     { 
 >         get;
 >         set;
 >     }
->     public string Url
->     {
->          get; 
->     }
+>
+>     public string Url { get; }
 > }
 > ```
 > defines an attribute class named `HelpAttribute` that has one positional parameter, `url`, and one named parameter, `Topic`. Although it is non-static and public, the property `Url` does not define a named parameter, since it is not read-write.
@@ -364,16 +363,12 @@ It is a compile-time error to use a single-use attribute class more than once on
 > [AttributeUsage(AttributeTargets.Class)]
 > public class HelpStringAttribute : Attribute
 > {
->     string value;
 >     public HelpStringAttribute(string value)
 >     {
->         this.value = value;
+>         Value = value;
 >     }
 >
->     public string Value
->     { 
->             get {...} 
->     }
+>     public string Value { get; }
 > }
 > [HelpString("Description of Class1")]
 > [HelpString("Another description of Class1")]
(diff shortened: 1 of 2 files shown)


jskeet:
Will review on Monday - busy weekend!

RexJaeschke (approved):

standard/attributes.md
```diff
@@ -109,17 +114,21 @@ Attribute classes can have ***positional parameters*** and ***named parameters**
 > ```csharp
 > using System;
 > [AttributeUsage(AttributeTargets.Class)]
-> public class HelpAttribute: Attribute
+> public class HelpAttribute : Attribute
 > {
->     public HelpAttribute(string url) { // url is a positional parameter
+>     public HelpAttribute(string url)
+>     { 
+>         // url is a positional parameter
```
jskeet: I think this comment should still be on the declaration line (so the brace is moved but not the comment)

standard/attributes.md
```diff
@@ -109,17 +114,21 @@ Attribute classes can have ***positional parameters*** and ***named parameters**
...
 >         ...
 >     }
->     public string Topic { // Topic is a named parameter
->         get {...}
->         set {...}
+>     public string Topic
+>     { // Topic is a named parameter
```
jskeet: Likewise, let's move this comment to the declaration, or better yet:

```csharp
// Topic is a named parameter
public string Topic { get; set; }
```

standard/attributes.md
```diff
@@ -109,17 +114,21 @@ Attribute classes can have ***positional parameters*** and ***named parameters**
...
+>         get;
+>         set;
 >     }
->     public string Url {
->          get {...} 
+>     public string Url
```
jskeet: Rewrite to `public string Url { get; }`?

standard/attributes.md
```diff
@@ -341,14 +360,18 @@ It is a compile-time error to use a single-use attribute class more than once on
 > *Example*: The example
 > ```csharp
 > using System;
+>
 > [AttributeUsage(AttributeTargets.Class)]
-> public class HelpStringAttribute: Attribute
+> public class HelpStringAttribute : Attribute
 > {
 >     string value;
```
jskeet: Remove the field entirely and make it a read-only autoprop?

standard/attributes.md
```diff
@@ -341,14 +360,18 @@ It is a compile-time error to use a single-use attribute class more than once on
...
->     public HelpStringAttribute(string value) {
+>     public HelpStringAttribute(string value)
+>     {
 >         this.value = value;
 >     }
->     public string Value { 
+>
+>     public string Value
+>     { 
 >             get {...}
```
jskeet: Even if we keep this on a line on its own, the indentation looks wrong - maybe it's using a tab instead of spaces?

standard/attributes.md
```diff
@@ -370,30 +393,40 @@ An expression `E` is an *attribute_argument_expression* if all of the following
...
->     set {...}
-> }
-> public Type P2 {
->     get {...}
->     set {...}
-> }
-> public object P3 {
->     get {...}
->     set {...}
+>     public int P1
```
jskeet: I'd make all of these one-liners.

standard/attributes.md
```diff
@@ -512,12 +550,14 @@ A method decorated with the `Conditional` attribute is a conditional method. Eac
...
+>
 > class Eg
 > {
 >     [Conditional("ALPHA")]
 >     [Conditional("BETA")]
->     public static void M() {
->     //...
+>     public static void M()
+>     {
+>         //...
```
jskeet: Add a space between the `//` and `...`? A pet peeve of mine is comments written without capitals and without the space ;)

standard/attributes.md
```diff
@@ -598,18 +645,20 @@ The use of conditional methods in an inheritance chain can be confusing. Calls m
 > // File `class1.cs`
 > using System;
 > using System.Diagnostics;
+>
 > class Class1
 > {
 >     [Conditional("DEBUG")]
->     public virtual void M() {
+>     public virtual void M() =>
```
jskeet: How would you feel about unwrapping the body of this onto the same line?

standard/attributes.md
```diff
@@ -679,13 +732,15 @@ If a program uses a type or member that is decorated with the `Obsolete` attribu
 > {
 >     public void F() {}
 > }
+>
 > class B
 > {
 >     public void F() {}
-> }
+> }>
```
jskeet: I suspect you meant there to be a line-break between the `}` and `>` :)

jskeet (approved):
Some nits, but otherwise great, thanks :)

standard/attributes.md
BillWagner: That fits!

PR #477, opened 2022-03-08
Improved wording re nested generic structs

RexJaeschke:
A class-related version of the proposed sentence exists for [class declarations](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/classes.md#1421-general), and it seems that the same intent applies to struct declaration as well, but was not so stated.

### DIFF
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -25,7 +25,7 @@ A *struct_declaration* consists of an optional set of *attributes* ([§21](attri
 
 A struct declaration shall not supply a *type_parameter_constraints_clauses* unless it also supplies a *type_parameter_list*.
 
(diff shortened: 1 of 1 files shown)


jskeet (approved):

KalleOlaviNiemitalo:
Should similar wording be added for interface declarations and delegate declarations, as well?

Not for enum declarations though.

jskeet:
Interface and delegate declarations within generic classes and structs - yes, well spotted. @RexJaeschke are you happy to find and add these?

RexJaeschke:
Done; see PR https://github.com/dotnet/csharpstandard/pull/480.

PR #478, opened 2022-03-08
Improve named parameter wording

RexJaeschke:
The change may seem pedantic, but the new wording makes it clear that *both* accessors are needed, and *both* must be public. Also, this makes it easier to allow init accessors instead of get accessors in V9.


jskeet (approved):

BillWagner (approved):
This also LGTM @RexJaeschke

PR #480, opened 2022-03-08
Improved wording re nested generic interfaces and delegates

RexJaeschke:
@KalleOlaviNiemitalo, @jskeet If I understand correctly, these are the changes you requested.

[And as I'm currently working on the spec for records, I'll add this in there as well.]

### DIFF
--- a/standard/delegates.md
+++ b/standard/delegates.md
@@ -32,7 +32,7 @@ It is a compile-time error for the same modifier to appear multiple times in a d
 
 A delegate declaration shall not supply any *type_parameter_constraints_clause*s unless it also supplies a *variant_type_parameter_list*.
 
-A delegate declaration that supplies a *variant_type_parameter_list* is a generic delegate declaration.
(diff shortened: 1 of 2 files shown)


jskeet (approved):
Yup, I believe this is fine.

Might be worth getting @MadsTorgersen to just check as well.

KalleOlaviNiemitalo:
Now I wonder whether this should be added to enum type declarations after all.

```C#
class C<T> {
    enum E {}
    interface I {}
}
```

`C<int>.E` and `C<long>.E` are distinct types, right? Just like 
`C<int>.I` and `G<long>.I`. Should `C<>.E` be considered a generic type, then? Which clauses in the standard are affected by whether the type is generic?

Fortunately, `class C<out T>` is not allowed, and interface and delegate types cannot have nested types, so the nested types cannot gain generic variance from the outer type.

KalleOlaviNiemitalo:
> interface and delegate types cannot have nested types

Or can they… sharplab.io lets me define an interface within an interface.

jskeet:
> `C<int>.E` and `C<long>.E` are distinct types, right?

Yes - you can't assign a value from one to the other. This is a bit of an odd situation, given that enums can't be generic in other ways. Will file a new issue.

> Or can they… sharplab.io lets me define an interface within an interface.

Not in C# 6 though, although the error message refers to "default interface implementation" which is a bit odd.

PR #481, opened 2022-03-08
initial test project

BillWagner:
Several changes here:

- Create a unit test project for the word converter.
- Add one test. That caused....
- Making a number of types `public` instead of `internal`
- Define GitHub workflow to run tests on PRs that update any tool. (This is aspirational, but there's probably good reason to add test projects for other tools in time)


BillWagner:
First commit runs the tests with an expected failure. The second commit fixes that test.

jskeet (approved):

PR #484, opened 2022-03-09
Better betterness

### LINKED ISSUE #157: V6 Feature: Improved overload resolution
What's new in C#6 at https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-6#improved-overload-resolution, says:

> This last feature is one you probably won't notice. There were constructs where the previous version of the C# compiler may have found some method calls involving lambda expressions ambiguous.

Does this involve something previously disallowed or unspecified, but now allowed? What, if anything, do we need to say in the spec re this?

### LINKED ISSUE #283: Bug in spec for better conversion from expression
@gafter commented on [Tue Sep 27 2016](https://github.com/dotnet/roslyn/issues/14117)

The spec for better conversion from expression doesn't match the implementation. The spec says

> 7.5.3.3 Better conversion from expression
> Given an implicit conversion C1 that converts from an expression E to a type T1, and an implicit conversion C2 that converts from an expression E to a type T2, C1 is a better conversion than C2 if E does not exactly match T2 and one of the following holds:
> • E exactly matches T1 (§7.5.3.4)
> • T1 is a better conversion target than T2 (§7.5.3.5)

but the implementation does

``` cs
            bool t1MatchesExactly = ExpressionMatchExactly(node, t1, ref useSiteDiagnostics);
            bool t2MatchesExactly = ExpressionMatchExactly(node, t2, ref useSiteDiagnostics);

            if (t1MatchesExactly)
            {
                if (!t2MatchesExactly)
                {
                    // - E exactly matches T1
                    return BetterResult.Left;
                } // NOTE NO ELSE HERE
            }
            else if (t2MatchesExactly)
            {
                // - E exactly matches T2
                return BetterResult.Right;
            }

            // - T1 is a better conversion target than T2
            return BetterConversionTarget(node, t1, conv1, t2, conv2, ref useSiteDiagnostics, out okToDowngradeToNeither);
```

Specifically, if both match exactly, then we test for a better conversion target. This is what is implemented:

> 7.5.3.3 Better conversion from expression
> Given an implicit conversion C1 that converts from an expression E to a type T1, and an implicit conversion C2 that converts from an expression E to a type T2, C1 is a better conversion than C2 if one of the following holds:
> • E does not exactly match T2 and E exactly matches T1 (§7.5.3.4)
> • E either exactly matches both T1 and T2, or exactly matches neither T1 and T2, and also T1 is a better conversion target than T2 (§7.5.3.5)

---

@gafter commented on [Tue Sep 27 2016](https://github.com/dotnet/roslyn/issues/14117#issuecomment-250006038)

This issue is preventing the proposed spec approach for `out var` overload resolution:

> an inferred “out var” argument exactly matches (see 7.5.3.4 Exactly matching Expression) every type. 

from having the effect of causing out var arguments to ignore the corresponding parameter position.

---

@gafter commented on [Tue Sep 27 2016](https://github.com/dotnet/roslyn/issues/14117#issuecomment-250013651)

This appears to be mooted in the ECMA version. But the ECMA version has no concept of exactly matching expression.

---

@AlekseyTs commented on [Tue Sep 27 2016](https://github.com/dotnet/roslyn/issues/14117#issuecomment-250019454)

It looks like the implementation is intentional, we would want the following scenario compiling without errors and using the second overload:

```
using System;
namespace ConsoleApplication2
{
    public static class Foo
    {
        public static U IfNotNull<T, U>(this T value, Func<T, U> selector)
        {
            System.Console.WriteLine("IfNotNull<T, U>(this T value, Func<T, U> selector");
            return value != null ? selector(value) : default(U);
        }
        public static U IfNotNull<T, U>(this T? source, Func<T, U> selector) where T : struct
        {
            System.Console.WriteLine("IfNotNull<T, U>(this T? source, Func<T, U> selector)");
            return source.HasValue ? selector(source.Value) : default(U);
        }
    }
    class Program
    {
        static void Main(string[] args)
        {
            double? val = null;
            var d1 = val.IfNotNull(v => v / 100);
            var d2 = Foo.IfNotNull(val, v => v / 100);
        }
    }
}
```

---

@gafter commented on [Tue Sep 27 2016](https://github.com/dotnet/roslyn/issues/14117#issuecomment-250019576)

Indeed. Moreover, we want to implement the ECMA spec instead of the old C# 4 spec, which is quoted above in the OP.

---

@gafter commented on [Tue Sep 27 2016](https://github.com/dotnet/roslyn/issues/14117#issuecomment-250019721)

Closing, as we are intending to use the ECMA spec as our base, and it does not have this problem.

---

@MadsTorgersen commented on [Tue Sep 27 2016](https://github.com/dotnet/roslyn/issues/14117#issuecomment-250030490)

We cannot use the ECMA spec as our base, because it does not reflect the changes for "better betterness" that were done in C# 6.0. Instead, the ECMA spec needs to embrace the Microsoft spec when it gets updated for C# 6.0.

I agree that there is a discrepancy between how "better betterness" was spec'ed and how it's implemented. The implementation is already in use, so unless there's a strong reason not to (I don't see any), we should go with what's implemented.

That means adjusting the spec a bit, so that better conversion target is checked both when _neither_ matches precisely, and when _both_ do. Something like this:

> 7.5.3.3 Better conversion from expression
> Given an implicit conversion C1 that converts from an expression E to a type T1, and an implicit conversion C2 that converts from an expression E to a type T2, C1 is a better conversion than C2 if E does not exactly match T2 and one of the following holds:
> • E exactly matches T1 and E does not exactly match T2 (§7.5.3.4)
> • E exactly matches both or neither of T1 and T2, and T1 is a better conversion target than T2 (§7.5.3.5)

---

@gafter commented on [Tue Sep 27 2016](https://github.com/dotnet/roslyn/issues/14117#issuecomment-250033918)

@MadsTorgersen 
I think you intended to delete

> E does not exactly match T2 and 

from the prefix to the bulleted list.

---

@gafter commented on [Tue Sep 27 2016](https://github.com/dotnet/roslyn/issues/14117#issuecomment-250035753)

@MadsTorgersen That reopen the question if how you would like to handle `out var` in the spec.

BillWagner:
Fixes #157
Fixes #283 

This takes the text from [Better betterness](https://github.com/dotnet/roslyn/blob/main/docs/specs/CSharp%206/Better%20Betterness.md) and then updates it for compatibility fixes, related to exact matches and method group conversions.

Reviewing each commit in order clarifies what changes came from the roslyn doc, and then those changes that followed after incorporating that initial spec. Commit messages link to dotnet/roslyn#6750 that explains the later changes.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -965,7 +965,7 @@ Given an expression `E` and a type `T`, `E` ***exactly matches*** `T` if one of 
 Given two different types `T₁` and `T₂`, `T₁` is a ***better conversion target*** than `T₂` if one of the following holds:
 
 - An implicit conversion from `T₁` to `T₂` exists
-- In case of a method group conversion (§10.6) for the corresponding argument, if a better conversion target (§7.5.3.5), is a delegate type that is not compatible (§19.4) with the single best method from the method group (§10.6), then neither delegate type is better.
+- In case of a method group conversion (§10.6) for the corresponding argument, if a better conversion target (§11.6.4.6), is a delegate type that is not compatible (§19.4) with the single best method from the method group (§10.6), then neither delegate type is better.
 - `T₁` is `Task<S₁>`, `T₂` is `Task<S₂>`, and `S₁` is a better conversion target than `S₂`
 - `T₁` is `S₁` or `S₁?` where `S₁` is a signed integral type, and `T₂` is `S₂` or `S₂?` where `S₂` is an unsigned integral type. Specifically:
   - `S₁` is `sbyte` and `S₂` is `byte`, `ushort`, `uint`, or `ulong`
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -945,12 +945,6 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
 
 #### 11.6.4.4 Better conversion from expression
 
-<!-- 
-> Section 7.5.3.3 is now 11.6.4.4
-> Section 7.5.3.4 is now 11.6.4.5
-> Section 7.5.3.5 is now 11.6.4.6
--->
-
 Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a better conversion than `C₂` if one of the following holds:
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` (§11.6.4.5)
 - `E` exactly matches both or neither of `T₁` and `T2`, and `T₁` is a better conversion target than `T2` (§11.6.4.6)
@@ -963,17 +957,15 @@ Given an expression `E` and a type `T`, `E` ***exactly matches*** `T` is one of 
 - `E` is an anonymous function, `T` is either a delegate type `D` or an expression tree type `Expression<D>` and one of the following holds:
   - An inferred return type `X` exists for `E` in the context of the parameter list of `D` (§11.6.3.12), and an identity conversion exists from `X` to the return type of `D`
   - Either `E` is non-async and `D` has a return type `Y` or `E` is async and  `D` has a return type `Task<Y>`, and one of the following holds:
-- The body of `E` is an expression that exactly matches `Y`
-- The body of `E` is a statement block where every return statement returns an expression that exactly matches `Y`
+    - The body of `E` is an expression that exactly matches `Y`
+    - The body of `E` is a statement block where every return statement returns an expression that exactly matches `Y`
 
 ### 11.6.4.6 Better conversion target
 
 Given two different types `T₁` and `T₂`, `T₁` is a ***better conversion target*** than `T₂` if:
 
 - An implicit conversion from `T₁` to `T₂` exists
-- `T₁` is either a delegate type `D₁` or an expression tree type `Expression<D₁>`, `T₂` is either a delegate type `D₂` or an expression tree type `Expression<D₂>`, `D₁` has a return type `S₁` and one of the following holds:
-  - `D₂` is void returning
-  - `D₂` has a return type `S₂`, and `S₁` is a better conversion target than `S₂`
+- In case of a method group conversion (§10.6) for the corresponding argument, if a better conversion target (§7.5.3.5), is a delegate type that is not compatible (§19.4) with the single best method from the method group (§10.6), then neither delegate type is better.
 - `T₁` is `Task<S₁>`, `T₂` is `Task<S₂>`, and `S₁` is a better conversion target than `S₂`
 - `T₁` is `S₁` or `S₁`? where `S₁` is a signed integral type, and `T₂` is `S₂` or `S₂`? where `S₂` is an unsigned integral type. Specifically:
   - `S₁` is `sbyte` and `S₂` is `byte`, `ushort`, `uint`, or `ulong`
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -945,34 +945,42 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
 
 #### 11.6.4.4 Better conversion from expression
 
-Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if at least one of the following holds:
+<!-- 
+> Section 7.5.3.3 is now 11.6.4.4
+> Section 7.5.3.4 is now 11.6.4.5
+> Section 7.5.3.5 is now 11.6.4.6
+-->
 
-- `E` has a type `S` and an identity conversion exists from `S` to `T₁` but not from `S` to `T₂`
-- `E` is not an anonymous function and `T₁` is a better conversion target than `T₂` ([§11.6.4.6](expressions.md#11646-better-conversion-target))
-- `E` is an anonymous function, `T₁` is either a delegate type `D₁` or an expression tree type `Expression<D₁>`, `T₂` is either a delegate type `D₂` or an expression tree type `Expression<D₂>` and one of the following holds:
-  - `D₁` is a better conversion target than `D₂`
-  - `D₁` and `D₂` have identical parameter lists, and one of the following holds:
-    - `D₁` has a return type `Y₁`, and `D₂` has a return type `Y₂`, an inferred return type `X` exists for `E` in the context of that parameter list ([§11.6.3.13](expressions.md#116313-inferred-return-type)), and the conversion from `X` to `Y₁` is better than the conversion from `X` to `Y₂`
-    - `E` is async, `D₁` has a return type `Task<Y₁>`, and `D₂` has a return type `Task<Y>`, an inferred return type `Task<X>` exists for `E` in the context of that parameter list ([§11.6.3.13](expressions.md#116313-inferred-return-type)), and the conversion from `X` to `Y₁` is better than the conversion from `X` to `Y₂`
-    - `D₁` has a return type `Y`, and `D₂` is void returning
+Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if `E` does not exactly match `T₂` and one of the following holds:
 
-#### 11.6.4.5 Better conversion from type
+- `E` exactly matches `T₁` (§11.6.4.5)
+- `T₁` is a better conversion target than `T₂` (§11.6.4.6)
 
-Given a conversion `C₁` that converts from a type `S` to a type `T₁`, and a conversion `C₂` that converts from a type `S` to a type `T₂`, `C₁` is a *better conversion* than `C₂` if at least one of the following holds:
+### 11.6.4.5 Exactly matching Expression
 
--   An identity conversion exists from `S` to `T₁` but not from `S` to `T₂`
--   `T₁` is a better conversion target than `T₂` ([§11.6.4.6](expressions.md#11646-better-conversion-target))
+Given an expression `E` and a type `T`, `E` ***exactly matches*** `T` is one of the following holds:
 
-#### 11.6.4.6 Better conversion target
+- `E` has a type `S`, and an identity conversion exists from `S` to `T`
+- `E` is an anonymous function, `T` is either a delegate type `D` or an expression tree type `Expression<D>` and one of the following holds:
+  - An inferred return type `X` exists for `E` in the context of the parameter list of `D` (§11.6.3.12), and an identity conversion exists from `X` to the return type of `D`
+  - Either `E` is non-async and `D` has a return type `Y` or `E` is async and  `D` has a return type `Task<Y>`, and one of the following holds:
+- The body of `E` is an expression that exactly matches `Y`
+- The body of `E` is a statement block where every return statement returns an expression that exactly matches `Y`
 
-Given two different types `T₁` and `T₂`, `T₁` is a better conversion target than `T₂` if at least one of the following holds:
+### 11.6.4.6 Better conversion target
 
-- An implicit conversion from `T₁` to `T₂` exists, and no implicit conversion from `T₂` to `T₁` exists
-- `T₁` is a signed integral type and `T₂` is an unsigned integral type. Specifically:
-  - `T₁` is `sbyte` and `T₂` is `byte`, `ushort`, `uint`, or `ulong`
-  - `T₁` is `short` and `T₂` is `ushort`, `uint`, or `ulong`
-  - `T₁` is `int` and `T₂` is `uint`, or `ulong`
-  - `T₁` is `long` and `T₂` is `ulong`
+Given two different types `T₁` and `T₂`, `T₁` is a ***better conversion target*** than `T₂` if:
+
+- An implicit conversion from `T₁` to `T₂` exists
+- `T₁` is either a delegate type `D₁` or an expression tree type `Expression<D₁>`, `T₂` is either a delegate type `D₂` or an expression tree type `Expression<D₂>`, `D₁` has a return type `S₁` and one of the following holds:
+  - `D₂` is void returning
+  - `D₂` has a return type `S₂`, and `S₁` is a better conversion target than `S₂`
+- `T₁` is `Task<S₁>`, `T₂` is `Task<S₂>`, and `S₁` is a better conversion target than `S₂`
+- `T₁` is `S₁` or `S₁`? where `S₁` is a signed integral type, and `T₂` is `S₂` or `S₂`? where `S₂` is an unsigned integral type. Specifically:
+  - `S₁` is `sbyte` and `S₂` is `byte`, `ushort`, `uint`, or `ulong`
+  - `S₁` is `short` and `S₂` is `ushort`, `uint`, or `ulong`
+  - `S₁` is `int` and `S₂` is `uint`, or `ulong`
+  - `S₁` is `long` and `S₂` is `ulong`
 
 #### 11.6.4.7 Overloading in generic classes
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -946,8 +946,9 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
 #### 11.6.4.4 Better conversion from expression
 
 Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a better conversion than `C₂` if one of the following holds:
+
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` (§11.6.4.5)
-- `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T2` (§11.6.4.6)
+- `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` (§11.6.4.6)
 
 #### 11.6.4.5 Exactly matching expression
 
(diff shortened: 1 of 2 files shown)


standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
-Given two different types `T₁` and `T₂`, `T₁` is a better conversion target than `T₂` if at least one of the following holds:
-
-- An implicit conversion from `T₁` to `T₂` exists, and no implicit conversion from `T₂` to `T₁` exists
-- `T₁` is a signed integral type and `T₂` is an unsigned integral type. Specifically:
-  - `T₁` is `sbyte` and `T₂` is `byte`, `ushort`, `uint`, or `ulong`
-  - `T₁` is `short` and `T₂` is `ushort`, `uint`, or `ulong`
-  - `T₁` is `int` and `T₂` is `uint`, or `ulong`
-  - `T₁` is `long` and `T₂` is `ulong`
+- An implicit conversion from `T₁` to `T₂` exists
+- In case of a method group conversion (§10.6) for the corresponding argument, if a better conversion target (§7.5.3.5), is a delegate type that is not compatible (§19.4) with the single best method from the method group (§10.6), then neither delegate type is better.
```
BillWagner: This bullet item still needs work (where it may be more than one bullet item)

standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
 
 #### 11.6.4.4 Better conversion from expression
 
-Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if at least one of the following holds:
+Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a better conversion than `C₂` if one of the following holds:
+- `E` exactly matches `T₁` and `E` does not exactly match `T₂` (§11.6.4.5)
+- `E` exactly matches both or neither of `T₁` and `T2`, and `T₁` is a better conversion target than `T2` (§11.6.4.6)
```
jskeet: Nit: T2 should be T₂.

More importantly, this looks different to the Roslyn doc, in terms of the second bullet and E exactly matching T₂.

Is this effectively a fix to what was written in the Roslyn doc?

standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
-- `E` is an anonymous function, `T₁` is either a delegate type `D₁` or an expression tree type `Expression<D₁>`, `T₂` is either a delegate type `D₂` or an expression tree type `Expression<D₂>` and one of the following holds:
-  - `D₁` is a better conversion target than `D₂`
-  - `D₁` and `D₂` have identical parameter lists, and one of the following holds:
-    - `D₁` has a return type `Y₁`, and `D₂` has a return type `Y₂`, an inferred return type `X` exists for `E` in the context of that parameter list ([§11.6.3.13](expressions.md#116313-inferred-return-type)), and the conversion from `X` to `Y₁` is better than the conversion from `X` to `Y₂`
-    - `E` is async, `D₁` has a return type `Task<Y₁>`, and `D₂` has a return type `Task<Y>`, an inferred return type `Task<X>` exists for `E` in the context of that parameter list ([§11.6.3.13](expressions.md#116313-inferred-return-type)), and the conversion from `X` to `Y₁` is better than the conversion from `X` to `Y₂`
-    - `D₁` has a return type `Y`, and `D₂` is void returning
+### 11.6.4.5 Exactly matching Expression
```
jskeet: Is there a reason for Expression to have a capital E here (it doesn't in the preceding section)

standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
 
-#### 11.6.4.5 Better conversion from type
+Given an expression `E` and a type `T`, `E` ***exactly matches*** `T` is one of the following holds:
 
-Given a conversion `C₁` that converts from a type `S` to a type `T₁`, and a conversion `C₂` that converts from a type `S` to a type `T₂`, `C₁` is a *better conversion* than `C₂` if at least one of the following holds:
+- `E` has a type `S`, and an identity conversion exists from `S` to `T`
+- `E` is an anonymous function, `T` is either a delegate type `D` or an expression tree type `Expression<D>` and one of the following holds:
+  - An inferred return type `X` exists for `E` in the context of the parameter list of `D` (§11.6.3.12), and an identity conversion exists from `X` to the return type of `D`
+  - Either `E` is non-async and `D` has a return type `Y` or `E` is async and  `D` has a return type `Task<Y>`, and one of the following holds:
```
jskeet, line 960: How does that work for `ValueTask<Y>` etc?

standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
+    - The body of `E` is an expression that exactly matches `Y`
+    - The body of `E` is a statement block where every return statement returns an expression that exactly matches `Y`
```
jskeet, line 962: Do we need "statement block" rather than just "block"?

standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
-  - `D₁` is a better conversion target than `D₂`
-  - `D₁` and `D₂` have identical parameter lists, and one of the following holds:
-    - `D₁` has a return type `Y₁`, and `D₂` has a return type `Y₂`, an inferred return type `X` exists for `E` in the context of that parameter list ([§11.6.3.13](expressions.md#116313-inferred-return-type)), and the conversion from `X` to `Y₁` is better than the conversion from `X` to `Y₂`
-    - `E` is async, `D₁` has a return type `Task<Y₁>`, and `D₂` has a return type `Task<Y>`, an inferred return type `Task<X>` exists for `E` in the context of that parameter list ([§11.6.3.13](expressions.md#116313-inferred-return-type)), and the conversion from `X` to `Y₁` is better than the conversion from `X` to `Y₂`
-    - `D₁` has a return type `Y`, and `D₂` is void returning
+### 11.6.4.5 Exactly matching Expression
 
-#### 11.6.4.5 Better conversion from type
+Given an expression `E` and a type `T`, `E` ***exactly matches*** `T` is one of the following holds:
```
jskeet: Nit: is => if

standard/expressions.md
jskeet: Should this be `####` rather than `###`?

standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
+- `E` has a type `S`, and an identity conversion exists from `S` to `T`
+- `E` is an anonymous function, `T` is either a delegate type `D` or an expression tree type `Expression<D>` and one of the following holds:
+  - An inferred return type `X` exists for `E` in the context of the parameter list of `D` (§11.6.3.12), and an identity conversion exists from `X` to the return type of `D`
+  - Either `E` is non-async and `D` has a return type `Y` or `E` is async and  `D` has a return type `Task<Y>`, and one of the following holds:
+    - The body of `E` is an expression that exactly matches `Y`
+    - The body of `E` is a statement block where every return statement returns an expression that exactly matches `Y`
 
--   An identity conversion exists from `S` to `T₁` but not from `S` to `T₂`
--   `T₁` is a better conversion target than `T₂` ([§11.6.4.6](expressions.md#11646-better-conversion-target))
+### 11.6.4.6 Better conversion target
```
jskeet: Again, I suspect this should be `####` (or 11.6.4.4 is wrong)

standard/expressions.md
jskeet: Right - "In case of a method group conversion" definitely seems like an odd start to this bullet.

standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
 
-#### 11.6.4.6 Better conversion target
+Given two different types `T₁` and `T₂`, `T₁` is a ***better conversion target*** than `T₂` if:
```
jskeet: We need something like "if at least one of the following holds" or "if all of the following hold"

standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
-- An implicit conversion from `T₁` to `T₂` exists, and no implicit conversion from `T₂` to `T₁` exists
-- `T₁` is a signed integral type and `T₂` is an unsigned integral type. Specifically:
-  - `T₁` is `sbyte` and `T₂` is `byte`, `ushort`, `uint`, or `ulong`
-  - `T₁` is `short` and `T₂` is `ushort`, `uint`, or `ulong`
-  - `T₁` is `int` and `T₂` is `uint`, or `ulong`
-  - `T₁` is `long` and `T₂` is `ulong`
+- An implicit conversion from `T₁` to `T₂` exists
+- In case of a method group conversion (§10.6) for the corresponding argument, if a better conversion target (§7.5.3.5), is a delegate type that is not compatible (§19.4) with the single best method from the method group (§10.6), then neither delegate type is better.
+- `T₁` is `Task<S₁>`, `T₂` is `Task<S₂>`, and `S₁` is a better conversion target than `S₂`
+- `T₁` is `S₁` or `S₁`? where `S₁` is a signed integral type, and `T₂` is `S₂` or `S₂`? where `S₂` is an unsigned integral type. Specifically:
```
jskeet: Should the question-mark be within the back-ticks in both occurrences of this?

jskeet (commented):
Some initial thoughts that didn't require actual understanding. I haven't tried to work out what situations this actually helps with yet :)

standard/expressions.md
BillWagner: > More importantly, this looks different to the Roslyn doc, in terms of the second bullet and E exactly matching T₂.
> 
> Is this effectively a fix to what was written in the Roslyn doc?

I think so. I added a note in the PR description relating to this. The first commit brings in the roslyn doc. Later commits address issues with that spec language.

standard/expressions.md
BillWagner: I think it will work the same. (Generalized async return types don't come until V7 though).

standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
-- `E` is not an anonymous function and `T₁` is a better conversion target than `T₂` ([§11.6.4.6](expressions.md#11646-better-conversion-target))
-- `E` is an anonymous function, `T₁` is either a delegate type `D₁` or an expression tree type `Expression<D₁>`, `T₂` is either a delegate type `D₂` or an expression tree type `Expression<D₂>` and one of the following holds:
-  - `D₁` is a better conversion target than `D₂`
-  - `D₁` and `D₂` have identical parameter lists, and one of the following holds:
-    - `D₁` has a return type `Y₁`, and `D₂` has a return type `Y₂`, an inferred return type `X` exists for `E` in the context of that parameter list ([§11.6.3.13](expressions.md#116313-inferred-return-type)), and the conversion from `X` to `Y₁` is better than the conversion from `X` to `Y₂`
-    - `E` is async, `D₁` has a return type `Task<Y₁>`, and `D₂` has a return type `Task<Y>`, an inferred return type `Task<X>` exists for `E` in the context of that parameter list ([§11.6.3.13](expressions.md#116313-inferred-return-type)), and the conversion from `X` to `Y₁` is better than the conversion from `X` to `Y₂`
-    - `D₁` has a return type `Y`, and `D₂` is void returning
```
Nigel-Ecma, line 957: In the current Standard we have delegate return type of `Y`, presumably non-void, being a better conversion than a return type of void. I don’t immediately see this case exists in the revised model, is this intentional? A breaking change?

standard/expressions.md
```diff
@@ -945,34 +945,33 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
+Given two different types `T₁` and `T₂`, `T₁` is a ***better conversion target*** than `T₂` if one of the following holds:
 
-- An implicit conversion from `T₁` to `T₂` exists, and no implicit conversion from `T₂` to `T₁` exists
-- `T₁` is a signed integral type and `T₂` is an unsigned integral type. Specifically:
-  - `T₁` is `sbyte` and `T₂` is `byte`, `ushort`, `uint`, or `ulong`
-  - `T₁` is `short` and `T₂` is `ushort`, `uint`, or `ulong`
-  - `T₁` is `int` and `T₂` is `uint`, or `ulong`
-  - `T₁` is `long` and `T₂` is `ulong`
+- An implicit conversion from `T₁` to `T₂` exists
+- In case of a method group conversion (§10.6) for the corresponding argument, if a better conversion target (§11.6.4.6), is a delegate type that is not compatible (§19.4) with the single best method from the method group (§10.6), then neither delegate type is better.
```
Nigel-Ecma: I think the comma in “conversion target (§11.6.4.6), is a delegate type” should not be there? And maybe the one in “method group (§10.6), then neither delegate” as well?

standard/expressions.md
Nigel-Ecma: I've now found the void case is in the source material [Better betterness](https://github.com/dotnet/roslyn/blob/main/docs/specs/CSharp%206/Better%20Betterness.md) so I guess this was unintentional?

BillWagner: This change is intentional, and it was a regression in the original Better Betterness spec. The issue is dotnet/roslyn#6750.  There is still work to do here, because I think this loses some other requirements.

BillWagner:
Adding a comment on the roadmap for much of what I've been reading to reach this point.

- [The better betterness spec](https://github.com/dotnet/roslyn/pull/7934) for C# 6.0. This includes the text for the compatibility fix, and an explanation of the new sections. This PR did not include any of the informative text from that document. We may want to consider some of that.
- [Breaking change due to better betterness](https://github.com/dotnet/roslyn/issues/6560). This describes the case for the later changes to the better betterness spec.
- [Implementation of the C# 6.0 rules](https://github.com/dotnet/roslyn/pull/7727).  This PR contains the changes for the compatibility fix. "Adjust Better Betterness rules around method group conversions in order to restore backward compatibility"
- [Additional change for C# 7.3](https://github.com/dotnet/roslyn/issues/8942) This issue describes another case regarding method group conversions. This was fixed in C# 7.3, as described in the issue. I created #499 to track that when we're working on the C# 7.x spec.

standard/expressions.md
BillWagner: ping @RexJaeschke  to check if "statement block" or "block" is better.  

From Jon: There are two other occurrences of "statement block" that should be changed.  Jon will create an issue for that change.

standard/expressions.md
```diff
@@ -945,34 +945,34 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
-  - `T₁` is `int` and `T₂` is `uint`, or `ulong`
-  - `T₁` is `long` and `T₂` is `ulong`
+- An implicit conversion from `T₁` to `T₂` exists
+- `T₁` is `Task<S₁>`, `T₂` is `Task<S₂>`, and `S₁` is a better conversion target than `S₂`
+- `T₁` is `S₁` or `S₁?` where `S₁` is a signed integral type, and `T₂` is `S₂` or `S₂?` where `S₂` is an unsigned integral type. Specifically:
+  - `S₁` is `sbyte` and `S₂` is `byte`, `ushort`, `uint`, or `ulong`
+  - `S₁` is `short` and `S₂` is `ushort`, `uint`, or `ulong`
+  - `S₁` is `int` and `S₂` is `uint`, or `ulong`
+  - `S₁` is `long` and `S₂` is `ulong`
+- `E` is the result of a method group conversion is a delegate creation expression (§11.7.15.6) whose argument is a method group and `T₁` is compatible (§19.4) with the single best method from the method group (§10.8)
```
jskeet: Add "or" after "method group conversion"?

standard/expressions.md
jskeet: Case that I think is broken by this:

```csharp
void M(Action action) {}
void M(Func<object> func) {}

void Test()
{
    // "xyz" doesn't *exactly match* object, so this should be ambiguous (according to the PR)
    M(() => "xyz");
}
```

Case that I think isn't:

```csharp
void M(Action action) {}
void M(Func<object> func) {}

void Test()
{
    // new object() exactly matches object, therefore Func<object> is preferred
    M(() => new object());
}
```

standard/expressions.md
BillWagner: I reworked the sentence a bit.

BillWagner:
Marking as ready for review, and adding Mads and Neal as reviewers.

Consensus from today's meeting was that this is close, and any future updates can be handled in a separate PR (for C# 6)

standard/expressions.md
```diff
@@ -945,34 +945,34 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
-  - `T₁` is `int` and `T₂` is `uint`, or `ulong`
-  - `T₁` is `long` and `T₂` is `ulong`
+- An implicit conversion from `T₁` to `T₂` exists
+- `T₁` is `Task<S₁>`, `T₂` is `Task<S₂>`, and `S₁` is a better conversion target than `S₂`
+- `T₁` is `S₁` or `S₁?` where `S₁` is a signed integral type, and `T₂` is `S₂` or `S₂?` where `S₂` is an unsigned integral type. Specifically:
+  - `S₁` is `sbyte` and `S₂` is `byte`, `ushort`, `uint`, or `ulong`
+  - `S₁` is `short` and `S₂` is `ushort`, `uint`, or `ulong`
+  - `S₁` is `int` and `S₂` is `uint`, or `ulong`
+  - `S₁` is `long` and `S₂` is `ulong`
+- `E` is a delegate creation expression (§11.7.15.6) resulting from the result of a method group conversion and `T₁` is compatible (§19.4) with the single best method from the method group (§10.8)
```
jskeet: Do we really only want this for delegate creation expressions such as `new Action(Method)` rather than plain method group conversions? I suspect it should actually cover both - or possibly *just* method group conversions, given that if there's a delegate creation expression we know the exact type anyway.

Separately, "resulting from the result" sounds redundant and says the same thing twice ;)

BillWagner: > Do we really only want this for delegate creation expressions such as `new Action(Method)` rather than plain method group conversions? I suspect it should actually cover both - or possibly _just_ method group conversions, given that if there's a delegate creation expression we know the exact type anyway.

It should only be plain method group conversions. The `new Action(Method)` is an exact matching expression for `Action`.

I'll reword the sentence.

standard/expressions.md
```diff
@@ -945,34 +945,34 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
...
-  - `T₁` is `int` and `T₂` is `uint`, or `ulong`
-  - `T₁` is `long` and `T₂` is `ulong`
+- An implicit conversion from `T₁` to `T₂` exists
+- `T₁` is `Task<S₁>`, `T₂` is `Task<S₂>`, and `S₁` is a better conversion target than `S₂`
+- `T₁` is `S₁` or `S₁?` where `S₁` is a signed integral type, and `T₂` is `S₂` or `S₂?` where `S₂` is an unsigned integral type. Specifically:
+  - `S₁` is `sbyte` and `S₂` is `byte`, `ushort`, `uint`, or `ulong`
+  - `S₁` is `short` and `S₂` is `ushort`, `uint`, or `ulong`
+  - `S₁` is `int` and `S₂` is `uint`, or `ulong`
+  - `S₁` is `long` and `S₂` is `ulong`
+- `E` is a delegate creation expression (§11.7.15.6) resulting from a method group conversion and `T₁` is compatible (§19.4) with the single best method from the method group (§10.8)
```
jskeet: Ping for any updates on this? It would be really good to get this PR done and merged this week, so we can get the Word version out next week.

jskeet: Ping for any updates on this? It would be really good to get this PR done and merged this week, so we can get the Word version out next week.

BillWagner: I'm ready to merge, if you are. 

I've pinged Mads offline for a review as well.

jskeet: But this still has "delegate creation expression" where I thought we *just* wanted it to handle method group conversions...

BillWagner: Gotcha. I misunderstood your earlier comment.

Update coming this morning....

BillWagner: Update made.

jskeet (approved):
Great - assuming I've remembered correctly that the final line was all I was nervous about, I'm happy with this :)

BillWagner:
I'll merge this on the analysis that it's better than what we had before. This area may still need some edits.

PR #485, opened 2022-03-10
Reformat section 11 (expressions)

jskeet:
In various places I've used more modern C# (particularly
automatically-implemented properties), but I haven't done so
absolutely everywhere.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -306,17 +306,11 @@ In both of the above cases, a cast expression can be used to explicitly convert 
 
 > *Example*: In the following code
 > ```csharp
-> decimal AddPercent(decimal x, double percent)
-> {
->     return x * (1.0 + percent / 100.0);
-> }
+> decimal AddPercent(decimal x, double percent) => x * (1.0 + percent / 100.0);
 > ```
 > a binding-time error occurs because a `decimal` cannot be multiplied by a `double`. The error is resolved by explicitly converting the second operand to `decimal`, as follows:
 > ```csharp
-> decimal AddPercent(decimal x, double percent)
-> {
->     return x * (decimal)(1.0 + percent / 100.0);
-> }
+> decimal AddPercent(decimal x, double percent) => x * (decimal)(1.0 + percent / 100.0);
 > ```
 > *end example*
 
@@ -587,10 +581,8 @@ The expressions of an argument list are always evaluated in textual order.
 > ```csharp
 > class Test
 > {
->     static void F(int x, int y = -1, int z = -2)
->     {
->         System.Console.WriteLine("x = {0}, y = {1}, z = {2}", x, y, z);
->     }
+>     static void F(int x, int y = -1, int z = -2) =>
+>         System.Console.WriteLine($"x = {x}, y = {y}, z = {z}");
 >
 >     static void Main()
 >     {
@@ -660,10 +652,7 @@ When a generic method is called without specifying type arguments, a ***type inf
 > {
 >     static Random rand = new Random();
 >
->     public static T Choose<T>(T first, T second)
->     {
->         return (rand.Next(2) == 0)? first: second;
->     }
+>     public static T Choose<T>(T first, T second) => rand.Next(2) == 0 ? first : second;
 > }
 > ```
 > it is possible to invoke the `Choose` method without explicitly specifying a type argument:
@@ -1666,17 +1655,17 @@ The preceding rules mean that instance methods take precedence over extension me
 > ```csharp
 > public static class C
 > {
->     public static void F(this int i) { Console.WriteLine("C.F({0})", i); }
->     public static void G(this int i) { Console.WriteLine("C.G({0})", i); }
->     public static void H(this int i) { Console.WriteLine("C.H({0})", i); }
+>     public static void F(this int i) => Console.WriteLine($"C.F({i})");
+>     public static void G(this int i) => Console.WriteLine($"C.G({i})");
+>     public static void H(this int i) => Console.WriteLine($"C.H({i})");
 > }
 >
 > namespace N1
 > {
 >     public static class D
 >     {
->         public static void F(this int i) { Console.WriteLine("D.F({0})", i); }
->         public static void G(this int i) { Console.WriteLine("D.G({0})", i); }
+>         public static void F(this int i) => Console.WriteLine($"D.F({i})");
+>         public static void G(this int i) => Console.WriteLine($"D.G({i})");
 >     }
 > }
 >
@@ -1686,7 +1675,7 @@ The preceding rules mean that instance methods take precedence over extension me
 >
 >     public static class E
 >     {
->         public static void F(this int i) { Console.WriteLine("E.F({0})", i); }
+>         public static void F(this int i) => Console.WriteLine($"E.F({i})");
 >     }
 >
 >     class Test
@@ -4745,7 +4734,7 @@ It is possible for anonymous function delegates to share some captured variables
 >     for (int i = 0; i < 3; i++)
 >     {
 >         int y = 0;
->         result[i] = () => Console.WriteLine("{0} {1}", ++x, ++y);
+>         result[i] = () => Console.WriteLine($"{++x} {++y}");
 >     }
 >     return result;
 > }
@@ -4811,7 +4800,7 @@ class Test
 {
     static void F()
     {
-        D d = () => { Console.WriteLine("test"); };
+        D d = () => Console.WriteLine("test");
     }
 }
 ```
@@ -5611,17 +5600,15 @@ class C<T> : C
 {
     public C<T> Where(Func<T,bool> predicate);
     public C<U> Select<U>(Func<T,U> selector);
-    public C<V> SelectMany<U,V>(Func<T,C<U>> selector,
-    Func<T,U,V> resultSelector);
+    public C<V> SelectMany<U,V>(Func<T,C<U>> selector, Func<T,U,V> resultSelector);
     public C<V> Join<U,K,V>(C<U> inner, Func<T,K> outerKeySelector,
-    Func<U,K> innerKeySelector, Func<T,U,V> resultSelector);
+    	Func<U,K> innerKeySelector, Func<T,U,V> resultSelector);
     public C<V> GroupJoin<U,K,V>(C<U> inner, Func<T,K> outerKeySelector,
-    Func<U,K> innerKeySelector, Func<T,C<U>,V> resultSelector);
+        Func<U,K> innerKeySelector, Func<T,C<U>,V> resultSelector);
     public O<T> OrderBy<K>(Func<T,K> keySelector);
     public O<T> OrderByDescending<K>(Func<T,K> keySelector);
     public C<G<K,T>> GroupBy<K>(Func<T,K> keySelector);
-    public C<G<K,E>> GroupBy<K,E>(Func<T,K> keySelector,
-    Func<T,E> elementSelector);
+    public C<G<K,E>> GroupBy<K,E>(Func<T,K> keySelector, Func<T,E> elementSelector);
 }
 
 class O<T> : C<T>
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -306,14 +306,16 @@ In both of the above cases, a cast expression can be used to explicitly convert 
 
 > *Example*: In the following code
 > ```csharp
-> decimal AddPercent(decimal x, double percent) {
->    return x * (1.0 + percent / 100.0);
+> decimal AddPercent(decimal x, double percent)
+> {
+>     return x * (1.0 + percent / 100.0);
 > }
 > ```
 > a binding-time error occurs because a `decimal` cannot be multiplied by a `double`. The error is resolved by explicitly converting the second operand to `decimal`, as follows:
 > ```csharp
-> decimal AddPercent(decimal x, double percent) {
->    return x * (decimal)(1.0 + percent / 100.0);
+> decimal AddPercent(decimal x, double percent)
+> {
+>     return x * (decimal)(1.0 + percent / 100.0);
 > }
 > ```
 > *end example*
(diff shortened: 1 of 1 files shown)


jskeet:
(Basically, please review any non-whitespace changes more carefully than others.)

Oh, and the LINQ expansions are probably a bit of a personal choice... I definitely find ".Select" together to be simpler than ". at the end of the previous line, then Select at the start of the next" but I may be the odd one out.

jskeet:
Not sure why the validator is failing here... will check what I've done to the grammar (there shouldn't be any changes...)

standard/expressions.md
```diff
@@ -2062,17 +2090,18 @@ When an initializer target refers to an indexer, the arguments to the indexer sh
...
 > public class Rectangle
 > {
->    Point p1 = new Point();
->    Point p2 = new Point();
->    public Point P1 { get { return p1; } }
->    public Point P2 { get { return p2; } }
+>     Point p1 = new Point();
+>     Point p2 = new Point();
+>     public Point P1 { get; } = new Point();
+>     public Point P2 { get; } = new Point();
```
RexJaeschke: As you have made the property auto-implementable, I presume you no longer need the named backing-store fields p1 and p2.

jskeet: Whoops, yes. Will fix.

standard/expressions.md
```diff
@@ -5542,7 +5659,7 @@ assignment
     ;
 
 assignment_operator
-    : '=' | '+=' | '-=' | '*=' | '/=' | '%=' | '&=' | '|=' | '^=' | '<<='
+    : '' | '+=' | '-=' | '*=' | '/=' | '%=' | '&=' | '|=' | '^=' | '<<='
```
RexJaeschke: This doesn't look right; the new first token appears to be empty!

RexJaeschke (approved):

standard/expressions.md
jskeet: I suspect this is what's causing the validation failure, and my guess is my finger slipped at some point. Thanks, will fix.

standard/expressions.md
```diff
@@ -306,14 +306,16 @@ In both of the above cases, a cast expression can be used to explicitly convert
 
 > *Example*: In the following code
 > ```csharp
-> decimal AddPercent(decimal x, double percent) {
->    return x * (1.0 + percent / 100.0);
+> decimal AddPercent(decimal x, double percent)
+> {
+>     return x * (1.0 + percent / 100.0);
 > }
```
BillWagner: What do you think of using expression bodied members more often?

```suggestion
> decimal AddPercent(decimal x, double percent) => x * (1.0 + percent / 100.0);
```

standard/expressions.md
```diff
@@ -306,14 +306,16 @@ In both of the above cases, a cast expression can be used to explicitly convert
...
 > ```
 > a binding-time error occurs because a `decimal` cannot be multiplied by a `double`. The error is resolved by explicitly converting the second operand to `decimal`, as follows:
 > ```csharp
-> decimal AddPercent(decimal x, double percent) {
->    return x * (decimal)(1.0 + percent / 100.0);
+> decimal AddPercent(decimal x, double percent)
+> {
+>     return x * (decimal)(1.0 + percent / 100.0);
 > }
```
BillWagner: Because this is a variation of the previous sample, adding the suggestion here as well. 

```suggestion
> decimal AddPercent(decimal x, double percent) => x * (decimal)(1.0 + percent / 100.0);
```

standard/expressions.md
```diff
@@ -585,14 +587,17 @@ The expressions of an argument list are always evaluated in textual order.
...
->    }
->    static void Main() {
->       int i = 0;
->       F(i++, i++, i++);
->       F(z: i++, x: i++);
->    }
+>     static void F(int x, int y = -1, int z = -2)
+>     {
+>         System.Console.WriteLine("x = {0}, y = {1}, z = {2}", x, y, z);
+>     }
```
BillWagner: Here I would consider both an expression bodied member and an interpolated string:

```suggestion
>     static void F(int x, int y = -1, int z = -2) => System.Console.WriteLine($"x = {x}, y = {y}, z = {z}");
```

standard/expressions.md
```diff
@@ -651,10 +658,12 @@ When a generic method is called without specifying type arguments, a ***type inf
...
->    static Random rand = new Random();
->    public static T Choose<T>(T first, T second) {
->       return (rand.Next(2) == 0)? first: second;
->    }
+>     static Random rand = new Random();
+>
+>     public static T Choose<T>(T first, T second)
+>     {
+>         return (rand.Next(2) == 0)? first: second;
+>     }
```
BillWagner: same.

Even if you reject this change, I would add a space before the `?` and `:`

```suggestion
>     public static T Choose<T>(T first, T second) => (rand.Next(2) == 0) ? first : second;
```

standard/expressions.md
```diff
@@ -1447,20 +1462,25 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
...
->    }
+>     public Color Color;              // Field Color of type Color
+>
+>     void F()
+>     {
+>         Color = Color.Black;         // Refers to Color.Black static member
+>         Color = Color.Complement();  // Invokes Complement() on Color fld
+>     }
+>
+>     static void G()
```
BillWagner, line 1469: Not recommending an expression bodied member her. I agree this looks better with the comment in place.

standard/expressions.md
```diff
@@ -1613,62 +1633,71 @@ The preceding rules mean that instance methods take precedence over extension me
...
 > In the example, `B`'s method takes precedence over the first extension method, and `C`'s method takes precedence over both extension methods.
 > ```csharp
 > public static class C
 > {
->    public static void F(this int i) { Console.WriteLine("C.F({0})", i); }
->    public static void G(this int i) { Console.WriteLine("C.G({0})", i); }
->    public static void H(this int i) { Console.WriteLine("C.H({0})", i); }
+>     public static void F(this int i) { Console.WriteLine("C.F({0})", i); }
+>     public static void G(this int i) { Console.WriteLine("C.G({0})", i); }
+>     public static void H(this int i) { Console.WriteLine("C.H({0})", i); }
```
BillWagner: I would make all three of these expression bodied members:

```suggestion
>     public static void F(this int i) => Console.WriteLine($"C.F({i})");
>     public static void G(this int i) => Console.WriteLine($"C.G({i})");
>     public static void H(this int i) => Console.WriteLine($"C.H({i})");
```

standard/expressions.md
```diff
@@ -1613,62 +1633,71 @@ The preceding rules mean that instance methods take precedence over extension me
...
 > {
->    public static class D
->    {
->       public static void F(this int i) { Console.WriteLine("D.F({0})", i); }
->       public static void G(this int i) { Console.WriteLine("D.G({0})", i); }
->    }
+>     public static class D
+>     {
+>         public static void F(this int i) { Console.WriteLine("D.F({0})", i); }
+>         public static void G(this int i) { Console.WriteLine("D.G({0})", i); }
```
BillWagner: similar comment:

```suggestion
>         public static void F(this int i) => Console.WriteLine($"D.F({i})"); }
>         public static void G(this int i) => Console.WriteLine($"D.G({i})"); }
```

standard/expressions.md
```diff
@@ -1613,62 +1633,71 @@ The preceding rules mean that instance methods take precedence over extension me
...
->          1.F();
->          2.G();
->          3.H();
->       }
->    }
+>     using N1;
+>
+>     public static class E
+>     {
+>         public static void F(this int i) { Console.WriteLine("E.F({0})", i); }
```
BillWagner: I'll stop adding the suggestions, but same comment on expression bodied members and string interpolation.

standard/expressions.md
```diff
@@ -4661,14 +4738,16 @@ It is possible for anonymous function delegates to share some captured variables
...
->    }
->    return result;
+> static D[] F()
+> {
+>     D[] result = new D[3];
+>     int x = 0;
+>     for (int i = 0; i < 3; i++)
+>     {
+>         int y = 0;
+>         result[i] = () => Console.WriteLine("{0} {1}", ++x, ++y);
```
BillWagner: string interpolation instead.

standard/expressions.md
```diff
@@ -5173,17 +5279,16 @@ from * in ( «e» ) . Select ( «x» => new { «x» , «y» = «f» } )
...
 > ```csharp
-> orders.
-> Select(o => new { o, t = o.Details.Sum(d => d.UnitPrice * d.Quantity) }).
-> Where(x => x.t >= 1000).
-> Select(x => new { x.o.OrderID, Total = x.t })
+> orders
+>     .Select(o => new { o, t = o.Details.Sum(d => d.UnitPrice * d.Quantity) })
+>     .Where(x => x.t >= 1000)
+>     .Select(x => new { x.o.OrderID, Total = x.t })
 > ```
```
BillWagner, line 5281: I agree with this style choice.

standard/expressions.md
```diff
@@ -5490,34 +5601,38 @@ A generic type `C<T>` supports the query-expression-pattern if its public member
...
-   Func<U,K> innerKeySelector, Func<T,C<U>,V> resultSelector);
-   public O<T> OrderBy<K>(Func<T,K> keySelector);
-   public O<T> OrderByDescending<K>(Func<T,K> keySelector);
-   public C<G<K,T>> GroupBy<K>(Func<T,K> keySelector);
-   public C<G<K,E>> GroupBy<K,E>(Func<T,K> keySelector,
-   Func<T,E> elementSelector);
+    public C<T> Where(Func<T,bool> predicate);
+    public C<U> Select<U>(Func<T,U> selector);
+    public C<V> SelectMany<U,V>(Func<T,C<U>> selector,
+    Func<T,U,V> resultSelector);
```
BillWagner: I think this line should either be indented, or combined with the previous line.

standard/expressions.md
```diff
@@ -5490,34 +5601,38 @@ A generic type `C<T>` supports the query-expression-pattern if its public member
...
+    public C<V> Join<U,K,V>(C<U> inner, Func<T,K> outerKeySelector,
+    Func<U,K> innerKeySelector, Func<T,U,V> resultSelector);
```
BillWagner: same: indent or combine with previous.

standard/expressions.md
```diff
@@ -5490,34 +5601,38 @@ A generic type `C<T>` supports the query-expression-pattern if its public member
...
+    public C<V> GroupJoin<U,K,V>(C<U> inner, Func<T,K> outerKeySelector,
+    Func<U,K> innerKeySelector, Func<T,C<U>,V> resultSelector);
```
BillWagner: same: indent or combine with previous.

standard/expressions.md
```diff
@@ -5490,34 +5601,38 @@ A generic type `C<T>` supports the query-expression-pattern if its public member
...
+    public O<T> OrderBy<K>(Func<T,K> keySelector);
+    public O<T> OrderByDescending<K>(Func<T,K> keySelector);
+    public C<G<K,T>> GroupBy<K>(Func<T,K> keySelector);
+    public C<G<K,E>> GroupBy<K,E>(Func<T,K> keySelector,
+    Func<T,E> elementSelector);
```
BillWagner: same: indent or combine with previous.

BillWagner (approved):
This looks great @jskeet 

I had a number of suggestions / nits here and there.

Consider them as we evolve our style, and then let's :shipit:

standard/expressions.md
jskeet: Agreed - I think these were more "places I missed" than deliberate choices :)

standard/expressions.md
jskeet: Yup, done.

standard/expressions.md
jskeet: I've broken it onto the next line (otherwise it's really long), but otherwise, yes.

standard/expressions.md
jskeet: Going with the expression-bodied member - but I've removed the brackets around the first operand of the conditional expression, too.

standard/expressions.md
jskeet: Agreed. (I did something similar elsewhere; I must have missed these ones.)

standard/expressions.md
jskeet: Done.

standard/expressions.md
jskeet: Done.

standard/expressions.md
jskeet: Done.

standard/expressions.md
jskeet: Eek, absolutely.

jskeet: (Have unwrapped some lines, and indented others.)

jskeet (commented):
Changes pushed in a new commit.

The only occurrences of `{0}` are now in the section about string interpolation, so hopefully that means we've found all of those. There may well be other places we can add more expression-bodied members - we can pick those off over time.

jskeet:
@BillWagner: If you could just check over the last commit to make sure I haven't done anything stupid, that would be great. If you're happy, we can then squash and merge.

BillWagner:
@jskeet I just looked at the latest commit.  I'll :shipit:

standard/expressions.md
```diff
@@ -1447,20 +1451,25 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
...
+>     void F()
+>     {
+>         Color = Color.Black;         // Refers to Color.Black static member
+>         Color = Color.Complement();  // Invokes Complement() on Color fld
+>     }
+>
+>     static void G()
+>     {
+>         Color c = Color.White;       // Refers to Color.White static member
+>     }
```
KalleOlaviNiemitalo, line 1472: Filed <https://github.com/dotnet/csharpstandard/issues/541> about the loss of `**` delimiters that are still referenced by the paragraph below the code sample.

PR #488, opened 2022-03-10
Reformat sections 12 and 13 (statements and namespaces)

jskeet:
When reformatting statements, there are two slightly unusual aspects:

- Some `if` statement bodies are not braced; this is deliberate as
  either it's to indicate an error which would not be an error in
  a block body, or it's to avoid making assumptions about reachability
  of a block body vs reachability of a statement
- I've unilaterally gone with a "2 space unindent" for labels. We can
  definitely revisit that

### DIFF
--- a/standard/statements.md
+++ b/standard/statements.md
@@ -36,7 +36,8 @@ The *embedded_statement* nonterminal is used for statements that appear within o
 
 > *Example*: The code
 > ```csharp
-> void F(bool b) {
+> void F(bool b)
+> {
 >    if (b)
 >       int i = 44;
 > }
@@ -53,12 +54,13 @@ If a statement can possibly be reached by execution, the statement is said to be
 
 > *Example*: In the following code
 > ```csharp
-> void F() {
->    Console.WriteLine("reachable");
->    goto Label;
->    Console.WriteLine("unreachable");
->    Label:
->    Console.WriteLine("reachable");
+> void F()
+> {
+>     Console.WriteLine("reachable");
+>     goto Label;
+>     Console.WriteLine("unreachable");
+>   Label:
+>     Console.WriteLine("reachable");
 > }
 > ```
 > the second invocation of Console.WriteLine is unreachable because there is no possibility that the statement will be executed. *end example*
@@ -69,16 +71,20 @@ A warning is reported if a statement other than *throw_statement*, *block*, or *
 > 
 > In the example
 > ```csharp
-> void F() {
->    const int i = 1;
->    if (i == 2) Console.WriteLine("unreachable");
+> void F()
+> {
+>     const int i = 1;
+>     if (i == 2)
+>         Console.WriteLine("unreachable");
 > }
 > ```
 > the Boolean expression of the `if` statement is a constant expression because both operands of the `==` operator are constants. As the constant expression is evaluated at compile-time, producing the value `false`, the `Console.WriteLine` invocation is considered unreachable. However, if `i` is changed to be a local variable
 > ```csharp
-> void F() {
->    int i = 1;
->    if (i == 2) Console.WriteLine("reachable");
+> void F()
+> {
+>     int i = 1;
+>     if (i == 2)
+>         Console.WriteLine("reachable");
 > }
 > ```
 > the `Console.WriteLine` invocation is considered reachable, even though, in reality, it will never be executed. *end note*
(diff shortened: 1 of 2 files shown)


standard/statements.md
```diff
@@ -172,18 +180,20 @@ Execution of an empty statement simply transfers control to the end point of the
...
->    ...
->    if (done) goto exit;
->    ...
->    exit: ;
+> void F()
+> {
+>     ...
+>     if (done) goto exit;
+>     ...
+>   exit: ;
```
RexJaeschke: Did you mean to *not* put the goto on its own line inside braces?

standard/statements.md
```diff
@@ -208,10 +218,13 @@ Labels have their own declaration space and do not interfere with other identifi
 
 > *Example*: The example
 > ```csharp
-> int F(int x) {
->     if (x >= 0) goto x;
+> int F(int x)
+> {
+>     if (x >= 0)
+>         goto x;
```
RexJaeschke: Same as previous comment.

standard/statements.md
jskeet: Nope, missed that one.

RexJaeschke (approved):

standard/statements.md
jskeet: Done - and moved the `;` to a line on its own too.

standard/statements.md
jskeet: Done.

standard/statements.md
```diff
@@ -69,16 +71,20 @@ A warning is reported if a statement other than *throw_statement*, *block*, or *
...
 > In the example
 > ```csharp
-> void F() {
->    const int i = 1;
->    if (i == 2) Console.WriteLine("unreachable");
+> void F()
+> {
+>     const int i = 1;
+>     if (i == 2)
+>         Console.WriteLine("unreachable");
```
BillWagner, line 78: I think there should be one more space here.

standard/statements.md
```diff
@@ -69,16 +71,20 @@ A warning is reported if a statement other than *throw_statement*, *block*, or *
...
 > the Boolean expression of the `if` statement is a constant expression because both operands of the `==` operator are constants. As the constant expression is evaluated at compile-time, producing the value `false`, the `Console.WriteLine` invocation is considered unreachable. However, if `i` is changed to be a local variable
 > ```csharp
-> void F() {
->    int i = 1;
->    if (i == 2) Console.WriteLine("reachable");
+> void F()
+> {
+>     int i = 1;
+>     if (i == 2)
+>         Console.WriteLine("reachable");
```
BillWagner, line 87: one additional space?

standard/statements.md
```diff
@@ -87,9 +93,11 @@ The *block* of a function member or an anonymous function is always considered r
...
 > *Example*: In the following code
 > ```csharp
-> void F(int x) {
->    Console.WriteLine("start");
->    if (x < 0) Console.WriteLine("negative");
+> void F(int x)
+> {
+>     Console.WriteLine("start");
+>     if (x < 0)
+>         Console.WriteLine("negative");
```
BillWagner, line 100: space?

standard/statements.md
```diff
@@ -843,13 +892,17 @@ The order in which `foreach` traverses the elements of an array, is as follows:
...
+>     static void Main()
+>     {
+>         double[,] values =
+>         {
 >             {1.2, 2.3, 3.4, 4.5},
 >             {5.6, 6.7, 7.8, 8.9}
 >         };
 >         foreach (double elementValue in values)
+>         {
 >             Console.Write("{0} ", elementValue);
```
BillWagner: Consider string interpolation:

```suggestion
>             Console.Write($"{elementValue} ");
```

standard/statements.md
```diff
@@ -990,20 +1052,29 @@ The target of a `goto` *identifier* statement is the labeled statement with the
...
+>                     {
 >                         goto done;
+>                     }
+>                 }
+>             }
 >             Console.WriteLine("{0} not found", str);
 >             continue;
->         done:
+>           done:
 >             Console.WriteLine("Found {0} at [{1}][{2}]", str, row, colm);
```
BillWagner: consider string interpolation:

```suggestion
>             Console.WriteLine($"Found {str} at [{row}][{colm}]");
```

BillWagner (approved):
This also looks great @jskeet 

I had a couple small suggestions, and then let's :shipit:

standard/statements.md
jskeet: If you mean before Console, I don't think so - it's indented by 4 spaces compared with the `if`, so 8 spaces compared with the `void F()` - ditto the examples below.

standard/statements.md
jskeet: Done.

standard/statements.md
jskeet: Done - and there are no more occurrences of `{0}` :)

jskeet (commented):
Fixed the string interpolation and repushed a new single commit. I'll merge now, as I'm *pretty* sure that the extra spaces aren't required. Happy to create a new PR if I'm wrong though :)

standard/statements.md
BillWagner: Could have been my eyes. You're correct.

PR #489, opened 2022-03-10
Manual smart quotes

Nigel-Ecma:
The file standard-library.md contains a large amount of HTML text which is not processed
by the smarten tool. This text originated in the CLR Standard and mostly has typographic
quotes already, but there were a few omissions and inconsistencies. This fixes those.

### DIFF
--- a/standard/standard-library.md
+++ b/standard/standard-library.md
@@ -529,7 +529,7 @@ property.</p></td>
 <p><code>f</code></p></td>
 <td><p><strong>Fixed-Point Format:</strong> Used for strings in the following
 form:</p>
-<p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;"[-]<em>m.dd...d</em>"</p>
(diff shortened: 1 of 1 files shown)


RexJaeschke (approved):

standard/standard-library.md
```diff
@@ -529,7 +529,7 @@ property.</p></td>
 <p><code>f</code></p></td>
 <td><p><strong>Fixed-Point Format:</strong> Used for strings in the following
 form:</p>
-<p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;"[-]<em>m.dd...d</em>"</p>
+<p>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[-]<em>m.dd...d</em></p>
```
BillWagner, line 532: Here, you removed the quotes but didn't add smart quotes. Was that intentional?

BillWagner (approved):
This LGTM @Nigel-Ecma 

I did have one question.

Nigel-Ecma:
@BillWagner: Yes those two were intentional; when I went to fix them I noticed an inconsistency in the original as the ones just above are not quoted so I just deleted them.

PR #492, opened 2022-03-14
More string interpolation

jskeet:
The only occurrences of `{0}` are now in expressions.md, within the
section about string interpolation.


BillWagner (approved):
This is great @jskeet 

You can :shipit: when ready.

PR #493, opened 2022-03-14
Unescape pipe escaping for table cells

### LINKED ISSUE #442: Word converter: backslash not being unescaped in table heading
In the table for 12.12.5 ("Nullable Boolean `&` and `|` operators") the `|` in the heading isn't unescaped.

(This is one of the very rare tables that *isn't* hand-crafted...)

jskeet:
Fixes #442.

Note that there's no test here. While we have a sample test for something simple, we don't really have test infrastructure for "here's some Markdown, and here's the expected converted Word document". The Word conversion code always creates an actual file, as far as I can see, which also hinders testing.

Some options:

- If the Word document is byte-for-byte reproducible, we could have "sample JSON, expected Word document" pairs
- We could easily write a unit test for just the code that's changed in this case, but it doesn't help in the longer term, as most changes really need the wider context

I think the first option is the way to go, but it will require a bit of refactoring to allow easy generation of a single-Markdown-file spec (possibly without TOC). I'll try that as a separate commit, still in this PR for now just for simplicity.
(I don't know what we do if the precise content changes each time...)


jskeet:
Second commit contains a rudimentary test framework. While crude, I think it's better than nothing - we can iterate over time.

BillWagner (approved):
The actual fix is great @jskeet 

Over time, I'd like to have our test code be more concise, but I agree that's a task for another day.

jskeet:
> Over time, I'd like to have our test code be more concise, but I agree that's a task for another day.

My hope is that the new infrastructure means we just need a single line of code and a pair of files for any change that is basically "Markdown x leads to XML y".

I suspect we could get the size of the XML file down if we remove all the bits that are just about the heading (which is required for parsing the Markdown) - I suggest we wait until we've got a few more files so we can see the commonality.

PR #496, opened 2022-03-15
Run smarten on PR merge

BillWagner:
Run the smart quoter when a PR is merged into the version release branch.

The strategy I took is to add the smarten app to the current workflow that runs when a PR is merged. That guarantees order, and ensures we have at most one automation PR to merge when a human-authored PR is merged.

Because the smarten executable is a .NET Framework based (Windows only) application, that rippled into a few other changes:

- The merge workflow now runs on Windows.
- Because the workflow runs on a Windows host, the default shell needs to specify bash.

I made the choice to run the smarten tool first of the automated transforms. That seems to be the best  choice.


BillWagner:
@Nigel-Ecma One interesting result from testing this locally. Both classes.md and expressions.md wouldn't round trip correctly, so a second file was created instead of overwriting the original. When I looked at the diffs, none of the quotes changed in either file. Is that consistent with your experience? If not, any ideas on how to fix?

tools/run-smarten.sh
```diff
@@ -0,0 +1,57 @@
...
+    "portability-issues"
+    "standard-library"
+    "documentation-comments"
+    "bibliography"
+    )
+
+# unpack the package to ./smarten
+tar -xvf ../.github/workflows/dependencies/EcmaTC49.Smarten.tar
+
+# On some files, smarten can't "round trip", so creates
```
jskeet: What exactly is meant by "can't round trip" here?

BillWagner: I'd have to dig deeper into the smarten source. On (almost) every file, the smarten_CL writes the same destination file as the source (as we want). One these two, it fails to do that, and writes a different file destination.

Those are the largest files in the standard. I'm suspecting it has something to do with the file size, but I haven't figured that out yet.

BillWagner: Well, after looking at the source, I don't know why those files are generated. These larger files are processed correctly, but an internal diagnostic fails, so the extra file is generated.

I think it can just be removed. I'd like to verify that with @Nigel-Ecma before we merge this.

Nigel-Ecma:
@BillWagner: No it’s not consistent :-( 

The uploaded smarten would roundtrip the then “standard” (so “standard” -> “pass1” -> “pass2” – first two differ, second two are equal).

Of course the contents of “standard” has had a few updates since then, I’ll go dig…

Nigel-Ecma:
Helpful information, narrows it down.

The smarten tool does a pre-check of round tripping its input using markdig only, no smartening, and if the output is not exactly the same as the input issues a warning and the extra file so someone (me) can go dig…

BillWagner:
> Helpful information, narrows it down. The smarten tool does a pre-check of round tripping its input using markdig only, no smartening, and if the output is not exactly the same as the input issues a warning and the extra file so someone (me) can go dig…

If it helps, the only diffs between the original and the round trip pass was additional whitespace.

BillWagner:
Thanks to @Nigel-Ecma 's work in #498, I've removed the extra script logic to move files if the round trip fails.

The script will leave the \*_(RoundTrip).md files in place in the PR if any of the standard source doesn't round-trip correctly. That should help us find any recurrence of issues.

I think this is ready to merge.

jskeet (approved):

PR #503, opened 2022-03-16
Fix tests by performing conversion in constructor

### LINKED ISSUE #502: "Long lines" test is failing
Not sure how we missed this - I suspect it's due to failing to rebase properly. Investigating.

(The tests which shouldn't generate warnings are fine; it's the test that *should* generate a warning that's failing.)

jskeet:
We really don't want to perform a new conversion every time
Paragraphs() is called, causing more warnings, more changes to the
context etc. Admittedly doing work in the constructor isn't great
either, but I'm okay with that for now.

Fixes #502


BillWagner (approved):
This LGTM @jskeet 

You can :shipit: when you're ready.

PR #505, opened 2022-03-17
Word converter: Rewrite list blocks to handle embedded code blocks

jskeet:
First commit is just #503; we can either merge that PR first and the first commit here will vanish, or merge this and abandon that PR.


BillWagner (approved):
This also looks good @jskeet 

You can :shipit: when ready.

jskeet:
Merging without waiting for the status checks, as GitHub actions are slow/flaky right now - I'll happily fix up any problems later.

PR #508, opened 2022-03-17
Fix notes

### LINKED ISSUE #507: Notes and examples are unlabeled
Looks like this is a regression caused by #495 somehow - although currently I'm failing to understand why.

I'm looking into it urgently.

jskeet:
Fixes #507.


jskeet:
(Rebased so it's now just one commit.)

BillWagner (approved):
This also LGTM @jskeet 

You can :shipit: when ready.

PR #509, opened 2022-03-17
Fix (and detect) broken lists

jskeet:
The first two commits are from #503 and #508.


BillWagner (approved):
This LGTM @jskeet 

You can :shipit: when ready.

Issue #441, opened 2022-01-20
Word converter: paragraph after table in list not handled properly


jskeet:
In 12.9.2 (multiplication operator) there's a table that's expected to be in a list, followed by a paragraph. The paragraph should be part of the list - followed by the next list item beginning "Decimal multiplication"

The Markdown isn't quite right here anyway, but after trying a few things I believe we'll still need more work to get it to format appropriately in Word. (Or we could rearrange the Markdown...)

jskeet:
We need to do *something* here - that may be rearranging the Markdown, or it may be fixing the tooling, but it needs to be sorted out for C#6.

BillWagner:
Adding myself as an assignee to remind myself to take a look at the markdown and see what options we have.

jskeet:
Not sure that this really needs to be blocking. It's not ideal, but it's not horrific as-is.

jskeet:
This is fixable without too much hassle. There's a prototype in #512. I'll fix it up when other PRs have been merged.

PR #512, opened 2022-03-17
Fix tables in lists

jskeet:
This is now ready for review.


jskeet:
Fixes #441.

jskeet:
Note: we should validate what happens to 11.9.6 when both this and #515 are merged (that's got an example in a triply-nested list). I hope it'll be okay, but it's worth checking.

BillWagner (approved):
This LGTM @jskeet 

Let's :shipit:

jskeet:
Thanks - will rebase #515 on this, as it's very likely to have merge conflicts.

PR #515, opened 2022-03-18
Indent lists in notes

### LINKED ISSUE #435: Word converter: List items in notes/examples are badly formatted
The end of 9.3.5 has a particularly thorny note, containing a list which itself contains an example. We basically don't format it properly at all.
We may be able to simplify that note a little, but we should definitely do a *better* job of supporting lists in notes.

jskeet:
Fixes #435 (and extract some constants which were duplicated in many places)

I've checked that this looks appropriate in 6.4.3, 8.3.5 and 17.6.2, which were all mentioned in that issue. Given that those ones are fine, I don't expect to see any problems elsewhere.

### DIFF
--- a/tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
+++ b/tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
@@ -23,6 +23,10 @@ namespace MarkdownConverter.Converter
         /// </summary>
         public const int MaximumCodeLineLength = 95;
 
+        private const int InitialIndentation = 540;
+        private const int ListLevelIndentation = 360;
+        private const int TableIndentation = 360;
+
         private static readonly Dictionary<char, char> SubscriptUnicodeToAscii = new Dictionary<char, char>
         {
             { '\u1d62', 'i' },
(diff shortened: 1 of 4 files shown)


jskeet:
Lists in notes now work... notes in lists are slightly trickier. We lose the extra indentation, but that's not so terrible.

jskeet:
We *also* lose the extra indentation for code blocks in notes. We could probably fix that relatively easily if we wanted to, but it would cramp the line lengths further.

jskeet:
(Rebase complete.)

tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
```diff
@@ -138,15 +147,39 @@ IEnumerable<OpenXmlCompositeElement> Paragraph2Paragraphs(MarkdownParagraph md)
...
                         yield return paragraph;
                     }
                     else if (element is Table table)
                     {
                         if (table.ElementAt(0) is TableProperties tableProperties)
                         {
                             tableProperties.TableIndentation ??= new TableIndentation();
-                            tableProperties.TableIndentation.Width = 540;
+                            // TODO: This will be incorrect if we ever have a table in a list in a note.
+                            // Let's just try not to do that.
```
BillWagner, line 181: LOLZ

BillWagner (approved):
This also LGTM @jskeet

PR #516, opened 2022-03-18
Replace "statement block" with "block" everywhere

### LINKED ISSUE #506: Replace "statement block" with "block"
In a few places, we refer to "statement block" when it would be simpler and clearer to just use "block".

jskeet:
Section 12.3 (blocks) just refers to blocks, e.g.:

> A *block* permits multiple statements to be written in contexts where a single statement is allowed.

So there's no need to use "statement block" anywhere else.

Fixes #506.

(Would like confirmation from Mads that I'm not missing something here...)


BillWagner (approved):
This LGTM @jskeet 

I agree on checking with @MadsTorgersen to be sure.

jskeet:
(Ideally we'd still get this into C# 6, in the revised submission, but it's not a huge deal if it slips into C# 7.)

jskeet:
Jon to rebase and then merge; Rex will apply to the Word doc.

jskeet:
@RexJaeschke: This is now merged, so please apply the changes in the Word draft.

RexJaeschke:
Done! These edits delete "statement" from all 12 occurrences of "statement block" I found in the Word spec.

PR #517, opened 2022-03-18
Note that an enum can be a generic type

### LINKED ISSUE #482: How do we describe enums in generic types?
This was raised in #480.

Consider:

```csharp
public class C<T>
{
    public enum E { Xyz }
}
```

That's entirely valid, and `C<string>.E` and `C<int>.E` are different types... even though an enum declaration itself can't have type parameters.

Is it reasonable to consider the declaration of `C<>.E` to be a "generic enum declaration"?

jskeet:
Also link to the constructed types clause for all kinds of type,
which clarifies why we classify nested types that don't themselves
declare type parameters as generic types.

Fixes #482.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -21,7 +21,7 @@ A *class_declaration* consists of an optional set of *attributes* ([§21](attrib
 
 A class declaration shall not supply a *type_parameter_constraints_clause*s unless it also supplies a *type_parameter_list*.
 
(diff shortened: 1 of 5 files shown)


Nigel-Ecma (approved):
LGTM

BillWagner (approved):
This also LGTM @jskeet 

Let's :shipit:

PR #518, opened 2022-03-18
Fix formatting issues around code

### LINKED ISSUE #511: Detect and fix long lines in examples
Section 11.9.6 (subtraction operator) has an example in a level 3 list item, causing wrapping.

We could just fix these by inspection (there probably aren't many) or change our long line detection to be indentation-aware (which will be harder, but more robust).

jskeet:
This was originally intended to find examples with over-long lines, but found a bunch of minor issues along the way.

Fixes #511 as far as I think we want to - I don't think it's worth going to extraordinary lengths to detect the issues.

(This was checked with the Word converter in the "latest possible" state, i.e. after applying all the other open PRs for code.)

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -641,7 +641,7 @@ The use of conditional methods in an inheritance chain can be confusing. Calls m
 >     public override void M()
 >     {
 >         Console.WriteLine("Class2.M executed");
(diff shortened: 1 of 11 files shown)


BillWagner (approved):
This also LGTM @jskeet 

Let's :shipit:

PR #521, opened 2022-03-18
Add a manual workflow for smarten,

BillWagner:
Smarten doesn't run on every PR now. It will only run on demand.

I also updated the package based on Nigel's recent updates. This didn't fix the problem Jon identified with indenting tables.

The update-on-merge workflow doesn't run smarten now.


jskeet (approved):
LGTM, thanks.

PR #522, opened 2022-03-18
configure markdown lint

BillWagner:
First configuration of markdown lint. We can determine which warnings we want to enable or disable from a first test run.

We can adjust what's reported by adding items from the [markdown lint rules](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md).

### DIFF
--- a/.github/workflows/markdownlint-problem-matcher.json
+++ b/.github/workflows/markdownlint-problem-matcher.json
@@ -1,0 +1,17 @@
+{
+    "problemMatcher": [
+        {
+            "owner": "markdownlint",
+            "pattern": [
+                {
+                    "regexp": "^([^:]*):(\\d+):?(\\d+)?\\s([\\w-\\/]*)\\s(.*)$",
+                    "file": 1,
+                    "line": 2,
+                    "column": 3,
+                    "code": 4,
+                    "message": 5
+                }
+            ]
+        }
+    ]
+}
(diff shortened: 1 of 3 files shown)


BillWagner:
@jskeet Here's a PR that turns on markdown lint in our repo.

Over time, it's probably a good idea to have it on with most warnings enable. I already turned off 2 by design: 

- Warning on inline HTML
- Warning on long lines

We use both of those intentionally. 

Take a look at the warnings. See which ones are most helpful to get our markdown source to work well with the word converter.

Then, we can figure out a plan.

jskeet (dismissed):
All the current warnings look good to me - thanks!

BillWagner:
@jskeet 

I'm think of this as a plan:

- I'll update this PR to disable every current warning (due to the noise).
- Once this is merged, I'll make PRs to re-enable each warning (or a set) and resolve each of those warnings.

jskeet:
Works for me :)

BillWagner:
@jskeet This is ready for final review. I want you to take a look because of some of rules changes:

> A couple note on the 4th commit:
>
> MD034 is for "raw URL link". The standard contains the link to the ECMA site, which is intentional. If it helps the word converter, that can be formatted using markdown's `[]()` syntax.
>
> MD30 and MD31 will be the two biggest jobs.
>
> Also, some of the warnings are generated from the markdown source in the test project. I think the right move will be to update those test sources once the associated rules are enabled.

jskeet:
I'll look at this on Monday if that's okay - my weekend is pretty busy.

jskeet (approved):
Yup, this feels like a good approach to me.

PR #523, opened 2022-03-18
Fix missing space in general-description.md

Nigel-Ecma:
Add missing space in general-description.md which also prevented a quote being smartened (found reviewing #504)


BillWagner (approved):
This LGTM @Nigel-Ecma 

I'll :shipit: now.

PR #525, opened 2022-03-21
Remove last vestiges of "antlr" language for code blocks

### LINKED ISSUE #432: Word converter: "antlr" code path is never used
In MarkdownSourceConverter, we initially handle "antlr" and "ANTLR" the same way... but then *only* use "antlr" to apply the "Grammar" style. When trying to make this case-insensitive, the converter falls over entirely.

I believe we use "ANTLR" exclusively now - so it might be worth a) deciding if we want the style; b) being case-sensitive about requiring "ANTLR".

jskeet:
All our grammar blocks use ANTLR instead.

Fixes #432.


BillWagner (approved):
This LGTM @jskeet 

I again appreciate seeing us remove more code. Let's :shipit:

PR #528, opened 2022-03-23
More markdown lint fixes

BillWagner:
- [MD022](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md022---headings-should-be-surrounded-by-blank-lines) - Headings should be surrounded by blank lines.
- [MD027](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md027---multiple-spaces-after-blockquote-symbol) - Multiple spaces not allowed after a blockquote symbol.
- [MD028](https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md028---blank-line-inside-blockquote) - blank line inside a blockquote.

Most of the changes are straightforward. The first commits did the bulk of the work, followed by some warnings and stray characters.

### DIFF
--- a/standard/arrays.md
+++ b/standard/arrays.md
@@ -168,20 +168,14 @@ An array initializer consists of a sequence of variable initializers, enclosed b
 
 The context in which an array initializer is used determines the type of the array being initialized. In an array creation expression, the array type immediately precedes the initializer, or is inferred from the expressions in the array initializer. In a field or variable declaration, the array type is the type of the field or variable being declared. When an array initializer is used in a field or variable declaration,
 
-> *Example*:
-> ```csharp
-> int[] a = {0, 2, 4, 6, 8};
-> ```
-> *end example*
+```csharp
+int[] a = {0, 2, 4, 6, 8};
+```
+it is simply shorthand for an equivalent array creation expression:
 
-It is simply shorthand for an equivalent array creation expression:
-
-> *Example*:
-> ```csharp
-> int[] a = new int[] {0, 2, 4, 6, 8};
-> ```
-> *end example*
-
+```csharp
+int[] a = new int[] {0, 2, 4, 6, 8};
+```
 For a single-dimensional array, the array initializer shall consist of a sequence of expressions, each having an implicit conversion to the element type of the array ([§10.2](conversions.md#102-implicit-conversions)). The expressions initialize array elements in increasing order, starting with the element at index zero. The number of expressions in the array initializer determines the length of the array instance being created.
 
 > *Example*: The array initializer above creates an `int[]` instance of length 5 and then initializes the instance with the following values:
--- a/standard/arrays.md
+++ b/standard/arrays.md
@@ -174,7 +174,7 @@ The context in which an array initializer is used determines the type of the arr
 > ```
 > *end example*
 
-> it is simply shorthand for an equivalent array creation expression:
+It is simply shorthand for an equivalent array creation expression:
 
 > *Example*:
 > ```csharp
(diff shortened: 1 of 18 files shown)


standard/arrays.md
```diff
@@ -174,7 +174,7 @@ The context in which an array initializer is used determines the type of the arr
 > ```
 > *end example*
 
-> it is simply shorthand for an equivalent array creation expression:
+It is simply shorthand for an equivalent array creation expression:
```
BillWagner: @RexJaeschke  

Can you look at this edit and suggest any needed changes?

It was a single sentence (fragment) in a block quote, but not included in either of the two examples. It doesn't seem to make sense that this is a normative statement, but it's currently not in a *note* or *example*.

What's your suggestion?

standard/arrays.md
```diff
@@ -232,5 +232,7 @@ When an array creation expression includes both explicit dimension lengths and a
 > int[] z = new int[3] {0, 1, 2, 3}; // Error, length/initializer mismatch
 > ```
 > Here, the initializer for `y` results in a compile-time error because the dimension length expression is not a constant, and the initializer for `z` results in a compile-time error because the length and the number of elements in the initializer do not agree. *end example*
+<!-- markdownlint-disable MD028 -->
 
+<!-- markdownlint-enable MD028 -->
```
BillWagner, line 231: @jskeet I added a number of these comments to disable a warning.

There are numerous locations in the standard where we have repeated *note* and *warning* blocks. But we want to enable this warning to prevent blank (non-block-quote) lines that split an example or note.

The word converter passed, but I wanted you to see this.

jskeet: Thanks for the shout-out... I'll double-check the Word converter after merging; if we need to add code to explicitly ignore these and avoid creating warnings, that should be easy to do.

jskeet: (If we can work out a different way of separating consecutive notes/warnings, that would be even better, of course. But this is a good start.)

standard/expressions.md
```diff
@@ -130,26 +130,28 @@ When an expression contains multiple operators, the ***precedence*** of the oper
...
->   [§11.12](expressions.md#1112-logical-operators)             | Logical AND                      | `&`
->   [§11.12](expressions.md#1112-logical-operators)             | Logical XOR                      | `^`
->   [§11.12](expressions.md#1112-logical-operators)             | Logical OR                       | `\|`
->   [§11.13](expressions.md#1113-conditional-logical-operators)             | Conditional AND                  | `&&`
->   [§11.13](expressions.md#1113-conditional-logical-operators)             | Conditional OR                   | `\|\|`
->   [§11.14](expressions.md#1114-the-null-coalescing-operator)             | Null coalescing                  | `??`
->   [§11.15](expressions.md#1115-conditional-operator)             | Conditional                      | `?:`
->   [§11.18](expressions.md#1118-assignment-operators) and [§11.16](expressions.md#1116-anonymous-function-expressions)  | Assignment and lambda expression | `=` `*=` `/=` `%=` `+=` `-=` `<<=` `>>=` `&=` `^=` `\|=` `=>`
+> |  **Subclause**      | **Category**                     | **Operators**
```
jskeet, line 138: We'll want to check this with the Word converter afterwards to make sure we don't have a blank first column. But that's fine.

jskeet (approved):
LGTM, thanks.

standard/arrays.md
RexJaeschke: Bill, I believe all this should be normative, and I just made changes to do that.

RexJaeschke (approved):

standard/expressions.md
BillWagner: I ran the converter locally. The table looks the same. I'll merge this one.

standard/arrays.md
BillWagner: > (If we can work out a different way of separating consecutive notes/warnings, that would be even better, of course. But this is a good start.)

I looked at the markdown lint docs, and I didn't find a better alternative. (Doesn't mean it doesn't exist, just that I didn't find it.)

PR #534, opened 2022-03-28
Format endnote and endexample markers

BillWagner:
The *end note* and *end example* marker for a note should be at the end of a single paragraph note. If a note spans multiple paragraphs (including notes that contain examples or lists), the *end note* or *end example* marker should be its own paragraph.

I searched for both *end note* (first commit) and *end example* (second commit) and made the updates to those that didn't follow the format we agreed on.

### DIFF
--- a/tools/MarkdownConverter/Spec/MarkdownSpec.cs
+++ b/tools/MarkdownConverter/Spec/MarkdownSpec.cs
@@ -247,6 +247,9 @@ namespace MarkdownConverter.Spec
         /// them as separate spans. We remove this later on when converting Markdown to Word. Note that this
         /// only works at the top level at the moment.
         /// </summary>
+        /// <remarks>
+        /// This may no longer be necessary now that we're using Markdown Lint to avoid this condition.
+        /// </remarks>
         private static string SeparateNotesAndExamples(string text) => text
             .Replace("*end note*\r\n\r\n> *", $"*end note*\r\n\r\n{NoteAndExampleFakeSeparator}\r\n\r\n> *")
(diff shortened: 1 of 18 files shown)


jskeet (commented):
Any idea which file is causing the Word converter to fail? I'm happy to pull your PR and run it locally if that would help.

BillWagner:
@jskeet  I'm looking into it this morning. I've already got this PR local, so I'll track it down.

BillWagner:
Found it! (actually two issues).

The good news is the word converter worked correctly, and helped find two additional formatting issues.

tools/MarkdownConverter/Spec/MarkdownSpec.cs
```diff
@@ -247,6 +247,9 @@ private static string RemoveBlockComments(string text, string file)
         /// them as separate spans. We remove this later on when converting Markdown to Word. Note that this
         /// only works at the top level at the moment.
         /// </summary>
+        /// <remarks>
+        /// This may no longer be necessary now that we're using Markdown Lint to avoid this condition.
```
jskeet, line 251: I think it probably *is* still necessary, when they're in separate paragraphs. For example:

```markdown
> *Example*:
>
> Some text in multiple paragraphs.
>
> *end example*

> *Note*:
>
> Some text in multiple paragraphs.
>
> *end note*
```

I *think* the two will be collapsed so we end up with an "end example Note:" paragraph without this. I can add a test for it once this PR is in, perhaps.

jskeet (approved):
Excellent... I think we're nearly ready to give Rex a document...

tools/MarkdownConverter/Spec/MarkdownSpec.cs
BillWagner: I added a comment in the code pointing to this discussion for reference.

PR #538, opened 2022-03-31
Remove markdown lint from the end of notes/examples

jskeet:


BillWagner:
There is one failing test now. Is that a false positive?

jskeet:
Ooh, no... that's odd. I'm out of time for today, but I'll look first thing tomorrow.

jskeet:
Ah - it's possible that this is a "difference between \r\n and \n" issue. That's just a guess though; will look into it properly tomorrow.

jskeet:
@BillWagner: Fixed now - it's an annoyance of the Markdown parser that even if we normalize the input, it then uses `Environment.NewLine` itself. I've amended the commit, and the test is now passing.

BillWagner (commented):
Thanks @jskeet 

This LGTM. Let's :shipit:

BillWagner (approved):

PR #542, opened 2022-04-06
Various Pre-v6 tweaks

RexJaeschke:
This mostly fixes putting *end example* at the end of a single-para example, or putting it on its own line with a multi-para example. Plus other formatting fixes found when proofing V6 Word.

I've made all these changes in the Word version.


jskeet (approved):
Thanks Rex!

BillWagner (approved):
This LGTM as well.

Let's :shipit:

PR #551, opened 2022-04-13
Fix comparison operator appearing outside code

### LINKED ISSUE #544: Potential problem with the Word generator
In 11. Expressions, in 11.11.3 Floating-point comparison operators, we have the following md:

> *Example*: If either of `x` and `y` is NaN, then `x` < `y` is `false`, but `!(x >= y)` is `true`.
> 
> *end example*

which is well formed. However, the Word generator turned that into the following: 

*Example*: If either of `x` and `y` is NaN, then `x` < \`y\` is \`false\`, but \`!(x >= y)`is`true\`.
*end example*

The problem seems to start at the "<" character.
 
BTW, I have since changed the md by pushing the *end example* up to the the end of the single-paragraph example instead if being in a new paragraph.

jskeet:
This confuses the Word converter, but is undesirable and
inconsistent anyway, so this change just puts it into the inline code.

Fixes #544


RexJaeschke:
Great minds think alike, Jon! That's exactly the manual fix I made to the Word spec.

RexJaeschke (approved):

PR #553, opened 2022-04-14
Change "better conversion target" to not require an expression

### LINKED ISSUE #549: 11.6.4.6 Better conversion target
This is a nit introduced in spec for language version 6.  In that version we have the definition

> Given an expression `E` and two types `T1` and `T2`, `T1` is a _**better conversion target**_ than `T2` for `E` if one of the following holds...

Where this concept is referenced elsewhere in the spec, we should use the modifier "for" to identify the type "E" which we are checking if it is a better conversion *for*.

However, that modifier is missing in the following two places:

- 11.6.4.4 Better conversion from expression
- 11.6.4.6 Better conversion target (yes, within the definition itself!)

I believe both are easily fixed.

### LINKED ISSUE #550: 11.6.4.4 Better conversion from expression
This is a minor nit introduced in the C# 6 version of the specification.

`11.6.4.4 Better conversion from expression` says

> Given an [implicit](https://github.com/dotnet/csharpstandard/issues/new#_Trm00093) [conversion](https://github.com/dotnet/csharpstandard/issues/new#_Trm00092) C1 that converts from an expression `E` to a type `T1`, and an [implicit](https://github.com/dotnet/csharpstandard/issues/new#_Trm00093) [conversion](https://github.com/dotnet/csharpstandard/issues/new#_Trm00092) `C2` that converts from an expression `E` to a type `T2`, `C1` is a better [conversion](https://github.com/dotnet/csharpstandard/issues/new#_Trm00092) than C2 if one of the following holds:

There is one issue with this:
- The phrase being defined is not in bold and italics.

jskeet:
This effectively moves the method group aspect up from 11.6.4.6 to
11.6.4.4.

Notes:

I've added "for conversion C1" to the "single best method from the
method group" part as the single best method may be different for C1
and C2 - there's no single best method from a method group without
a specific method group conversion being considered

The rule doesn't refer to T2/C2, but I don't think it has to...
because it's valid for both conversions to be better than the other.
That's already the case in better conversion target, as types T1 and
T2 could both have implicit conversions to each other.

I think this means we could also remove "and E does not exactly
match T2" in the first rule of 11.6.4.4, but I haven't gone that far
in this PR.

Fixes #549
Fixes #550

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1004,7 +1004,7 @@ Given an implicit conversion `C₁` that converts from an expression `E` to a ty
 
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` ([§11.6.4.5](expressions.md#11645-exactly-matching-expression))
 - `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§11.6.4.6](expressions.md#11646-better-conversion-target))
-- `E` is a method group ([§11.2](expressions.md#112-expression-classifications)), and `T₁` is compatible ([§19.4](delegates.md#194-delegate-compatibility)) with the single best method from the method group for conversion `C₁`
+- `E` is a method group ([§11.2](expressions.md#112-expression-classifications)), `T₁` is compatible ([§19.4](delegates.md#194-delegate-compatibility)) with the single best method from the method group for conversion `C₁`, and `T₂` is not compatible with the single best method from the method group for conversion `C₂`
 
 #### 11.6.4.5 Exactly matching expression
 
@@ -1021,7 +1021,7 @@ Given an expression `E` and a type `T`, `E` ***exactly matches*** `T` if one of 
 
 Given two types `T₁` and `T₂`, `T₁` is a ***better conversion target*** than `T₂` if one of the following holds:
 
-- An implicit conversion from `T₁` to `T₂` exists
+- An implicit conversion from `T₁` to `T₂` exists and no implicit conversion from `T₂` to `T₁` exists
 - `T₁` is `Task<S₁>`, `T₂` is `Task<S₂>`, and `S₁` is a better conversion target than `S₂`
 - `T₁` is `S₁` or `S₁?` where `S₁` is a signed integral type, and `T₂` is `S₂` or `S₂?` where `S₂` is an unsigned integral type. Specifically:
   - `S₁` is `sbyte` and `S₂` is `byte`, `ushort`, `uint`, or `ulong`
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1000,10 +1000,11 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
 
 #### 11.6.4.4 Better conversion from expression
 
-Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a better conversion than `C₂` if one of the following holds:
+Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if one of the following holds:
 
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` ([§11.6.4.5](expressions.md#11645-exactly-matching-expression))
 - `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§11.6.4.6](expressions.md#11646-better-conversion-target))
+- `E` is a method group ([§11.2](expressions.md#112-expression-classifications)), and `T₁` is compatible ([§19.4](delegates.md#194-delegate-compatibility)) with the single best method from the method group for conversion `C₁`
 
 #### 11.6.4.5 Exactly matching expression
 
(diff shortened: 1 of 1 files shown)


jskeet:
It's still not *entirely* clear to me whether this is sufficient though. Consider:

```csharp
interface I1 { }
interface I2 { }
class C : I1, I2 { }

class Program
{
    static void Main()
    {
        M(X);
    }

    static void M(Func<C, int> func) { }
    static void M(Func<string, string> func) { }

    static int X(I1 x) => 0;
    static string X(I2 x) => "";
}
```

This compiles in the C# 10 compiler, even though the method group conversion for `X` to `Func<C, int>` doesn't have a "single best method" in terms of simple overload resolution: `X(new C());` fails to compile, for example. I'm hoping that this is a matter of a further improvement to 10.8 post-C#-6 which means that the return type of the method is taken into consideration when finding the "single best method".

BillWagner:
> It's still not _entirely_ clear to me whether this is sufficient though. Consider:
>
> ...
>
> This compiles in the C# 10 compiler, even though the method group conversion for `X` to `Func<C, int>` doesn't have a "single best method" in terms of simple overload resolution: `X(new C());` fails to compile, for example. I'm hoping that this is a matter of a further improvement to 10.8 post-C#-6 which means that the return type of the method is taken into consideration when finding the "single best method".

I believe the behavior you describe maps to this issue https://github.com/dotnet/roslyn/issues/8942  which is addressed in C# 7.3. I made #499 to make sure we don't lose track of this one.

jskeet:
> I believe the behavior you describe maps to this issue https://github.com/dotnet/roslyn/issues/8942 which is addressed in C# 7.3. I made https://github.com/dotnet/csharpstandard/issues/499 to make sure we don't lose track of this one.

Excellent! Glad to hear that's not required for C# 6 :)

BillWagner (approved):
I like this strategy. I do think it addresses the final nits well. Great job!

RexJaeschke:
@jskeet, @BillWagner While one of the three reviewers has OK'd this, I wasn't comfortable with assuming the others would agree without change, so I did *not* put these edits in the Word document that I just submitted to Ecma. I can do that in the final draft that we submit late May.

BillWagner:
*Note*: I've changed the default branch to `draft-v7`. This is the only open PR on dtaft-v6. When this is merged, we'll need to push this change (and any others) into V7.

standard/expressions.md
```diff
@@ -1000,10 +1000,11 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
 
 #### 11.6.4.4 Better conversion from expression
 
-Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a better conversion than `C₂` if one of the following holds:
+Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if one of the following holds:
 
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` ([§11.6.4.5](expressions.md#11645-exactly-matching-expression))
 - `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§11.6.4.6](expressions.md#11646-better-conversion-target))
+- `E` is a method group ([§11.2](expressions.md#112-expression-classifications)), and `T₁` is compatible ([§19.4](delegates.md#194-delegate-compatibility)) with the single best method from the method group for conversion `C₁`
```
MadsTorgersen: We probably need to say that T2 is _not_ compatible

MadsTorgersen:
```suggestion
- `E` is a method group ([§11.2](expressions.md#112-expression-classifications)), `T₁` is compatible ([§19.4](delegates.md#194-delegate-compatibility)) with the single best method from the method group for conversion `C₁`, and `T₂` is not compatible with the single best method from the method group for conversion `C₂`
```

MadsTorgersen (changes requested):
Almost there! I made one suggested change to make sure the new bullet in Better Conversion actually differentiates T1 from T2.

Otherwise excellent!

standard/expressions.md
jskeet: I don't think we actually do.

As per the original PR comment:

> The rule doesn't refer to T2/C2, but I don't think it has to...
> because it's valid for both conversions to be better than the other.
> That's already the case in better conversion target, as types T1 and
> T2 could both have implicit conversions to each other.

So basically if E is a method group and both T1 and T2 are compatible with the best single method from the method group conversion, then each is better than the other and it's ambiguous.

Now I'd agree that it's a bit weird for both conversions to be better than each other, but given that that's *already* the case elsewhere, I figured I might as well go with the flow.

But I'm happy to be corrected if I'm missing something :)

MadsTorgersen: It looks like you're right, that better conversion target _also_ currently allows two types to both be better than one another. However, I think we should fix that too. I seem to remember having done that before, at some point, but who knows which version of spec out there that was in!

Here's why I think we should not allow two conversions to both be better than the other. In 11.6.4.3 Better Function Member we say:

> Given an argument list `A` with a set of argument expressions `{E₁, E₂, ..., Eᵥ}` and two applicable function members `Mᵥ` and `Mₓ` with parameter types `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂, ..., Qᵥ}`, `Mᵥ` is defined to be a ***better function member*** than `Mₓ` if
>
> - for each argument, the implicit conversion from `Eᵥ` to `Qᵥ` is not better than the implicit conversion from `Eᵥ` to `Pᵥ`, and
> - for at least one argument, the conversion from `Eᵥ` to `Pᵥ` is better than the conversion from `Eᵥ` to `Qᵥ`.

The first bullet will fail incredibly often if a type is better than itself, for instance!

So I think we should fix both this method group case, as I proposed, and also the implicit conversion case of better conversion target.

jskeet: Works for me. I'll accept your suggestion (tomorrow) and try to fix the implicit conversion case at the same time.

MadsTorgersen (approved):

jskeet:
@RexJaeschke: Please apply this to the Word version.

@BillWagner: Are you happy to cherry pick or merge this into draft-v7?

BillWagner:
> Are you happy to cherry pick or merge this into draft-v7?

@jskeet I'll cherry-pick it.

RexJaeschke:
@jskeet Changed in Word.

BillWagner:
@jskeet PR is #559 

(PR because of our branch protection rules. I couldn't just cherry-pick and push.

Issue #345, opened 2021-06-29
v7.x Q.s re Generalized Async Methods


RexJaeschke:
My notes re the addition of this feature are contained in the attached Word document ([Generalized_async_return_types.docx](https://github.com/dotnet/csharpstandard/files/6733946/Generalized_async_return_types.docx)). While I have come up with quite a few specific edits, I really don’t have enough to make a decent Draft PR. All the things marked in red or yellow need attention.

jskeet:
(Labeling this as a feature in that it's a request for more details *for* a feature.)

gafter:
@ljw1004 Your help on this issue would be most welcome!

RexJaeschke:
The points raised by the referred-to Word document have either been resolved or logged as comments in PR #556 (which contains the proposed changes), so this Issue can be closed.

PR #556, opened 2022-04-14
C# V7: Generalized async return types

RexJaeschke:
## Questions re 14.15 Async Functions|14.15.1 General: 

A. Re `System.Runtime.CompilerServices.AsyncMethodBuilder`, in the absence of a CLI spec update, presumably, we’ll need to spec this attribute type. Or at least update §C.3, “Standard Library Types not defined in ISO/IEC 23271” to include this new Attribute type.

B. §11.8.8.2, “Awaitable expressions,” allows for the possibility of the awaiter implementing `ICriticalNotifyCompletion`. What, if anything, should we say about awaiter/builder types being required/permitted to implement ICriticalNotifyCompletion?

Bill W.: Maybe. We (the ECMA committee) stuck with just INotifyCompletion. As such, this may be an implementation detail. Let’s discuss in committee.

C. Where are the semantics of the TaskBuilderType methods defined? Is there some sort of IBuilderMembers interface that an awaiter should implement, or does the awaiter simply have to declare each of those members?

Bill W.: My first thought is to follow the example of LINQ query expressions: A type must follow a “shape”, but what those methods actually do isn’t spelled out in detail in the spec.

D. Is there some spec for the V6 state-machine code generation for `Task` and `Task<T>` that I can adapt for all task types?

Bill W. I don’t think so, but from my comment above, we may not need to add this.

E: In which section should the spec for state-machine code generation go? In the async-return spec it’s listed under “Execution.”

Bill W. Do you think we need something like this: [§12.17.3](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#12173-query-expression-translation) Query expression translation? That’s my first thought.

## General question:

F. How/where do I merge in the text from the async-return spec section “Overload Resolution?” Maybe in 11.6.4.5, “Exactly matching expression.”

@gafter On yesterday's TG2 call you said you'd like to run this PR by an expert. I've spec'd all that I can for this feature, and listed the open issues above.


gafter:
I've asked @ljw1004 to look at this.

BillWagner:
The PR for local functions will be merged before this one.

This PR should be updated per https://github.com/dotnet/csharpstandard/pull/104#discussion_r864956697 to address other allowed return types in async local functions.

BillWagner:
See #345 for more questions on this feature

(Rex closed 345 on 2023-02-19.)

BillWagner:
[link to feature speclet](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.0/task-types.md)

BillWagner:
> Re `System.Runtime.CompilerServices.AsyncMethodBuilder`, in the absence of a CLI spec update, presumably, we’ll need to spec this attribute type. Or at least update §C.3, “Standard Library Types not defined in ISO/IEC 23271” to include this new Attribute type.

I added the declarations in §C.3 for `System.Runtime.CompilerServices.AsyncMethodBuilder`. 

> §11.8.8.2, “Awaitable expressions,” allows for the possibility of the awaiter implementing `ICriticalNotifyCompletion`. What, if anything, should we say about awaiter/builder types being required/permitted to implement `ICriticalNotifyCompletion`?

> C. Where are the semantics of the TaskBuilderType methods defined? Is there some sort of IBuilderMembers interface that an awaiter should implement, or does the awaiter simply have to declare each of those members?

> D. Is there some spec for the V6 state-machine code generation for Task and Task<T> that I can adapt for all task types?
> E: In which section should the spec for state-machine code generation go? In the async-return spec it’s listed under “Execution.”

I addressed all of these by adding the text for the semantics of the `TaskBuilderType` members. There isn't an interface, but the compiler expects certain semantics.  That text is in a new clause under §14.15. It specifies what calls the compiler generates to create an instance of the builder type and what methods will be  called for that builder type. 

> F. How/where do I merge in the text from the async-return spec section “Overload Resolution?” Maybe in 11.6.4.5, “Exactly matching expression.”

I added the first rule to "Exactly matching expression". The second and third I added to "Better conversion target".

BillWagner:
No further comments for any invited reviewers. 

@jskeet  Can we look at this one for the upcoming meeting as well?

jskeet:
Yup, absolutely. Please add the label - it's fiddly to do on my phone...

standard/classes.md
```diff
@@ -5172,12 +5172,80 @@ A method ([§14.6](classes.md#146-methods)) or anonymous function ([§11.17](exp
...
 
-The exact definition of the task types is implementation-defined, but from the language’s point of view, a task type is in one of the states *incomplete*, *succeeded* or *faulted*. A *faulted* task records a pertinent exception. A *succeeded* `Task<T>` records a result of type `T`. Task types are awaitable, and tasks can therefore be the operands of await expressions ([§11.8.8](expressions.md#1188-await-expressions)).
+An async method returning a task type is said to be ***task-returning***.
 
+The exact definition of the task types is implementation-defined, but from the language’s point of view, a task type is in one of the states *incomplete*, *succeeded* or *faulted*. A *faulted* task records a pertinent exception. A *succeeded* `«TaskType»<T>` records a result of type `T`. Task types are awaitable, and tasks can therefore be the operands of await expressions ([§11.8.8](expressions.md#1188-await-expressions)).
+
+> *Example*: The task type `MyTask<T>` is associated to the builder type `MyTaskMethodBuilder<T>` and the awaiter type `Awaiter<T>`:
```
```suggestion
> *Example*: The task type `MyTask<T>` is associated with the builder type `MyTaskMethodBuilder<T>` and the awaiter type `Awaiter<T>`:
```

standard/classes.md
```diff
@@ -5172,12 +5172,80 @@ A method ([§14.6](classes.md#146-methods)) or anonymous function ([§11.17](exp
...
 
-The *return_type* of an async method shall be either `void` or a ***task type***. The task types are `System.Threading.Tasks.Task` and types constructed from `System.Threading.Tasks.Task<T>`. For the sake of brevity, in this clause these types are referenced as `Task` and `Task<T>`, respectively. An async method returning a task type is said to be ***task-returning***.
+The *return_type* of an async method shall be either `void` or a ***task type***. For an async method that returns a value, a task type shall be generic. For an async method that does not return a value, a task type shall not be generic. Such types are referred to in this specification as `«TaskType»<T>` and `«TaskType»`, respectively. (The Standard library types `System.Threading.Tasks.Task<T>` and types constructed from `System.Threading.Tasks.Task` are task types.)
+A task type shall be a class or struct type that is associated with a ***builder type*** via the attribute `System.Runtime.CompilerServices.AsyncMethodBuilder`. Such types are referred to in this specification as `«TaskBuilderType»<T>` and `«TaskBuilderType»`.
```
jskeet: Should we call this task builder type instead of builder type? It feels too general at the moment.

standard/classes.md
```diff
@@ -5172,12 +5172,80 @@ A method ([§14.6](classes.md#146-methods)) or anonymous function ([§11.17](exp
...
+> {
+>     public void OnCompleted(Action completion) { ... }
+>     public bool IsCompleted { get; }
+>     public T GetResult() { ... }
+> }
+> ```
+>
+> *end example*
+
+A builder type is a class or struct type that corresponds to a specific task type. A builder type can have at most one type parameter and cannot not be nested in a generic type. A builder type shall have the following public methods (for non-generic builder types, `SetResult` has no parameters):
```
```suggestion
A builder type is a class or struct type that corresponds to a specific task type. A builder type can have at most one type parameter and cannot be nested in a generic type. A builder type shall have the following public members (for non-generic builder types, `SetResult` has no parameters):
```

(Changes methods to members due to the `Task` property.)

jskeet: It feels like this should be in "Task-type builder pattern" section (possibly with a link from here, although after moving there isn't much between this location and where it will be anyway).

jskeet: Do the members actually need to be public, if the task builder type is only being used in the same assembly as its declaration? (I'd expect them to only have to be accessible at the point of use.)

standard/classes.md
```diff
@@ -5172,12 +5172,80 @@ A method ([§14.6](classes.md#146-methods)) or anonymous function ([§11.17](exp
...
+}
+```
+  
 An async function has the ability to suspend evaluation by means of await expressions ([§11.8.8](expressions.md#1188-await-expressions)) in its body. Evaluation may later be resumed at the point of the suspending await expression by means of a ***resumption delegate***. The resumption delegate is of type `System.Action`, and when it is invoked, evaluation of the async function invocation will resume from the await expression where it left off. The ***current caller*** of an async function invocation is the original caller if the function invocation has never been suspended or the most recent caller of the resumption delegate otherwise.
 
+### §task-builder-pattern Task-type builder pattern
+
+The compiler generates code that uses the «TaskBuilderType» to implement the semantics of suspending and resuming the evaluation of the async function. The uses the «TaskBuilderType» as follows:
+
+- `Builder.Create()` is invoked to create an instance of the «TaskBuilderType», named `builder` in this list.
```
```suggestion
- `«TaskBuilderType».Create()` is invoked to create an instance of the «TaskBuilderType», named `builder` in this list.
```

standard/classes.md
```diff
@@ -5172,12 +5172,80 @@ A method ([§14.6](classes.md#146-methods)) or anonymous function ([§11.17](exp
...
+  - The builder must call `stateMachine.MoveNext()` either in `Start()` or after `Start()` has returned to advance the state machine.
+- After `Start()` returns, the `async` method invokes `builder.Task` for the task to return from the async method.
+- Each call to `stateMachine.MoveNext()` will advance the state machine.
+- If the state machine completes successfully, `builder.SetResult()` is called, with the method return value, if any.
+- Otherwise, if an exception, `e` is thrown in the state machine, `builder.SetException(e)` is called.
+- If the state machine reaches an `await expr` expression, `expr.GetAwaiter()` is invoked.
+- If the awaiter implements `ICriticalNotifyCompletion` and `IsCompleted` is false, the state machine invokes `builder.AwaitUnsafeOnCompleted(ref awaiter, ref stateMachine)`.
+  - `AwaitUnsafeOnCompleted()` should call `awaiter.UnsafeOnCompleted(action)` with an `Action` that calls `stateMachine.MoveNext()` when the awaiter completes.
+- Otherwise, the state machine invokes `builder.AwaitOnCompleted(ref awaiter, ref stateMachine)`.
+  - `AwaitOnCompleted()` should call `awaiter.OnCompleted(action)`.
```
```suggestion
  - `AwaitOnCompleted()` should call `awaiter.OnCompleted(action)` with an `Action` that calls `stateMachine.MoveNext()` when the awaiter completes.
```

jskeet: (For symmetry with the unsafe version.)

standard/classes.md
```diff
@@ -5172,12 +5172,80 @@ A method ([§14.6](classes.md#146-methods)) or anonymous function ([§11.17](exp
...
+- `SetStateMachine(IAsyncStateMachine)` may be called by the compiler-generated `IAsyncStateMachine` implementation to identify the instance of the builder associated with a state machine instance, particularly for cases where the state machine is implemented as a value type.
+  - If the builder calls `stateMachine.SetStateMachine(stateMachine)`, the `stateMachine` will call `builder.SetStateMachine(stateMachine)` on the *builder instance associated with* `stateMachine`.
+
+> *Note:* The generated code is equivalent to the code generated for async methods that return `Task`, `Task<T>`, or `void`. The difference is, for those well known types, the *builder types* are also known to the compiler. In other words, the semantics for a *builder-type* should match the semantics of the known builder types for `Task`, `Task<T>`, or `void` *end note*
```
jskeet: Is this actually an implementation detail? (It would be reasonable to have an entirely different pattern for the implementation-specific known builder types.) I guess "the semantics" should be the same in the end, but there could be quite different members available.

standard/conversions.md
```diff
@@ -737,18 +737,17 @@ Given a user-defined conversion operator that converts from a non-nullable value
...
-- If the body of `F` is a block, and *either* `D` has a void return type *or* `F` is async and `D` has the return type `Task`, then when each parameter of `F` is given the type of the corresponding parameter in `D`, the body of `F` is a valid block (w.r.t [§12.3](statements.md#123-blocks)) in which no `return` statement specifies an expression.
-- If the body of `F` is an expression, and *either* `F` is non-async and `D` has a non-`void` return type `T`, *or* `F` is async and `D` has a return type `Task<T>`, then when each parameter of `F` is given the type of the corresponding parameter in `D`, the body of `F` is a valid expression (w.r.t [§11](expressions.md#11-expressions)) that is implicitly convertible to `T`.
-- If the body of `F` is a block, and *either* `F` is non-async and `D` has a non-void return type `T`, *or* `F` is async and `D` has a return type `Task<T>`, then when each parameter of `F` is given the type of the corresponding parameter in `D`, the body of `F` is a valid block (w.r.t [§12.3](statements.md#123-blocks)) with a non-reachable end point in which each return statement specifies an expression that is implicitly convertible to `T`.
+
```
jskeet: Remove this blank line? The next four bullets are intended to be in the same list, right?

standard/conversions.md
```diff
@@ -737,18 +737,17 @@ Given a user-defined conversion operator that converts from a non-nullable value
...
+- If the body of `F` is an expression, and *either* `D` has a void return type *or* `F` is async and `D` has the return type `«TaskType»` (§14.15.1](classes.md#14151-general)), then when each parameter of `F` is given the type of the corresponding parameter in `D`, the body of `F` is a valid expression (w.r.t [§11](expressions.md#11-expressions)) that would be permitted as a *statement_expression* ([§12.7](statements.md#127-expression-statements)).
```
jskeet: I'm not sure about "has the return type `«TaskType»`" - would "has a `«TaskType»` return type" be better? It feels like `«TaskType»` isn't really a single type.

jskeet: (I note that in the generic version below, it's "has a return type `«TaskType»<T>`" which is half way between the type...)

jskeet (commented):
A few comments - I'm okay to go with a somewhat vague wording for `«TaskType»` in the hope that we tighten it up before *final* submission.

(There are a few other changes that should be easier to make or reject.)

standard/classes.md
BillWagner: Correct. They must be accessible.

standard/classes.md
```diff
@@ -5172,12 +5172,80 @@ A method ([§14.6](classes.md#146-methods)) or anonymous function ([§11.17](exp
...
+> {
+>     public void OnCompleted(Action completion) { ... }
+>     public bool IsCompleted { get; }
+>     public T GetResult() { ... }
+> }
+> ```
+>
+> *end example*
+
+A builder type is a class or struct type that corresponds to a specific task type. A builder type can have at most one type parameter and cannot be nested in a generic type. A builder type shall have the following public members (for non-generic builder types, `SetResult` has no parameters):
```
BillWagner: Per previous discussion:

```suggestion
A builder type is a class or struct type that corresponds to a specific task type. A builder type can have at most one type parameter and cannot be nested in a generic type. A builder type shall have the following accessible members (for non-generic builder types, `SetResult` has no parameters):
```

standard/classes.md
BillWagner: I tend to agree. As a note, it's not normative. Do you think we should remove it?

jskeet: I think that's best at least for now, yes.

standard/classes.md
```diff
@@ -5172,12 +5172,80 @@ A method ([§14.6](classes.md#146-methods)) or anonymous function ([§11.17](exp
...
+- Otherwise, the state machine invokes `builder.AwaitOnCompleted(ref awaiter, ref stateMachine)`.
+  - `AwaitOnCompleted()` should call `awaiter.OnCompleted(action)` with an `Action` that calls `stateMachine.MoveNext()` when the awaiter completes.
+- `SetStateMachine(IAsyncStateMachine)` may be called by the compiler-generated `IAsyncStateMachine` implementation to identify the instance of the builder associated with a state machine instance, particularly for cases where the state machine is implemented as a value type.
+  - If the builder calls `stateMachine.SetStateMachine(stateMachine)`, the `stateMachine` will call `builder.SetStateMachine(stateMachine)` on the *builder instance associated with* `stateMachine`.
+
+> *Note:* The generated code is equivalent to the code generated for async methods that return `Task`, `Task<T>`, or `void`. The difference is, for those well known types, the *task builder types* are also known to the compiler. In other words, the semantics for a *task-builder-type* should match the semantics of the known task builder types for `Task`, `Task<T>`, or `void` *end note*
```
BillWagner: Remove note per feedback:

```suggestion
```

standard/conversions.md
BillWagner: Good catch. Updated both.

standard/classes.md
BillWagner: I like it. Made that change.

BillWagner:
@jskeet Async return types is rebased and ready for the upcoming meeting.

standard/conversions.md
```diff
@@ -767,18 +767,16 @@ Given a user-defined conversion operator that converts from a non-nullable value
...
-- If the body of `F` is a block, and *either* `F` is non-async and `D` has a non-void return type `T`, *or* `F` is async and `D` has a return type `Task<T>`, then when each parameter of `F` is given the type of the corresponding parameter in `D`, the body of `F` is a valid block (w.r.t [§13.3](statements.md#133-blocks)) with a non-reachable end point in which each return statement specifies an expression that is implicitly convertible to `T`.
+- If the body of `F` is an expression, and *either* `D` has a void return type *or* `F` is async and `D` has the return type `«TaskType»` (§15.15.1](classes.md#15151-general)), then when each parameter of `F` is given the type of the corresponding parameter in `D`, the body of `F` is a valid expression (w.r.t [§12](expressions.md#12-expressions)) that would be permitted as a *statement_expression* ([§13.7](statements.md#137-expression-statements)).
```
```suggestion
- If the body of `F` is an expression, and *either* `D` has a void return type *or* `F` is async and `D` has a `«TaskType»` return type  (§15.15.1](classes.md#15151-general)), then when each parameter of `F` is given the type of the corresponding parameter in `D`, the body of `F` is a valid expression (w.r.t [§12](expressions.md#12-expressions)) that would be permitted as a *statement_expression* ([§13.7](statements.md#137-expression-statements)).
```

standard/conversions.md
```diff
@@ -767,18 +767,16 @@ Given a user-defined conversion operator that converts from a non-nullable value
...
+- If the body of `F` is an expression, and *either* `F` is non-async and `D` has a non-`void` return type `T`, *or* `F` is async and `D` has a `«TaskType»<T>` return type ([§15.15.1](classes.md#15151-general)), then when each parameter of `F` is given the type of the corresponding parameter in `D`, the body of `F` is a valid expression (w.r.t [§12](expressions.md#12-expressions)) that is implicitly convertible to `T`.
+- If the body of `F` is a block, and *either* `F` is non-async and `D` has a non-void return type `T`, *or* `F` is async and `D` has a return type `«TaskType»<T>`, then when each parameter of `F` is given the type of the corresponding parameter in `D`, the body of `F` is a valid statement block (w.r.t [§13.3](statements.md#133-blocks)) with a non-reachable end point in which each return statement specifies an expression that is implicitly convertible to `T`.
```
```suggestion
- If the body of `F` is a block, and *either* `F` is non-async and `D` has a non-void return type `T`, *or* `F` is async and `D` has a `«TaskType»<T>` return type, then when each parameter of `F` is given the type of the corresponding parameter in `D`, the body of `F` is a valid statement block (w.r.t [§13.3](statements.md#133-blocks)) with a non-reachable end point in which each return statement specifies an expression that is implicitly convertible to `T`.
```

jskeet (approved):

jskeet (approved):

jskeet:
Approved on 2023-05-17 meeting.

PR #559, opened 2022-04-27
Make "better conversions" one way

BillWagner:
This prevents "T1 is better than T2, and T2 is also better than T1"
(for both targets and types).

/cc @jskeet 

This has your cherry-picked commit, and nothing else.

We should *not* squash merge this.

(PR because of our branch protection rules)

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1004,6 +1004,7 @@ Given an implicit conversion `C₁` that converts from an expression `E` to a ty
 
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` ([§11.6.4.5](expressions.md#11645-exactly-matching-expression))
 - `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§11.6.4.6](expressions.md#11646-better-conversion-target))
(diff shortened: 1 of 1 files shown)


standard/expressions.md
```diff
@@ -1004,6 +1004,7 @@ Given an implicit conversion `C₁` that converts from an expression `E` to a ty
 
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` ([§11.6.4.5](expressions.md#11645-exactly-matching-expression))
 - `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§11.6.4.6](expressions.md#11646-better-conversion-target))
+- `E` is a method group ([§11.2](expressions.md#112-expression-classifications)), `T₁` is compatible ([§19.4](delegates.md#194-delegate-compatibility)) with the single best method from the method group for conversion `C₁`, and `T₂` is not compatible with the single best method from the method group for conversion `C₂`
```
jskeet, line 1007: This was moved from line 1031, which needs to be removed too... and line 1021 needs changing too.
I suspect you only cherry-picked the second commit from #553 - I should have highlighted that there were two commits there, sorry!

BillWagner: That's fine. I'll close this, and file it as "too smart by half".

I'll open a new PR that merges V6 into V7. That way, we also ensure any other changes we lost track of are current.

BillWagner:
Closed in favor of #560

PR #566, opened 2022-05-04
Update documentation-comments.md

stesee:
Added missing '/'

PR #567, opened 2022-05-04
Fixes to documentation comments

jskeet:
One based on #566, but also a typo in the included XML. I originally
thought we should delete StringList, but I assume the intention is
actually to indicate that the relevant piece of XML is included and
other pieces are ignored.

(Once this is merged, I'll make the same XML change in draft-v7.)

Rex: please update Word doc with these changes. (We need to work out when to stop doing this, of course.)

### DIFF
--- a/standard/documentation-comments.md
+++ b/standard/documentation-comments.md
@@ -221,7 +221,7 @@ and the external file “docs.xml” had the following contents:
     </class>
     <class name="StringList">
         <summary>
-            Contains a list of integers.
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
Good catch @jskeet 

I'll approve. (Waiting to merge until @RexJaeschke sees this.

RexJaeschke (approved):

RexJaeschke:
Changes made in Word V6.

PR #574, opened 2022-05-10
Clarified note about strings in comments

svick:
The potential string literal `"D */ "` is not wholly contained inside a comment in the example code, so I think it's confusing to say it "appears" there. Making it explicit that it's the opening `"` character that is relevant should make this note clearer.

Also, made a tiny grammar edit in the following paragraph.


Nigel-Ecma (approved):
LGTM

jskeet (approved):

gafter (approved):

PR #580, opened 2022-05-17
Tweak return-value description for partial methods

RexJaeschke:
As a result of the features we added in V6, we have

```ANTLR
method_body
    : block
    | '=>' null_conditional_invocation_expression ';'
    | '=>' expression ';'
    | ';'
    ;
```

When we added the two "=>" rules, we missed updating the narrative describing *method_body* for partial methods. It currently mentions the semicolon and *block* cases only, but it needs to include all four possibilities. Rather than enumerate the possibilities, I've simply used "semicolon" and "other than semicolon."


jskeet (commented):
Good catch - that works for me. (Would be good to get other perspectives though.)

BillWagner (approved):
This LGTM as well.

Let's :shipit:

PR #584, opened 2022-05-25
Enable smarten again

BillWagner:
Turn on the config that runs smarten on each PR merge.

Note the changes in expressions.md.  Our smarten tool out-dents that table. There are two possible fixes: 

1. replace the HTML table with a markdown table. Issue: Will the Word converter work?
2. Fix the smarten code. Issue: That's not quite so simple.

Ping @jskeet @Nigel-Ecma for thoughts.

### DIFF
--- a/.github/workflows/update-on-merge.yaml
+++ b/.github/workflows/update-on-merge.yaml
@@ -37,12 +37,12 @@ jobs:
       with:
         java-version: 15.0
 
-#    - name: Smarten quotes
-#      id: smarten-quote
-#      run: | 
-#        cd tools
-#        ./run-smarten.sh
-#      shell: bash
+    - name: Smarten quotes
+      id: smarten-quote
+      run: | 
+        cd tools
+        ./run-smarten.sh
+      shell: bash
 
     - name: Renumber sections
       id: renumber-sections
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -3212,7 +3212,6 @@ The predefined multiplication operators are listed below. The operators all comp
 
   The product is computed according to the rules of IEC 60559 arithmetic. The following table lists the results of all possible combinations of nonzero finite values, zeros, infinities, and NaNs. In the table, `x` and `y` are positive finite values. `z` is the result of `x * y`, rounded to the nearest representable value. If the magnitude of the result is too large for the destination type, `z` is infinity. Because of rounding, `z` may be zero even though neither `x` nor `y` is zero.
   
-  <!-- Custom Word conversion: multiplication -->
   |       | `+y`  | `-y`  | `+0`  | `-0`  | `+∞`  | `-∞`  | `NaN` |
   | :---- | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
   | `+x`  | `+z`  | `-z`  | `+0`  | `-0`  | `+∞`  | `-∞`  | `NaN` |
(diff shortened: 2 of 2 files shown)


.github/workflows/update-on-merge.yaml
```diff
@@ -37,12 +37,12 @@ jobs:
...
-#      run: | 
-#        cd tools
-#        ./run-smarten.sh
-#      shell: bash
+    - name: Smarten quotes
+      id: smarten-quote
+      run: | 
+        cd tools
+        ./run-smarten.sh
+      shell: bash
```
BillWagner, line 45: This is the change that will run smarten on each PR merge.

standard/expressions.md
```diff
@@ -3214,95 +3214,95 @@ The predefined multiplication operators are listed below. The operators all comp
   
   <!-- Custom Word conversion: multiplication -->
   <table>
-  <!-- md equivalent:   ` `   | **`+y`**  | **`-y`**  | **`+0`**  | **`-0`**  | **`+∞`**  | **`-∞`**  | **`NaN`** -->
```
BillWagner, line 3217: This change must be reverted. See notes in the PR description for actions needed.

standard/expressions.md
```diff
@@ -3478,95 +3478,95 @@ The predefined remainder operators are listed below. The operators all compute t
...
+    <td><code>NaN</code></td>
+    <td><code>NaN</code></td>
+    <td><code>NaN</code></td>
+    <td><code>NaN</code></td>
+    <td><code>NaN</code></td>
+    <td><code>NaN</code></td>
+    <td><code>NaN</code></td>
+  </tr>
+</table>
 - Decimal remainder:
```
jskeet, line 3325: I believe if we add a blank line after the table, the Word converter should be happy.

standard/expressions.md
jskeet: Does it actually need to be reverted? The Word converter uses the "Custom Word conversion: multiplication" comment above to put the table in. I think we're probably okay to lose the "md equivalent" comment... or we could potentially have a separate comment with the markdown equivalents if we want.

(We could try using the Markdown equivalent directly, of course - I don't remember the details of why we've got an HTML table here.)

BillWagner: I'll run two experiments:

1. Remove the MD equivalent comment. See if it round-trips.
2. Use a markdown table, and check the word and GH rendering.

BillWagner: OK. I ran the markdown table experiment. That works, and the output is better on GH. The Word version is unchanged, and smarten works without any issues.

I think this is the solution we should use. I'm wondering if we should remove the code fences around these tables. It looks kind of ugly.

jskeet: Interesting. I don't remember now what the problems we had were previously...

standard/expressions.md
```diff
@@ -3213,96 +3213,15 @@ The predefined multiplication operators are listed below. The operators all comp
...
-      <td><code>NaN</code></td>
-      <td><code>NaN</code></td>
-      <td><code>NaN</code></td>
-      <td><code>NaN</code></td>
-      <td><code>NaN</code></td>
-      <td><code>NaN</code></td>
-      <td><code>NaN</code></td>
-    </tr>
-  </table>
+  |       | `+y`  | `-y`  | `+0`  | `-0`  | `+∞`  | `-∞`  | `NaN` |
```
jskeet, line 3215: Please could you remove the Custom Word conversion comment? I *suspect* it's now not being used, but it's slightly confusing to have it there :)

standard/expressions.md
BillWagner: Speculation from looking at the table source: It was not correctly formed markdown and that caused issues with our other tools.

The badly formed parts were that the table was missing the opening `|` on the first column, and the closing `|` on the last column. The first row wasn't formatted as a header row. (Meaning the 2nd row wasn't specifying the justification).  

In time, it may be worth looking at the other tables where we have HTML and see if we can use properly formatted markdown tables for those as well.

standard/expressions.md
BillWagner: I did do that, and the Word output was fine. That's now pushed.

jskeet (approved):
Thanks, that's great :)

standard/expressions.md
gewarren: @BillWagner These tables are getting flagged in the build report for dotnet/docs because there's a column without a heading. You can resolve it by bolding everything in the first column to show that it's a data matrix.

https://buildapi.docs.microsoft.com/Output/PullRequest/a3bda507-3390-de91-8f7f-26f90f4e5fc8/202206090300048116-29770/BuildReport?accessString=7e8b720a7ea2f64577ecd0387de1f9e8f766ed8d1f35267ed986663930190107

BillWagner: Thanks. I'll handle this as part of #585, which I'm planning on this week.

PR #587, opened 2022-05-29
Fix ANTLR representation of Fixed-size buffer declarations

neNerower:
Fixes the problem described in issue #583.

### DIFF
--- a/standard/unsafe-code.md
+++ b/standard/unsafe-code.md
@@ -863,7 +863,7 @@ Fixed-size buffers are only permitted in struct declarations and may only occur 
 ```ANTLR
 fixed_size_buffer_declaration
     : attributes? fixed_size_buffer_modifier* 'fixed' buffer_element_type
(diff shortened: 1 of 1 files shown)


Nigel-Ecma (approved):
@neNerower – Again thanks!

BillWagner:
ping @jskeet I'm happy merging this. Do you want to take a look

jskeet:
If Nigel's happy, I'm happy - I don't know ANTLR well enough to know whether introducing a new rule is better or worse than inlining it.

neNerower:
@jskeet 
It's not about changing the style, but about fixing the inaccuracy of the grammar, which made it possible to interpret incorrect statements as valid.
Nigel is aware of all the details, you can also find more information in issue #583.

jskeet:
I understand the change was semantic as well - but I assume it *could* have been done inline as well. I'm fine either way.

Issue #582, opened 2022-05-23
We use the term "module" but never define it


RexJaeschke:
I've been writing the formal spec for a V9 feature, module initialization. As I was reading the MS proposal, it occurred to me that I didn't know what the term *module* meant, esp. as something separate from *assembly*. So, I looked at the C# spec.

We use the term *module* nine times, all in the context of specifying a global attribute that applies to a module (or assembly). However, we never define that term.

We do define some terms around *assembly*, however:

> - **application**: assembly with an entry point
> - **assembly**: one or more files output by the compiler as a result of program compilation [we use this term 27 times]
> - **class library**: assembly that can be used by other assemblies

In theory, a spec should contain everything to allow someone to build a clean-room implementation, but I doubt ours does with regard to module and assembly, especially if we are to *not* rely on .NET-CLI being "under the hood."

According to the CLI spec, Partition I.5, “Terms and definitions”:

> - **assembly**: A configured set of loadable code modules and other resources that together implement a unit of functionality.
> - **library**: A repository for a set of types, which are grouped into one or more assemblies. A library can also contain modifications to types defined in other libraries. For example, a library can include additional methods, interfaces, and exceptions for types defined in other libraries.
> - **module**: A single file containing content that can be executed by the VES. 
> - **Virtual Execution System (VES)**: This system implements and enforces the CTS model. The VES is responsible for loading and running programs written for the CLI. It provides the services needed to execute managed code and data using the metadata to connect separately generated modules together at runtime. The VES is also known as the Execution Engine.

The CLI spec II.6.1, "Overview of modules, assemblies, and files," draws a picture of the relationship between these three things and has accompanying narrative, but I find it hard-going to read and to map onto a conforming C# implementation.

The [site](https://www.infoq.com/news/2020/01/CSharp-Module-Initializers/) says:

> A [“module” in the .NET CLR sense](https://blogs.msdn.microsoft.com/junfeng/2004/07/15/multimodule-assemblies/) is a file containing IL code. An “assembly” is [a] logical unit consisting of one or more modules, one of which is designated as the head assembly. **Most .NET languages are designed to only create single-module/single-file assemblies. So, for most developers, the terms are interchangeable.**

What, if anything, more can/should we say about modules (and maybe assemblies)?

svick:
This may be irrelevant to the spec, but note that .Net Core does not support multi-module assemblies, see https://github.com/dotnet/runtime/issues/10598.

gafter:
[...]

PR #593, opened 2022-06-09
Add definition for module

RexJaeschke:
This addresses Issue https://github.com/dotnet/csharpstandard/issues/582.

### DIFF
--- a/standard/terms-and-definitions.md
+++ b/standard/terms-and-definitions.md
@@ -34,6 +34,8 @@ For the purposes of this specification, the following definitions apply. Other t
   - exceptional condition reported during program execution
 - **implementation**
   - particular set of software (running in a particular translation environment under particular control options) that performs translation of programs for, and supports execution of methods in, a particular execution environment
+- **module**
+  - the contents of an assembly produced by a compiler.  Some implementations may have facilities to produce assemblies that contain more than one module.  The behavior in such situations is outside the scope of this specification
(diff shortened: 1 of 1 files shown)


standard/terms-and-definitions.md
```diff
@@ -34,6 +34,8 @@ For the purposes of this specification, the following definitions apply. Other t
   - exceptional condition reported during program execution
 - **implementation**
   - particular set of software (running in a particular translation environment under particular control options) that performs translation of programs for, and supports execution of methods in, a particular execution environment
+- **module**
+  - the contents of an assembly produced by a compiler.  Some implementations may have facilities to produce assemblies that contain more than one module.  The behavior in such situations is outside the scope of this specification
```
jskeet, line 38: Add a period at the end given that it's a third complete sentence?

jskeet (approved):
Basically fine. (Personally I don't like double spaces, but it's no big deal.)

standard/terms-and-definitions.md
RexJaeschke: @jskeet None of the entries in this chapter are terminated with periods, so I was being consistent. In any event, the style of each entry was initially based on ISO rules, which require that each definition be able to replace directly an instance of the term. Of course, that doesn't work grammatically if a definition contains multiple sentences. So, for now, I'll leave it as is.

PR #595, opened 2022-06-10
fix column headers.

BillWagner:
See https://github.com/dotnet/csharpstandard/pull/584#discussion_r893034446

All remaining HTML tables (Function members, and string formatting). These should remain as HTML tables.

/cc @gewarren


gewarren (commented):
Thanks Bill. What about the tables on lines 3266 and 3315?

gewarren (approved):
Thank you!

RexJaeschke (approved):

PR #601, opened 2022-07-19
C# 7.x: ref struct

### LINKED ISSUE #267: Draft C# 7.2 language spec for ref structs, missing clause on iterators defined on ref structs?
The compiler forbids iterator methods on ref structs, e.g.

```cs
public ref struct RefStruct 
{
    public IEnumerator<int> GetEnumerator()
    {
        yield return 1;
    }
}
```

This makes sense, however I can't see any wording in the spec which explicitly forbids it. The closest it gets is:

> Neither a ref local, nor a local of a ref struct type may be in scope at the point of a yield return statement or an await expression.

I suspect it simply didn't consider that `this` is always in scope for this case, and can be a ref struct. It also doesn't cover fields which are themselves ref structs (although `this` being a ref struct implicitly covers this case). Perhaps it could do with a tweak to cover this case?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: d07ca1ab-2178-2c31-ceda-773ee5bd03b4
* Version Independent ID: b815bff8-7b47-d637-c10a-96c5e411e2d1
* Content: [Compile time safety for ref-like types - C# 7.2 specification proposals](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-7.2/span-safety#feedback)
* Content Source: [proposals/csharp-7.2/span-safety.md](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-7.2/span-safety.md)
* Product: **dotnet-csharp**
* Technology: **csharp-spec**
* GitHub Login: @dotnet-bot
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 52135](https://dev.azure.com/msft-skilling/Content/_workitems/edit/52135)

### LINKED ISSUE #334: v7.x Q.s re ref struct
This issue is intended to help complete the (Draft) PR #333.

## A. Terminology, "execution stack"

As drafted, the PR states 

> The `ref` modifier indicates that the *struct_declaration* declares a type whose instances are allocated on the **execution stack**.

The term "execution stack" occurs in the spec in the following places:

> exceptions.md: `System.StackOverflowException` -- Thrown when the **execution stack** is exhausted by having too many pending calls; typically indicative of very deep or unbounded recursion.

> expressions.md:  - `E` is evaluated. If this evaluation causes an exception, then no further steps are executed. For an instance constructor, this evaluation consists of allocating storage (typically from an **execution stack**) for the new object. ...

The term "call stack" occurs three times in unsafe.md.

The term "stack" on its own appears in unsafe.md ("... to allocate a buffer of 16 characters on the **stack**".)

**Q1. Should we consolidate the various kinds of uses of "stack"?**

**Q2.  Is it a spec requirement that a ref struct instance be on the stack, or could it be in some other "fixed" place?**

Bill thought perhaps we might consider borrowing words from "23.4 Fixed and moveable variables." That section currently contains the following:

> The address-of operator (§23.6.5) and the fixed statement (§23.7) divide variables into two categories: Fixed variables and moveable variables.
> Fixed variables reside in storage locations that are unaffected by operation of the garbage collector. (Examples of fixed variables include local variables, value parameters, and variables created by dereferencing pointers.) On the other hand, moveable variables reside in storage locations that are subject to relocation or disposal by the garbage collector. (Examples of moveable variables include fields in objects and elements of arrays.)

Perhaps we could massage this and move it out of the unsafe.md chapter, as it applies to non-unsafe things as well, and then use "fixed" rather than "stack-based."

## B. What to include in the spec w.r.t "safe-to-escape" scopes

The primary starter document for the ref struct specification is https://github.com/dotnet/csharplang/blob/standard-proposals/proposals/csharp-7.2/span-safety.md (hereinafter called the span-safety spec). This document is broken into the following main sections: Introduction and Explanations, Draft language specification, and Language Constraints, the last two of which contain standards-like wording.

This paper has a lot of background/rationale/tutorial material. It talks about expressions and lvalues “escaping to a scope” and draws some parallels to the ref locals feature. My reaction was that this is interesting background with some implementation details, but how much of it belongs in a standard? As a result, I posed the following question:

**Presumably, the whole idea of “escaping” and related scopes needs to be in the spec. Do you agree?**

Jared Parsons’ (MS) response> This impacts type / memory safety hence I would lean towards it being included in the spec. 

**Should we have a unification approach to “all-things-ref”? That is, if the answer to Q1 above is to have a general intro and specifics spread around the spec, should they factor in the use of ref with arguments/parameters (the only use of ref prior to V7)? If so, should the existing ref arg/param text be adjusted (if necessary) to use any new terminology?**

Separately, the spec currently uses “reference” in a number of unrelated ways: variable reference, reference type, reference argument/parameter, and then there is ref. It seems confusing to use “reference” in any discussion of ref; perhaps “alias” would be better.

Jared Parsons’ response> I agree we need to get crisp now because ref values are a lot more prominent in the language than they used to be. Consistency around when we are referring to a ref value vs. an alias would likely make the spec a lot more readable. 

**Q3. The bottom line is "How much do we need to include in the spec w.r.t Scopes and Escaping?"**

## C. Lvalues and Rvalues

The span-safety-spec mentioned above makes considerable use of the (C-language) terms *lvalue* and *rvalue*. As C# does not traffic in those terms, we'd need to use something like *variable reference* and *value*, respectively.

---
[Associated WorkItem - 52133](https://dev.azure.com/msft-skilling/Content/_workitems/edit/52133)

gafter:
Split out `ref struct` from #333.

This still needs to safety rules to be incorporated.  See https://github.com/dotnet/csharpstandard/issues/334 and https://github.com/dotnet/csharplang/blob/standard-proposals/proposals/csharp-7.2/span-safety.md#draft-language-specification .  Those safety rules are for ref return, ref locals, and ref structs.

Fixes #267 
Fixes #334

### DIFF
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -64,7 +64,7 @@ When an instance of a readonly struct is passed to a method, its `this` is treat
 
 ### §ref-modifier-new-clause Ref modifier
 
-The `ref` modifier indicates that the *struct_declaration* declares a type whose instances are allocated on the execution stack. The `ref` modifier declares that instances may contain ref-like fields, and may not be copied out of its safe-scope (§safe-scope-rules). The rules for determining the safe scope of a `ref struct` are described in §safe-scope-rules.
+The `ref` modifier indicates that the *struct_declaration* declares a type whose instances are allocated on the execution stack. The `ref` modifier declares that instances may contain ref-like fields, and may not be copied out of its safe-context (§safe-context-rules). The rules for determining the safe context of a `ref struct` are described in §safe-context-rules.
 
 It is a compile-time error if a ref struct type is used in any of the following contexts:
 
@@ -502,65 +502,67 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
 
 > *Note*: This access restriction means that constructors in structs containing automatically implemented properties often need an explicit constructor initializer where they would not otherwise need one, to satisfy the requirement of all fields being definitely assigned before any function member is invoked or the constructor returns. *end note*
 
-### §safe-scope-rules Safe scope constraint for `ref struct` types
+### §safe-context-rules Safe context constraint for `ref struct` types
 
-#### safe-scope-rules-general General
+#### §safe-context-rules-general General
 
-At compile-time, each expression whose type is a ref struct is associated with a scope where that instance is safe, its ***safe-scope***. The safe-scope is a scope, enclosing an expression, which it is safe for the value to escape to. If that scope is the entire function member, we say that the value is ***safe-to-return*** from the function member.
+At compile-time, each expression whose type is a ref struct is associated with a context where that instance is safe, its ***safe-context***. The safe-context is a context, enclosing an expression, which it is safe for the value to escape to. If that context is the entire function member, we say that the value is ***safe-to-return*** from the function member.
 
-The safe-scope records which scope a ref struct may be copied into. Given an assignment from an expression `E1` with a safe-scope `S1`, to an expression `E2` with safe-scope `S2`, it is an error if `S2` is a wider scope than `S1`.
+The safe-context records which context a ref struct may be copied into. Given an assignment from an expression `E1` with a safe-context `S1`, to an expression `E2` with safe-context `S2`, it is an error if `S2` is a wider context than `S1`.
 
-There are three different safe-scope values, the same as the ref-safe-scope values defined for reference variables (§ref-span-safety-escape-scopes): block, function-member, and caller-scope. An expression `e1` of a ref struct type is constrained by its safe-scope as follows:
+There are three different safe-context values, the same as the ref-safe-context values defined for reference variables (§ref-safe-contexts): declaration-block, function-member, and caller-context. When an expression has the safe-context of caller-context, that expression is safe-to-return. An expression `e1` of a ref struct type is constrained by its safe-context as follows:
 
-- For a return statement `return e1`, the safe-scope of `e1` must be calling-method.
-- For an assignment `e1 = e2` the safe-scope of `e2` must be at least as wide a scope as the safe-scope of `e1`.
+- For a return statement `return e1`, the safe-context of `e1` must be calling-method. The expression `e1` must be safe-to-return.
+- For an assignment `e1 = e2` the safe-context of `e2` must be at least as wide a context as the safe-context of `e1`.
 
-For a method invocation if there is a `ref` or `out` argument of a `ref struct` type (including the receiver unless the type is `readonly`), with safe-scope `S1`, then no argument (including the receiver) may have a narrower safe-scope than `S1`.
+For a method invocation if there is a `ref` or `out` argument of a `ref struct` type (including the receiver unless the type is `readonly`), with safe-context `S1`, then no argument (including the receiver) may have a narrower safe-context than `S1`.
 
-Any expression whose compile-time type is not a ref struct has a safe-scope of calling-method.
+Any expression whose compile-time type is not a ref struct has a safe-context of calling-method.
 
-A `default` expression, for any type, has safe-scope of calling-method.
+A `default` expression, for any type, has safe-context of calling-method.
 
-#### safe-scope-rules-parameter Parameter safe scope
+#### safe-context-rules-parameter Parameter safe context
 
-A formal parameter of a ref struct type, including the `this` parameter of an instance method, has a safe-scope of calling-method.
+A formal parameter of a ref struct type, including the `this` parameter of an instance method, has a safe-context of calling-method.
 
-#### safe-scope-rules-local Local variable safe scope
+#### safe-context-rules-local Local variable safe context
 
-A local variable of a ref struct type has a safe-scope as follows:
+A local variable of a ref struct type has a safe-context as follows:
 
-- If the variable is an iteration variable of a `foreach` loop, then the variable's safe-scope is the same as the safe-scope of the `foreach` loop's expression.
-- Otherwise if the variable's declaration has an initializer then the variable's safe-scope is the same as the safe-scope of that initializer.
-- Otherwise the variable is uninitialized at the point of declaration and has a safe-scope of the calling method.
+- If the variable is an iteration variable of a `foreach` loop, then the variable's safe-context is the same as the safe-context of the `foreach` loop's expression.
+- Otherwise if the variable's declaration has an initializer then the variable's safe-context is the same as the safe-context of that initializer.
+- Otherwise the variable is uninitialized at the point of declaration and has a safe-context of the calling method.
 
-#### safe-scope-rules-field Field safe scope
+#### §safe-context-rules-field Field safe context
 
-A reference to a field `e.F`, where the type of `F` is a ref struct type, has a safe-scope that is the same as the safe-scope of `e`.
+A reference to a field `e.F`, where the type of `F` is a ref struct type, has a safe-context that is the same as the safe-context of `e`.
 
-#### safe-scope-rules-operator Operators
+#### §safe-context-rules-operator Operators
 
-The application of a user-defined operator is treated as a method invocation (§safe-scope-method-invocation).
+The application of a user-defined operator is treated as a method invocation (§safe-context-method-invocation).
 
-For an operator that yields a value, such as `e1 + e2` or `c ? e1 : e2`, the safe-scope of the result is the narrowest scope among the safe-scopes of the operands of the operator. As a consequence, for a unary operator that yields a value, such as `+e`, the safe-scope of the result is the safe-scope of the operand.
+For an operator that yields a value, such as `e1 + e2` or `c ? e1 : e2`, the safe-context of the result is the narrowest context among the safe-contexts of the operands of the operator. As a consequence, for a unary operator that yields a value, such as `+e`, the safe-context of the result is the safe-context of the operand.
 
-#### §safe-scope-method-invocation Method and property invocation
+> *Note*: The first operand of a conditional operator is a `bool`, so its safe-context is caller-context. It follows that the resulting safe-context is the narrowest safe-context of the second and third operand. *end note*
 
-A value resulting from a method invocation `e1.M(e2, ...)` has safe-scope of the smallest of the following scopes:
+#### §safe-context-method-invocation Method and property invocation
+
+A value resulting from a method invocation `e1.M(e2, ...)` has safe-context of the smallest of the following contexts:
 
 - calling-method.
-- The safe-scope of all argument expressions (including the receiver).
+- The safe-context of all argument expressions (including the receiver).
 
 A property invocation (either `get` or `set`) is treated as a method invocation of the underlying method by the above rules.
 
-#### safe-scope-rules-stackalloc `stackalloc`
+#### §safe-context-rules-stackalloc `stackalloc`
 
-The result of a stackalloc expression has safe-scope of current-method. It is not safe-to-return from the method.
+The result of a stackalloc expression has safe-context of current-method. It is not safe-to-return from the method.
 
-#### safe-scope-rules-constructor Constructor invocations
+#### §safe-context-rules-constructor Constructor invocations
 
 A `new` expression that invokes a constructor obeys the same rules as a method invocation that is considered to return the type being constructed.
 
-In addition the safe-scope is the smallest of the safe-scopes of all arguments and operands of all object initializer expressions, recursively, if any initializer is present.
+In addition the safe-context is the smallest of the safe-contexts of all arguments and operands of all object initializer expressions, recursively, if any initializer is present.
 
 > *Note*: These rules rely on `Span<T>` not having a constructor of the following form:
 >
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -560,7 +560,7 @@ The result of a stackalloc expression has safe-scope of current-method. It is no
 
 A `new` expression that invokes a constructor obeys the same rules as a method invocation that is considered to return the type being constructed.
 
-In addition the safe-scope is no wider than the smallest of the safe-scopes of all arguments and operands of all object initializer expressions, recursively, if any initializer is present.
+In addition the safe-scope is the smallest of the safe-scopes of all arguments and operands of all object initializer expressions, recursively, if any initializer is present.
 
 > *Note*: These rules rely on `Span<T>` not having a constructor of the following form:
 >
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -506,7 +506,7 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
 
 #### safe-scope-rules-general General
 
-At compile-time, each expression whose type is a ref struct is associated with a scope where that instance is safe, its ***safe-scope***. The ***safe-scope*** is a scope, enclosing an expression, to which it is safe for the value to escape to. If that scope is the entire function member, we say that the value is ***safe-to-return*** from the function member.
+At compile-time, each expression whose type is a ref struct is associated with a scope where that instance is safe, its ***safe-scope***. The safe-scope is a scope, enclosing an expression, to which it is safe for the value to escape to. If that scope is the entire function member, we say that the value is ***safe-to-return*** from the function member.
 
 The safe-scope records which scope a ref struct may be copied into. Given an assignment from an expression `E1` with a safe-scope `S1`, to an expression `E2` with safe-scope `S2`, it is an error if `S2` is a wider scope than `S1`.
 
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -75,12 +75,12 @@ It is a compile-time error if a ref struct type is used in any of the following 
 - An async method.
 - An iterator.
 
-> *Note*: A `ref struct` shall not declare `async` methods nor use a `yield return` or `yield break` statement because the implicit `this` parameter cannot be used in those contexts. *end note*
+> *Note*: A `ref struct` shall not declare `async` instance methods nor use a `yield return` or `yield break` statement within an instance method, because the implicit `this` parameter cannot be used in those contexts. *end note*
 
 - There is no conversion from a `ref struct` type to the type `object` or the type `System.ValueType`.
-- A `ref struct` type shall not be declared to implement any interface
+- A `ref struct` type shall not be declared to implement any interface.
 - An instance method declared in `object` or in `System.ValueType` but not overridden in a `ref struct` type shall not be called with a receiver of that `ref struct` type.
-- An instance method of a `ref struct` type shall not be captured by method conversion to a delegate type.
+- An instance method of a `ref struct` type shall not be captured by method group conversion to a delegate type.
 - A ref struct shall not be captured by a lambda expression or a local function.
 
 These constraints ensure that no variable of `ref struct` type refers to stack memory or variables that are no longer valid.
@@ -506,16 +506,16 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
 
 #### safe-scope-rules-general General
 
-At compile-time, each expression whose type is a ref struct is associated with a scope where that instance is safe, its ***safe-scope***. The safe-scope is a scope, enclosing an expression, to which it is safe for the value to escape to. If that scope is the entire function member, we say that the value is ***safe-to-return*** from the function member.
+At compile-time, each expression whose type is a ref struct is associated with a scope where that instance is safe, its ***safe-scope***. The safe-scope is a scope, enclosing an expression, which it is safe for the value to escape to. If that scope is the entire function member, we say that the value is ***safe-to-return*** from the function member.
 
 The safe-scope records which scope a ref struct may be copied into. Given an assignment from an expression `E1` with a safe-scope `S1`, to an expression `E2` with safe-scope `S2`, it is an error if `S2` is a wider scope than `S1`.
 
-There are three different safe-scope values, the same as the ref-safe-scope values defined for reference variables (§ref-span-safety-escape-scopes): block, function-member, and calling-method. An expression `e1` of a ref struct type is constrained by its safe-scope as follows:
+There are three different safe-scope values, the same as the ref-safe-scope values defined for reference variables (§ref-span-safety-escape-scopes): block, function-member, and caller-scope. An expression `e1` of a ref struct type is constrained by its safe-scope as follows:
 
 - For a return statement `return e1`, the safe-scope of `e1` must be calling-method.
 - For an assignment `e1 = e2` the safe-scope of `e2` must be at least as wide a scope as the safe-scope of `e1`.
 
-For a method invocation if there is a `ref` or `out` argument of a `ref struct` type (including the receiver unless the type is `readonly`), with safe-scope `E1`, then no argument (including the receiver) may have a narrower safe-scope than `E1`.
+For a method invocation if there is a `ref` or `out` argument of a `ref struct` type (including the receiver unless the type is `readonly`), with safe-scope `S1`, then no argument (including the receiver) may have a narrower safe-scope than `S1`.
 
 Any expression whose compile-time type is not a ref struct has a safe-scope of calling-method.
 
@@ -530,12 +530,12 @@ A formal parameter of a ref struct type, including the `this` parameter of an in
 A local variable of a ref struct type has a safe-scope as follows:
 
 - If the variable is an iteration variable of a `foreach` loop, then the variable's safe-scope is the same as the safe-scope of the `foreach` loop's expression.
-- A local of `ref struct` type uninitialized at the point of declaration has a safe-scope of calling method.
+- A local of `ref struct` type which is uninitialized at the point of declaration has a safe-scope of calling method.
 - Otherwise the variable's declaration requires an initializer. The variable's safe-scope is the same as the safe-scope of its initializer.
 
 #### safe-scope-rules-field Field safe scope
 
-A a reference to a field, `e.F` where `F` is a ref struct type, has a safe-scope that is the same as the safe-scope of `e`.
+A reference to a field `e.F`, where the type of `F` is a ref struct type, has a safe-scope that is the same as the safe-scope of `e`.
 
 #### safe-scope-rules-operator Operators
 
@@ -550,7 +550,7 @@ A value resulting from a method invocation `e1.M(e2, ...)` has safe-scope of the
 - calling-method.
 - The safe-scope of all argument expressions (including the receiver).
 
-A property invocation (either `get` or `set`) it treated as a method invocation of the underlying method by the above rules.
+A property invocation (either `get` or `set`) is treated as a method invocation of the underlying method by the above rules.
 
 #### safe-scope-rules-stackalloc `stackalloc`
 
@@ -565,7 +565,7 @@ In addition the safe-scope is no wider than the smallest of the safe-scopes of a
 > *Note*: These rules rely on `Span<T>` not having a constructor of the following form:
 >
 > ```csharp
-> public Span<T>(ret T p)
+> public Span<T>(ref T p)
 > ```
 >
 > Such a constructor makes instances of `Span<T>` used as fields indistinguishable from a `ref` field. The safety rules described in this document depend on `ref` fields not being a valid construct in C# or .NET. *end note*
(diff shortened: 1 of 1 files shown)


standard/structs.md
```diff
@@ -49,6 +51,22 @@ It is a compile-time error for the same modifier to appear multiple times in a s
...
+
+It is a compile-time error if a ref struct type is used in any of the following contexts:
+
+- As the element type of an array.
+- As the declared type of a field of a class or a non-ref struct.
+- To implement an interface.
+- Being boxed to `System.ValueType` or `System.Object`.
+- As a type argument.
+- An async method.
+- An iterator.
```
KalleOlaviNiemitalo: Does "a ref struct type is used" include this case:

```C#
using System;
using System.Threading.Tasks;

class Widget
{
}

static class Class1
{
    static async Task M(Task t)
    {
        Widget w = stackalloc int[] { 42 };
        await t;
    }
}
```

.NET SDK 6.0.101 compiles it without errors. The Span\<int\> becomes a local variable in the MoveNext method of the state machine.

__Originally posted by @KalleOlaviNiemitalo in <https://github.com/dotnet/csharpstandard/pull/333#discussion_r775953793>__

gafter: Yes.  The implementation is more permissive than required.

BillWagner:
Rex had written several questions in #334  Reviewers should check that those are addressed.

standard/structs.md
KalleOlaviNiemitalo: Ah, so "It is a compile-time error" does not mean "The compiler shall issue an error".

BillWagner:
@gafter 

Do you want me to pick up this one? I'm working on updating the language for #213, #219, #238, and possibly #240 for the next meeting. I'll also be adding the sections on ref safety.

If so, I can add it to my list and get this one ready as well.

gafter:
@BillWagner Yes, please.  If you do, please try to address #267 as well.

BillWagner:
@jskeet @gafter 

I added a new clause to define the ref safety rules as they apply to ref structs. This should be ready for our next meeting.

BillWagner:
> @BillWagner Yes, please. If you do, please try to address #267 as well.

The normative language was already present. I added a note to make it more obvious.

standard/structs.md
```diff
@@ -16,18 +16,20 @@ A *struct_declaration* is a *type_declaration* ([§14.7](namespaces.md#147-type-
...
 
 A struct declaration shall not supply a *type_parameter_constraints_clauses* unless it also supplies a *type_parameter_list*.
 
 A struct declaration that supplies a *type_parameter_list* is a generic struct declaration. Additionally, any struct nested inside a generic class declaration or a generic struct declaration is itself a generic struct declaration, since type arguments for the containing type shall be supplied to create a constructed type ([§8.4](types.md#84-constructed-types)).
 
+A struct declaration that includes a `ref` keyword may not have a *struct_interfaces* part.
```
jskeet: Shall not rather than may not, for clarity? One for Rex.

standard/structs.md
```diff
@@ -60,6 +62,26 @@ A readonly struct has the following constraints:
 
 When an instance of a readonly struct is passed to a method, its `this` is treated like an `in` argument/parameter, which disallows write access to any instance fields (except by constructors).
 
+### §ref-modifier-new-clause Ref modifier
+
+The `ref` modifier indicates that the *struct_declaration* declares a type whose instances are allocated on the execution stack. The `ref` modifier declares that instances may contain ref-like fields, and must obey the *ref_safe_scope* rules. Instances of the ref struct may not be copied out of the *ref_safe_scope* of any of its ref-like fields. The rules for determining the *safe_scope* of a `ref struct` are described in §ref-like-field-initialization.
```
jskeet: This has both ref safe scope and safe scope... Needs slight rework with the recent changes?

standard/structs.md
```diff
@@ -60,6 +62,26 @@ A readonly struct has the following constraints:
...
+
+- As the element type of an array.
+- As the declared type of a field of a class or a non-ref struct.
+- To implement an interface.
+- Being boxed to `System.ValueType` or `System.Object`.
+- As a type argument.
+- An async method.
+- An iterator.
+
+> *Note*: A `ref struct` may not declare `async` methods nor use a `yield return` or `yield break` statement because the implicit `this` parameter cannot be used in those contexts. *end note*
```
jskeet: More "may not" that should be shall not, I think.

standard/structs.md
```diff
@@ -476,3 +498,19 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+
+An instance of a `ref struct` `S` requires an initializing ref expression, `ref e` for any ref like fields. The *ref_safe_scope* of `e` determines the *ref_safe_scope* for `S`.
+
+- If `e` is a local variable, the *ref_safe_scope* of `S` is *block*.
+- If `e` is a formal parameter, the *ref_safe_scope* of `S` is *calling_method*. This includes when `e` is the `this` parameter.
+- If `e` designates a reference to a field, `e1.F`, the *ref_safe_scope* of `S` is the same as the *ref_safe_scope* of `e1`.
+- If `e` designates a reference to a conditional, `ref c ? ref e1 : ref e2`, the *ref_safe_scope* of `S` is the narrowest scope among the *ref_safe_scopes* of `e1` and `e2`.
+- If `e` is the result of a method invocation, `e1.M(e2, ...)`, the *ref_safe_scope* of `S` is the narrowest of the *ref_safe_scope*s of all argument expressions, including the receiver.
+
+For an assignment `e1 = e2`, if the type of `e1` is a `ref struct` type, then the *ref_safe_scope* of `e1` is the *ref_safe_scope* of `e2`.
```
jskeet: Not sure how assignment matters here. Would like to hear more in the meeting.

jskeet (approved):
I think this is basically fine, with a couple of notes. I may well be missing subtleties though.

standard/structs.md
```diff
@@ -476,3 +498,19 @@ Static constructors for structs follow most of the same rules as for classes. Th
 Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-implemented-properties)) use hidden backing fields, which are only accessible to the property accessors.
 
 > *Note*: This access restriction means that constructors in structs containing automatically implemented properties often need an explicit constructor initializer where they would not otherwise need one, to satisfy the requirement of all fields being definitely assigned before any function member is invoked or the constructor returns. *end note*
+
+### §ref-like-field-initialization Ref safe scopes
+
+A `ref struct` may contain *ref_like field*s. A *ref_like field* is a *reference_variable* (§ref-span-safety). An instance of a `ref struct` may not be copied outside the *ref_safe_scope* (§ref-span-safety-escape-scopes) of any of its ref-like fields. The *ref_safe_scope* of a ref struct is determined from its initializing expression.
+
+The default value of ref-like fields is the value of the `default` expression, a null reference. The *ref_safe_scope* of a null reference is *caller_scope*. A `ref struct` initialized to its default value can be copied to the *caller_scope*.
```
Nigel-Ecma: “The default value of ref-like fields is the value of the `default` expression, a null reference.” does not read well. Is “…, which is a null reference" the intention? Or is it “The default value of ref-like fields, and the value of the `default` expression, is a null reference.”? Should there be an xref to the default literal/operator? Sorry, only questions not answers!

Nigel-Ecma (changes requested):

standard/structs.md
BillWagner: Pint @RexJaeschke again.

standard/structs.md
BillWagner: Using a modified version of another example:

```csharp
    ReadOnlySpan<char> span1;
    ReadOnlySpan<char> span3;
    if (x is not null)
    {
        ReadOnlySpan<char> span2 = x.AsSpan(); // span2 ref_safe_scope is *method*.
        span1 = span2;     // span1 ref_safe_scope is *method* (same as span2)
        span3 = span1; // span3 ref_safe_scope is method (same as span1)
    }
}
```

standard/structs.md
jskeet: Action from 2023-04-19: let's use shall not or "it is an error to" etc.

standard/structs.md
```diff
@@ -60,6 +62,26 @@ A readonly struct has the following constraints:
...
+### §ref-modifier-new-clause Ref modifier
+
+The `ref` modifier indicates that the *struct_declaration* declares a type whose instances are allocated on the execution stack. The `ref` modifier declares that instances may contain ref-like fields, and must obey the *ref_safe_scope* rules. Instances of the ref struct may not be copied out of the *ref_safe_scope* of any of its ref-like fields. The rules for determining the *ref_safe_scope* of a `ref struct` are described in §ref-like-field-initialization.
+
+It is a compile-time error if a ref struct type is used in any of the following contexts:
+
+- As the element type of an array.
+- As the declared type of a field of a class or a non-ref struct.
+- To implement an interface.
+- Being boxed to `System.ValueType` or `System.Object`.
```
BillWagner: Should this include `dynamic` as well?

standard/structs.md
jskeet: Action from 2023-04-19: we should just keep a note of this in the non-conformance doc (which is not yet merged/decided on). No change required to this PR.

standard/structs.md
jskeet: Action from 2023-04-19: As before, @BillWagner to change to anything unambiguous.

standard/structs.md
jskeet: Discussion in meeting: a *variable* `v` has a *ref_safe_scope*, and if `v` can be assigned variable references which have different *ref_safe_scopes*, it would be odd for `v`'s *ref_safe_scope* to change.

As per https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.2/span-safety.md#language-constraints, the assignment part should be a constraint on what *can* be assigned, but a determination of the *ref_safe_scope* of the LHS.

jskeet: Also noted from @gafter: this would be safe_to_escape not ref_safe_to_escape in the previous way of explaining things.

Action 2023-04-19: @BillWagner to consider - and ping others where appropriate.

gafter: These should be language constraints, as per https://github.com/dotnet/csharplang/blob/main/proposals/csharp-7.2/span-safety.md#language-constraints, not inference rules.

jskeet (commented):
Result from meeting: this is definitely not right at the moment, but we're not sure how hard it will be to fix. @BillWagner has the lead.

standard/structs.md
BillWagner: See dotnet/docs#35096 to capture this issue.

standard/structs.md
BillWagner: Update per meeting notes.

RexJaeschke:
@BillWagner In #### safe-scope-rules-general General

> At compile-time, each expression whose type is a ref struct is associated with a scope where that instance is safe, its ***safe-scope***. The ***safe-scope*** is a scope, enclosing an expression, to which it is safe for the value to escape to. If that scope is the entire function member, we say that the value is ***safe-to-return*** from the function member.

the term ***safe-scope*** is defined twice w.r.t term formatting. Remove the formatting from the second occurrence.

BillWagner:
> @BillWagner In #### safe-scope-rules-general General
> 
> > At compile-time, each expression whose type is a ref struct is associated with a scope where that instance is safe, its _**safe-scope**_. The _**safe-scope**_ is a scope, enclosing an expression, to which it is safe for the value to escape to. If that scope is the entire function member, we say that the value is _**safe-to-return**_ from the function member.
> 
> the term _**safe-scope**_ is defined twice w.r.t term formatting. Remove the formatting from the second occurrence.

Thanks Rex. Updated.

standard/structs.md
```diff
@@ -60,6 +62,29 @@ A readonly struct has the following constraints:
...
+It is a compile-time error if a ref struct type is used in any of the following contexts:
+
+- As the element type of an array.
+- As the declared type of a field of a class or a non-ref struct.
+- Being boxed to `System.ValueType` or `System.Object`:
+- As a type argument.
+- An async method.
+- An iterator.
+
+> *Note*: A `ref struct` shall not declare `async` methods nor use a `yield return` or `yield break` statement because the implicit `this` parameter cannot be used in those contexts. *end note*
```
jskeet:
```suggestion
> *Note*: A `ref struct` shall not declare `async` instance methods nor use a `yield return` or `yield break` statement within an instance method, because the implicit `this` parameter cannot be used in those contexts. *end note*
```

(Or something similar - basically, static iterator blocks and static async methods are fine.)

standard/structs.md
```diff
@@ -60,6 +62,29 @@ A readonly struct has the following constraints:
...
+
+- There is no conversion from a `ref struct` type to the type `object` or the type `System.ValueType`.
+- A `ref struct` type shall not be declared to implement any interface
```
jskeet:
```suggestion
- A `ref struct` type shall not be declared to implement any interface.
```

standard/structs.md
```diff
@@ -60,6 +62,29 @@ A readonly struct has the following constraints:
...
+- An instance method declared in `object` or in `System.ValueType` but not overridden in a `ref struct` type shall not be called with a receiver of that `ref struct` type.
+- An instance method of a `ref struct` type shall not be captured by method conversion to a delegate type.
```
jskeet:
```suggestion
- An instance method of a `ref struct` type shall not be captured by method group conversion to a delegate type.
```

(Maybe?)

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
 Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-implemented-properties)) use hidden backing fields, which are only accessible to the property accessors.
 
 > *Note*: This access restriction means that constructors in structs containing automatically implemented properties often need an explicit constructor initializer where they would not otherwise need one, to satisfy the requirement of all fields being definitely assigned before any function member is invoked or the constructor returns. *end note*
+
+### §safe-scope-rules Safe scope constraint for `ref struct` types
+
+#### safe-scope-rules-general General
+
+At compile-time, each expression whose type is a ref struct is associated with a scope where that instance is safe, its ***safe-scope***. The safe-scope is a scope, enclosing an expression, to which it is safe for the value to escape to. If that scope is the entire function member, we say that the value is ***safe-to-return*** from the function member.
```
jskeet: Not sure what "where that instance is safe" means. Safe to use? Perhaps "where that instance can be used safely"?

jskeet:
```suggestion
At compile-time, each expression whose type is a ref struct is associated with a scope where that instance is safe, its ***safe-scope***. The safe-scope is a scope, enclosing an expression, which it is safe for the value to escape to. If that scope is the entire function member, we say that the value is ***safe-to-return*** from the function member.
```

(Or remove the "to" after "escape" - but we don't need or want both.)

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+
+The safe-scope records which scope a ref struct may be copied into. Given an assignment from an expression `E1` with a safe-scope `S1`, to an expression `E2` with safe-scope `S2`, it is an error if `S2` is a wider scope than `S1`.
+
+There are three different safe-scope values, the same as the ref-safe-scope values defined for reference variables (§ref-span-safety-escape-scopes): block, function-member, and calling-method. An expression `e1` of a ref struct type is constrained by its safe-scope as follows:
```
jskeet:
```suggestion
There are three different safe-scope values, the same as the ref-safe-scope values defined for reference variables (§ref-span-safety-escape-scopes): block, function-member, and caller-scope. An expression `e1` of a ref struct type is constrained by its safe-scope as follows:
```

(And the same below.)

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+
+- For a return statement `return e1`, the safe-scope of `e1` must be calling-method.
+- For an assignment `e1 = e2` the safe-scope of `e2` must be at least as wide a scope as the safe-scope of `e1`.
+
+For a method invocation if there is a `ref` or `out` argument of a `ref struct` type (including the receiver unless the type is `readonly`), with safe-scope `E1`, then no argument (including the receiver) may have a narrower safe-scope than `E1`.
```
jskeet:
```suggestion
For a method invocation if there is a `ref` or `out` argument of a `ref struct` type (including the receiver unless the type is `readonly`), with safe-scope `S1`, then no argument (including the receiver) may have a narrower safe-scope than `S1`.
```

(Just because we normally use E for the expression, and S for its scope.)

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+#### safe-scope-rules-parameter Parameter safe scope
+
+A formal parameter of a ref struct type, including the `this` parameter of an instance method, has a safe-scope of calling-method.
+
+#### safe-scope-rules-local Local variable safe scope
+
+A local variable of a ref struct type has a safe-scope as follows:
+
+- If the variable is an iteration variable of a `foreach` loop, then the variable's safe-scope is the same as the safe-scope of the `foreach` loop's expression.
+- A local of `ref struct` type uninitialized at the point of declaration has a safe-scope of calling method.
```
jskeet:
```suggestion
- A local of `ref struct` type which is uninitialized at the point of declaration has a safe-scope of calling method.
```

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+- Otherwise the variable's declaration requires an initializer. The variable's safe-scope is the same as the safe-scope of its initializer.
+
+#### safe-scope-rules-field Field safe scope
+
+A a reference to a field, `e.F` where `F` is a ref struct type, has a safe-scope that is the same as the safe-scope of `e`.
```
jskeet:
```suggestion
A reference to a field `e.F`, where the type of `F` is a ref struct type, has a safe-scope that is the same as the safe-scope of `e`.
```

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+
+#### safe-scope-rules-operator Operators
+
+The application of a user-defined operator is treated as a method invocation (§safe-scope-method-invocation).
+
+For an operator that yields a value, such as `e1 + e2` or `c ? e1 : e2`, the safe-scope of the result is the narrowest scope among the safe-scopes of the operands of the operator. As a consequence, for a unary operator that yields a value, such as `+e`, the safe-scope of the result is the safe-scope of the operand.
```
jskeet: Is that the case for the conditional operator? I wouldn't expect the first operand to be relevant there, but I could be wrong.

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+
+#### §safe-scope-method-invocation Method and property invocation
+
+A value resulting from a method invocation `e1.M(e2, ...)` has safe-scope of the smallest of the following scopes:
+
+- calling-method.
+- The safe-scope of all argument expressions (including the receiver).
+
+A property invocation (either `get` or `set`) it treated as a method invocation of the underlying method by the above rules.
```
jskeet:
```suggestion
A property invocation (either `get` or `set`) is treated as a method invocation of the underlying method by the above rules.
```

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+#### safe-scope-rules-constructor Constructor invocations
+
+A `new` expression that invokes a constructor obeys the same rules as a method invocation that is considered to return the type being constructed.
+
+In addition the safe-scope is no wider than the smallest of the safe-scopes of all arguments and operands of all object initializer expressions, recursively, if any initializer is present.
+
+> *Note*: These rules rely on `Span<T>` not having a constructor of the following form:
+>
+> ```csharp
+> public Span<T>(ret T p)
```
jskeet:
```suggestion
> public Span<T>(ref T p)
```

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+
+The result of a stackalloc expression has safe-scope of current-method. It is not safe-to-return from the method.
+
+#### safe-scope-rules-constructor Constructor invocations
+
+A `new` expression that invokes a constructor obeys the same rules as a method invocation that is considered to return the type being constructed.
+
+In addition the safe-scope is no wider than the smallest of the safe-scopes of all arguments and operands of all object initializer expressions, recursively, if any initializer is present.
+
+> *Note*: These rules rely on `Span<T>` not having a constructor of the following form:
```
jskeet, line 567: This is an odd note, being Span-specific. What happens if someone creates a `MySpan<T>` type with a constructor like that?

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+
+#### safe-scope-rules-stackalloc `stackalloc`
+
+The result of a stackalloc expression has safe-scope of current-method. It is not safe-to-return from the method.
+
+#### safe-scope-rules-constructor Constructor invocations
+
+A `new` expression that invokes a constructor obeys the same rules as a method invocation that is considered to return the type being constructed.
+
+In addition the safe-scope is no wider than the smallest of the safe-scopes of all arguments and operands of all object initializer expressions, recursively, if any initializer is present.
```
jskeet: Can we be more specific than "no wider", e.g. just "is the smallest of the safe-scopes [...]"?

jskeet (commented):
Most of these are trivial.

standard/structs.md
BillWagner: it doesn't matter, but it doesn't hurt either. The first operand must be a `bool`. Its safe scope is calling-method.  I couldn't wordsmith language that read reasonably well. Perhaps a note on the conditional operator's first operand?

standard/structs.md
BillWagner: We have to cross that bridge in C# 11. Before that, `MySpan<T>` can't have a `ref` field.

standard/structs.md
```diff
@@ -60,6 +62,29 @@ A readonly struct has the following constraints:
...
+
+> *Note*: A `ref struct` shall not declare `async` instance methods nor use a `yield return` or `yield break` statement within an instance method, because the implicit `this` parameter cannot be used in those contexts. *end note*
+
+- There is no conversion from a `ref struct` type to the type `object` or the type `System.ValueType`.
+- A `ref struct` type shall not be declared to implement any interface.
+- An instance method declared in `object` or in `System.ValueType` but not overridden in a `ref struct` type shall not be called with a receiver of that `ref struct` type.
+- An instance method of a `ref struct` type shall not be captured by method group conversion to a delegate type.
+- A ref struct shall not be captured by a lambda expression or a local function.
+
+These constraints ensure that no variable of `ref struct` type refers to stack memory or variables that are no longer valid.
```
Nigel-Ecma: I think this sentence could be parsed (and I think I read it was the wrong one):

> These constraints ensure that no variable of `ref struct` type refers to (stack memory) or (variables that are no longer valid).

or as:

> These constraints ensure that no variable of `ref struct` type refers to (stack memory or variables) that are no longer valid.

Suggest clarify.

standard/structs.md
```diff
@@ -60,6 +62,29 @@ A readonly struct has the following constraints:
 
 When an instance of a readonly struct is passed to a method, its `this` is treated like an `in` argument/parameter, which disallows write access to any instance fields (except by constructors).
 
+### §ref-modifier-new-clause Ref modifier
+
+The `ref` modifier indicates that the *struct_declaration* declares a type whose instances are allocated on the execution stack. The `ref` modifier declares that instances may contain ref-like fields, and may not be copied out of its safe-scope (§safe-scope-rules). The rules for determining the safe scope of a `ref struct` are described in §safe-scope-rules.
```
Nigel-Ecma: "execution stack", "stack memory" (used below) – need to address issued raised by @RexJaeschke in #334.

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+
+A formal parameter of a ref struct type, including the `this` parameter of an instance method, has a safe-scope of calling-method.
+
+#### safe-scope-rules-local Local variable safe scope
+
+A local variable of a ref struct type has a safe-scope as follows:
+
+- If the variable is an iteration variable of a `foreach` loop, then the variable's safe-scope is the same as the safe-scope of the `foreach` loop's expression.
+- A local of `ref struct` type which is uninitialized at the point of declaration has a safe-scope of calling method.
+- Otherwise the variable's declaration requires an initializer. The variable's safe-scope is the same as the safe-scope of its initializer.
```
Nigel-Ecma: There is a bit of repetition here and unusual phrasing, maybe try:

```suggestion
A local variable of a ref struct type has a safe-scope as follows:

- If the variable is an iteration variable of a `foreach` loop, then the variable's safe-scope is the same as the safe-scope of the `foreach` loop's expression.
- Otherwise if the variable's declaration has an initializer then the variable's safe-scope is the same as the safe-scope of that initializer.

```

Note: I omitted that the final option means the variable is not initialised, if that is desired the final bullet can be:

```
- Otherwise the variable is uninitialized at the point of declaration and has a safe-scope of the calling method.
```

standard/structs.md
Nigel-Ecma: It is the safe-scope *of the result* so the first operand (`c`) is not relevant *per se* as the safe-scope is determined statically and not based on which alternative is selected at runtime. I wouldn't wordsmith this into the normative text, but a note could be added if desired.

Nigel-Ecma (changes requested):

standard/structs.md
BillWagner: updated in next commit.

standard/structs.md
BillWagner: #238 includes the changes that allow `stackalloc` in safe code. The first sentence of that new clause reads:

> A stack allocation initializer is a *local_variable_initializer* (§9.2.8) that allocates a block of memory from the call stack.

Ping @RexJaeschke Where should we add this definition? I'll add it to either this PR or #238

MadsTorgersen: Should "ref struct" be introduced as a term?

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
 Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-implemented-properties)) use hidden backing fields, which are only accessible to the property accessors.
 
 > *Note*: This access restriction means that constructors in structs containing automatically implemented properties often need an explicit constructor initializer where they would not otherwise need one, to satisfy the requirement of all fields being definitely assigned before any function member is invoked or the constructor returns. *end note*
+
+### §safe-scope-rules Safe scope constraint for `ref struct` types
```
BillWagner: Here, as in #742, change from "scope" to "context".

standard/structs.md
```diff
@@ -476,3 +501,71 @@ Static constructors for structs follow most of the same rules as for classes. Th
...
+
+#### safe-scope-rules-general General
+
+At compile-time, each expression whose type is a ref struct is associated with a scope where that instance is safe, its ***safe-scope***. The safe-scope is a scope, enclosing an expression, which it is safe for the value to escape to. If that scope is the entire function member, we say that the value is ***safe-to-return*** from the function member.
+
+The safe-scope records which scope a ref struct may be copied into. Given an assignment from an expression `E1` with a safe-scope `S1`, to an expression `E2` with safe-scope `S2`, it is an error if `S2` is a wider scope than `S1`.
+
+There are three different safe-scope values, the same as the ref-safe-scope values defined for reference variables (§ref-span-safety-escape-scopes): block, function-member, and caller-scope. An expression `e1` of a ref struct type is constrained by its safe-scope as follows:
```
BillWagner: Associate *safe-to-return* with *caller-scope*.

standard/structs.md
BillWagner: Decision to add in #238.  See https://github.com/dotnet/csharpstandard/pull/238/files#r1197026105

standard/structs.md
BillWagner: I added a note, as this has come up multiple times.

PR #604, opened 2022-08-04
remove prohibition of unmanaged constructed types

RexJaeschke:


BillWagner:
rebased on the latest draft v8 branch on 9/26/2023

BillWagner (approved):
This LGTM @RexJaeschke 

I'm happy to merge at the next meeting.

jskeet:
The title talks about unmanaged constructed types - the text seems to be about pointer types. (It looks like constructed types still count as managed.)
I think I need a bit more context to understand this change.

standard/types.md
```diff
@@ -701,3 +701,4 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
 - `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`.
 - Any *enum_type*.
 - Any user-defined *struct_type* that is not a constructed type and contains instance fields of *unmanaged_type*s only.
+- In unsafe code ([§23.2](unsafe-code.md#232-unsafe-contexts)), any *pointer_type* ([§23.3](unsafe-code.md#233-pointer-types)).
```
BillWagner: Suggested update (Assuming the type parameter must have the unmanaged constraint):

```suggestion
- Any user-defined *struct_type* that contains instance fields of *unmanaged_type*s or where all arguments for all type parameters in a closed constructed types are of *unmanaged_type*s only.
- In unsafe code ([§23.2](unsafe-code.md#232-unsafe-contexts)), any *pointer_type* ([§23.3](unsafe-code.md#233-pointer-types)).
```

BillWagner (commented):
Added one possible suggestion.

standard/types.md
Nigel-Ecma: It is not clear that the `unmanaged` constraint is required, just that the provided type is unmanaged.

Have fun with the below:

- With the `where` clause the current VS set to 8.0
  - rejects the 3rd & 4th `CheckUnmanaged()` calls at compile time – they are not unmanaged types
  - accepts the second even though there is no `unmanaged` constraint on `Coords<>`
- However take out the `where` and it runs and reports only one of the types is managed, which is a little worrying! But then I copied the test code from StackOverflow…

```
using System;
using System.Reflection;

namespace UnmanagedConstructed
{
    public struct Coords<T>
    {
        public T X;
        public T Y;
    }

    public static class UnmanagedTypeExtensions
    {
        // U and IsUnmanaged copied from StackOverflow...
        class U<T> where T : unmanaged { }
        public static bool IsUnmanaged(this Type t)
        {
            try { typeof(U<>).MakeGenericType(t); return true; }
            catch (Exception) { return false; }
        }
    }

    public class UnmanagedTypes
    {
        public static void Main()
        {
            CheckUnmanaged<int>();
            CheckUnmanaged<Coords<int>>();
            CheckUnmanaged<object>();           // invalid if below where clause is uncommented
            CheckUnmanaged<Coords<object>>();   // ditto
        }

        private static void CheckUnmanaged<T>() // where T : unmanaged
        {
            Console.WriteLine($"{typeof(T)} is {(typeof(T).IsUnmanaged() ? "unmanaged" : "managed")}");
        }
    }
}
```

Nigel-Ecma (changes requested):
One comment added in context as its in the range of the changes.

Lines 515, 617 & 699 probably also need to be changed. Also need to consider the *unmanaged_type* grammar (line 693), should it now allow a type parameter with the unmanaged constraint?

standard/types.md
jskeet: After discussion, we agree with Nigel's analysis, and think we don't need to refer to type parameters, but we want an example (potentially like Nigel's, but simpler - just the method with an `unmanaged` constraint).

jskeet:
(@jskeet to have another pass at this - in this PR.)

jskeet:
And look for other places, e.g. 515.

jskeet:
Also (try to) remove the grammar as type parameters can be unmanaged - instead, make references to it specify that the type must be unmanaged. (Do this in a separate commit.)

standard/types.md
BillWagner: @jskeet Is someone assigned to create the example? (I can't remember from the last meeting)

jskeet: @BillWagner: Me - the example along with the other changes to this PR. If I can just find the time...

jskeet: I've been having another look at this, and trying to get my head back into where we were in May... and I've failed. I'll mark this as "meeting discuss" again to see if we can get it right together *during* the next meeting.

I think there's some subtlety around type parameters though - because a type parameter with no constraints won't count as as unmanaged, but a type parameter with the unmanaged constraint will:

```csharp
class Program
{
    static void Main()
    {
        Check1<int>();
        Check2<int>();
    }

    static void Check1<T>()
    {
        CheckUnmanaged<T>(); // Error, because T *might* not be unmanaged
    }

    static void Check2<T>() where T : unmanaged
    {
        CheckUnmanaged<T>();
    }

    static void CheckUnmanaged<T>() where T : unmanaged
    {
    }
}
```

jskeet:
We should talk about this again - I'm finding it hard to get my head back into it, so when we talk, let's make verbose notes.

jskeet:
The latest merge has undone all the changes here... suggest we revert the last commit.

BillWagner:
OK, I went through the merge, tried to revert and that failed to make any sensible changes (the only change left was the revert of the links for the "pointer" restriction.

Tried a rebase as well, and the same issue remained. 

I've tried looking at a reflog and I can't find the actual changes there either.

BillWagner:
Here's the original proposal as an issue in C# lang:  dotnet/csharplang#1504

And the championed issue is here: dotnet/csharplang#1744

Finally, this issue contains some notes and details: dotnet/csharplang#1937

standard/types.md
```diff
@@ -715,11 +715,11 @@ unmanaged_type
     ;
 ```
 
-An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_parameter*, or a constructed type, and contains no instance fields whose type is not an *unmanaged_type*. In other words, an *unmanaged_type* is one of the following:
+An *unmanaged_type* is any type that isn’t a *reference_type*, or a *type_parameter*, and contains no instance fields whose type is not an *unmanaged_type*. In other words, an *unmanaged_type* is one of the following:
```
```suggestion
An *unmanaged_type* is any type that is neither a *reference_type* nor a *type_parameter* that is not constrained to be unmanaged, and contains no instance fields whose type is not an *unmanaged_type*. In other words, an *unmanaged_type* is one of the following:
```

jskeet: Extra example to consider:

```csharp
Foo<T1,T2>
    where T1 : T2
    where T2 : unmanaged
```

jskeet: That gives:

> error CS8379: Type parameter 'T2' has the 'unmanaged' constraint so 'T2' cannot be used as a constraint for 'T1'

jskeet: (Changed the suggestion to avoid using the `unmanaged` constraint.)

standard/types.md
jskeet: > However take out the where and it runs and reports only one of the types is managed, which is a little worrying! But then I copied the test code from StackOverflow…

We believe that `Coords<object>` being reported as unmanaged is just a runtime bug - it really *is* a managed type, and so this is incorrect. But out of the scope of the spec.

standard/types.md
```diff
@@ -715,11 +715,11 @@ unmanaged_type
...
 
 - `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`.
 - Any *enum_type*.
-- Any user-defined *struct_type* that is not a constructed type and contains instance fields of *unmanaged_type*s only.
+- Any user-defined *struct_type* that contains instance fields of *unmanaged_type*s only.
```
jskeet, line 722:
```suggestion
- Any user-defined *struct_type* that contains instance fields of *unmanaged_type*s only, provided there are no type parameters that are not constrained to be unmanaged.
```

standard/types.md
```diff
@@ -715,11 +715,11 @@ unmanaged_type
...
 - In unsafe code ([§23.2](unsafe-code.md#232-unsafe-contexts)), any *pointer_type* ([§23.3](unsafe-code.md#233-pointer-types)).
```
```suggestion
- Any type parameter which is constrained to be unmanaged.
- Any *pointer_type* ([§23.3](unsafe-code.md#233-pointer-types)).
```

MadsTorgersen (approved):

gafter (approved):

jnm2 (approved):

Nigel-Ecma (approved):

standard/types.md
jnm2: The `IsUnmanaged` runtime check using `MakeGenericType` in the sample above does not match the compiler's behavior, and the runtime decided this was by design (https://github.com/dotnet/runtime/issues/12420#issuecomment-479983685) citing the original 7.3 proposal, which is quoted there as saying:

> The unmanaged constraint is not enforced by CLR, only by the language.

jnm2: Ah. Closing the loop on this, the reason that one fails and not the other has nothing to do with `unmanaged`. It's just the runtime checking the implicit `struct` constraint that gets emitted when `unmanaged` is used. The runtime totally ignores `unmanaged` as a concept.

PR #606, opened 2022-08-08
Add support for Async Streams

RexJaeschke:
There is some overlap between this PR and PR #672 in 13.14, "The using statement" that will require reconciliation when the second of these two PRs is merged. (PR 606 was spec'd first.)


standard/classes.md
```diff
@@ -5087,3 +5089,43 @@ When the body of the async function terminates, the return task is moved out of
...
+>     {
+>         await Task.Delay(100);
+>         yield return i;
+>     }
+> }
+> ```
+>
+> *end example*
+
+Consumption of an asynchronous stream requires an `await foreach` to enumerate the elements of the stream. Adding the `await` keyword requires the method that enumerates the asynchronous stream to be declared with the `async` modifier and to return a type allowed for an async method.
```
KalleOlaviNiemitalo: "requires" seems too strong; it is certainly possible to enumerate the items of an asynchronous stream by calling IAsyncEnumerator\<T\>.MoveNextAsync explicitly, without `await foreach`. I thought perhaps "consumption" is a term that means only `foreach` loops, but I didn't find the term used with such a meaning anywhere else in this repo.

standard/standard-library.md
```diff
@@ -373,12 +373,39 @@ The following types, including the members listed, must be defined in a conformi
 
 A conforming implementation may provide `Task.GetAwaiter()` and `Task<T>.GetAwaiter()` as extension methods.
 
+> **Note to TG2 reviewers:** The .NET library actually declares `GetAsyncEnumerator` to take one argument, as follows:
+
+```csharp
+IAsyncEnumerator<T> GetAsyncEnumerator(System.Threading.CancellationToken cancellationToken = default);
+```
+
+> However, as we have not mentioned type `System.Threading.CancellationToken` in any other threading contexts in the spec, it has been omitted from the requirements, as if there was no argument expected.
```
KalleOlaviNiemitalo: In my opinion, `System.Threading.CancellationToken` and `System.Runtime.CompilerServices.EnumeratorCancellationAttribute` should be added to the standard.

standard/standard-library.md
```diff
@@ -373,12 +373,39 @@ The following types, including the members listed, must be defined in a conformi
...
+
 ```csharp
 namespace System
 {
+    public interface IAsyncDisposable
+    {
+        ValueTask DisposeAsync();
```
KalleOlaviNiemitalo, line 387: `ValueTask` and `ValueTask<TResult>` are not yet in the standard.

standard/statements.md
```diff
@@ -1488,6 +1519,8 @@ While a mutual-exclusion lock is held, code executing in the same execution thre
 
 ## 12.14 The using statement
 
+(Almost all of the text in this subclause was written prior to the addition of async streams; the text applies to async streams as well, unless stated otherwise, usually in parenthetic text. Asynchronous stream creation and usage is discussed in §asynchronous-streams.)
+
 The `using` statement obtains one or more resources, executes a statement, and then disposes of the resource.
 
 ```ANTLR
```
KalleOlaviNiemitalo: The _using_statement_ grammar needs to allow `await using`.

standard/statements.md
```diff
@@ -1501,13 +1534,13 @@ resource_acquisition
...
+A `using` statement is translated into three parts: acquisition, usage, and disposal. Usage of the resource is implicitly enclosed in a `try` statement that includes a `finally` clause. This `finally` clause disposes of the resource. If a `null` resource is acquired, then no call to `Dispose` (`DisposeAsync` for async streams) is made, and no exception is thrown. If the resource is of type `dynamic` it is dynamically converted through an implicit dynamic conversion ([§10.2.10](conversions.md#10210-implicit-dynamic-conversions)) to `IDisposable` (`IAsyncDisposable` for async streams) during acquisition in order to ensure that the conversion is successful before the usage and disposal.
 
 A `using` statement of the form
```
KalleOlaviNiemitalo, line 1900: Needs similar lowering rules for `await using`, particularly that the result of `DisposeAsync` is awaited rather than discarded.

KalleOlaviNiemitalo:
A comparison that doesn't include so many unrelated changes: <https://github.com/dotnet/csharpstandard/compare/draft-v7...RexJaeschke:csharpstandard:async-streams>

RexJaeschke:
This PR needs two files to be added to additional-files in the tools folder (which does not yet exist here until draft-v8 has been rebased): Support1AsyncStreams.cs and Support1AsyncStreams.cs.

```csharp
// Support1AsyncStreams.cs 
partial class Class1
{
    static async Task Main()
    {
        await M();
        await Task.Delay(2000);
    }

    static async Task M()
    {
        await foreach (var number in GenerateSequence())
        {
            Console.WriteLine(number);
        }
    }
}

// Support2AsyncStreams.cs 
using System.Threading.Tasks;

partial class Class1
{
    static async Task Main()
    {
        await M();
        await Task.Delay(2000);
    }

    static async IAsyncEnumerable<int> GenerateSequence()
    {
        for (int i = 0; i <= 5; i++)
        {
            await Task.Delay(100);
            yield return i;
        }
    }
}
```

standard/classes.md
```diff
@@ -5125,11 +5127,11 @@ An enumerator object’s `Current` property is affected by `yield return` statem
 
 When an enumerator object is in the **suspended** state, the value of `Current` is the value set by the previous call to `MoveNext`. When an enumerator object is in the **before**, **running**, or **after** states, the result of accessing `Current` is unspecified.
 
-For an iterator with a yield type other than `object`, the result of accessing `Current` through the enumerator object’s `IEnumerable` implementation corresponds to accessing `Current` through the enumerator object’s `IEnumerator<T>` implementation and casting the result to `object`.
+For an iterator with a yield type other than `object`, the result of accessing `Current` through the enumerator object’s `IEnumerable` implementation corresponds to accessing `Current` through the enumerator object’s `IEnumerator<T>` (`IAsyncEnumerator<T>` for async streams) implementation and casting the result to `object`.
 
 #### 14.14.5.4 The Dispose method
 
-The `Dispose` method is used to clean up the iteration by bringing the enumerator object to the **after** state.
+The `Dispose` (`DisposeAync` for async streams) method is used to clean up the iteration by bringing the enumerator object to the **after** state.
```
KalleOlaviNiemitalo: typo, DisposeA<ins>s</ins>ync

standard/standard-library.md
KalleOlaviNiemitalo: If the C# standard defines IAsyncEnumerable\<T\> with `IAsyncEnumerator<T> GetAsyncEnumerator();`, and a user-defined class implements IAsyncEnumerable\<T\> by explicit interface implementation:

```csharp
using System;
class Class1 : IAsyncEnumerable<int> {
    IAsyncEnumerator<int> IAsyncEnumerable<int>.GetAsyncEnumerator() {
        throw new NotImplementedException();
    }
}
```

then Class1 will not be portable between the standard and the .NET implementation, as the method signature will not match.

standard/standard-library.md
KalleOlaviNiemitalo: They were added to the C# 7.x draft in <https://github.com/dotnet/csharpstandard/pull/556>.

BillWagner:
rebased on the latest draft v8 branch on 09/23/2023

standard/classes.md
```diff
@@ -5205,6 +5205,8 @@ For a discussion of the behavior when an exception is thrown from a finalizer, s
 
 ### 15.14.1 General
 
+(Almost all of the text in this subclause and its siblings was written prior to the addition of async streams; the text applies to async streams as well, unless stated otherwise, usually in parenthetic text. Asynchronous stream creation and usage is discussed in §asynchronous-streams.)
```
jskeet: Is the intention to keep this part? It looks a little odd at the moment. I suspect it should at least be a note instead of just in parentheses.

standard/classes.md
```diff
@@ -5213,41 +5215,41 @@ When a function member is implemented using an iterator block, it is a compile-t
 
 ### 15.14.2 Enumerator interfaces
 
-The ***enumerator interfaces*** are the non-generic interface `System.Collections.IEnumerator` and all instantiations of the generic interface `System.Collections.Generic.IEnumerator<T>`. For the sake of brevity, in this subclause and its siblings these interfaces are referenced as `IEnumerator` and `IEnumerator<T>`, respectively.
+The ***enumerator interfaces*** are the non-generic interface `System.Collections.IEnumerator` and all instantiations of the generic interface `System.Collections.Generic.IEnumerator<T>` (`System.Collections.Generic.IAsyncEnumerator<T>` for async streams). For the sake of brevity, in this subclause and its siblings these interfaces are referenced as `IEnumerator` and `IEnumerator<T>` (`IAsyncEnumerator<T>` for async streams), respectively.
```
jskeet: I'm personally finding this style really hard to read. I'd be happy to try to get this PR into a shape we want, merge it, and then I can create a new PR to change the style - assuming others feel the same way.

standard/classes.md
```diff
@@ -5286,11 +5288,11 @@ An enumerator object’s `Current` property is affected by `yield return` statem
 
 When an enumerator object is in the **suspended** state, the value of `Current` is the value set by the previous call to `MoveNext`. When an enumerator object is in the **before**, **running**, or **after** states, the result of accessing `Current` is unspecified.
 
-For an iterator with a yield type other than `object`, the result of accessing `Current` through the enumerator object’s `IEnumerable` implementation corresponds to accessing `Current` through the enumerator object’s `IEnumerator<T>` implementation and casting the result to `object`.
+For an iterator with a yield type other than `object`, the result of accessing `Current` through the enumerator object’s `IEnumerable` implementation corresponds to accessing `Current` through the enumerator object’s `IEnumerator<T>` (`IAsyncEnumerator<T>` for async streams) implementation and casting the result to `object`.
 
 #### 15.14.5.4 The Dispose method
 
-The `Dispose` method is used to clean up the iteration by bringing the enumerator object to the **after** state.
+The `Dispose` (`DisposeAync` for async streams) method is used to clean up the iteration by bringing the enumerator object to the **after** state.
```
```suggestion
The `Dispose` (`DisposeAsync` for async streams) method is used to clean up the iteration by bringing the enumerator object to the **after** state.
```

standard/classes.md
jskeet: Added as a suggestion, thanks.

BillWagner: adding suggestion:

```suggestion
The `Dispose` (`DisposeAsync` for async streams) method is used to clean up the iteration by bringing the enumerator object to the **after** state.
```

standard/statements.md
```diff
@@ -1109,25 +1111,25 @@ If the *foreach_statement* contains both or neither `ref` and `readonly`, the it
 
 The iteration variable corresponds to a local variable with a scope that extends over the embedded statement. During execution of a `foreach` statement, the iteration variable represents the collection element for which an iteration is currently being performed. If the iteration variable denotes a read-only variable, a compile-time error occurs if the embedded statement attempts to modify it (via assignment or the `++` and `--` operators) or pass it as a `ref` or `out` parameter.
 
-In the following, for brevity, `IEnumerable`, `IEnumerator`, `IEnumerable<T>` and `IEnumerator<T>` refer to the corresponding types in the namespaces `System.Collections` and `System.Collections.Generic`.
+In the following, for brevity, `IEnumerable` and `IEnumerator` refer to the corresponding types in the namespace `System.Collections`, and `IEnumerable<T>`,  `IEnumerator<T>`, and `IAsyncEnumerator<T>` refer to the corresponding types in the namespace `System.Collections.Generic`.
```
BillWagner: I don't think a non-generic `IAsyncEnumerable` and `IAsyncEnumerator` exist. We should re-write this so it doesn't imply that they do.

standard/statements.md
```diff
@@ -1109,25 +1111,25 @@ If the *foreach_statement* contains both or neither `ref` and `readonly`, the it
...
 - Otherwise, determine whether the type `X` has an appropriate `GetEnumerator` method:
-  - Perform member lookup on the type `X` with identifier `GetEnumerator` and no type arguments. If the member lookup does not produce a match, or it produces an ambiguity, or produces a match that is not a method group, check for an enumerable interface as described below. It is recommended that a warning be issued if member lookup produces anything except a method group or no match.
+  - Perform member lookup on the type `X` with identifier `GetEnumerator` (`GetAsyncEnumerator` for async streams) and no type arguments. If the member lookup does not produce a match, or it produces an ambiguity, or produces a match that is not a method group, check for an enumerable interface as described below. It is recommended that a warning be issued if member lookup produces anything except a method group or no match.
```
BillWagner: This paragraph seems to imply a non-generic GetAsyncEnumerator

standard/statements.md
```diff
@@ -1109,25 +1111,25 @@ If the *foreach_statement* contains both or neither `ref` and `readonly`, the it
...
   - Member lookup is performed on `E` with the identifier `Current` and no type arguments. If the member lookup produces no match, the result is an error, or the result is anything except a public instance property that permits reading, an error is produced and no further steps are taken.
-  - Member lookup is performed on `E` with the identifier `MoveNext` and no type arguments. If the member lookup produces no match, the result is an error, or the result is anything except a method group, an error is produced and no further steps are taken.
-  - Overload resolution is performed on the method group with an empty argument list. If overload resolution results in no applicable methods, results in an ambiguity, or results in a single best method but that method is either static or not public, or its return type is not `bool`, an error is produced and no further steps are taken.
+  - Member lookup is performed on `E` with the identifier `MoveNext` (`MoveNextAsync` for async streams) and no type arguments. If the member lookup produces no match, the result is an error, or the result is anything except a method group, an error is produced and no further steps are taken.
```
BillWagner: Again, this shouldn't apply to async streams, because of no non-generic types.

BillWagner (commented):
This is close, but I think there are a few updates to make.

standard/classes.md
jskeet: This feels like an odd paragraph in general. In particular, the "Adding the `await` keyword" part feels ambiguous - adding it to what? I think we can make that a lot clearer, probably taking language from earlier in 15.15. (The program text for a method can include `await` without it being asynchronous, due to local functions of course, so some care is required.)

standard/classes.md
```diff
@@ -5416,3 +5418,52 @@ When the body of the async function terminates, the return task is moved out of
...
+>     await foreach (var number in GenerateSequence())
+>     {
+>         Console.WriteLine(number);
+>     }
+> }
+> ```
+>
+> *end example*
+
+A method can both consume and produce an asynchronous stream.
```
jskeet: Do we need this sentence? I don't see the equivalent for synchronous iterators.

standard/standard-library.md
jskeet: I agree that we should add this to the standard and make everything match up. We could probably do it in a second PR, *if* that's easier.

standard/standard-library.md
jskeet: @KalleOlaviNiemitalo: Can we resolve this conversation now, or do you have remaining concerns on this one?

standard/statements.md
```diff
@@ -1094,12 +1094,14 @@ The end point of a `for` statement is reachable if at least one of the following
 
 ### 13.9.5 The foreach statement
 
+(Almost all of the text in this subclause was written prior to the addition of async streams; the text applies to async streams as well, unless stated otherwise, usually in parenthetic text. Asynchronous stream creation and usage is discussed in §asynchronous-streams.)
```
jskeet: Another bit to remove, I suggest.

standard/classes.md
jskeet: Why do we need to introduce the term "stream" here? I would prefer us to be consistent between synchronous and asynchronous terminology unless there's an inherent reason for them to differ.

standard/statements.md
jskeet: I don't think so - it's just saying that the `GetEnumerator()` and `GetAsyncEnumerator()` methods aren't generic to start (which is correct for both).

standard/statements.md
jskeet: Again, I think it's about the `MoveNext` and `MoveNextAsync` methods not being generic.

standard/statements.md
```diff
@@ -1109,25 +1111,25 @@ If the *foreach_statement* contains both or neither `ref` and `readonly`, the it
...
-  - Overload resolution is performed on the method group with an empty argument list. If overload resolution results in no applicable methods, results in an ambiguity, or results in a single best method but that method is either static or not public, or its return type is not `bool`, an error is produced and no further steps are taken.
+  - Member lookup is performed on `E` with the identifier `MoveNext` (`MoveNextAsync` for async streams) and no type arguments. If the member lookup produces no match, the result is an error, or the result is anything except a method group, an error is produced and no further steps are taken.
+  - Overload resolution is performed on the method group with an empty argument list. If overload resolution results in no applicable methods, results in an ambiguity, or results in a single best method but that method is either static or not public, or its return type is not `bool`, an error is produced, and no further steps are taken.
```
jskeet, line 1133: The return type of `MoveNextAsync` is `ValueTask<bool>` rather than `bool`.

standard/statements.md
```diff
@@ -1309,6 +1311,33 @@ The order in which `foreach` traverses the elements of an array, is as follows:
 >  
 > *end example*
 
+An async enumerator may optionally expose a `DisposeAsync` method that may be invoked with no arguments and that returns something that can be `await`ed and whose `GetResult()` returns `void`.
```
jskeet, line 1364: I suspect there's a cleaner way of expressing this via task types, e.g. "with a return type that is a non-generic task type".

standard/statements.md
```diff
@@ -1797,6 +1826,8 @@ While a mutual-exclusion lock is held, code executing in the same execution thre
 
 ## 13.14 The using statement
 
+(Almost all of the text in this subclause was written prior to the addition of async streams; the text applies to async streams as well, unless stated otherwise, usually in parenthetic text. Asynchronous stream creation and usage is discussed in §asynchronous-streams.)
```
jskeet: And another bit to remove

standard/standard-library.md
KalleOlaviNiemitalo: I don't have remaining concerns.

standard/classes.md
Nigel-Ecma: It does feel a little odd. As an off the cuff suggestion:
- swap the order of §15.14 Iterators and §15.15 Async Functions
- rename "Iterators” to “Iterators and Async Streams”
- rewrite this new intro paragraph to say the clause covers synchronous and async iterators, the latter usually termed async streams.
- remove the parentheses and merge the content into the surrounding text

Of course this isn't the only “odd” thing we’ve done, e.g. property & indexer accessors with the “read exchanging terms” instruction.

standard/classes.md
Nigel-Ecma: As an example of what my opening off the cuff suggestion would produce this para would become:

```suggestion
The ***enumerator interfaces*** are the non-generic interface `System.Collections.IEnumerator` and all instantiations of the generic interfaces `System.Collections.Generic.IEnumerator<T>` and `System.Collections.Generic.IAsyncEnumerator<T>`. For the sake of brevity, in this subclause and its siblings these interfaces are referenced as `IEnumerator`, `IEnumerator<T>` and `IAsyncEnumerator<T>` respectively.
```

Nigel-Ecma: I think changing the name is a non-starter (flashbacks to NNRT here ;-)) and there is some convention in using the term “stream”. If my first suggestion above is followed the stream/async iterator stuff could be covered in §asynchronous-streams which will occur before the §15.15 Iterators and Async Streams clause.

standard/classes.md
```diff
@@ -5416,3 +5418,52 @@ When the body of the async function terminates, the return task is moved out of
 If the return type of the async function is `void`, evaluation differs from the above in the following way: Because no task is returned, the function instead communicates completion and exceptions to the current thread’s ***synchronization context***. The exact definition of synchronization context is implementation-dependent, but is a representation of “where” the current thread is running. The synchronization context is notified when evaluation of a `void`-returning async function commences, completes successfully, or causes an uncaught exception to be thrown.
 
 This allows the context to keep track of how many `void`-returning async functions are running under it, and to decide how to propagate exceptions coming out of them.
+
+### §asynchronous-streams Asynchronous stream creation and usage
+
+A stream can be created and consumed asynchronously.
```
Nigel-Ecma: The definition of “stream” needs to be added here, elsewhere in the Standard we have streams of Unicode characters and tokens, uses of `StreamReader` etc.

See previous comments for how this might be done; it may even be decided to move this section into the (now following) §15.15 Iterators and Async Streams clause.

standard/standard-library.md
Nigel-Ecma: Let’s do it in *this* PR

Nigel-Ecma (dismissed):
Like @jskeet I think this needs some reorg/rewriting to cover synchronous and asynchronous iterators more uniformly. Also need to address the omissions reported by @KalleOlaviNiemitalo .

standard/classes.md
```diff
@@ -5416,3 +5418,52 @@ When the body of the async function terminates, the return task is moved out of
 If the return type of the async function is `void`, evaluation differs from the above in the following way: Because no task is returned, the function instead communicates completion and exceptions to the current thread’s ***synchronization context***. The exact definition of synchronization context is implementation-dependent, but is a representation of “where” the current thread is running. The synchronization context is notified when evaluation of a `void`-returning async function commences, completes successfully, or causes an uncaught exception to be thrown.
 
 This allows the context to keep track of how many `void`-returning async functions are running under it, and to decide how to propagate exceptions coming out of them.
+
+### §asynchronous-streams Asynchronous stream creation and usage
```
jskeet: Do we need this section at all? (Really not needed in normative text?)

standard/classes.md
```diff
@@ -5416,3 +5418,52 @@ When the body of the async function terminates, the return task is moved out of
...
+
+A stream can be created and consumed asynchronously.
+
+A method that creates an asynchronous stream has the following characteristics:
+
+- It's declared with the `async` modifier.
+- It returns an `IAsyncEnumerable<T>`.
```
jskeet: Or `IAsyncEnumerator<T>`.

standard/classes.md
jskeet: (All of the bits in the bullet points can be included in the iterator section.)

jskeet:
Summary of discussions:

- Try to reword to avoid the brackets
- Avoid using "stream" in normative text, but have a note
- Merge the "Asynchronous stream creation and usage" into the iterator section
- Reorder async and iterators so that async comes first
- Defer the CancellationToken support until after we've got the structure right (but as a fast-follow - definitely in C# 8)

jskeet:
It looks like there's still quite a bit to do here - but it's not clear to me that Rex is the right person to do this.

For discussion at today's call (if there's time): who should take this on? (I *may* have time, if we decide I'm the most appropriate person bearing in mind other PRs...)

standard/classes.md
BillWagner: I also think we should continue to use the term "async stream". It fits well with the feature, and "stream" is the conventional term.

I also like the idea of some reordering. That may be a follow-up PR?

standard/classes.md
```diff
@@ -5416,3 +5418,52 @@ When the body of the async function terminates, the return task is moved out of
...
+- It contains a `yield return` statement to return successive elements in the asynchronous stream.
```
BillWagner: We need to add that the argument to `yield return` must be implicitly convertible to the type argument for the return.

BillWagner: If we don't combine this section with the enumerator section, we should add that `yield break` is can be used.

standard/classes.md
BillWagner: More later, but I think a good way to address this is to look at the statements section, and add a new subclause for "The await foreach" statement, rather than modifying the foreach statement section.

standard/classes.md
BillWagner: I don't think it's needed.

standard/statements.md
BillWagner: I'd make this two sub-clauses:  One for "foreach", and one for "await foreach". 

The grammar is similar, and I'll defer to Nigel on if one or two grammar productions makes more sense.

The section on "MoveNext" and "MoveNextAsync" becomes more clear as two distinct sections: One for a synchronous "move next" and one for an asynchronous value-task based "move next asynchronously"

standard/statements.md
BillWagner: That is in #672.

See Line 419 and 473, and 1864 of statements.md in that PR .

standard/classes.md
jskeet: > and "stream" is the conventional term

But stream has had a very specific meaning in .NET since 1.0, which *isn't* what's being described here.

We talk about "enumeration", "collection" and "iteration" elsewhere - couldn't we continue to use that terminology here? I'd prefer "async iterable" or "async enumerable" over "async stream".

To put it another way - if an `IAsyncEnumerable<T>` is an async stream, isn't it reasonable for someone to assume that an `IEnumerable<T>` is a stream? If you agree with that premise, then consider: if someone asked you to write a method accepting a stream parameter, what would you write:

```csharp
public void M(Stream stream)
```

or

```csharp
public void M<T>(IEnumerable<T> stream)
```

? (While C# and .NET aren't the same, it's reasonable to expect that 99.9% of readers of the C# standard are familiar with .NET, and with System.IO.Stream.)

jskeet: Revisiting this, I'm getting the same vibe as before. Having multiple parenthesized bits in *so* many sentences is really distracting.

I think I'd prefer to define terms (e.g. "advancing method" for MoveNext/MoveNextAsync), define how those terms map to each of sync and async enumerables, then refer to them appropriately. Would welcome other perspectives though.

standard/classes.md
```diff
@@ -5407,11 +5409,11 @@ An enumerator object’s `Current` property is affected by `yield return` statem
 
 When an enumerator object is in the **suspended** state, the value of `Current` is the value set by the previous call to `MoveNext`. When an enumerator object is in the **before**, **running**, or **after** states, the result of accessing `Current` is unspecified.
 
-For an iterator with a yield type other than `object`, the result of accessing `Current` through the enumerator object’s `IEnumerable` implementation corresponds to accessing `Current` through the enumerator object’s `IEnumerator<T>` implementation and casting the result to `object`.
+For an iterator with a yield type other than `object`, the result of accessing `Current` through the enumerator object’s `IEnumerable` implementation corresponds to accessing `Current` through the enumerator object’s `IEnumerator<T>` (`IAsyncEnumerator<T>` for async streams) implementation and casting the result to `object`.
```
jskeet: I suspect this paragraph should just not talk about async streams at all, as there's no equivalent of `IEnumerable`, right?

standard/classes.md
jskeet: Additionally, this presupposes that the *only* way of creating an asynchronous stream is via an async iterator block. What if I have my own implementation of `IAsyncEnumerable<T>` and I construct that manually?

standard/statements.md
jskeet: I'd hope that await foreach could be described more simply than foreach, as it doesn't need to worry about arrays or the non-generic IEnumerable.

Does `await foreach` perform the same sort of implicit cast as `foreach`? (So can I do `IAsyncEnumerable<object> collection = ...; await foreach(string x in collection)`?) I suspect so, but haven't checked.

standard/statements.md
jskeet: Not a problem if we're creating a new section :)

standard/statements.md
```diff
@@ -1796,6 +1825,8 @@ While a mutual-exclusion lock is held, code executing in the same execution thre
...
```
jskeet: @BillWagner Similar to the foreach discussion, should we have an "await using" statement?

jskeet (commented):
The big things I think need addressing here are:

- Whether or not to use the term "stream" (I suspect we'll have some passionate disagreement on this one)
- Splitting "the foreach statement" into two sections (foreach and await foreach)
- Splitting "the using statement" into two sections (using and await using)
- Removing the "asynchronous stream creation and usage" section (moving content elsewhere where necessary)
- Ideally avoiding so much parenthetical text in 15.14
- Adding CancellationToken and EnumeratorCancellationAttribute to the standard library

I'm on the fence as to whether we do that by merging as-is (with just minor tweaks) or say "let's create a second PR that does those things".

standard/statements.md
BillWagner: I did test, and yes, this is legal:

```csharp
public static async Task Example()
{
    await foreach (string number in GetStringsAsync())
    {
        Console.WriteLine(number);
    }
}
public static async IAsyncEnumerable<object> GetStringsAsync()
{
    for (int i = 0; i < 10; i++)
    {
        await Task.Delay(100); // Simulate some asynchronous work
        yield return i.ToString();
    }
}
```

standard/statements.md
BillWagner: Verified to be sure: there aren't non-generic AsyncEnumerable types.

standard/statements.md
BillWagner: I like that idea (in both places.) Admittedly a bit more duplication, but I think it reads more clearly.

Whichever we decide, we should do the same in both locations.

standard/statements.md
BillWagner: I copied this comment to #672, so we don't lose this information.

BillWagner (commented):
I agree with Jon's concepts of that most important tasks.

One factor: I didn't find any new concerns with this PR. I only made new comments on the issues that already exist.

jskeet:
Suggested terminology:

- "async enumerable" for "something that implements `IAsyncEnumerable<T>`"
- "async iterator" (or "async iterator block") for "a block with yield statements which produces an async enumerable"

jskeet:
As well as the terminology above, discussion of the structure suggested:

- Separate out `await foreach` from `foreach` as there's a lot in `foreach` that doesn't apply (due to arrays and non-generic types)
- Separate out `await using` from `using` (but we could go either way on this)
- Keep the iterator block section as combined sync and async, but try to move all the differences into one introductory part

Additionally, Nigel has a suggestion right at the top of `classes.md` to consider, in terms of ordering.

@RexJaeschke and @BillWagner to work on this.

standard/statements.md
jnm2: Would the term "awaitable types" help avoid conflating this with either the particular type System.Threading.Tasks.Task, or with the separate "task-like type" which is a concept on the production side allowed as a return type of an async method (e.g. ValueTask<T>)?

RexJaeschke:
On the 2025-05-14 TG2 call:

Meeting recording time: 86-minute mark

**Action**:   Bill and Rex will work on terminology above and discussion of the structure, as stated in Bill’s [comment](https://github.com/dotnet/csharpstandard/pull/606#issuecomment-2881631576).

standard/standard-library.md
BillWagner: I've added both declarations in my next commit. 

For `CancellationToken`, I've included all public members. We should remove those not referenced in the standard once we agree on other language for the feature.

standard/statements.md
BillWagner: Proposal for June meeting:  Decide if we want a single clause for **The `using` statement** to include `using` and `await` using, or two clauses:  **The `using` statement** and **The `await using` statement**. 

This PR and #672 both contain changes to the current using statement. Once we merge both, it will be much simpler to split this clause without losing any important edits. (Should that be the decision)

standard/statements.md
BillWagner: See lines 2018 - 2051 in #672

standard/statements.md
```diff
@@ -1119,14 +1129,14 @@ The compile-time processing of a `foreach` statement first determines the ***col
...
   - Member lookup is performed on `E` with the identifier `Current` and no type arguments. If the member lookup produces no match, the result is an error, or the result is anything except a public instance property that permits reading, an error is produced and no further steps are taken.
-  - Member lookup is performed on `E` with the identifier `MoveNext` and no type arguments. If the member lookup produces no match, the result is an error, or the result is anything except a method group, an error is produced and no further steps are taken.
-  - Overload resolution is performed on the method group with an empty argument list. If overload resolution results in no applicable methods, results in an ambiguity, or results in a single best method but that method is either static or not public, or its return type is not `bool`, an error is produced and no further steps are taken.
+  - Member lookup is performed on `E` with the identifier `MoveNext`and no type arguments. If the member lookup produces no match, the result is an error, or the result is anything except a method group, an error is produced and no further steps are taken.
```
Nigel-Ecma: Looks like a space was accidentally removed:

```suggestion
  - Member lookup is performed on `E` with the identifier `MoveNext` and no type arguments. If the member lookup produces no match, the result is an error, or the result is anything except a method group, an error is produced and no further steps are taken.
```

standard/statements.md
```diff
@@ -1093,22 +1093,32 @@ The end point of a `for` statement is reachable if at least one of the following
...
-In the following, for brevity, `IEnumerable`, `IEnumerator`, `IEnumerable<T>` and `IEnumerator<T>` refer to the corresponding types in the namespaces `System.Collections` and `System.Collections.Generic`.
+The compile-time processing of a `foreach` statement first determines the ***collection type***, ***enumerator type*** and ***iteration type*** of the expression. The processing for a `foreach` statement is detailed in §sync-foreach and the process for an `await foreach` is detailed in §async-foreach.
+
+> *Note*: If *expression* has the value `null`, a `System.NullReferenceException` is thrown at run-time. *end note*
+
+An implementation is permitted to implement a given *foreach_statement* differently; e.g., for performance reasons, as long as the behavior is consistent with the above expansion.
+
+#### §sync-foreach Synchronous foreach
 
 The compile-time processing of a `foreach` statement first determines the ***collection type***, ***enumerator type*** and ***iteration type*** of the expression. This determination proceeds as follows:
```
Nigel-Ecma, line 1123: Most of this has just been said in 1115 just above.

standard/statements.md
```diff
@@ -1308,6 +1314,80 @@ The order in which `foreach` traverses the elements of an array, is as follows:
 >
 > *end example*
 
+#### §async-foreach await foreach
+
+The compile-time processing of a `foreach` statement first determines the ***collection type***, ***enumerator type*** and ***iteration type*** of the expression. The processing for a `foreach` statement is detailed in §sync-foreach and the process for an `await foreach` is detailed in §async-foreach.
```
Nigel-Ecma, line 1319: This is stated in 1115 & 1123. Revising the split of material between General, the sync, and the async subclauses might improve matters.

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+
+## 15.14 Synchronous and asynchronous iterators
+
+### 15.14.1 General
+
+A function member ([§12.6](expressions.md#126-function-members)) or local function (§13.6.4) implemented using an iterator block ([§13.3](statements.md#133-blocks)) is called an ***iterator***. An iterator block may be used as the body of a function member as long as the return type of the corresponding function member is one of the enumerator interfaces ([§15.14.2](classes.md#15142-enumerator-interfaces)) or one of the enumerable interfaces ([§15.14.3](classes.md#15143-enumerable-interfaces)).
+
+An async function (§15.15) implemented using an iterator block ([§13.3](statements.md#133-blocks)) is called an ***asynchronous iterator***. An asynchronous iterator block may be used as the body of a function member as long as the return type of the corresponding function member is the asynchronous enumerator interfaces ([§15.14.2](classes.md#15142-enumerator-interfaces)) or the asynchronous enumerable interfaces ([§15.14.3](classes.md#15143-enumerable-interfaces)).
```
jskeet, line 5476:
```suggestion
An async function (§15.15) implemented using an iterator block ([§13.3](statements.md#133-blocks)) is called an ***asynchronous iterator***. An asynchronous iterator block may be used as the body of a function member as long as the return type of the corresponding function member is one of the asynchronous enumerator interfaces ([§15.14.2](classes.md#15142-enumerator-interfaces)) or one of the asynchronous enumerable interfaces ([§15.14.3](classes.md#15143-enumerable-interfaces)).
```

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+
+An iterator block may occur as a *method_body*, *operator_body* or *accessor_body*, whereas events, instance constructors, static constructors and finalizer shall not be implemented as synchronous or asynchronous iterators.
+
+When a function member or local function is implemented using an iterator block, it is a compile-time error for the parameter list of the function member to specify any `in`, `out`, or `ref` parameters, or an parameter of a `ref struct` type.
```
jskeet, line 5480:
```suggestion
When a function member or local function is implemented using an iterator block, it is a compile-time error for the parameter list of the function member to specify any `in`, `out`, or `ref` parameters, or a parameter of a `ref struct` type.
```

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+
+### 15.14.2 Enumerator interfaces
+
+The ***enumerator interfaces*** are the non-generic interface `System.Collections.IEnumerator` and all instantiations of the generic interfaces `System.Collections.Generic.IEnumerator<T>`.
```
jskeet, line 5484: Is "instantiations" what we use elsewhere for this? Quite possibly so - just checking. (When I have more time I may try to find a reference myself...)

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+- The yield type of an iterator that returns an `IEnumerator<T>`, `IAsyncEnumerator<T>`, `IEnumerable<T>`, or `IAsyncEnumerable<T>` is `T`.
+
+### 15.14.5 Enumerator objects
+
+#### 15.14.5.1 General
+
+When a function member or local function returning an enumerator interface type is implemented using an iterator block, invoking the function does not immediately execute the code in the iterator block. Instead, an ***enumerator object*** is created and returned. This object encapsulates the code specified in the iterator block, and execution of the code in the iterator block occurs when the enumerator object’s `MoveNext` or `MoveNextAsync` method is invoked. An enumerator object has the following characteristics:
+
+- It implements `System.IDisposable`, `IEnumerator` and `IEnumerator<T>`, or `System.IAsyncDisposable` and `IAsyncEnumerator<T>`, where `T` is the yield type of the iterator.
+- It is initialized with a copy of the argument values (if any) and instance value passed to the function member.
```
jskeet, line 5512: Do we need "or local function" here? (There may be other places too.)

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+
+The following subclauses describe the required behavior of the member to advance the enumerator, retrieve the current value from the enumerator, and dispose of resources used by the enumerator. These are defined in the following members for synchronous and asynchronous enumerators, respectively:
+
+- To advance the enumerator: `MoveNext` and `MoveNextAsync`.
+- To retrieve the current value: `Current`.
+- To dispose of resources: `Dispose` and `DisposeAsync`.
+
+Enumerator objects do not support the `IEnumerator.Reset` method. Invoking this method causes a `System.NotSupportedException` to be thrown.
+
+Synchronous and asynchronous iterator blocks differ in that asynchronous iterator members return task types and may be awaited.
```
jskeet, line 5527: This feels odd to me - I suspect we don't need it at all, and if we *do* need it we should move it to the "general" part, I suspect.

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+
+#### 15.14.5.2 Advance the enumerator
+
+The `MoveNext` and `MoveNextAsync` methods of an enumerator object encapsulates the code of an iterator block. Invoking the `MoveNext` or `MoveNextAsync` method executes code in the iterator block and sets the `Current` property of the enumerator object as appropriate.
```
jskeet, line 5531:
```suggestion
The `MoveNext` and `MoveNextAsync` methods of an enumerator object encapsulate the code of an iterator block. Invoking the `MoveNext` or `MoveNextAsync` method executes code in the iterator block and sets the `Current` property of the enumerator object as appropriate.
```

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+
+- If the state of the enumerator object is **before**, invoking `MoveNext`:
+  - Changes the state to **running**.
+  - Initializes the parameters (including `this`) of the iterator block to the argument values and instance value saved when the enumerator object was initialized.
+  - Executes the iterator block from the beginning until execution is interrupted (as described below).
+- If the state of the enumerator object is **running**, the result of invoking `MoveNext` is unspecified.
+- If the state of the enumerator object is **suspended**, invoking MoveNext:
+  - Changes the state to **running**.
+  - Restores the values of all local variables and parameters (including `this`) to the values saved when execution of the iterator block was last suspended.  
+    > *Note*: The contents of any objects referenced by these variables may have changed since the previous call to `MoveNext`. *end note*
```
jskeet, line 5545: I'm not sure whether this note is necessary.

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+  - Resumes execution of the iterator block immediately following the yield return statement that caused the suspension of execution and continues until execution is interrupted (as described below).
+- If the state of the enumerator object is **after**, invoking `MoveNext` returns false.
+
+When `MoveNext` executes the iterator block, execution can be interrupted in four ways: By a `yield return` statement, by a `yield break` statement, by encountering the end of the iterator block, and by an exception being thrown and propagated out of the iterator block.
```
jskeet, line 5549: Do we need to add a fifth way for asynchronous iterators, namely "awaiting something awaitable that hasn't completed yet"?

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+An iterator produces a sequence of values, all of the same type. This type is called the ***yield type*** of the iterator.
+
+- The yield type of an iterator that returns `IEnumerator` or `IEnumerable` is `object`.
+- The yield type of an iterator that returns an `IEnumerator<T>`, `IAsyncEnumerator<T>`, `IEnumerable<T>`, or `IAsyncEnumerable<T>` is `T`.
+
+### 15.14.5 Enumerator objects
+
+#### 15.14.5.1 General
+
+When a function member or local function returning an enumerator interface type is implemented using an iterator block, invoking the function does not immediately execute the code in the iterator block. Instead, an ***enumerator object*** is created and returned. This object encapsulates the code specified in the iterator block, and execution of the code in the iterator block occurs when the enumerator object’s `MoveNext` or `MoveNextAsync` method is invoked. An enumerator object has the following characteristics:
```
jskeet, line 5509:
```suggestion
When a function member or local function returning an enumerator interface type or asynchronous enumerator interface type is implemented using an iterator block, invoking the function does not immediately execute the code in the iterator block. Instead, an ***enumerator object*** is created and returned. This object encapsulates the code specified in the iterator block, and execution of the code in the iterator block occurs when the enumerator object’s `MoveNext` or `MoveNextAsync` method is invoked. An enumerator object has the following characteristics:
```

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+  - Changes the state to **running**.
+  - Executes any finally blocks as if the last executed `yield return` statement were a `yield break` statement. If this causes an exception to be thrown and propagated out of the iterator body, the state of the enumerator object is set to **after** and the exception is propagated to the caller of the `Dispose` method.
+  - Changes the state to **after**.
+- If the state of the enumerator object is **after**, invoking `Dispose` has no affect.
+
+### 15.14.6 Enumerable objects
+
+#### 15.14.6.1 General
+
+When a function member or local function returning an enumerable interface type is implemented using an iterator block, invoking the function member does not immediately execute the code in the iterator block. Instead, an ***enumerable object*** is created and returned.
```
jskeet, line 5594:
```suggestion
When a function member or local function returning an enumerable interface type or asynchronous enumerable interface type is implemented using an iterator block, invoking the function member does not immediately execute the code in the iterator block. Instead, an ***enumerable object*** is created and returned.
```

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+
+The enumerable object’s `GetEnumerator` or `GetAsyncEnumerator` method returns an enumerator object that encapsulates the code specified in the iterator block, and execution of the code in the iterator block occurs when the enumerator object’s `MoveNext` or `MoveNextAsync` method is invoked. An enumerable object has the following characteristics:
+
+- It implements `IEnumerable` and `IEnumerable<T>` or `IAsyncEnumerable<T>`, where `T` is the yield type of the iterator.
```
jskeet, line 5598: We should probably clarify that the bracketing here is (`IEnumerable` and `IEnumerable<T>`) or (`IAsyncEnumerable<T>`). I don't have a suggestion for that right now.

standard/classes.md
```diff
@@ -5584,3 +5466,146 @@ When the body of the async function terminates, the return task is moved out of
...
+
+An enumerable object may implement more interfaces than those specified above.
+
+> *Note*: For example, an enumerable object may also implement `IEnumerator` and `IEnumerator<T>`, enabling it to serve as both an enumerable and an enumerator. Typically, such an implementation would return its own instance (to save allocations) from the first call to `GetEnumerator`. Subsequent invocations of `GetEnumerator`, if any, would return a new class instance, typically of the same class, so that calls to different enumerator instances will not affect each other. It cannot return the same instance even if the previous enumerator has already enumerated past the end of the sequence, since all future calls to an exhausted enumerator must throw exceptions. *end note*
```
jskeet, line 5605: I don't think this last sentence is true. Earlier we have:

- If the state of the enumerator object is **after**, invoking `MoveNext` returns false.

(And that Dispose has no effect, and that Current is unspecified...)

standard/statements.md
jskeet: I think this comment is now obsolete, as this is only about synchronous foreach. We should check that and resolve the comment in the meeting if so.

standard/statements.md
jskeet: I think just deleting this paragraph would be fine - there may be more split revision to do, but I *think* in this case it's just a matter of removing the paragraph here and at 1123. We'd need to replace "This determination" with "The determination of the ***collection type***, ***enumerator type*** and ***iteration type*** of the expression" as well though (in both places).

standard/statements.md
```diff
@@ -1308,6 +1314,80 @@ The order in which `foreach` traverses the elements of an array, is as follows:
...
+
+This determination proceeds as follows:
+
+- Determine whether the type `X` has an appropriate `GetAsyncEnumerator` method:
+  - Perform member lookup on the type `X` with identifier `GetAsyncEnumerator` and no type arguments. If the member lookup does not produce a match, or it produces an ambiguity, or produces a match that is not a method group, check for an enumerable interface as described below. It is recommended that a warning be issued if member lookup produces anything except a method group or no match.
+  - Perform overload resolution using the resulting method group and an empty argument list. If overload resolution results in no applicable methods, results in an ambiguity, or results in a single best method but that method is either static or not public, check for an enumerable interface as described below. It is recommended that a warning be issued if overload resolution produces anything except an unambiguous public instance method or no applicable methods.
+  - If the return type `E` of the `GetAsyncEnumerator` method is not a class, struct or interface type, an error is produced and no further steps are taken.
```
jskeet, line 1326: Do delegate types count as classes for this purpose? Likewise do enums count as structs? Is this basically trying to prohibit pointers and void? Anything else?

standard/statements.md
```diff
@@ -1308,6 +1314,80 @@ The order in which `foreach` traverses the elements of an array, is as follows:
...
+  - Member lookup is performed on `E` with the identifier `Current` and no type arguments. If the member lookup produces no match, the result is an error, or the result is anything except a public instance property that permits reading, an error is produced and no further steps are taken.
+  - Member lookup is performed on `E` with the identifier `MoveNextAsync` and no type arguments. If the member lookup produces no match, the result is an error, or the result is anything except a method group, an error is produced and no further steps are taken.
```
jskeet, line 1328: In both of these bullets, we have two different error conditions, but we describe them separately: "If X happens, an error is produced; if Y happens, an error is produced". Can't we have "If X or Y happens, an error is produced"?

standard/statements.md
```diff
@@ -1809,13 +1889,13 @@ resource_acquisition
     ;
 ```
 
-A ***resource*** is a class or struct that implements the `System.IDisposable` interface, which includes a single parameterless method named `Dispose`. Code that is using a resource can call `Dispose` to indicate that the resource is no longer needed.
+A ***resource*** is a class or struct that implements the `System.IDisposable` (`IAsyncDisposable` for async streams) interface, which includes a single parameterless method named `Dispose` (`DisposeAsync` for async streams). Code that is using a resource can call `Dispose` to indicate that the resource is no longer needed.
```
jskeet, line 1892: I think all the async stream bits need reworking here and below.

jskeet: Possibly just define "async resources" instead? Although that's a bit weird. I do wonder whether we should drop "resource" as a term entirely, and talk about disposables and asynchronous disposables, akin to enumerators and asynchronous enumerators, and enumerables and async enumerables. Basically name these by what you can do, not any expected meaning.

jskeet (commented):
I generally like the structure of this *much, much* more - thank you for the hard work! Plenty of comments left, but hopefully we can resolve a lot of them in the meeting.

BillWagner (approved):

jskeet (approved):

standard/classes.md
BillWagner: Fixed in #672

standard/classes.md
BillWagner: Fixed in #672, I simplified it just to the interfaces listed, because that does include all derived interfaces and so on (based on the text)

standard/classes.md
BillWagner: Addressed in #672

standard/classes.md
BillWagner: Updated to use "function" rather than "function member" in all pertinent locations in #672

standard/classes.md
BillWagner: The refactoring in #672 removed this

standard/classes.md
BillWagner: fixed in other edits in #672

standard/classes.md
BillWagner: Removed in #672

standard/classes.md
BillWagner: I added a note in #672, because that can be derived from the fact that `MoveNextAsync` may include `await` expressions.

standard/classes.md
BillWagner: Added in #672

standard/classes.md
BillWagner: Removed that sentence in #672 as it's in a note anyway...

standard/statements.md
BillWagner: Reworked in #672

standard/statements.md
BillWagner: Updated in #672 to use an "awaitable return type" (with xref) and the *await_expression* classified as `bool` (with xref)

standard/statements.md
BillWagner: I tried that rewrite, and the language was clumsier because one is a method group and the other is a property. Leaving as is.

BillWagner (commented):
All comments tagged as resolved are updated in commit e7a46eb6c4269b90ecf5f9c97a634b0a5f05614c

standard/statements.md
BillWagner: It can be implied that they don't, because later steps require the returned type to have a suitable `MoveNext` and `Current` members (sync / async variants as required.

standard/classes.md
BillWagner: Addressed in #672

PR #608, opened 2022-08-08
Add Support for Name shadowing

RexJaeschke:
---
[Associated WorkItem - 250822](https://dev.azure.com/msft-skilling/Content/_workitems/edit/250822)


BillWagner:
rebased on 09/26/2023

@RexJaeschke Oddly enough, additional changes in basic-concepts.md appeared after rebasing. Let me know if we need to revert some of those changes.

standard/statements.md
```diff
@@ -477,7 +477,7 @@ The *type* and *constant_expression* of a local constant declaration shall follo
 
 The value of a local constant is obtained in an expression using a *simple_name* ([§12.8.4](expressions.md#1284-simple-names)).
 
-The scope of a local constant is the block in which the declaration occurs. It is an error to refer to a local constant in a textual position that precedes the end of its *constant_declarator*. Within the scope of a local constant, it is a compile-time error to declare another local variable, local function or constant with the same name.
+The scope of a local constant is the block in which the declaration occurs. It is an error to refer to a local constant in a textual position that precedes the end of its *constant_declarator*.
```
KalleOlaviNiemitalo, line 482: …except in `nameof`?

```csharp
class C {
    void M() {
        const string s = nameof(s);
    }
}
```

standard/basic-concepts.md
```diff
@@ -81,7 +81,7 @@ There are several different types of declaration spaces, as described in the fol
...
+- Each method declaration, property declaration, property accessor declaration, indexer declaration, indexer accessor declaration, operator declaration, instance constructor declaration, anonymous function, and local function creates a new declaration space called a ***local variable declaration space***. Names are introduced into this declaration space through formal parameters (*fixed_parameter*s and *parameter_array*s) and *type_parameter*s. The set accessor for a property or an indexer introduces the name `value` as a formal parameter. The body of the function member, anonymous function, or local function, if any, is considered to be nested within the local variable declaration space. It is an error for a local variable declaration space and a nested local variable declaration space to contain elements with the same name. When a local variable declaration space and a nested local variable declaration space contain elements with the same name, within the scope of the nested local name, the outer local name is hidden ([§7.7.1](basic-concepts.md#771-general)) by the nested local name.
```
KalleOlaviNiemitalo: Should this be removed now:

> It is an error for a local variable declaration space and a nested local variable declaration space to contain elements with the same name.

RexJaeschke:
My latest commit is intended to address @BillWagner's [comment](https://github.com/dotnet/csharpstandard/pull/608#issuecomment-1735857484). It probably also addresses @KalleOlaviNiemitalo 's [comment](https://github.com/dotnet/csharpstandard/pull/608#discussion_r1337528349).

jskeet (commented):
I'm concerned that we need to be careful not to remove more restrictions than we want. My understanding is that shadowing is only permitted in a very few situations - you still can't write:

```csharp
public void M()
{
    int x = 10;
    if (x == 10)
    {
        int x = 5;
        Console.WriteLine(x);
    }
}
```

... right? That may already be covered - I'll try to check before the meeting - but we need to make sure that it *is* still covered.

Nigel-Ecma (commented):
Kalle's observation on `nameof` is either a bug or special case handling of `nameof` in declarations which is probably not clear in the Standard. If the former it has no impact here, if the latter it also has no impact here as it is the wrong place to describe it so an issue should be raised to deal with it.

Apart from that this looks good

jskeet:
> That may already be covered - I'll try to check before the meeting - but we need to make sure that it _is_ still covered.

Nigel's pretty convinced that this isn't a problem - the onus is on me to convince myself too.

standard/statements.md
jskeet: I'll spin this off into a separate issue tomorrow.

standard/basic-concepts.md
jskeet: This is what makes me nervous. We want this to be invalid:

```csharp
void M()
{
    int x = 10;
    {
        int x = 5;
    }
}
```

... but this to be valid:

```csharp
void M()
{
    int x = 10;
    LocalFunction();

    void LocalFunction()
    {
        int x = 5;
    }
}
```

standard/statements.md
jskeet: Filed #1111.

BillWagner:
closing and reopening for a fresh CI build (test examples should pass now)

Nigel-Ecma (approved):
Triggered in error, my last review stands – approved!

jskeet:
Closed and reopened to check the test again... then we can merge on green. (I've created #1131 to check my earlier concern.)

jskeet:
Tests are still failing - leaving assigned to @BillWagner to check up on. (Feel free to assign it back to me Bill.)

PR #609, opened 2022-08-08
Add support for Null-coalescing assignment

RexJaeschke:


standard/expressions.md
```diff
@@ -5814,7 +5814,22 @@ When a property or indexer declared in a *struct_type* is the target of an assig
...
 
-An operation of the form `x «op»= y` is processed by applying binary operator overload resolution ([§11.4.5](expressions.md#1145-binary-operator-overload-resolution)) as if the operation was written `x «op» y`. Then,
+For operator `??=`, given `a ??= b`, where `A` is the type of `a`, `B` is the type of `b`, and `A0` is the underlying type of `A`, if `A` is a nullable value type:
+
+1. If `A` does not exist or is a non-nullable value type, a compile-time error occurs.
+2. If `B` is not implicitly convertible to `A` or `A0` (if `A0` exists), a compile-time error occurs.
```
KalleOlaviNiemitalo: The expression `b` can be implicitly convertible to `A` or `A0` even if its type `B` is not.

```C#
byte? e = null;
e ??= 0;
```

Perhaps also lambda expressions and interpolated strings.

KalleOlaviNiemitalo:
`??=` needs its own definite-assignment rules in variables.md. It cannot be treated like other compound assignment expressions because the right-hand operand is conditionally evaluated.

standard/expressions.md
Nigel-Ecma: Hmm, implicit integer literal value-based conversions... so you're thinking of something along the lines of:

> If `B` is not implicitly convertible to `A` or `A0` (if `A0` exists) _or_ if `b` is a constant expression of type `int` and its value is not implicitly  convertible to `A` or `A0` (if `A0` exists), then a compile-time error occurs.

I think that wording captures both int literal to smaller integral types and int literal zero to enum types.

Ouch, I can't but wonder how many times that might be missed in the Standard...

This might require an issue of its own rather than fixing it when it is spotted – unless someone already knows it is allowed for wherever needed?

BillWagner:
rebased on the latest draft-v8 branch on 09/26/2023

gafter:
I don't remember why this is assigned to me. For writing the definite assignment text? For general review? I couldn't find anything in the minutes.

BillWagner:
@gafter 

> I don't remember why this is assigned to me. For writing the definite assignment text? For general review? I couldn't find anything in the minutes.

Going by memory, but I think it was a general reviewer and to make any suggested edits.

standard/expressions.md
BillWagner: The last sentence of [§12.23](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#1223-constant-expressions) should cover this everywhere:

> An implicit constant expression conversion ([§10.2.11](conversions.md#10211-implicit-constant-expression-conversions)) permits a constant expression of type `int` to be converted to `sbyte`, `byte`, `short`, `ushort`, `uint`, or `ulong`, provided the value of the constant expression is within the range of the destination type.

standard/expressions.md
```diff
@@ -6671,7 +6672,22 @@ The ref assignment operator shall not read the storage location referenced by th
...
+3. If `A0` exists and `B` is implicitly convertible to `A0`, and `B` is not dynamic, then the type of `a ??= b` is `A0`. `a ??= b` is evaluated at runtime as:
+
+   ```csharp
+   var tmp = a.GetValueOrDefault();
+   if (!a.HasValue) { tmp = b; a = tmp; }
+   tmp
+   ```
+
+   Except that `a` is only evaluated once.
+4. Otherwise, the type of `a ??= b` is `A`. `a ??= b` is evaluated at runtime as `a ?? (a = b)`, except that `a` is only evaluated once.
```
BillWagner: Can this be shortened to something as simple as:

> Operator `x ??= y` is processed the same as the expression `(x ?? (x = y))` following the process in §12.15.

BillWagner (approved):
I think this is good to go, with a couple discussion items.

BillWagner:
> `??=` needs its own definite-assignment rules in variables.md. It cannot be treated like other compound assignment expressions because the right-hand operand is conditionally evaluated.

I'd like to make a separate issue for this (and to check other impacts for definite assignment in C# 8).

standard/expressions.md
```diff
@@ -6475,7 +6475,7 @@ The `=` operator is called the ***simple assignment operator***. It assigns the
...
 
-The assignment operators other than the `=` and `= ref` operators are called the ***compound assignment operators***. These operators perform the indicated operation on the two operands, and then assign the resulting value to the variable, property, or indexer element given by the left operand. The compound assignment operators are described in [§12.21.4](expressions.md#12214-compound-assignment).
+The assignment operators other than the `=` operator are called the ***compound assignment operators***. These operators perform the indicated operation on the two operands, and then except for `??=`, assign the resulting value to the variable, property, or indexer element given by the left operand. For the `??=` operator, only if the left-operand is `null`, is the right-operand evaluated and the result assigned to the variable, property, or indexer element given by the left operand. The compound assignment operators are described in [§12.21.4](expressions.md#12214-compound-assignment).
```
Nigel-Ecma: Re-instate `= ref`:

```suggestion
The assignment operators other than the `=` and `= ref` operators are called the ***compound assignment operators***. These operators perform the indicated operation on the two operands, and then except for `??=`, assign the resulting value to the variable, property, or indexer element given by the left operand. For the `??=` operator, only if the left-operand is `null`, is the right-operand evaluated and the result assigned to the variable, property, or indexer element given by the left operand. The compound assignment operators are described in [§12.21.4](expressions.md#12214-compound-assignment).
```

standard/expressions.md
```diff
@@ -6671,7 +6672,22 @@ The ref assignment operator shall not read the storage location referenced by th
...
+   ```csharp
+   var tmp = a.GetValueOrDefault();
+   if (!a.HasValue) { tmp = b; a = tmp; }
+   tmp
+   ```
+
+   Except that `a` is only evaluated once.
+4. Otherwise, the type of `a ??= b` is `A`. `a ??= b` is evaluated at runtime as `a ?? (a = b)`, except that `a` is only evaluated once.
+
+Otherwise, an operation of the form `x «op»= y` is processed by applying binary operator overload resolution ([§12.4.5](expressions.md#1245-binary-operator-overload-resolution)) as if the operation was written `x «op» y`. Then
```
Nigel-Ecma: Following @BillWagner’s excellent suggestion:

```suggestion
An operation of the form `x ??= y` has the same meaning as `x ?? (x = y)`, except that `x` is only evaluated once.

Otherwise, an operation of the form `x «op»= y` is processed by applying binary operator overload resolution ([§12.4.5](expressions.md#1245-binary-operator-overload-resolution)) as if the operation was written `x «op» y`. Then
```

This avoids dealing here with any issues of definite assignment or conversions as they are covered by the semantics of `x ?? (x = y)`.

Nigel-Ecma (changes requested):
Two changes suggests, on adding back `= ref` and the other based @BillWagner’s excellent suggestion.

standard/expressions.md
Nigel-Ecma: @BillWagner – Excellent 👍! Review based on this and another fix just added.

standard/expressions.md
jskeet: I think the "except for `??=`" is in the wrong place here. My suggestion would be change the second sentence to:

> Except for `??=`, these operators perform the indicated operation on the two operands, and then assign the resulting value to the variable, property, or indexer element given by the left operand.

Alternatively we could separate this into bullet points.

standard/expressions.md
jskeet: Can we not just use `b` instead of `B` here, so we're using "conversion from expression" instead of "conversion from type"? That might also make things easier with expressions which don't have a type, too:

```csharp
using System;

delegate void MyAction(int x);

class Program
{
    static void Main()
    {
        MyAction? d = null;
        d ??= x => Console.WriteLine(x);
    }
}
```

jskeet (commented):
I'm mostly happy with this - nits that we can either fix in the meeting or create separate issues for, but unless I'm missing something big I think we should be able to merge the PR during the meeting.

standard/expressions.md
BillWagner: I wanted to see Jon's suggestion of bullets here so we can review in the meeting. On reading it, I think the edit to the second sentence avoids duplication better.

```suggestion
The assignment operators other than the `=` and `= ref` operator are called the ***compound assignment operators***. These operators are processed as follows:

- For the `??=` operator, only if the left-operand is `null`, is the right-operand evaluated and the result assigned to the variable, property, or indexer element given by the left operand.
- Otherwise, the indicated operation is performed on the two operands, and then assign the resulting value to the variable, property, or indexer element given by the left operand. The compound assignment operators are described in [§12.21.4](expressions.md#12214-compound-assignment).
```

standard/expressions.md
BillWagner: This suggestion handles most of the issue. @jnm2 and @jskeet to review and update to handle the case of `a ??= b` where `a` is `int?` and `b` is `int`.

standard/expressions.md
```diff
@@ -6671,7 +6675,22 @@ The ref assignment operator shall not read the storage location referenced by th
...
 
-An operation of the form `x «op»= y` is processed by applying binary operator overload resolution ([§12.4.5](expressions.md#1245-binary-operator-overload-resolution)) as if the operation was written `x «op» y`. Then,
+For operator `??=`, given `a ??= b`, where `A` is the type of `a`, `B` is the type of `b`, and `A0` is the underlying type of `A`, if `A` is a nullable value type:
+
+1. If `A` does not exist or is a non-nullable value type, a compile-time error occurs.
+2. If `B` is not implicitly convertible to `A` or `A0` (if `A0` exists), a compile-time error occurs.
```
```suggestion
2. If `b` is not implicitly convertible to `A` or `A0` (if `A0` exists), a compile-time error occurs.
```

standard/expressions.md
```diff
@@ -6671,7 +6675,22 @@ The ref assignment operator shall not read the storage location referenced by th
...
+3. If `A0` exists and `B` is implicitly convertible to `A0`, and `B` is not dynamic, then the type of `a ??= b` is `A0`. `a ??= b` is evaluated at runtime as:
+
+   ```csharp
+   var tmp = a.GetValueOrDefault();
+   if (!a.HasValue) { tmp = b; a = tmp; }
+   tmp
+   ```
+
+   Except that `a` is only evaluated once.
+4. Otherwise, the type of `a ??= b` is `A`. `a ??= b` is evaluated at runtime as `a ?? (a = b)`, except that `a` is only evaluated once.
```
jnm2: The type of `a ??= b` is not always `A`. When `A` is a nullable value type and the type of `b` is a non-nullable value type, the type of `a ??= b` is the non-nullable form of `A`.

```cs
class C
{
    public void M(int? a, byte b)
    {
        int x1 = (a ??= b); // 'int'
        int x2 = (a ?? (a = b)); // 'int?' (fails unless x2 is changed to 'int?')
        int x3 = (a ?? (a = b).Value); // 'int'
    }
}
```

standard/expressions.md
jskeet: Resolving this having changed `B` to `b`.

standard/expressions.md
```diff
@@ -6475,7 +6475,10 @@ The `=` operator is called the ***simple assignment operator***. It assigns the
...
 
-The assignment operators other than the `=` and `= ref` operators are called the ***compound assignment operators***. These operators perform the indicated operation on the two operands, and then assign the resulting value to the variable, property, or indexer element given by the left operand. The compound assignment operators are described in [§12.21.4](expressions.md#12214-compound-assignment).
+The assignment operators other than the `=` and `= ref` operator are called the ***compound assignment operators***. These operators are processed as follows:
+
+- For the `??=` operator, only if the left-operand is `null`, is the right-operand evaluated and the result assigned to the variable, property, or indexer element given by the left operand.
+- Otherwise, the indicated operation is performed on the two operands, and then assign the resulting value to the variable, property, or indexer element given by the left operand. The compound assignment operators are described in [§12.21.4](expressions.md#12214-compound-assignment).
```
```suggestion
- Otherwise, the indicated operation is performed on the two operands, and then the resulting value is assigned to the variable, property, or indexer element given by the left operand. The compound assignment operators are described in [§12.21.4](expressions.md#12214-compound-assignment).
```

RexJaeschke:
On the 2025-05-14 TG2 call: 

Meeting recording time: 69-minute mark

**Action**: Joseph and Jon to look further at the comments re casting, conversion, and such

standard/expressions.md
jskeet: Just looking - this is what point 3 above is about (although it should be a conversion from `b` rather than `B`).

standard/expressions.md
jnm2: Point 3 expresses exactly that case. We didn't think of anything better than keeping the original form here, except for two things:

- Removing references to `B` since we need to support expressions that don't have a type, like `null`
- Simplifying the evaluation part of point 3 to parallel the evaluation part of point 4 more closely.

jnm2:
```suggestion
For operator `??=`, given `a ??= b`, where `A` is the type of `a` and `A0` is the underlying type of `A`, if `A` is a nullable value type:

1. If `A` does not exist or is a non-nullable value type, a compile-time error occurs.
2. If `b` is not implicitly convertible to `A` or `A0` (if `A0` exists), a compile-time error occurs.
3. If `A0` exists and `b` is implicitly convertible to `A0`, and the type of `b` is not dynamic, then the type of `a ??= b` is `A0`. `a ??= b` is evaluated at runtime as `a ?? (A0)(a = b)`, except that `a` is only evaluated once.
4. Otherwise, the type of `a ??= b` is `A`. `a ??= b` is evaluated at runtime as `a ?? (a = b)`, except that `a` is only evaluated once.

Otherwise, an operation of the form `x «op»= y` is processed by applying binary operator overload resolution ([§12.4.5](expressions.md#1245-binary-operator-overload-resolution)) as if the operation was written `x «op» y`. Then
```

standard/expressions.md
```diff
@@ -6671,7 +6675,22 @@ The ref assignment operator shall not read the storage location referenced by th
...
 
-An operation of the form `x «op»= y` is processed by applying binary operator overload resolution ([§12.4.5](expressions.md#1245-binary-operator-overload-resolution)) as if the operation was written `x «op» y`. Then,
+For operator `??=`, given `a ??= b`, where `A` is the type of `a`, `B` is the type of `b`, and `A0` is the underlying type of `A`, if `A` is a nullable value type:
+
+1. If `A` does not exist or is a non-nullable value type, a compile-time error occurs.
```
jnm2: A compile-time error occurs also for pointer types.

standard/expressions.md
jnm2: Ah, I see. I updated the other thread per our email conversation.

standard/expressions.md
```suggestion
1. If `A` does not exist or is a non-nullable value type or pointer type, a compile-time error occurs.
```

standard/expressions.md
```diff
@@ -6671,7 +6675,22 @@ The ref assignment operator shall not read the storage location referenced by th
...
+3. If `A0` exists and `B` is implicitly convertible to `A0`, and `B` is not dynamic, then the type of `a ??= b` is `A0`. `a ??= b` is evaluated at runtime as:
+
+   ```csharp
+   var tmp = a.GetValueOrDefault();
+   if (!a.HasValue) { tmp = b; a = tmp; }
+   tmp
+   ```
+
+   Except that `a` is only evaluated once.
+4. Otherwise, the type of `a ??= b` is `A`. `a ??= b` is evaluated at runtime as `a ?? (a = b)`, except that `a` is only evaluated once.
```
```suggestion
3. If `A0` exists and `b` is implicitly convertible to `A0`, and `B` is not dynamic, then the type of `a ??= b` is `A0`.
4. Otherwise, the type of `a ??= b` is `A`.

`a ??= b` is evaluated at runtime as `(T) (a ?? (a = b))`, where `T` is the type of the expression as determined above, except that `a` is only evaluated once.
```

standard/expressions.md
jnm2: Resolved with https://github.com/dotnet/csharpstandard/pull/609#discussion_r2104064589

Nigel-Ecma: Is there any case where `A` does not exist? This is the LHS of an assignment operator and from §12.21.1 we have the "left operand of an assignment shall be an expression classified as a variable, or, except for `= ref`, a property access, an indexer access, an event access or a tuple" .

This same change to pointer type probably needs to be made to §12.15 1st bullet which currently says "If A exists and is an unmanaged type ([§8.8](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/types.md#88-unmanaged-types)) or known to be a non-nullable value type, a compile-time error occurs." as unmanaged includes pointers and nullable value types

standard/expressions.md
Nigel-Ecma: Stating “is evaluated at runtime as” makes it a requirement, elsewhere the phrases used are “is (semantically) equivalent to” and “has the same meaning as” and something similar should be used here. (But see seperate question following first.)

Nigel-Ecma: Question: After the changes proposed to points 1 thru 4 do they all combine to:

> `a ??= b` is equivalent to `(T) (a ?? (a = b))` where `T` is the type of `a ?? b`.

Is this simply wrong, or is there some case this does not cover?

jskeet: The intention is that 1-4 now cover validity and expression type, so the evaluation can be covered separately. The complexity of 3 being evaluated in one way and 4 being evaluated in a different way felt unnecessary

Nigel-Ecma: @jskeet I’m asking whether *all* of 1-4 can be removed and replaced by the suggestion, which I now see I missed “only once” from so it should be:

> `a ??= b` is equivalent to `(T) (a ?? (a = b))` where `T` is the type of `a ?? b`, except that `a` is evaluated only once.

Of something like that. This follows the other assignment operators in pushing the details to the existing non-assignment operator and equals, avoid repetition and having to keep two places in sync.

Except of course if the equivalence does not hold…

jskeet: Ah - the problem is that we need to establish `T` beforehand, and that's what 3 and 4 do... with the "well if we can be sure it's not null, the result type is not null" part being the bit that's slightly unusual.

standard/expressions.md
```diff
@@ -6475,7 +6475,10 @@ The `=` operator is called the ***simple assignment operator***. It assigns the
 
 The operator `= ref`  is called the ***ref assignment operator***. It makes the right operand, which shall be a *variable_reference* ([§9.5](variables.md#95-variable-references)), the referent of the reference variable designated by the left operand. The ref assignment operator is described in [§12.21.3](expressions.md#12213-ref-assignment).
 
-The assignment operators other than the `=` and `= ref` operators are called the ***compound assignment operators***. These operators perform the indicated operation on the two operands, and then assign the resulting value to the variable, property, or indexer element given by the left operand. The compound assignment operators are described in [§12.21.4](expressions.md#12214-compound-assignment).
+The assignment operators other than the `=` and `= ref` operator are called the ***compound assignment operators***. These operators are processed as follows:
+
+- For the `??=` operator, only if the left-operand is `null`, is the right-operand evaluated and the result assigned to the variable, property, or indexer element given by the left operand.
```
Nigel-Ecma: The left operand cannot be the literal `null` (unlike for `??`), so this should maybe say something about the value being null to avoid confusion .

standard/expressions.md
jnm2: > Question: After the changes proposed to points 1 thru 4 do they all combine to:
> 
> > `a ??= b` is equivalent to `(T) (a ?? (a = b))` where `T` is the type of `a ?? b`.
> 
> Is this simply wrong, or is there some case this does not cover?

I think this is a great simplification that covers all cases except when `b` is dynamic.

```cs
void M(int? a, dynamic b)
{
    var x = a ?? b; // type: dynamic
    var y = a ??= b; // type: 'int?'
} 
```

So how about:
> `a ??= b` is equivalent to `(T) (a ?? (a = b))`, where `T` is the type of `a` when the type of `b` is dynamic and otherwise `T` is the type of `a ?? b`, except that `a` is evaluated only once.

jnm2: The nice thing is that this special case for dynamic makes good sense. `a ??= b` should differ from `a ?? b` here because unlike `a ?? b`, the resulting "combined value" is passing back _through_ `a`.

standard/expressions.md
```suggestion
- For the `??=` operator, only if the value of the left-operand is `null`, is the right-operand evaluated and the result assigned to the variable, property, or indexer element given by the left operand.
```

standard/expressions.md
jskeet: Resolving as we're trying to get rid of this already, and it's harder to see what's going on while this conversation is present - we can unresolve if the imminent suggestion is rejected.

standard/expressions.md
jskeet: Again, resolving for clarity - but basically the plan is to use @jnm2's last suggestion.

```suggestion
`a ??= b` is equivalent to `(T) (a ?? (a = b))`, except that `a` is evaluated only once, where `T` is the type of `a` when the type of `b` is dynamic and otherwise `T` is the type of `a ?? b`.
```

We might want to tweak the "Otherwise" in the following paragraph, but it's at worst "okay".

standard/expressions.md
jskeet: Addressed in a suggestion below.

standard/expressions.md
jnm2: Moving the 'except' sounds fine to me.

jskeet: Updated now.

jskeet:
We now have a couple more suggestions on this PR. I'd considered trying to get this merged before the next meeting, but I think it'll be fine to look at it briefly in person - I do expect to get this merged during the meeting though, unless we suddenly come up with nasty new situations.

Nigel-Ecma (approved):
LGTM

jskeet (approved):
No suggestions or comments left - let's merge it.

PR #611, opened 2022-08-11
Finalize V6.

BillWagner:
Change readme text to reflect that V6 is now the standard text.

Similar changes to #610.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,9 @@ This project has adopted the code of conduct defined by the Contributor Covenant
 
 ## C# Language Specification
 
-### C# 6.0 draft
+### C# 6.0 standard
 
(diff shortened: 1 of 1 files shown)


RexJaeschke (approved):

KalleOlaviNiemitalo:
I didn't notice the ECMA approval of C# 6 before now.

<https://github.com/dotnet/csharpstandard/releases> might not make sense if you're going to keep editing administrivia, but would it be useful to post and pin an announcement issue?

BillWagner:
@KalleOlaviNiemitalo 

> https://github.com/dotnet/csharpstandard/releases might not make sense if you're going to keep editing administrivia, but would it be useful to post and pin an announcement issue?

Well, we don't *intend* to keep updating administrivia for V6 now that ECMA has approved it. I plan on making a release once I change the branch name.

PR #612, opened 2022-08-15
Fix typos

pkulikov:
Will it be possible to publish that fix as well?
This is in [11.7.18 The checked and unchecked operators](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#11718-the-checked-and-unchecked-operators) section


BillWagner (approved):
Thanks @pkulikov 

This LGTM.

I'll :shipit: now.

BillWagner:
Tagging @RexJaeschke @jskeet 

If we want to patch the V6 standard, we should add these fixes.

PR #615, opened 2022-08-30
Add bolding to data matrix tables

gewarren:
This is necessary to prevent build suggestions when these files are pulled into the dotnet/docs repo.

"Column headers are required for tables. Add appropriate header text, or remove any extra table cells that cause an extra column."

@BillWagner 

Contributes to https://github.com/dotnet/docs/issues/30399.


Nigel-Ecma (approved):
This fixes an inconsistency with the previous three tables for the multiplicative operators so approved. What benefits or otherwise it has on "build suggestions" I've no idea.

Issue #614, opened 2022-08-17
Should we have a way of compiling/testing the code in the standard?


jskeet:
It would be nice if we could validate that code in the standard compiles (where it's meant to) and has the output we claim it has.

This comes with all kinds of challenges, such as where "extra" code (which is described briefly but not provided in the spec) should live, how we test it etc.

This issue is basically for discussion of what we want to do, and how we might achieve it.

svick:
One option is to copy what .Net docs do: have a regular project (with .csproj and .cs files) for a group of snippets, where the code looks e.g. [like this](https://github.com/dotnet/docs/blob/194ad61/docs/core/tutorials/snippets/with-visual-studio-6-0/csharp/Program.cs):

```c#
namespace HelloWorld;

public class HelloWorld
{
    public static void Main(string[] args)
    {
        // <MainMethod>
        Console.WriteLine("What is your name?");
        var name = Console.ReadLine();
        var currentDate = DateTime.Now;
        Console.WriteLine($"{Environment.NewLine}Hello, {name}, on {currentDate:d} at {currentDate:t}!");
        Console.Write($"{Environment.NewLine}Press any key to exit...");
        Console.ReadKey(true);
        // </MainMethod>
    }
}
```

The relevant part of the code is then included in the text using [non-standard Markdown](https://docs.microsoft.com/en-us/contribute/code-in-docs#in-repo-snippet-references), [like this](https://github.com/dotnet/docs/blob/194ad61/docs/core/tutorials/with-visual-studio.md?plain=1#L104):

```
:::code language="csharp" source="./snippets/with-visual-studio-6-0/csharp/Program.cs" id="MainMethod":::
```

This has the advantage that it should make things easy when displaying the spec on the .Net docs site and that there are some existing tools that help with this (mainly VS Code integration). I don't know what effect this would have on the rest of the tooling.

jskeet:
@svick: Possibly. Although personally, for ease of editing and reading the raw Markdown spec, I think I'd like it to be the other way round in this case - that the raw code is in the spec, and then it's extracted when we want to run it.

markjulmar:
@jskeet The approach laid out by @svick is related to how docs.microsoft.com processes and publishes Markdown content - so it relies on a Markdig extension built by the docs team. It would only work cleanly if this repo were published to docs and not intended to be read directly from GitHub. I don't believe that is the case - is that correct?

jskeet:
[...]

PR #623, opened 2022-09-09
First pass at tooling to allow examples to be tested

jskeet:
Very rough and ready, and will need to be expanded (e.g. for expected exceptions) but it's enough to at least start a discussion.

Not yet tested on non-Windows platforms.

Towards #614.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -68,6 +68,7 @@ When a non-abstract class is derived from an abstract class, the non-abstract cl
 
 > *Example*: In the following code
 >
+> <!-- Example: {template:"standalone-lib", name:"AbstractMethodImplementation"} -->
 > ```csharp
 > abstract class A
 > {
@@ -186,6 +187,7 @@ When a *class_type* is included in the *class_base*, it specifies the direct bas
 
 > *Example*: In the following code
 >
+> <!-- Example: {template:"standalone-lib", name:"DirectBaseClass"} -->
 > ```csharp
 > class A {}
 > class B : A {}
@@ -199,6 +201,7 @@ For a constructed class type, including a nested type declared within a generic 
 
 > *Example*: Given the generic class declarations
 >
+> <!-- Example: {template:"standalone-lib", name:"GenericBaseClass", replaceEllipsis:true} -->
 > ```csharp
 > class B<U,V> {...}
 > class G<T> : B<string,T[]> {...}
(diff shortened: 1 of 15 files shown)


jskeet:
Note that I've deliberately chosen Json.NET as it's pleasantly (in this case) lax about JSON - so we don't need to add quotes around the property keys, for example. Nothing is set in stone though.

Sample output:

```text
Deleting old output subdirectories (13)
Loaded 2 templates
Loaded 13 examples
Processing example AbstractMethodImplementation from classes.md:72-89
Processing example DirectBaseClass from classes.md:191-193
Processing example GenericBaseClass from classes.md:205-207
Processing example TypeParameterUsedAsBaseClass from classes.md:220-230
Processing example RecursiveBaseClassSpecification from classes.md:244-250
Processing example DirectBaseClasses from classes.md:262-266
Processing example SelfBaseClass from classes.md:280-281
Processing example CircularBaseClass1 from classes.md:287-290
Processing example CircularBaseClass2 from classes.md:296-301
Processing example NestedClassDependency from classes.md:313-317
Processing example DeriveFromSealedClass from classes.md:328-330
Processing example TypeParameterSubstitution from classes.md:859-882
Processing example PropertyReservedSignatures from classes.md:1295-1324
Finished example extraction.

Testing AbstractMethodImplementation from classes.md:72-89
Testing DirectBaseClass from classes.md:191-193
Testing GenericBaseClass from classes.md:205-207
Testing TypeParameterUsedAsBaseClass from classes.md:220-230
Testing RecursiveBaseClassSpecification from classes.md:244-250
Testing DirectBaseClasses from classes.md:262-266
Testing SelfBaseClass from classes.md:280-281
Testing CircularBaseClass1 from classes.md:287-290
Testing CircularBaseClass2 from classes.md:296-301
Testing NestedClassDependency from classes.md:313-317
Testing DeriveFromSealedClass from classes.md:328-330
Testing TypeParameterSubstitution from classes.md:859-882
Testing PropertyReservedSignatures from classes.md:1295-1324

Tests: 13
Failures: 0
```

BillWagner (approved):
This is a good start @jskeet 

Let's :shipit:

jskeet:
Just before we merge, could you double check (maybe perhaps with @RexJaeschke or @Nigel-Ecma as well) that the few changes I've made *inside the examples* are okay? I think just around line 224 in classes.md would be the bit to check.

RexJaeschke (approved):

Nigel-Ecma (commented):
I don't see any issue with the renaming in the code ~224 in classes.md

As to the rest I've more questions than answers at present. I haven't looked at the tool's code. However I noted in emails that this hasn't been tested cross-platform yet – is that important for this? As Microsoft error codes are embedded in the MD this isn't going to be cross-compiler, but I'd say that's fine for a tool we're running internally. I do wonder though whether we should be embedding Microsoft specific error codes in the Standard's MD source, or at least documenting that these are not part of the Standard _per se_ but just there for TG2 internal tooling?

jskeet:
> However I noted in emails that this hasn't been tested cross-platform yet – is that important for this? 

It will be, yes - because we'll want to run it as a GitHub action. I don't *anticipate* it being a problem, but we'll want to test it.

> I do wonder though whether we should be embedding Microsoft specific error codes in the Standard's MD source, or at least documenting that these are not part of the Standard per se but just there for TG2 internal tooling?

I suggest we document in a README either in the top level or under `standard/`. I don't think we need to refer to it in the standard itself, given that it's only relevant for those viewing the source. (There are other comments that are only present for our Word converter already, of course.)

PR #631, opened 2022-10-13
Remove duplicate conditional-compilation example

RexJaeschke:
Bill, I chose to remove the first occurrence, but you might check that it the best solution.


BillWagner (approved):
This LGTM @RexJaeschke 

I'll :shipit: now.

PR #632, opened 2022-10-14
Fix/enhance some examples in variables.md

RexJaeschke:
Label `LABEL:` is not followed by a statement, and needs to be; null statement added
Enhanced 2xdelegate examples (by adding a parent class wrapper), so they are a better fit for the test harness


BillWagner (approved):
This LGTM @RexJaeschke 

I'll :shipit: now.

PR #633, opened 2022-10-14
Enhance some examples in conversions.md

RexJaeschke:
Tweak 3 examples to make it easier to get them through the test harness.

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -23,19 +23,25 @@ Some conversions in the language are defined from expressions to types, others f
 >
 > <!-- IncompleteExample: {template:"standalone-lib", name:"Conversions2", expectedErrors:["CS0246"], ignoredWarnings:["CS0219"]} -->
 > ```csharp
-> enum Color { Red, Blue, Green }
->
-> // The expression 0 converts implicitly to enum types
-> Color c0 = 0;
->
-> // Other int expressions need explicit conversion
-> Color c1 = (Color)1;
->
-> // Conversion from null expression (no type) to String
-> String x = null;
->
-> // Conversion from lambda expression to delegate type
-> Func<int, int> square = x => x * x;
+> class A
+> {
+> 	enum Color { Red, Blue, Green }
+> 
+> 	void M()
+> 	{
+> 		// The expression 0 converts implicitly to enum types
+> 		Color c0 = 0;
+> 
+> 		// Other int expressions need explicit conversion
+> 		Color c1 = (Color)1;
+> 
+> 		// Conversion from null expression (no type) to string
+> 		string x = null;
+> 
+> 		// Conversion from lambda expression to delegate type
+> 		Func<int, int> square = x => x * x;
+> 	}
+> }
 > ```
 >
 > *end example*
@@ -217,29 +223,31 @@ Boxing a value of a *nullable_value_type* produces a null reference if it is the
 >
 > will output the string “Box contains an `int`” on the console.
 >
-> A boxing conversion implies making a copy of the value being boxed. This is different from a conversion of a *reference_type* to type `object`, in which the value continues to reference the same instance and simply is regarded as the less derived type `object`. For example, given the declaration
+> A boxing conversion implies making a copy of the value being boxed. This is different from a conversion of a *reference_type* to type `object`, in which the value continues to reference the same instance and simply is regarded as the less derived type `object`. For example, the following 
 >
 > <!-- IncompleteExample: {template:"standalone-lib", name:"BoxingConversions4", expectedErrors:["x","x"], expectedWarnings:["x","x"]} -->
 > ```csharp
 > struct Point
 > {
 >     public int x, y;
->
+> 
 >     public Point(int x, int y)
 >     {
 >         this.x = x;
 >         this.y = y;
 >     }
 > }
-> ```
->
-> the following statements
->
-> ```csharp
-> Point p = new Point(10, 10);
-> object box = p;
-> p.x = 20;
-> Console.Write(((Point)box).x);
+>  
+> class A
+> {
+>     void M() 
+>     {
+>         Point p = new Point(10, 10);
+>         object box = p;
+>         p.x = 20;
+>         Console.Write(((Point)box).x);
+>     }
+> }
 > ```
 >
 > will output the value 10 on the console because the implicit boxing operation that occurs in the assignment of `p` to `box` causes the value of `p` to be copied. Had `Point` been declared a `class` instead, the value 20 would be output because `p` and `box` would reference the same instance.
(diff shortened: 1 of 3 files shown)


jskeet (commented):
Hmm. Not sure about change the example here. Or maybe just move the enum to the top-level instead of it being nested?

standard/conversions.md
```diff
@@ -21,19 +21,25 @@ Some conversions in the language are defined from expressions to types, others f
...
->
-> // Other int expressions need explicit conversion
-> Color c1 = (Color)1;
->
-> // Conversion from null expression (no type) to String
-> String x = null;
->
-> // Conversion from lambda expression to delegate type
-> Func<int, int> square = x => x * x;
+> class A
```
jskeet: Note: I think we're using 4-space indents, aren't we?

BillWagner:
closing and reopening to force a build.

jskeet (commented):
Again, let's leave this open until we've decided whether to create a template that has a class in so that we don't need the changes.

RexJaeschke:
@jskeet ready for another look by you

jskeet:
Will look tomorrow, thanks.

jskeet:
The branch is currently 4 commits behind draft-v7, so first I'm going to rebase it onto draft-v7, to make sure we don't get collisions.

standard/conversions.md
```diff
@@ -25,17 +25,23 @@ Some conversions in the language are defined from expressions to types, others f
 > ```csharp
 > enum Color { Red, Blue, Green }
 >
-> // The expression 0 converts implicitly to enum types
-> Color c0 = 0;
+> class A
```
jskeet: The introduction of an extra class and method here still bothers me.

If we have a new template of "library code in a class" we could stick with the original code - the statements would be introducing instance fields, and the enum would be nested. A bit weird, but legal.

Any thoughts on that?

standard/conversions.md
```diff
@@ -217,29 +223,31 @@ Boxing a value of a *nullable_value_type* produces a null reference if it is the
...
-> A boxing conversion implies making a copy of the value being boxed. This is different from a conversion of a *reference_type* to type `object`, in which the value continues to reference the same instance and simply is regarded as the less derived type `object`. For example, given the declaration
+> A boxing conversion implies making a copy of the value being boxed. This is different from a conversion of a *reference_type* to type `object`, in which the value continues to reference the same instance and simply is regarded as the less derived type `object`. For example, the following
 >
 > <!-- IncompleteExample: {template:"standalone-lib", name:"BoxingConversions4", expectedErrors:["x","x"], expectedWarnings:["x","x"]} -->
 > ```csharp
 > struct Point
 > {
 >     public int x, y;
->
+>
```
jskeet: I suspect this will cause a trailing whitespace warning.

standard/conversions.md
```diff
@@ -217,29 +223,31 @@ Boxing a value of a *nullable_value_type* produces a null reference if it is the
 >
 > will output the string “Box contains an `int`” on the console.
 >
-> A boxing conversion implies making a copy of the value being boxed. This is different from a conversion of a *reference_type* to type `object`, in which the value continues to reference the same instance and simply is regarded as the less derived type `object`. For example, given the declaration
+> A boxing conversion implies making a copy of the value being boxed. This is different from a conversion of a *reference_type* to type `object`, in which the value continues to reference the same instance and simply is regarded as the less derived type `object`. For example, the following
 >
 > <!-- IncompleteExample: {template:"standalone-lib", name:"BoxingConversions4", expectedErrors:["x","x"], expectedWarnings:["x","x"]} -->
```
jskeet: I think this can be just "Example" now?

jskeet: (I'm fixing it to be so.)

standard/conversions.md
jskeet: (Removing.)

standard/conversions.md
jskeet: Will try this in a new commit...

jskeet: Yes, this works. I'm not sure it's going to be a great long-term solution, but I think we'll want the new template anyway.

standard/conversions.md
```diff
@@ -739,44 +747,56 @@ Specifically, an anonymous function `F` is compatible with a delegate type `D`
...
+>     void M1()
+>     {
+>         D d1 = delegate { };                         // Ok
+>         D d2 = delegate() { };                       // Error, signature mismatch
+>         D d3 = delegate(long x) { };                 // Error, signature mismatch
+>         D d4 = delegate(int x) { };                  // Ok
+>         D d5 = delegate(int x) { return; };          // Ok
+>         D d6 = delegate(int x) { return x; };        // Error, return type mismatch
+>     }
+>     delegate void E(out int x);
```
jskeet: I think all of this can benefit from the "code-in-class-lib" template introduced for the earlier example.

jskeet (commented):
I've added another commit which addresses my own comments.

Please take a look, and if you're happy with this approach of "you send a PR, I amend it" then I'll do the same for the other PR.

standard/conversions.md
jskeet: (Done - which significantly reduces the changes required for this example.)

jskeet: I'd missed that the methods had been introduced as well - I've removed those, so it's declaring fields instead of locals. Seems to work! That's in the final commit.

RexJaeschke:
@jskeet @BillWagner If I've understood all comments correctly, Jon has fixed all the issues he raised. If so, it's ready for your review.

jskeet:
Will have a look on Monday.

standard/conversions.md
```diff
@@ -211,16 +211,17 @@ Boxing a value of a *nullable_value_type* produces a null reference if it is the
...
 > }
 > ```
 >
 > will output the string “Box contains an `int`” on the console.
 >
-> A boxing conversion implies making a copy of the value being boxed. This is different from a conversion of a *reference_type* to type `object`, in which the value continues to reference the same instance and simply is regarded as the less derived type `object`. For example, given the declaration
+> A boxing conversion implies making a copy of the value being boxed. This is different from a conversion of a *reference_type* to type `object`, in which the value continues to reference the same instance and simply is regarded as the less derived type `object`. For example, the following
 >
-> <!-- IncompleteExample: {template:"standalone-lib", name:"BoxingConversions4", expectedErrors:["x","x"], expectedWarnings:["x","x"]} -->
+> <!-- Example: {template:"standalone-lib", name:"BoxingConversions4"} -->
```
jskeet, line 224: We could convert this to standalone-console and test the output. But we can do that later.

jskeet (approved):
I've made a slight change to one example so that it passes, and there's one thing we might want to change later on, but I'll merge now to avoid this getting stale.

PR #641, opened 2022-10-19
Enhance some examples in documentation-comments.md

RexJaeschke:
Here are the kinds of changes I made to help with the test harness:
- changed all `// ...` to `...`, as the extractor turns these into `/*...*/`
- in D.4.3, separated the list of strings from the source that produced them by putting the list in a `console` fence instead
- In the full definition of class Point, I tweaked the `Equals` method to explicitly show I wanted reference-equality rather than value-equality, and added the `GetHashCode` method to remove some warnings
- Added a `Point` or `Myclass` wrapper class around numerous sets of class members to make them compile; also added some namespace qualifiers

### DIFF
--- a/standard/documentation-comments.md
+++ b/standard/documentation-comments.md
@@ -105,7 +105,7 @@ This tag provides a mechanism to indicate that a fragment of text within a descr
 /// </summary>
 public class Point
 {
-    // ...
+    ...
 }
 ```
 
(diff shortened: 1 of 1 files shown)


standard/documentation-comments.md
```diff
@@ -166,8 +166,8 @@ where
 **Example:**
 
 ```csharp
-class MasterFileFormatCorruptException: System.Exception { ...}
-class MasterFileLockedOpenException: System.Exception { ...}
+class MasterFileFormatCorruptException: System.Exception { ... }
```
jskeet: Could you add spaces before the colons as well?

standard/documentation-comments.md
```diff
@@ -304,16 +309,19 @@ where
...
 ```csharp
-/// <summary>This is the entry point of the Point class testing program.
-/// <para>
-/// This program tests each method and operator, and
-/// is intended to be run after any non-trivial maintenance has
-/// been performed on the Point class.
-/// </para>
-/// </summary>
-public static void Main() 
+public class Point
```
jskeet, line 319: Maybe rename this to PointTest? (It's likely not to be in Point itself.)

jskeet (approved):
Happy to go with this for now. We could create a template with a Point class if we wanted. (I'd also be happy to create a template that embedded the main example in a class whose name was specified in the metadata...)

standard/documentation-comments.md
RexJaeschke: Done

PR #642, opened 2022-10-19
Enhance/correct some examples in unsafe-code.md

RexJaeschke:
When using the test harness, I found the following:
- Fixed an error introduced when we converted to `$""` strings
- Added an unsafe wrapper around some top-level statements
- Formatted some code
- Corrected an ill-formed character literal

### DIFF
--- a/standard/unsafe-code.md
+++ b/standard/unsafe-code.md
@@ -291,7 +291,8 @@ When one pointer type is converted to another, if the resulting pointer is not c
 >     int* pi = (int*)pv;
 >     int i = *pi; // undefined
 >     *pi = 123456; // undefined
-> }> ```
+> }
+> ```
 >
 > *end example*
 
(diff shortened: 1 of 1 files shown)


standard/unsafe-code.md
```diff
@@ -283,13 +283,15 @@ When one pointer type is converted to another, if the resulting pointer is not c
...
-> ```
+> unsafe static void M()
+> {
+>     char c = 'A';
+>     char* pc = &c;
+>     void* pv = pc;
+>     int* pi = (int*)pv;
+>     int i = *pi; // undefined
+>     *pi = 123456; // undefined
+> }> ```
```
jskeet: You need a line break after `}`.

standard/unsafe-code.md
```diff
@@ -821,8 +826,8 @@ A `char*` value produced by fixing a string instance always points to a null-ter
 > 
 >     unsafe static void F(char* p)
 >     {
->         for (int i = 0; p[i] != '\\0'; ++i)
->             Console.WriteLine(p[i]);
+>         for (int i = 0; p[i] != '\0'; ++i)
+>             System.Console.WriteLine(p[i]);
```
jskeet, line 832: Please add braces around this loop body.

jskeet (approved):
Happy to merge this as-is.

PR #643, opened 2022-10-22
Enhance/correct some examples in classes.md

RexJaeschke:
As a result of using the test harness, I made the following edits:

- used `System` namespace prefixes (later, removed these)
- replaced all en-dashes with hyphens
- added a wrapper class
- changed instance method to static
- removed backticks and other spurious characters from inside examples and console output
- added accessors
- fixed escape sequences

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -2222,7 +2222,7 @@ Output parameters are typically used in methods that produce multiple return val
 >     static void Main()
 >     {
 >         string dir, name;
->         SplitPath("c:\\Windows\\System\\hello.txt", out dir, out name);
+>         SplitPath(@"c:\Windows\System\hello.txt", out dir, out name);
 >         Console.WriteLine(dir);
 >         Console.WriteLine(name);
 >     }
@@ -4320,8 +4320,8 @@ The `true` and `false` unary operators require pair-wise declaration. A compile-
 > public class IntVector
 > {
 >     public IntVector(int length) {...}
->     public int Length { get { /* ... */ } }                          // Read-only property
->     public int this[int index] { get { /* ... */ } set { /*...*/} }  // Read-write indexer
+>     public int Length { get { ... } }                       // Read-only property
+>     public int this[int index] { get { ... } set { ... } }  // Read-write indexer
 >
 >     public static IntVector operator++(IntVector iv)
 >     {
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -1013,8 +1013,6 @@ A type declared within a class or struct is called a ***nested type***. A type t
 >
 > <!-- Example: {template:"standalone-lib", name:"NestedTypes"} -->
 > ```csharp
-> using System;
->
 > class A
 > {
 >     class B
@@ -1085,7 +1083,6 @@ A nested type may hide ([§7.7.2.2](basic-concepts.md#7722-hiding-through-nestin
 >
 > <!-- Example: {template:"standalone-console", name:"Hiding", expectedOutput:["Derived.M.F"]} -->
 > ```csharp
-> using System;
 > class Base
 > {
 >     public static void M()
@@ -1126,9 +1123,6 @@ A nested type and its containing type do not have a special relationship with re
 >
 > <!-- Example: {template:"standalone-console", name:"ThisAccess", expectedOutput:["123"]} -->
 > ```csharp
->
-> using System;
->
 > class C
 > {
 >     int i = 123;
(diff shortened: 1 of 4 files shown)


standard/classes.md
```diff
@@ -874,10 +874,10 @@ All members of a generic class can use type parameters from any enclosing class,
 >     static void Main()
 >     {
 >         C<int> x1 = new C<int>(1);
->         Console.WriteLine(x1.f1);              // Prints 1
+>         System.Console.WriteLine(x1.f1);              // Prints 1
```
jskeet: I definitely want to hold off on this until we've had a discussion about it.

standard/classes.md
```diff
@@ -2177,7 +2180,7 @@ Output parameters are typically used in methods that produce multiple return val
 >     static void Main()
 >     {
 >         string dir, name;
->         SplitPath("c:\\\Windows\\\\System\\\\hello.txt", out dir, out name);
+>         SplitPath("c:\\Windows\\System\\hello.txt", out dir, out name);
```
jskeet: Might be worth using a verbatim string literal here? `SplitPath(@"c:\Windows\System\hello.txt", out dir, out name);`

standard/classes.md
```diff
@@ -4223,8 +4227,8 @@ The `true` and `false` unary operators require pair-wise declaration. A compile-
 > public class IntVector
 > {
 >     public IntVector(int length) {...}
->     public int Length { ... }           // Read-only property
->     public int this[int index] { ... }  // Read-write indexer
+>     public int Length { get { /* ... */ } }                          // Read-only property
+>     public int this[int index] { get { /* ... */ } set { /*...*/} }  // Read-write indexer
```
jskeet: Add space around the second ellipsis for consistency?

standard/classes.md
```diff
@@ -4223,8 +4227,8 @@ The `true` and `false` unary operators require pair-wise declaration. A compile-
 > public class IntVector
 > {
 >     public IntVector(int length) {...}
->     public int Length { ... }           // Read-only property
->     public int this[int index] { ... }  // Read-write indexer
+>     public int Length { get { /* ... */ } }                          // Read-only property
```
jskeet: It's slightly annoying that we can't just keep using ... for this. We may be able to add some "advanced configuration" to make it work... I'm fine with doing it like this for now, so long as we're consistent so we can find everywhere that needs it, and revert later on.

jskeet (commented):
Glad to see this is finding issues. I'd like to hold off merging this PR until we've discussed the style convention of namespaces though.

standard/classes.md
RexJaeschke: Also remove all `/*` and `*/`, as they are added by the extractor.

standard/classes.md
RexJaeschke: I think we can here. I just failed to remove the comment delimiters when pasting the source back in from the post-extraction source

jskeet (commented):
Looks good other than the points for discussion.

standard/classes.md
RexJaeschke: removed qualifier and marked "Incomplete"

RexJaeschke:
@jskeet ready for another look by you

RexJaeschke:
@jskeet @BillWagner I see in the TG2 meeting minutes from 2022-11-02, "We decided to include in the appropriate extraction templates a set of three or four commonly used namespaces (such as System, System.Collections, System.Collections.Generic, and System.Io), and to not use fully qualified names unless they are needed to support discussions on the use of such names."

I just searched the md files for explicit `using` namespace directives, and found the following:

```csharp
using System;
using System.Collections;
using System.Diagnostics;
using System.IO;
using System.Reflection;
using System.Runtime.InteropServices;
using System.Security.Permissions;
using System.Text;
using System.Threading;
```

I added this list to the top of (my private version of) the templates standalone-console, standalone-lib, and code-in-main. I then tested a few examples that needed one or more of these.

Now when these templates are used with an example that already explicitly has one or more of these directives, I get

> `warning CS0105: The using directive for '<namespace name>' appeared previously in this namespace`

If we use this approach, we can look at all the examples having explicit `using` directives, and if those are *not necessary* for explaining the example, we can remove them; otherwise, we leave them in and add `ignoredWarnings:["CS0105"]` to the metadata. And, of course, we remove *all* fully qualified prefixes from examples where they are not necessary for explaining the example.

How does that sound?

jskeet:
Assuming I've followed you correctly while cooking dinner, that sounds fine.

RexJaeschke:
@jskeet This is ready for review. Once this has been merged, I'll go over the other md files and remove the using <namespace> directives and explicit namespace prefixes, as well.

jskeet (approved):

PR #645, opened 2022-10-23
Enhance/correct some examples in expressions.md

RexJaeschke:
As a result of using the test harness, I made the following edits:

- used `System` namespace
- added a wrapper class [later, undid this]
- combined consecutive parts of code and text in a single part that would compile [later, undid this]


standard/expressions.md
```diff
@@ -71,11 +71,16 @@ Static binding takes place at compile-time, whereas dynamic binding takes place
 > *Example*: The following illustrates the notions of static and dynamic binding and of binding-time:
 >
 > ```csharp
-> object o = 5;
-> dynamic d = 5;
-> Console.WriteLine(5); // static binding to Console.WriteLine(int)
-> Console.WriteLine(o); // static binding to Console.WriteLine(object)
-> Console.WriteLine(d); // dynamic binding to Console.WriteLine(int)
+> using System;
```
jskeet: I'm not sure whether this change actually makes it more readable.

We can definitely create a template that has some common using directives and then puts the code from the example into a method in a library. I think I'd prefer that over adding the method declaration explicitly in this example, but I'm willing to be persuaded.

standard/expressions.md
```diff
@@ -680,23 +685,25 @@ When arguments are omitted from a function member with corresponding optional pa
...
 When a generic method is called without specifying type arguments, a ***type inference*** process attempts to infer type arguments for the call. The presence of type inference allows a more convenient syntax to be used for calling a generic method, and allows the programmer to avoid specifying redundant type information.
 
-> *Example*: Given the method declaration:
+> *Example*: Given the method declaration and use, it is possible to invoke the `Choose` method without explicitly specifying a type argument:
 >
 > ```csharp
 > class Chooser
 > {
->     static Random rand = new Random();
+>     static System.Random rand = new System.Random();
```
jskeet: I think it would be more idiomatic to have a `using System` directive.

standard/expressions.md
```diff
@@ -888,22 +897,25 @@ The ***inferred return type*** is determined as follows:
...
+> using System;
+>
+> class A
 > {
->    return f2(f1(value));
+>     static Z F<X,Y,Z>(X value, Func<X,Y> f1, Func<Y,Z> f2)
+>     {
+>         return f2(f1(value));
+>     }
+>     static void M()
```
jskeet, line 915: Add a blank line between methods.

standard/expressions.md
```diff
@@ -888,22 +897,25 @@ The ***inferred return type*** is determined as follows:
...
 > and the result is of type `IEnumerable<string>`.
 >
-> The following example demonstrates how anonymous function type inference allows type information to “flow” between arguments in a generic method invocation. Given the method:
+> The following example demonstrates how anonymous function type inference allows type information to “flow” between arguments in a generic method invocation. Given the following method and invocation:
 >
 > ```csharp
-> static Z F<X,Y,Z>(X value, Func<X,Y> f1, Func<Y,Z> f2)
+> using System;
+>
+> class A
```
jskeet, line 908: Again, I suspect we could have a library template for "there's already a class declaration, put the given code inside it".

standard/expressions.md
```diff
@@ -1039,27 +1051,27 @@ Given two types `T₁` and `T₂`, `T₁` is a ***better conversion target*** th
...
+> interface I2<T> { ... }
 >
-> class G1<U>
+> abstract class G1<U>
 > {
->     int F1(U u);               // Overload resulotion for G<int>.F1
->     int F1(int i);             // will pick non-generic
->     void F2(I1<U> a);          // Valid overload
->     void F2(I2<U> a);
+>     abstract public int F1(U u);               // Overload resulotion for G<int>.F1
```
jskeet: Let's fix the typo for resolution at the same time :)

standard/expressions.md
```diff
@@ -1039,27 +1051,27 @@ Given two types `T₁` and `T₂`, `T₁` is a ***better conversion target*** th
...
+>     abstract public int F1(int i);             // will pick non-generic
+>     abstract public void F2(I1<U> a);          // Valid overload
```
jskeet: I'd be tempted to add a blank line between each set of names (so both F1s together, then both F2s, and likewise for F3, F4, F5, F6 in G2.)

standard/expressions.md
```diff
@@ -2127,23 +2141,25 @@ A member initializer that specifies a collection initializer after the equals si
 
 When an initializer target refers to an indexer, the arguments to the indexer shall always be evaluated exactly once. Thus, even if the arguments end up never getting used (e.g., because of an empty nested initializer), they are evaluated for their side effects.
 
-> *Example*: The following class represents a point with two coordinates:
+> *Example*: The following shows a class that represents a point with two coordinates, and teh creation and initializatiob of an instance:
```
jskeet: teh => the
initializiob => initialization

standard/expressions.md
```diff
@@ -2152,27 +2168,37 @@ When an initializer target refers to an indexer, the arguments to the indexer sh
 > Point a = __a;
 > ```
 >
-> where `__a` is an otherwise invisible and inaccessible temporary variable. The following class represents a rectangle created from two points:
+> where `__a` is an otherwise invisible and inaccessible temporary variable. 
+>
+> The following class ahows a rectangle created from two points, and the creation and initialization of a rectangle instance:
```
jskeet: ahows => shows

jskeet: And probably rectangle => `Rectangle` at the end of the sentence

standard/expressions.md
```diff
@@ -2152,27 +2168,37 @@ When an initializer target refers to an indexer, the arguments to the indexer sh
...
 >     public Point P1 { get; set; }
 >     public Point P2 { get; set; }
 > }
-> ```
->
-> An instance of `Rectangle` can be created and initialized as follows:
 >
-> ```csharp
-> Rectangle r = new Rectangle
+> class A
```
jskeet: This is a trickier one in terms of avoiding declaring an extra class and method - but having *some* top-level code. Would like to discuss this in our next meeting.

standard/expressions.md
```diff
@@ -2189,27 +2215,34 @@ When an initializer target refers to an indexer, the arguments to the indexer sh
...
 > {
->     P1 = { X = 0, Y = 1 },
->     P2 = { X = 2, Y = 3 }
-> };
+>     static void M()
+>     {
+>         Rectangle r = new Rectangle
+>         {
+>             P1 = new Point { X = 0, Y = 1 },
```
jskeet: You've changed the code so that it doesn't demonstrate what it's meant to... these should be

```csharp
P1 = { X = 0, Y = 1 },
P2 = { X = 2, Y = 3 }
```

standard/expressions.md
```diff
@@ -2189,27 +2215,34 @@ When an initializer target refers to an indexer, the arguments to the indexer sh
...
+>     public int X { get; set; }
+>     public int Y { get; set; }
 > }
-> ```
 >
-> the following construct can be used to initialize the embedded `Point` instances instead of assigning new instances:
+> public class Rectangle
+> {
+>     public Point P1 { get; } = new Point();
+>     public Point P2 { get; } = new Point();}
```
jskeet: Line break before the }

standard/expressions.md
```diff
@@ -3832,7 +3868,7 @@ Unless one of these conditions is true, a binding-time error occurs.
 >    {
 >       if (x == null)
 >       {
->           throw new ArgumentNullException();
+>           throw new System.ArgumentNullException();
```
jskeet: We should probably discuss what convention we want for explicit use of namespaces vs using directives.

standard/expressions.md
RexJaeschke: Examples as they existed *before* I proposed changes, are inconsistent; some have a `System.` prefix and some have a `using System;`.

Some observations:
1. Following our style, using the `using <namespace>;` approach adds 2 lines, that one plus a blank line.
2. In a few cases, we use a very long prefix, such as `System.Collections.Generic.`, which is a bit intrusive in code, and may be better suited as a `using <namespace>;`.

jskeet (commented):
Thanks - let's leave the PR open until we've discussed it next week. (I'll have a whole agenda item about this...)

standard/expressions.md
RexJaeschke: removed using directive

standard/expressions.md
RexJaeschke: removed name qualifiers

standard/expressions.md
RexJaeschke: I left it as I proposed. The example is still marked "incomplete," so we can revisit it later.

standard/expressions.md
RexJaeschke: @jskeet Did we get any direction on how to handle this?

standard/expressions.md
RexJaeschke: removed qualifier, and marked as "Incomplete"

standard/expressions.md
jskeet: I would suggest creating additional files for Point and Rectangle, and specifying them as required.

RexJaeschke: Done! I removed the redundant, hard-coded 1x`Point` and 2x`Rectangle` type declarations, replacing them with the following additions to the "additional-files" folder: Point.cs, Rectangle.cs, and RectangleWithConstructorPointAllocation.cs, and using the appropriate `additionalFiles:` in the examples' metadata.

RexJaeschke:
@jskeet I believe this is ready for your review.

standard/expressions.md
```diff
@@ -687,7 +690,7 @@ When arguments are omitted from a function member with corresponding optional pa
 
 When a generic method is called without specifying type arguments, a ***type inference*** process attempts to infer type arguments for the call. The presence of type inference allows a more convenient syntax to be used for calling a generic method, and allows the programmer to avoid specifying redundant type information.
 
-> *Example*: Given the method declaration:
+> *Example*: Given the method declaration and use, it is possible to invoke the `Choose` method without explicitly specifying a type argument:
```
jskeet: Hmm... this doesn't quite read right to me, as "it is possible" suggests that it's possible to add more code to invoke the Choose method without explicitly specifying a type argument.

Is it worth just removing the sentence entirely, leaving just "Example:"? I can't remember whether every example has a description.

standard/expressions.md
```diff
@@ -1059,27 +1070,31 @@ Given two types `T₁` and `T₂`, `T₁` is a ***better conversion target*** th
...
+> interface I2<T> { ... }
 >
-> class G1<U>
+> abstract class G1<U>
 > {
->     int F1(U u);               // Overload resulotion for G<int>.F1
->     int F1(int i);             // will pick non-generic
->     void F2(I1<U> a);          // Valid overload
->     void F2(I2<U> a);
+>     abstract public int F1(U u);               // Overload resolution for G<int>.F1
```
jskeet: I'd reorder all of these to "public abstract" - and maybe make the top-level class public as well?
(I do find all the "clutter" of public abstract to be unfortunate. I'm not sure what a better solution would be.)

standard/expressions.md
```diff
@@ -1059,27 +1070,31 @@ Given two types `T₁` and `T₂`, `T₁` is a ***better conversion target*** th
...
+>     abstract public void F3(U u, V v);         // Valid, but overload resolution for
+>     abstract public void F3(V v, U u);         // G2<int,int>.F3 will fail
+>
+>     abstract public void F4(U u, I1<V> v);     // Valid, but overload resolution for
+>     abstract public void F4(I1<V> v, U u);     // G2<I1<int>,int>.F4 will fail
+>
+>     abstract public void F5(U u1, I1<V> v2);   // Valid overload
+>     abstract public void F5(V v1, U u2);
+>
+>     abstract public void F6(ref U u);          // valid overload
```
```suggestion
>     abstract public void F6(ref U u);          // Valid overload
```

standard/expressions.md
```diff
@@ -2150,7 +2167,7 @@ A member initializer that specifies a collection initializer after the equals si
 
 When an initializer target refers to an indexer, the arguments to the indexer shall always be evaluated exactly once. Thus, even if the arguments end up never getting used (e.g., because of an empty nested initializer), they are evaluated for their side effects.
 
-> *Example*: The following class represents a point with two coordinates:
+> *Example*: The following shows a class that represents a point with two coordinates, and the creation and initialization of an instance:
```
jskeet: Revert this change, given that the immediate example doesn't initialize an instance?

standard/expressions.md
```diff
@@ -2235,7 +2257,7 @@ When an initializer target refers to an indexer, the arguments to the indexer sh
 > };
 > ```
 >
-> which has the same effect as
+> This has the same effect as
```
jskeet, line 2255: Should this have metadata so that we compile it?

standard/expressions.md
```diff
@@ -2287,33 +2309,37 @@ A collection initializer consists of a sequence of element initializers, enclose
...
 >     public string Name { get; set; }
 >     public List<string> PhoneNumbers { get; } = new List<string>();
 > }
-> ```
->
-> A `List<Contact>` can be created and initialized as follows:
 >
-> ```csharp
-> var contacts = new List<Contact>
+> class A
```
jskeet, line 2318: I wonder whether we should have a convention of `class Example` or similar?

standard/expressions.md
```diff
@@ -2359,18 +2385,18 @@ Except in an unsafe context ([§22.2](unsafe-code.md#222-unsafe-contexts)), the
 
 If an array creation expression of the first form includes an array initializer, each expression in the expression list shall be a constant and the rank and dimension lengths specified by the expression list shall match those of the array initializer.
 
-In an array creation expression of the second or third form, the rank of the specified array type or rank specifier shall match that of the array initializer. The individual dimension lengths are inferred from the number of elements in each of the corresponding nesting levels of the array initializer. Thus, the expression
+In an array creation expression of the second or third form, the rank of the specified array type or rank specifier shall match that of the array initializer. The individual dimension lengths are inferred from the number of elements in each of the corresponding nesting levels of the array initializer. Thus, the initializer expression in the following declaration
 
 <!-- IncompleteExample: {template:"standalone-console", name:"ArrayCreationExpressions1"} -->
 ```csharp
-new int[,] {{0, 1}, {2, 3}, {4, 5}}
+var a = new int[,] {{0, 1}, {2, 3}, {4, 5}}
```
jskeet: I'd expect this and the example below to require semi-colons at the end.

jskeet: (But then they could be marked as complete?)

standard/expressions.md
```diff
@@ -4097,7 +4122,7 @@ Note that some conversions, such as user defined conversions, are not possible w
 >     }
 >
 >     public T G<T>(object o)
->         where T : Attribute
+>         where T : System.Attribute
```
jskeet: Do we need this change, if standalone-lib is going to have using directives?

standard/expressions.md
```diff
@@ -4692,16 +4720,29 @@ When not captured, there is no way to observe exactly how often a local variable
...
->     return result;
+>         D[] result = new D[3];
+>         int x;
+>         for (int i = 0; i < 3; i++)
+>         {
+>             x = i * 2 + 1;
+>             result[i] = () => Console.WriteLine(x);
+>         }
+>         return result;
+> }
```
```suggestion
>     }
```

standard/expressions.md
```diff
@@ -4828,11 +4882,13 @@ The simplest form of an anonymous function is one that captures no outer variabl
...
 <!-- UsingsExample: {template:"standalone-lib", name:"AnonFunctionImplementationExample1"} -->
 ```csharp
+delegate void D();
+
 class Test
 {
     static void F()
     {
-        D d = () => Console.WriteLine("test");
+        D d = () => System.Console.WriteLine("test");
```
jskeet: Again, revert this?

jskeet: (And a few changes below, at lines 4911, 4928, 4950, 4992, 5014, 5055.)

standard/expressions.md
```diff
@@ -5715,6 +5785,7 @@ A generic type `C<T>` supports the query-expression-pattern if its public member
 
 <!-- IncompleteExample: {template:"standalone-lib", name:"Query-expressionPattern", expectedErrors:["x","x"], expectedWarnings:["x","x"]} -->
 ```csharp
+delegate R Func<T1,R>(T1 arg1);
```
jskeet, line 5782: Do we need either of these declarations now, if we've got using directives for System?

standard/expressions.md
RexJaeschke: OK; I removed the lead-in sentence.

standard/expressions.md
RexJaeschke: Done

standard/expressions.md
RexJaeschke: Done

standard/expressions.md
RexJaeschke: Done

standard/expressions.md
RexJaeschke: Added metadata placeholder `<!-- Untested$Example: { ...`, so this will get processed in the next round.

standard/expressions.md
RexJaeschke: I don't have an opinion on this.

standard/expressions.md
RexJaeschke: Done, and will "complete" them in the next round.

standard/expressions.md
RexJaeschke: reversed.

standard/expressions.md
RexJaeschke: Done

standard/expressions.md
RexJaeschke: All reversed

standard/expressions.md
RexJaeschke: I'll review this in the next round when I try to complete this example.

RexJaeschke:
@jskeet I've fixed all the errors and deferred a couple of things to the next round, at which time I'll factor in the new templates and additional-files. Once this PR has been merged, I'll start that next round.

standard/expressions.md
```diff
@@ -72,11 +72,14 @@ Static binding takes place at compile-time, whereas dynamic binding takes place
 >
 > <!-- Example: {template:"standalone-console", name:"BindingTime", expectedOutput:["5","5","5"]} -->
 > ```csharp
-> object o = 5;
-> dynamic d = 5;
-> Console.WriteLine(5); // static binding to Console.WriteLine(int)
-> Console.WriteLine(o); // static binding to Console.WriteLine(object)
-> Console.WriteLine(d); // dynamic binding to Console.WriteLine(int)
+> static void M()
```
jskeet: Do we need this declaration? If we used `code-in-main` as the template, could we basically leave the example as it was?

standard/expressions.md
```diff
@@ -859,6 +864,8 @@ The ***inferred return type*** is determined as follows:
 >
 > <!-- Example: {template:"standalone-lib", name:"InferredReturnType1"} -->
 > ```csharp
+> using System.Collections.Generic;
```
jskeet: Is this now redundant?

standard/expressions.md
```diff
@@ -2287,33 +2310,37 @@ A collection initializer consists of a sequence of element initializers, enclose
 
 The collection object to which a collection initializer is applied shall be of a type that implements `System.Collections.IEnumerable` or a compile-time error occurs. For each specified element in order, normal member lookup is applied to find a member named `Add`. If the result of the member lookup is not a method group, a compile-time error occurs. Otherwise, overload resolution is applied with the expression list of the element initializer as the argument list, and the collection initializer invokes the resulting method. Thus, the collection object shall contain an applicable instance or extension method with the name `Add` for each element initializer.
 
-> *Example*:The following class represents a contact with a name and a list of phone numbers:
+> *Example*:The following shows a class that represents a contact with a name and a list of phone numbers, and the creation and initialization of a `List<Contact>`:
 >
 > <!-- Example: {template:"standalone-lib", name:"CollectionInitializers2"} -->
 > ```csharp
+> using System.Collections.Generic;
```
jskeet: Remove unnecessary using directive?

standard/expressions.md
```diff
@@ -3888,7 +3914,7 @@ Unless one of these conditions is true, a binding-time error occurs.
 <!-- markdownlint-disable MD028 -->
 
 <!-- markdownlint-enable MD028 -->
-> *Example*: The following example checks whether an argument of an unconstrained type parameter type is `null`.
+> *IncompleteExample*: The following example checks whether an argument of an unconstrained type parameter type is `null`.
```
jskeet: I don't think this was where you intended to insert `Incomplete`.

standard/expressions.md
```diff
@@ -4507,7 +4533,7 @@ Anonymous functions in an argument list participate in type inference and overlo
 >
 > <!-- Example: {template:"standalone-lib", name:"OverloadResolution1"} -->
 > ```csharp
-> class ItemList<T> : List<T>
+> class ItemList<T> : System.Collections.Generic.List<T>
```
jskeet: Do we need this qualification any more, with the addition of the using directives?

jskeet (commented):
Just a last few comments now.

standard/expressions.md
RexJaeschke: Agreed; reverted.

standard/expressions.md
RexJaeschke: Agreed; removed

standard/expressions.md
RexJaeschke: agreed; removed

standard/expressions.md
RexJaeschke: agreed; removed

standard/expressions.md
RexJaeschke: agreed; removed

RexJaeschke:
@jskeet ready for your review. I'm getting a markdownlinterror, but I don't see that as being relating to the md file content.

RexJaeschke:
@jskeet I think we're good to go with this.

jskeet (approved):

Issue #646, opened 2022-10-26
Example Extractor and Tester Tools Status


RexJaeschke:
***To add your input to the discussion of any category in this issue, do NOT create a new comment; instead, locate the discussion thread comment for that category and add your contribution to the end using the “edit” option.***

The ExampleExtractor tool extracts source code for individual examples based on the metadata in an HTML comment that precedes each example’s code block. (If an example has no such comment, or the initial comment metadata is not recognized, that example’s code is not extracted.)

While most examples have metadata that allows them to be extracted, compiled, and optionally executed, without error, some cannot yet be handled completely by the extraction and testing tools. This issue tracks the status of the different categories of support currently provided by, or needed from, the tool. Once support for a category has been provided, that category’s checkbox should be checked.

Issues relating to specific metadata categories are addressed under **Categories**. General issues are addressed under **General Issues and Observations**.

> There is a [User Guide](https://github.com/dotnet/csharpstandard/blob/draft-v7/tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md) and [grammar](https://github.com/dotnet/csharpstandard/blob/draft-v7/tools/ExampleExtractor/ExampleAnnotationGrammar.g4) for the Extraction Metadata format.

## Categories

Each category shows the number of occurrences of examples of that category in all md files as of February 14, 2023.

* [x] [**Complete example**](https://github.com/dotnet/csharpstandard/issues/646#issuecomment-1291988683): (471)
* [x] [**Incomplete example**](https://github.com/dotnet/csharpstandard/issues/646#issuecomment-1291991717): (8)
* [x] [**Implementation-defined behavior**](https://github.com/dotnet/csharpstandard/issues/646#issuecomment-1291993550): (0)
* [x] [**Detecting intended exceptions**](https://github.com/dotnet/csharpstandard/issues/646#issuecomment-1291995068): (0)
* [x] [**Multifile examples**](https://github.com/dotnet/csharpstandard/issues/646#issuecomment-1291996664): (2)
* [x] [**Example needing review**](https://github.com/dotnet/csharpstandard/issues/646#issuecomment-1291998426): (1)
* [x] [**«…» delimiters used for emphasis**](https://github.com/dotnet/csharpstandard/issues/646#issuecomment-1292000778): (0)
* [x] [**Undefined behavior**](https://github.com/dotnet/csharpstandard/issues/646#issuecomment-1292002911): (0)
[...]

PR #647, opened 2022-10-26
Add template with code in a Main method

jskeet:
This will allow for examples which are just statements. (Also adds example in 16.1 Arrays to show how to use it.)

Towards #646

### DIFF
--- a/standard/arrays.md
+++ b/standard/arrays.md
@@ -10,6 +10,7 @@ The element type of an array can itself be an array type ([§16.2.1](arrays.md#1
 
 > *Example*:
 >
+> <!-- Example: {template:"code-in-main", name:"PascalArrayDeclarations"} -->
 > ```csharp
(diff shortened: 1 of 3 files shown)


RexJaeschke:
@jskeet, I'm confused! This isn't one of the scenarios I reported in the new issue. In fact, I was able to get that example (and other examples like it) to build by using the template standalone-console, and taking advantage of the ability to have statements at the top level in recent compiler versions.

jskeet:
It may not be necessary if we're happy to use a later compiler to test C# 7, but it still feels like it's part of fixing "incomplete example". (I don't have the full context of what doesn't work as we can't see your code with the metadata, unless you've pushed that to a branch I'm not aware of. So I'm going on descriptions basically.)

RexJaeschke (approved):
Yes, it sounds like a good idea to have this new template.

PR #649, opened 2022-10-26
Remove chevrons from example code when extracting it

jskeet:
The Color Color problem code needed a tiny change to avoid the ellipsis needing to return anything, but then it works.


RexJaeschke (approved):
Looks good to me, Jon.

PR #650, opened 2022-10-26
Allow unsafe code in examples

jskeet:
Towards #646.

### DIFF
--- a/standard/unsafe-code.md
+++ b/standard/unsafe-code.md
@@ -39,6 +39,7 @@ unsafe_statement
 
 > *Example*: In the following code
 >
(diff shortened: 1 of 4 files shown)


RexJaeschke (changes requested):
That's a nice, simple solution, Jon! However, now that we have template "code-in-main", we should enable unsafe mode there as well, no? I'm not sure we have any unsafe top-level examples, but I see no reason to preclude them.

jskeet:
Yes - didn't do it before as they were independent PRs, but I'll do it today.

jskeet:
Added now - PTAL and merge if you're happy.

RexJaeschke (approved):

PR #652, opened 2022-10-27
Add example-extractor annotation to Chapters 6, 7, 8, 9, and 10

RexJaeschke:
@jskeet, for now, I'm ignoring the new template, code-in-main. Once I've moved all my annotation to the master, I'll go back and identify which examples should be changed from standalone-console to code-in-main, and handle that in a separate PR.

### DIFF
--- a/standard/basic-concepts.md
+++ b/standard/basic-concepts.md
@@ -77,6 +77,7 @@ The textual order in which names are declared is generally of no significance. I
 
 > *Example*: The declaration space of a namespace is “open ended”, and two namespace declarations with the same fully qualified name contribute to the same declaration space. For example
 >
+> <!-- Example: {template:"standalone-lib", name:"Declarations1", replaceEllipsis:true} -->
 > ```csharp
 > namespace Megacorp.Data
 > {
@@ -279,6 +280,7 @@ The accessibility domain of a nested member `M` declared in a type `T` within 
 <!-- markdownlint-enable MD028 -->
 > *Example*: In the following code
 >
+> <!-- Example: {template:"standalone-lib", name:"AccessibilityDomains",ignoredWarnings:["CS0169","CS0649"]} -->
 > ```csharp
 > public class A
 > {
@@ -326,6 +328,7 @@ As described in [§7.4](basic-concepts.md#74-members), all members of a base cla
 
 > *Example*: In the following code
 >
+> <!-- Example: {template:"standalone-lib", name:"AccessibilityDomainsNot",expectedErrors:["CS0122"],ignoredWarnings:["CS0414"]} -->
 > ```csharp
 > class A
 > {
@@ -365,6 +368,7 @@ In addition to these forms of access, a derived class can access a protected ins
 
 > *Example*: In the following code
 >
+> <!-- Example: {template:"standalone-lib", name:"ProtectedAccess1",expectedErrors:["CS1540"]} -->
 > ```csharp
 > public class A
 > {
@@ -395,6 +399,7 @@ In addition to these forms of access, a derived class can access a protected ins
 <!-- markdownlint-enable MD028 -->
 > *Example*:
 >
+> <!-- Example: {template:"standalone-lib", name:"ProtectedAccess2"} -->
 > ```csharp
 > class C<T>
 > {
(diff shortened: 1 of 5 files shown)


standard/basic-concepts.md
```diff
@@ -668,6 +680,7 @@ Name hiding through nesting can occur as a result of nesting namespaces or types
 
 > *Example*: In the following code
 >
+> <!-- Example: {template:"standalone-lib",name:"HidingNesting1", ignoredWarnings:["CS0219"],ignoredWarnings:["CS0414"]} -->
```
jskeet: I suspect second of these `ignoredWarnings` will override (replace) the first. If you want to ignore multiple warnings, specify them in the array.

```suggestion
> <!-- Example: {template:"standalone-lib",name:"HidingNesting1", ignoredWarnings:["CS0219", "CS0414"]} -->
```

standard/basic-concepts.md
```diff
@@ -935,6 +953,7 @@ The behavior of the garbage collector can be controlled, to some degree, via sta
 
 > *Example*: Since the garbage collector is allowed wide latitude in deciding when to collect objects and run finalizers, a conforming implementation might produce output that differs from that shown by the following code. The program
 >
+> <!-- Example: {template:"standalone-console",name:"MemoryManagement1", expectedOutput:["??ordering"]} -->
```
jskeet: I'd expected to see this as UndefinedExample - isn't the idea that you're adding examples with the metadata described in #646, so we can easily identify what can be fixed with tooling support for each type of problem?

standard/conversions.md
```diff
@@ -725,6 +737,7 @@ Specifically, an anonymous function `F` is compatible with a delegate type `D`
 
 > *Example*: The following examples illustrate these rules:
 >
+> <!-- Example: {template:"standalone-lib", name:"AnonymousFunctionsConv1", expectedErrors:["CS1593","CS1661","CS1678","CS8030","CS1688","CS1661","CS1676","CS1643","CS0126","CS0029","CS1662","CS0029","CS1662"]} -->
```
jskeet: If we find that lines are getting very long, I can look at how hard it would be to allow multi-line comments with metadata...

standard/lexical-structure.md
```diff
@@ -187,6 +192,7 @@ A ***delimited comment*** begins with the characters `/*` and ends with the cha
 
 > *Example*: The example
 >
+> <!-- Example: {template:"standalone-console",name:"HelloWorld1",expectedOutput:[ "hello, world"]} -->
```
jskeet:
```suggestion
> <!-- Example: {template:"standalone-console",name:"HelloWorld1",expectedOutput:["hello, world"]} -->
```

standard/lexical-structure.md
```diff
@@ -208,6 +214,7 @@ A ***single-line comment*** begins with the characters `//` and extends to the
 
 > *Example*: The example
 >
+> <!-- Example: {template:"standalone-console",name:"HelloWorld2",expectedOutput:[ "hello, world"]} -->
```
jskeet:
```suggestion
> <!-- Example: {template:"standalone-console",name:"HelloWorld2",expectedOutput:["hello, world"]} -->
```

standard/lexical-structure.md
```diff
@@ -927,6 +938,7 @@ Each string literal does not necessarily result in a new string instance. When t
 
 > *Example*: For instance, the output produced by
 >
+> <!-- Example: {template:"standalone-console",name:"ObjectReferenceEquality",expectedOutput:["True"]} -->
```
jskeet, line 941: Note to self: we're currently inconsistent about whether to include spaces between metadata values.
Let's not *try* to get consistency yet; I can easily write tooling to make everything consistent later.

jskeet (commented):
Mostly fine - it would be good if you could get it into a state where the test shell script passes, as that would prove that all the problems have metadata like IncompleteExample. (At the moment the ones with expected output of "??ordering" would fail, I believe.)

standard/basic-concepts.md
RexJaeschke: My mistake; I've made it ImplementationDefinedExample.

jskeet (approved):
These now either work or will after #655

PR #653, opened 2022-10-27
Add example-extractor annotation to Chapters 11, 12, and 13

RexJaeschke:


standard/expressions.md
```diff
@@ -682,6 +687,7 @@ When a generic method is called without specifying type arguments, a ***type inf
 
 > *Example*: Given the method declaration:
 >
+> <!-- Example: {template:"standalone-lib", name:"TypeInference", expectedErrors:["x","x"], expectedWarnings:["x","x"]} -->
```
jskeet: Presumably this doesn't pass, so please give it an appropriate "problem" prefix, e.g. IncompleteExample.

standard/expressions.md
```diff
@@ -890,6 +897,7 @@ The ***inferred return type*** is determined as follows:
 >
 > The following example demonstrates how anonymous function type inference allows type information to “flow” between arguments in a generic method invocation. Given the method:
 >
+> <!-- Example: {template:"standalone-lib", name:"InferredReturnType2", replaceEllipsis:true, expectedOutput:["x", "x", "x"], expectedErrors:["x","x"], expectedWarnings:["x","x"]} -->
```
jskeet: And again - I won't note further ones, but basically I'd like the shell script to pass on your machine before we merge.

I'd also note that if we're expecting output, standalone-lib is the wrong template :)

standard/expressions.md
```diff
@@ -3852,6 +3890,7 @@ For an operation of the form `x == y` or `x != y`, if any applicable `operat
 <!-- markdownlint-enable MD028 -->
 > *Example*: The example
 >
+> <!-- Example: {template:"standalone-console", name:"ReferenceTypeEqualityOperators2", expectedOutput:["True","False","False","False"]} -->
```
jskeet: Note to self: it would be lovely to be able to say "derive the expected output from the subsequent console output section of the spec". Once all the examples are *working*, we can see whether we're consistently using the same fencing etc.

standard/statements.md
```diff
@@ -327,6 +335,7 @@ In the context of a local variable declaration, the identifier `var` acts as a c
 
 > *Example*: The following are incorrect implicitly typed local variable declarations:
 >
+> <!-- Example: {template:"standalone-console", name:"LocalVariableDecls1", expectedErrors:["CS0818","CS0820","CS0815","CS8917","CS0841"], expectedWarnings:["CS0168"]} -->
```
jskeet: Ignore the warning? (Once everything is in, we should probably perform an audit that all the expected warnings correspond to descriptions in the text.)

standard/statements.md
```diff
@@ -1064,6 +1089,7 @@ The order in which `foreach` traverses the elements of an array, is as follows:
 
 > *Example*: The following example prints out each value in a two-dimensional array, in element order:
 >
+> <!-- Example: {template:"standalone-console", name:"ForeachStatement2", replaceEllipsis:true, expectedOutput:["1.2","2.3","3.4","4.5","5.6","6.7","7.8","8.9"]} -->
```
jskeet: This is expecting each value on a different line - I would expect this to fail at the moment.

standard/statements.md
```diff
@@ -1700,6 +1731,7 @@ using (ResourceType rN = eN)
 
 > *Example*: The example below creates a file named log.txt and writes two lines of text to the file. The example then opens that same file for reading and copies the contained lines of text to the console.
 >
+> <!-- Example: {template:"standalone-console", name:"UsingStatement", expectedOutput:["This is line one","This is line two"], expectedErrors:["x","x"], expectedWarnings:["x","x"]} -->
```
jskeet: There's no output from this example, because it's writing to a text file.

jskeet (commented):
As with the other PR, let's make sure all the tests pass before merging. (Soon we should add a GitHub action that runs the tests.)

standard/statements.md
RexJaeschke: agreed

standard/expressions.md
RexJaeschke: agreed

standard/expressions.md
RexJaeschke: agreed

standard/statements.md
RexJaeschke: fixed

standard/statements.md
RexJaeschke: agreed

jskeet:
@RexJaeschke: Please review the final commit. I've used "UsingsExample" for where we're missing using directives. (Effectively that's a subset of IncompleteExample, I guess.)

jskeet (approved):
With my final commit in place, I'm happy to merge this when Rex is happy.

RexJaeschke:
@jskeet I'm happy with this.

PR #659, opened 2022-10-30
Create ExtractorAndTesterUsersGuide

RexJaeschke:
I wrote a detailed User's Guide for annotating examples for extraction and testing. As it uses ANTLR grammar notation, I've also provided that grammar (which validates).

I might have to tweak the links later.


tools/ExampleExtractor/ExampleAnnotationGrammar.g4
```diff
@@ -0,0 +1,78 @@
+// grammar understood by the ExampleExtractor tool when processing HTML comments containing example annotations.
```
jskeet, line 1: I suspect users would find it easier to understand that it's just JSON, and then give the valid properties within the JSON. Let's discuss in the meeting though, as I'm biased.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
```diff
@@ -0,0 +1,483 @@
...
+    | replace_ellipsis
+    | expected_errors
+    | expected_warnings
+    | ignored_warnings
+    | expected_output
+    | expected_exception
+    ;
+```
+
+While an *example_annotation* must precede its corresponding example, it need not be immediately prior. The two can be separated by text and/or HTML comments. However, if multiple *example_annotation*s precede the same code block, the second and subsequent *example_annotation*s are ignored.
```
jskeet, line 53: This is an accurate description of the current state of affairs. We should discuss whether or not it's what we actually want though!

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
```diff
@@ -0,0 +1,483 @@
...
+
+Arbitrary horizontal whitespace is permitted between any two adjacent tokens.
+
+If multiple *annotation_directive*s contain the same name, the right-most directive prevails.
```
jskeet: I think it would be better if it just failed though - I'll see if there's an option in the JSON parser that would enable that.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
```diff
@@ -0,0 +1,483 @@
...
+>
+> <!-- Example: { … } -->
+> ```csharp
+> // source code goes here
+> ```
+>
+> *end example*
+````
+
+In this case, the *example_annotation* is preceded by `> ` to match the example’s formatting. However, in the following case, it is not, as no \*Example\* is present:
```
jskeet, line 72: Out of interest, have you run across these while annotating? I didn't... the processor supports it, but maybe we don't need it.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
```diff
@@ -0,0 +1,483 @@
...
+    new int[] {1, 2, 1},
+    new int[] {1, 3, 3, 1}
+};
+    }
+}
+````
+
+### Example Names
+
+An annotation’s *name* is the name of the resulting test file. This *annotation_directive* is required.
```
jskeet: And must be unique across the whole standard. (If that isn't enforced at the moment, I should enforce it.) You've included this later, but I think I'd put it up front here.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
```diff
@@ -0,0 +1,483 @@
...
+    : 'name' ':' test_filename
+    ;
+test_filename
+    : JSON_string_value
+    ;
+```
+
+The ExampleExtractor tool processes all md files in a given input folder, and writes out all corresponding test files to a given output folder. As such, ***test_filenames*** **must be unique across the whole C# spec.** (If they are not, each subsequently generated same-named file will overwrite the previous one!)
+
+In the following example, the class-library source-code is written to file “NestedClassDependency”:
```
jskeet:
```suggestion
In the following example, the class-library source-code is written to directory “NestedClassDependency”:
```

(The filename is part of the template; it's only the directory name that's changed by the name. If I've remembered correctly, at least...)

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
```diff
@@ -0,0 +1,483 @@
...
+
+## Other Information
+
+### Unsafe Code
+
+All templates support compilation of examples containing unsafe code.
+
+### Examples Containing Pseudo-Code
+
+Some examples contain C# grammar-rule names, which show a general format rather than source code that will compile. Such names are fenced using the following notation: `«rule_name»`. These examples do *not* have *example_annotation*s.
```
jskeet, line 554: These typically aren't "examples" though - they're parts of the standard explaining how it processes things. That's why I'm not too bothered by it not working with the example parser :)

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
```diff
@@ -0,0 +1,483 @@
...
+If the runtime behavior is undefined, and the annotation is incomplete, use 
+
+````
+<!-- UndefinedExample: { … } -->
+````
+
+If the example involves multiple source files, use
+````
+<!-- RequiresSeparateFilesExample: { … } -->
+````
```
jskeet, line 638: It may be worth adding that this grep command can then be used to find all such examples:

```sh
 grep -E '<!-- [A-Za-z]+Example' standard/*.md
```

jskeet (approved):
I'm happy for this to go in as-is, but I do think using ANTLR is less useful than just treating it as JSON (which is is).

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
jskeet: I've got local code that makes it fail now. Hoping to push that later today...

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
jskeet: There are examples like this in documentation comments.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
```diff
@@ -0,0 +1,483 @@
...
+    : 'expectedErrors' ':' '[' cs_number (',' cs_number)* ']'
+    ;
+cs_number
+    : JSON_string_value      // of the form "CSnnnn"
+    ;
+```
+
+The `nnnn` in `"CSnnnn"` is a 4-digit number (with leading zeros, if necessary) [as used by Microsoft’s compiler](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/) for the error message expected.
+
+Error *cs_number*s shall appear in the order in which they are output to the console during compilation. When the compiler produces both error and warning messages, they may be interspersed; however, that has no impact on the checking order of this annotation directive’s list.
```
jskeet, line 300: It turns out that the reporting order isn't deterministic, so I'm changing it to be ordered by line number then error ID.

jskeet: (That requires changing a few examples, but that's okay.)

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
```diff
@@ -0,0 +1,483 @@
...
+    new int[] {1},
+    new int[] {1, 1},
+    new int[] {1, 2, 1},
+    new int[] {1, 3, 3, 1}
+};
+    }
+}
+````
+
+### Example Names
```
jskeet, line 159: It would be good to recommend that example names are valid C# identifiers. (I've been renaming some as I go along.)

At the moment it's not strictly required, but I can see it being useful later, potentially (e.g. for namespaces).

tools/ExampleExtractor/ExampleAnnotationGrammar.g4
RexJaeschke: On the 2022-11-2 TG2 call, we agreed to keep the ANTLR notation, and Jon will add more words about JSON in a separate PR.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
RexJaeschke: OK; let's do that after we merge this PR.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
RexJaeschke: OK

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
RexJaeschke: OK, have moved this to the opening para.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
RexJaeschke: OK, fixed.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
jskeet: This was merged as part of #660, so the user guide can be updated to indicate that an error will occur at extraction time.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
RexJaeschke: Agreed.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
RexJaeschke: Done

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
RexJaeschke: @jskeet I'm not quite sure how I should change the words here.

tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md
RexJaeschke: Done

PR #660, opened 2022-10-31
Improvements to the examples tester

jskeet:
Now supports command line options:

--quiet: Only report failures
--example: Only run the given named example
--source: Only run examples from the named Markdown file

Other changes:

- Briefer output when extracting examples (one line per Markdown file)
- Detect and report duplicate examples names
- Report actual warnings/errors when they don't match (with line numbers relative to the example)
- Trim the end of actual output, to avoid having to include trailing spaces
- Order errors/warnings by line number then error code, to make the ordering deterministic

Examples have been amended accordingly.


jskeet:
There are some larger changes to come here, but I wanted to get these small ones out of the way first.

BillWagner (approved):
These all LGTM @jskeet 

Let's :shipit:

PR #661, opened 2022-10-31
Allow output to be inferred from a section after the example instead of stated explicitly

jskeet:
This allows us to test that the claimed output in the standard is *really* the output. In various cases, this wasn't actually the case...

We still need ExpectedOutput at the moment for cases where the standard doesn't have the expected output in a triple-backtick-fenced block, but we could potentially aim to improve that over time.

@RexJaeschke: I've added you to the review list just to make you aware of the functionality :)

### DIFF
--- a/tools/ExampleExtractor/Example.cs
+++ b/tools/ExampleExtractor/Example.cs
@@ -1,10 +1,16 @@
 ﻿using Newtonsoft.Json;
-using System.Net.Http.Json;
 
 namespace ExampleExtractor;
 
 internal class Example
 {
+    /// <summary>
+    /// The maximum number of lines that can occur between the end of an example and the ```console
+    /// line that marks the start of the expected output, when <see cref="ExampleMetadata.InferOutput"/>
+    /// is true.
+    /// </summary>
+    private const int MaximumConsoleOutputDistance = 8;
+
     private const string ExampleCommentPrefix = "<!-- Example: ";
     private const string CommentSuffix = " -->";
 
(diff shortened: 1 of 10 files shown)


RexJaeschke:
@jskeet. Once this PR has been merged, I'll update the user guide and grammar (in my pending PR) to reflect this new option.

tools/ExampleExtractor/Example.cs
```diff
@@ -87,6 +87,27 @@ private static IEnumerable<Example> LoadExamplesFromFile(string markdownFile)
...
 
+            if (metadata.InferOutput)
+            {
+                if (metadata.ExpectedOutput is not null)
+                {
+                    throw new InvalidOperationException($"Example {metadata.Name} has both ${nameof(metadata.InferOutput)} and {nameof(metadata.ExpectedOutput)}");
+                }
+                int openingConsoleLine = FindLineEnding(closingLine + 1, "```console");
+                // We expect the output to appear very shortly after the example.
+                if (openingConsoleLine > closingLine + 10)
```
BillWagner: A couple points:

- `10` should be a declared constant, not a literal.
- This may be longer than we need. Because our hard line breaks are at paragraphs, maybe it could be as small as 3. It might be worth experimenting.

Or, did you already test that theory?

BillWagner (approved):
This is nice @jskeet 

I had one comment that may affect its resilience over time. But it's ready to :shipit:

tools/ExampleExtractor/Example.cs
jskeet: Happy to declare a constant. I haven't tried it with smaller values. I think we should at least allow:

```text
> ```csharp
> ...
>```
>
> In this example, something happens.
>
> The result is therefore:
>
> ```console
> ...
> ```

jskeet: Will test when I get a chance, and reduce to 5 if that works.

jskeet: The `DelegateInvocation` example requires an 8 line maximum. That allows three paragraphs between the example and the expected output. I've set the constant value to 8 for the moment.

(I've amended the existing commit.)

Given your previous comment Bill, I'm going to assume you're okay with this, and merge. (I have another change that then needs to be rebased, and I'll create yet another PR...)

PR #662, opened 2022-11-01
Allow additional files to be copied as part of template extraction

jskeet:
This can fix more than just the doc comments file, but that was the one that most obviously benefitted.

(If we create Point, Rectangle and Control files, I think that will help classes.md a lot.)

### DIFF
--- a/standard/documentation-comments.md
+++ b/standard/documentation-comments.md
@@ -711,7 +711,7 @@ IDs:
 
 **Fields** are represented by their fully qualified name.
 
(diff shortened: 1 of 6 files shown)


RexJaeschke (approved):

jskeet:
Thanks Rex - will wait for Bill as well due to there being more code changes than some other features.

BillWagner (approved):
This LGTM @jskeet 

Let's :shipit:

PR #665, opened 2022-11-02
Allow ellipsis replacement to be customized

jskeet:
The replacements are performed in order, using the default of /* ... */ for null values or at the end of the list. This is primarily targeted at non-void method bodies and property getters.

### DIFF
--- a/standard/basic-concepts.md
+++ b/standard/basic-concepts.md
@@ -481,7 +481,7 @@ The following accessibility constraints exist:
 <!-- markdownlint-enable MD028 -->
 > *Example*: Likewise, in the following code
 >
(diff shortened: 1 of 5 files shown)


jskeet:
I'm not as convinced by this change as by the others - I'm *very* open to hearing about other suggestions.

RexJaeschke (approved):
@jskeet I'm OK with this proposal. That said, the draft meeting minutes I just circulated do *not* mention our discussion of this topic; should they? I don't recall TG2 had a strong opinion.

jskeet:
@RexJaeschke: Let's see what Bill thinks. I don't think folks had a strong view in the meeting, but they didn't have much time to think. I guess we could always include it to start with, but then change to a different mechanism later if we think of one.

BillWagner (approved):
This LGTM @jskeet 

Let's :shipit:

If one day, future us find a better solution, we can implement it. Until then, I think our future selves will be happy with this plan.

PR #668, opened 2022-11-07
Add tooling to reformat example metadata

jskeet:
I'm very happy to debate the precise format here, but it feels like consistency is useful :)


RexJaeschke (approved):

BillWagner (approved):
THis also LGTM @jskeet 

Let's :shipit:

PR #671, opened 2022-11-24
Add support for constraints on override methods and explicitly implemented interface methods

RexJaeschke:


BillWagner:
rebased on latest draft-v8 branch on 09/26-2023

standard/classes.md
```diff
@@ -2602,7 +2602,7 @@ A compile-time error occurs unless all of the following are true for an override
...
+- If any *type_parameter_constraints_clause*s are present, each of their *type_parameter_constraints* shall be `class` or `struct`, and for the constraint `class` in the override shall correspond to a type parameter in the base method that is known to be a non-nullable reference type. Any type parameter that has the `struct` constraint in the override shall correspond to a type parameter in the base method that is known to be a non-nullable value type. In the absence of *type_parameter_constraints_clause*s, the constraints are inherited from the overridden base method, and for the constraint `class` a parameter type `T?` is interpreted as `System.Nullable<T>`. Constraints that are type parameters in the overridden method may be replaced by type arguments in the inherited constraint. This can lead to constraints that are not valid when explicitly specified, such as value types or sealed types.
```
KalleOlaviNiemitalo: I don't understand this:

> for the constraint `class` a parameter type `T?` is interpreted as `System.Nullable<T>`

If there is a *type_parameter_constraints_clause* `where T : class`, then `System.Nullable<T>` is not even valid.  Should that instead say "for the constraint `struct`"?

standard/interfaces.md
```diff
@@ -666,8 +666,30 @@ It is not possible to access an explicit interface member implementation through
...
+> }
+>
+> class C : I
+> {
+>     void I.Foo<T>(T? value) where T : class { }
+>     void I.Foo<T>(T? value) where T : struct { }
+> }
+> ```
+>
+> Without the type parameters in the implementing methods, the compiler won’t know which method is being implemented. *end example*
```
jskeet:
```suggestion
> Without the type parameter constraints in the implementing methods, the compiler won’t know which method is being implemented. *end example*
```

jskeet (commented):
I'm trying to understand why this is only valid for class and struct constraints - is it because you can only overload on the basis of those constraints? If so, I wonder whether a note might be useful here.

Aside from Kalle's point and a suggestion from me (which might be incorrect, given my uncertainty about the feature) I think we should be able to get this merged.

BillWagner:
rebased on the dotnet/draft-v8 branch.

standard/classes.md
BillWagner: Making a suggestion:

```suggestion
- If any *type_parameter_constraints_clause*s are present, each of their *type_parameter_constraints* shall be `class` or `struct`, and for the constraint `class` in the override shall correspond to a type parameter in the base method that is known to be a non-nullable reference type. Any type parameter that has the `struct` constraint in the override shall correspond to a type parameter in the base method that is known to be a non-nullable value type. In the absence of *type_parameter_constraints_clause*s, the constraints are inherited from the overridden base method, and for the constraint `struct` a parameter type `T?` is interpreted as `System.Nullable<T>`. Constraints that are type parameters in the overridden method may be replaced by type arguments in the inherited constraint. This can lead to constraints that are not valid when explicitly specified, such as value types or sealed types.
```

standard/classes.md
```diff
@@ -2777,8 +2777,29 @@ A compile-time error occurs unless all of the following are true for an override
...
+> {
+>     public override void Foo<T>(T? value) where T : class { }
+>     public override void Foo<T>(T? value) where T : struct { }
+> }
+> ```
+>
+> Without the type parameters in the overriding methods, the compiler won’t know which base method is being overridden. *end example*
 
-An override declaration can access the overridden base method using a *base_access* ([§12.8.15](expressions.md#12815-base-access)).
+An override declaration can access the overridden base method using a *base_access* ([§12.8.14](expressions.md#12814-base-access)).
```
BillWagner: It is now 12.8.15. I think this is a merge/rebase issue:

```suggestion
An override declaration can access the overridden base method using a *base_access* ([§12.8.15](expressions.md#12815-base-access)).
```

standard/interfaces.md
```diff
@@ -660,8 +660,30 @@ It is not possible to access an explicit interface member implementation through
 
 It is a compile-time error for an explicit interface member implementation to include any modifiers ([§15.6](classes.md#156-methods)) other than `extern` or `async`.
 
-It is a compile-time error for an explicit interface method implementation to include *type_parameter_constraints_clause*s. The constraints for a generic explicit interface method implementation are inherited from the interface method.
+If *type_parameter_constraints_clause*s are present in an explicit interface method implementation, each of their *type_parameter_constraints* shall be `class` or `struct`, and for the constraint `class` in the implementation shall correspond to a type parameter in the interface method that is known to be a non-nullable reference type. Any type parameter that has the `struct` constraint in the implementation shall correspond to a type parameter in the interface method that is known to be a non-nullable value type. In the absence of *type_parameter_constraints_clause*s, the constraints for an explicit interface method implementation are inherited from the interface method, and for the constraint `class` a parameter type `T?` is interpreted as `System.Nullable<T>`.
```
BillWagner: Same as above for class -> struct:

```suggestion
If *type_parameter_constraints_clause*s are present in an explicit interface method implementation, each of their *type_parameter_constraints* shall be `class` or `struct`, and for the constraint `class` in the implementation shall correspond to a type parameter in the interface method that is known to be a non-nullable reference type. Any type parameter that has the `struct` constraint in the implementation shall correspond to a type parameter in the interface method that is known to be a non-nullable value type. In the absence of *type_parameter_constraints_clause*s, the constraints for an explicit interface method implementation are inherited from the interface method, and for the constraint `struct` a parameter type `T?` is interpreted as `System.Nullable<T>`.
```

standard/classes.md
```diff
@@ -2777,8 +2777,29 @@ A compile-time error occurs unless all of the following are true for an override
...
+>     public virtual void Foo<T>(T? value) where T : struct { }
+> }
+> class B: A
+> {
+>     public override void Foo<T>(T? value) where T : class { }
+>     public override void Foo<T>(T? value) where T : struct { }
+> }
+> ```
+>
+> Without the type parameters in the overriding methods, the compiler won’t know which base method is being overridden. *end example*
```
BillWagner: Echoing Jon's comment on line 683 in interfaces.md:

```suggestion
> Without the type parameter constraints in the overriding methods, the compiler won’t know which base method is being overridden. *end example*
```

BillWagner (approved):
This LGTM, modulo a few small comments.

standard/classes.md
```diff
@@ -2159,7 +2159,7 @@ A generic method is a method whose declaration includes a *type_parameter_list*.
 
 A generic *method_declaration* for an explicit interface member implementation shall not have any *type_parameter_constraints_clause*s; the declaration inherits any constraints from the constraints on the interface method.
 
-Similarly, a method declaration with the `override` modifier shall not have any *type_parameter_constraints_clause*s and the constraints of the method’s type parameters are inherited from the virtual method being overridden.
+Similarly, in the absence of *type_parameter_constraints_clause*s, a method declaration with the `override` modifier inherits any constraints from the virtual method being overridden. The *member_name* specifies the name of the method.
```
Nigel-Ecma: First it is no longer “Similarly” with this change, but I’m not convinced this should start “Differently” either just that its not the former.

This does not seem to require the same number of constraints, or constraints for the same type parameters, and indeed a certain compiler does allow a 2 parameter/2 constraints method to be overridden with just 1 constraint.

Some semantics are required here:

- How do the original and overridden constraint combine, or not?
- If there are constraints on *N* type parameters in the original and constraints for *M* type parameters in the overide, *M < N*, are the non-overridden constraints inherited or not?

standard/classes.md
```diff
@@ -2748,7 +2748,7 @@ A compile-time error occurs unless all of the following are true for an override
...
+- If any *type_parameter_constraints_clause*s are present, each of their *type_parameter_constraints* shall be `class` or `struct`, and for the constraint `class` in the override shall correspond to a type parameter in the base method that is known to be a non-nullable reference type. Any type parameter that has the `struct` constraint in the override shall correspond to a type parameter in the base method that is known to be a non-nullable value type. In the absence of *type_parameter_constraints_clause*s, the constraints are inherited from the overridden base method, and for the constraint `class` a parameter type `T?` is interpreted as `System.Nullable<T>`. Constraints that are type parameters in the overridden method may be replaced by type arguments in the inherited constraint. This can lead to constraints that are not valid when explicitly specified, such as value types or sealed types.
```
Nigel-Ecma: After the typo first reported by @KalleOlaviNiemitalo is fixed, I still don’t see the purpose of this, what am I missing?

Apparently you can add `T : class` in the override if T is already known to be a reference type, this would seem to widen the accepted types if the original constraint was to some specific class type.

For struct types its a bit different, as you cannot constrain to a particular struct type in the first place but can require a struct and a specific interface and drop the interface in the overide? The answer is yes in a certain compiler  but the semantics are?

Consider:

```csharp
class A { }
class B : A { }
struct S : I { }
struct R { }
interface I { }
class C
{
    public virtual void M<T, U>(T x, U y) where T : B where U : struct, I
    {

    }
}
class D : C
{
    public override void M<T, U>(T x, U y) where T : class where U : struct
    {

    }
}

static void ConstrainedOverride()
{
    var c = new C();
    var d = new D();

    c.M(new B(), new S()); // OK
    d.M(new A(), new R()); // rejected (1) No conversion from A -> B, (2) No conversion R -> S
    c.M(new A(), new R()); // rejected with same errors
}
```

So the text is missing the intended semantics, the LDM spec on this is empty, and sorry I don't have time right now to read the compiler source to see what it is implementing – which may or may not be the intended semantics of course…

Nigel-Ecma (changes requested):
The semantics are missing.

standard/classes.md
BillWagner: I checked in with @MadsTorgersen and he pointed me here: https://github.com/dotnet/csharplang/blob/main/meetings/2019/LDM-2019-04-03.md#ambiguous-implementationsoverrides-with-generic-methods-and-nrts

Excerpt:

> ### Ambiguous implementations/overrides with generic methods and NRTs
>
> ```C#
> interface I
> {
>     void Foo<T>(T value) where T : class;
>     void Foo<T>(T? value) where T : struct;
> }
> 
> class C : I
> {
>     void I.Foo<T>(T? value) { }
> }
> ```
> 
> This was valid code in C# 7 and would always match the Nullable overload because T? could only mean Nullable\<T>. Now, however, `T?` is ambiguous. The ambiguity also cannot be fixed because we do not allow you to write the constraints.
> 
> Possible fixes:
> 
> 1. The existing code means the same thing as before, namely that `T?` always means `Nullable<T>`.
If you would like to use the nullable class form, we allow exactly one constrain on the
override/implementation: `where T : class`.
> 1. Same as (1), but also allow the constraint `where T : struct`.

The resolution was to choose option 2.

standard/classes.md
BillWagner: Not spec language yet, but here is my understanding of the rules and the semantics:

- A method with `override` inherits the type parameter constraints from the `virtual` member declaration. 
- A method with `override` may include the `class` constraint on a type parameter if the `virtual` method declaration declares the `class` constraint or a base class constraint
- A method with `override` may include the `struct` constraint on a type parameter if the `virtual` method declaration declares the `struct` or `unmanaged` constraint.

A constraint added to a method with `override` clarifies the meaning of a nullable type parameter (`T?`) to be either a nullable value type or a nullable reference type. In other words, it doesn't allow or restrict valid types for the type argument.

standard/interfaces.md
BillWagner: See [comment](https://github.com/dotnet/csharpstandard/pull/671/files#r2087100688)

standard/classes.md
Nigel-Ecma: @BillWagner – Thanks for doing the research!

Nigel-Ecma: After Bill’s research I’ve dug some more and found:

1. The ability to add constraints is added in C# 8, but
2. Allowing `T?` for nullable reference types requires C# 9

So the example from the LDM is not an issue in C# 8, `T?` means nullable value type and you can override both methods without adding any constraints. If you try to add `where T : class` it errors and advises to switch to C# 9.

Is there a case in C# 8 where the constraint enables something not possible without it?

So it seems this feature needs to be split in two: v8 – add the constraints but give no reason why you might use them; v9 – add that the class constraint can be applied to `T?` and provide an example of disambiguation.

Or am I still missing something here?

jnm2: I believe this is still all in langversion 8. There's no way to override this method while retaining the nullability annotation in `List<T?>` without adding `where T : class`:

```cs
using System.IO;

class A
{
    public virtual void M<T>(List<T> param) where T : Stream
    {
    }
}

class B : A
{
    public override void M<T>(List<T?> param) where T : class
    {
    }
}
```

jnm2: Specifically this is not accurate:

> Allowing T? for nullable reference types requires C# 9

C# 8 allows `T?` on generic type parameters so long as they are constrained to `class`, `struct`, or a class constraint. What C# 9 does is allow `T?` on other generic type parameters.

jnm2 (dismissed):

standard/classes.md
jskeet: I think the "a method declaration with the `override` modifier inherits any constraints from the virtual method being overridden" is incorrect - Joseph's very first example (at about 20:45 UTC) looked like it was *not* inheriting the constraint.

RexJaeschke:
On the 2025-05-14 TG2 call: 

Meeting recording time: 55-minute mark

**Action**: Jon will further investigate Nigel’s test code.
**Action**: Nigel and Joseph will review the remaining issues.

standard/classes.md
jnm2: Here's how I'd put it: Constraints are inherited whether or not `where T : class`/`struct` appears on the overriding/implementing method. Adding `where T : class`/`struct` to an overriding/implementing method doesn't actually constrain anything, besides being required to be _compatible_ with the base/interface method. It simply controls the meaning of `T?` in the signature so that the signature lines up with the intended base member being overridden or implemented.

In absence of `where T : class`, `T?` in the signature always means `System.Nullable<T>`. This is required for backward compatibility as demonstrated at <https://github.com/dotnet/csharplang/blob/main/meetings/2019/LDM-2019-04-03.md#ambiguous-implementationsoverrides-with-generic-methods-and-nrts>.

Overall, what this feature does is enable you to use `T?` in a signature when it is a nullable _reference_ type instead of `Nullable<T>`. In order to switch `T?` from meaning `Nullable<T>` to meaning a nullable reference type, you must add `where T : class`.

Then for consistency, they allow `where T : struct` as well, even though it isn't functionally necessary.

jnm2: How's this wording? GitHub won't let me do a suggestion across deleted lines, but I think we need to update the prior statement as well, so this replaces both this paragraph and the preceding paragraph:

> A generic *method_declaration* with the `override` modifier or for an explicit interface member implementation shall not have any *type_parameter_constraints_clause*s other than the `class` constraint or the `struct` constraint. Whether or not such clauses are present, the constraints of the method’s type parameters are always inherited from the virtual method being overridden or the interface method being implemented. When such a clause is present, it must be compatible with the constraint on the same type parameter of the base or interface member, and it determines the meaning of `T?` in the signature of the overriding/implementing method in the following way:
>
> - If a `class` constraint is present on the override or explicit implementation for a type parameter `T`, it changes the meaning of `T?` in the method signature to signify a nullable reference type rather than `System.Nullable<T>`.
> - If a `struct` constraint is present, or if no constraint is present, `T?` in the method signature signifies `System.Nullable<T>` as in previous versions of the language.

jnm2: However, line 2751 below and line 663 in `interfaces.md` seem to cover this already.

Is there a way we can avoid repeating it?

standard/classes.md
jnm2: > In the absence of *type_parameter_constraints_clause*s, the constraints are inherited from the overridden base method

This part needs to change. The constraints are still inherited either way. Constraints on the overriding member are about choosing a base member, not about specifying standalone constraints. (https://github.com/dotnet/csharpstandard/pull/671/files?diff=unified&w=0#r2118047007)

image

See also <https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBLANgHwAEAGAAkIEYA6ASQHkBuAWAChXCAmUgIVYG9WpIeQDMpAG7YoGAK4BDXORSkAYhAgAeACoA+ABRaJCmTACUpAO4ALGLFKGQpAMoZYcgLaDhAlsNIBfVkC2Fk5SABFSR14WHz9CMQhxWyhsABMYJVV1bX1DcWMzSxs7B3IOLyE44WD/IA=>

jskeet:
@jnm2 and @Nigel-Ecma Do we think there'll be more to review by the next meeting?

jnm2:
@jskeet Not that I'm aware of!

jskeet:
@RexJaeschke: Any thoughts on whether we should try to revisit this next week or leave it?

RexJaeschke:
@jskeet Let's spend a minute on the call to see if anything new has been posted between now and then and/or if we can note new concerns or action items before moving to another topic.

standard/classes.md
Nigel-Ecma: Suggestion from @BillWagner committed.
Last comment from @jnm2 subsumed by their comment https://github.com/dotnet/csharpstandard/pull/671#discussion_r2118047007 below
Marking thread as resolved.

standard/classes.md
Nigel-Ecma: The confusion over C#8/9 was caused by VS messing up, everything covered applies in C#8.
Rest of this conversation is either resolved or dealt with elsewhere, marking as Resolved.

Nigel-Ecma: Immediately before the paragraph at line 2743 which introduces this list:

> A compile-time error occurs unless all of the following are true for an override declaration:

insert the paragraph:

> The overriding method inherits any *type_parameter_constraints_clause*s of the overridden base method.

This statement used to be in this bullet point but is not a “compile-time error unless…”, it is just a fact ;-)

Now reword this bullet point:

```suggestion
- A *type_parameter_constraints_clause* may only consist of the `class` or `struct` *primary_constraint*s applied to *type_parameter*s which are known according to the inherited constraints to be either non-nullable reference or non-nullable value types respectively. Any type of the form `T?` in the overriding method’s signature, where `T` is a type parameter, is interpreted as follows:
  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
  - If either there is no added constraint, or a `struct` is added, for the type parameter `T` then `T?` is a nullable value type.
 > *Note:* The purpose of these constraints is to disambiguate which method is being overridden when the signature includes nullable types and there is more than one candidate method distinguished only by type parameter constraints. *end note*
```

standard/classes.md
```diff
@@ -2159,7 +2159,7 @@ A generic method is a method whose declaration includes a *type_parameter_list*.
 
 A generic *method_declaration* for an explicit interface member implementation shall not have any *type_parameter_constraints_clause*s; the declaration inherits any constraints from the constraints on the interface method.
```
Nigel-Ecma:
```suggestion
A generic *method_declaration*, either with an `override` modifier, or for an explicit interface member implementation, inherits type parameter constraints from the overridden method or interface member respectively. Such declarations may only have *type_parameter_constraints_clause*s containing the *primary_constraint*s `class` and `struct`, the meaning of which in this context is defined in [§15.6.5](classes.md#1565-override-methods) and [§18.6.2](interfaces.md#1862-explicit-interface-member-implementations) for overriding methods and explicit interface implementations respectively.
```

standard/interfaces.md
Nigel-Ecma:
```suggestion
An explicit interface method implementation inherits any type parameter constraints from the interface.

A *type_parameter_constraints_clause* on an explicit interface method implementation may only consist of the `class` or `struct` *primary_constraint*s applied to *type_parameter*s which are known according to the inherited constraints to be either non-nullable reference or non-nullable value types respectively. Any type of the form `T?` in the signature of the explicit interface method implementation, where `T` is a type parameter, is interpreted as follows:
  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
  - If either there is no added constraint, or a `struct` is added, for the type parameter `T` then `T?` is a nullable value type.
> *Note:* The purpose of these constraints is to disambiguate which interface method is being implemented when the signature includes nullable types and there is more than one candidate interface member distinguished only by type parameter constraints. *end note*
```

Nigel-Ecma (dismissed):
The changes in this review hopefully address all the remaining issues raised and details discovered regarding the meaning of `class` and `struct` constraints added to overriding method and explicit interface member implementations.

standard/classes.md
BillWagner: This change looks great. Just as an FYI to save time, when you add xrefs, you can just use the `§18.6.2` without link formatting, and the tooling will handle the rest.

standard/classes.md
BillWagner: Looking at the comments, Nigel's suggestion does address all the comments. Let's take it and merge it.

jskeet: With Nigel's suggestion (2160), we still end up with this line (2162) as well though - is that intended?

Nigel-Ecma: @jskeet:

> With Nigel's suggestion (2160), we still end up with this line (2162) as well though - is that intended?

Nope it was an error. I can't seem to get Github to allow me to now delete the blank line between paras but the following suggestion gets rid of the para itself:

```suggestion
```

standard/classes.md
jnm2: > which are known according to the inherited constraints to be either non-nullable reference or non-nullable value types respectively.

They don't have to be known to be non-nullable reference types. For instance, here the inherited constraint is known to be a _nullable_ reference type:

```cs
class Base
{
    public virtual void M<T>() where T : class? { }
}

class Derived : Base
{
    public override void M<T>() where T : class
    {
    }
}
```

I also don't think we need to mention "non-nullable" for the value types, either, because a type parameter can't be known to be a _nullable_ value type.

Suggestion:

> to be either reference or value types respectively.

jnm2: > > *Note:* The purpose of these constraints is to disambiguate which method is being overridden when the signature includes nullable types and there is more than one candidate method distinguished only by type parameter constraints. *end note*

I'm not loving the "disambiguation" angle to describe the purpose, or the noted dilemma of "which method." The purpose still remains just as concrete when there is only one method, thus no question of "which method."

The purpose is to make it possible to use a nullable form of a reference-typed type parameter inherited from the base method. That is really the only thing that this feature allows, and it is just as important even when there would be no ambiguity to disambiguate.

standard/classes.md
jnm2: Looks great to me too! Since the first line was originally my wording, I took the liberty to edit it in the suggestion block in Nigel's comment: `with either` -> `either with`, and added commas.

standard/interfaces.md
jnm2: My suggestions on https://github.com/dotnet/csharpstandard/pull/671#discussion_r2131458898 would also be relevant here.

standard/classes.md
jnm2: Other than that, looks great to me!

jnm2 (dismissed):
Responding to @Nigel-Ecma's most recent suggestions and review request:
Looks excellent to me, with just two sticking points to discuss.

standard/classes.md
jnm2: > Is there a way we can avoid repeating it?

@Nigel-Ecma talked me out of this via email, so I'm happy with the state of things being repeated.

standard/classes.md
Nigel-Ecma: @jnm2 – re “Disambiguation”

I’m going to disagree with you on this one, disambiguation I believe is the origin of this feature – and that makes sense.

In code such as:

```cs
class Base { public virtual void M<T>(T? x) where T : class { } }

class Derived : Base { public override void M<T>(T? x) { } }
```

There is no need to require a constraint on the override, there is only one possible virtual method to override, that method constrains `T?` to be a nullable reference type, and constraints are inherited, so ipso facto the `T?` in the override is determined to be a nullable reference type.

That a certain compiler gets this wrong by apparently assuming the override’s `T?` is a nullable value type *before* it considers the inherited constraints and then reports a mismatch is not something we should standardise – this is the kind of bug/feature that is fixable without any backward compatibility issues so it *could* get fixed (the probability that it will is another matter ;-))

That said adding the *Note* is in no way required, it is just informative rationale.

standard/classes.md
```diff
@@ -2748,7 +2747,7 @@ A compile-time error occurs unless all of the following are true for an override
...
+- If any *type_parameter_constraints_clause*s are present, each of their *type_parameter_constraints* shall be `class` or `struct`, and for the constraint `class` in the override shall correspond to a type parameter in the base method that is known to be a non-nullable reference type. Any type parameter that has the `struct` constraint in the override shall correspond to a type parameter in the base method that is known to be a non-nullable value type. In the absence of *type_parameter_constraints_clause*s, the constraints are inherited from the overridden base method, and for the constraint `struct` a parameter type `T?` is interpreted as `System.Nullable<T>`. Constraints that are type parameters in the overridden method may be replaced by type arguments in the inherited constraint. This can lead to constraints that are not valid when explicitly specified, such as value types or sealed types.
```
Nigel-Ecma:
```suggestion
- A *type_parameter_constraints_clause* may only consist of the `class` or `struct` *primary_constraint*s applied to *type_parameter*s which are known according to the inherited constraints to be either reference or value types respectively. Any type of the form `T?` in the overriding method’s signature, where `T` is a type parameter, is interpreted as follows:
  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
  - If either there is no added constraint, or a `struct` is added, for the type parameter `T` then `T?` is a nullable value type.
```

standard/interfaces.md
```diff
@@ -660,8 +660,30 @@ It is not possible to access an explicit interface member implementation through
 
 It is a compile-time error for an explicit interface member implementation to include any modifiers ([§15.6](classes.md#156-methods)) other than `extern` or `async`.
 
-It is a compile-time error for an explicit interface method implementation to include *type_parameter_constraints_clause*s. The constraints for a generic explicit interface method implementation are inherited from the interface method.
+If *type_parameter_constraints_clause*s are present in an explicit interface method implementation, each of their *type_parameter_constraints* shall be `class` or `struct`, and for the constraint `class` in the implementation shall correspond to a type parameter in the interface method that is known to be a non-nullable reference type. Any type parameter that has the `struct` constraint in the implementation shall correspond to a type parameter in the interface method that is known to be a non-nullable value type. In the absence of *type_parameter_constraints_clause*s, the constraints for an explicit interface method implementation are inherited from the interface method, and for the constraint `struct` a parameter type `T?` is interpreted as `System.Nullable<T>`.
```
Nigel-Ecma:
```suggestion
An explicit interface method implementation inherits any type parameter constraints from the interface.

A *type_parameter_constraints_clause* on an explicit interface method implementation may only consist of the `class` or `struct` *primary_constraint*s applied to *type_parameter*s which are known according to the inherited constraints to be either reference or value types respectively. Any type of the form `T?` in the signature of the explicit interface method implementation, where `T` is a type parameter, is interpreted as follows:
  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
  - If either there is no added constraint, or a `struct` is added, for the type parameter `T` then `T?` is a nullable value type.
```

Nigel-Ecma (commented):
@jnm2 – I *think* I managed to beat GitHub into submission, incorporate the revised changes, and this is now a clean commit to review before we take off Draft. I’ve remove both the informative rationale notes.

standard/classes.md
```diff
@@ -2748,7 +2748,9 @@ A compile-time error occurs unless all of the following are true for an override
...
+- A *type_parameter_constraints_clause* may only consist of the `class` or `struct` *primary_constraint*s applied to *type_parameter*s which are known according to the inherited constraints to be either reference or value types respectively. Any type of the form `T?` in the overriding method’s signature, where `T` is a type parameter, is interpreted as follows:
+  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
+  - If either there is no added constraint, or a `struct` is added, for the type parameter `T` then `T?` is a nullable value type.
```
jnm2:
```suggestion
  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
  - If either there is no added constraint, or a `struct` constraint is added, for the type parameter `T` then `T?` is a nullable value type.
```

standard/interfaces.md
```diff
@@ -660,7 +660,12 @@ It is not possible to access an explicit interface member implementation through
...
+An explicit interface method implementation inherits any type parameter constraints from the interface.
+
+A *type_parameter_constraints_clause* on an explicit interface method implementation may only consist of the `class` or `struct` *primary_constraint*s applied to *type_parameter*s which are known according to the inherited constraints to be either reference or value types respectively. Any type of the form `T?` in the signature of the explicit interface method implementation, where `T` is a type parameter, is interpreted as follows:
+
+- If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
+- If either there is no added constraint, or a `struct` is added, for the type parameter `T` then `T?` is a nullable value type.
```
jnm2:
```suggestion
- If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
- If either there is no added constraint, or a `struct` constraint is added, for the type parameter `T` then `T?` is a nullable value type.
```

standard/classes.md
jnm2: > this is the kind of bug/feature that is fixable without any backward compatibility issues

There is a backward compatibility issue preventing fixing this- the one that motivated this feature in the first place. <https://github.com/dotnet/csharplang/blob/main/meetings/2019/LDM-2019-04-03.md#ambiguous-implementationsoverrides-with-generic-methods-and-nrts>:

> ```cs
> interface I
> {
>     void Foo<T>(T value) where T : class;
>     void Foo<T>(T? value) where T : struct;
> }
> 
> class C : I
> {
>     void I.Foo<T>(T? value) { }
> }
> ```
> 
> This was valid code in C# 7 and would always match the Nullable overload because T? could only mean Nullable\<T>. Now, however, `T?` is ambiguous. The ambiguity also cannot be fixed because we do not allow you to write the constraints.

In order to keep this valid C# 7 code compiling, there _must_ be a rule that says we assume `T?` to be a nullable value type unless you step in with an explicit `where T : class`. This assumption of the current compiler is therefore not a bug, and not fixable.

Therefore I'm not sure how we can really say the compiler is getting this wrong. And so, I'm not sure how we can say there's no need to require `where T : class` when there is only one base method. And so, I'm not sure we can say this really comes down to disambiguation.

In a world where we didn't have this backward-compatibility burden, the rule could be changed to auto-resolve if there is just one base method, and to be ambiguous if there are two base methods. But there's even more to this: I see the constraints-for-disambiguation-only paradigm as a problematic design. One benefit of the current design is that you can understand whether or not `T?` means `System.Nullable<T>` or reference-typed `T?` even if you are unable to resolve the reference to the base method or class. So you know whether `someT.Value` is referring to the `Value` member of `System.Nullable<T>` versus a `Value` property on an interface that `T` is constrained to. Or let's say you're reading code on GitHub or some similar platform, outside an IDE, and you want to know what `T?` means without navigating to another file to see the base method declaration.

This benefit would be lost if this were merely about disambiguation.

standard/classes.md
```diff
@@ -2777,6 +2779,27 @@ A compile-time error occurs unless all of the following are true for an override
...
+>     public virtual void Foo<T>(T? value) where T : struct { }
+> }
+> class B: A
+> {
+>     public override void Foo<T>(T? value) where T : class { }
+>     public override void Foo<T>(T? value) where T : struct { }
+> }
+> ```
+>
+> Without the type parameter constraints in the overriding methods, the compiler won’t know which base method is being overridden. *end example*
```
jnm2: I feel that these two explanatory lines (one here, one in the interfaces example) are misleading. 

It's not about not knowing which base method is being overridden; it's about `T?` implying `where T : struct` unless you take control and manually specify `where T : class`. (Or later, in C# 9, `where T : default` as a third meaning.)

See <https://github.com/dotnet/csharpstandard/pull/671#discussion_r2151047975> for a more in-depth comment.

jnm2 (changes requested):
I'm very happy with all of this, except I'm worried about the explanatory line at the end of each example being misleading as to what is really going on.

standard/classes.md
jnm2: In retrospect, a "perfect language" may have made it illegal to use `T?` in an override or explicit implementation unless you do have `where T : struct`/`class`. (Or in C# 9, `where T : default` becomes a third distinct option.)

jnm2: There's a wrinkle I was missing which Nigel pointed out, which is that backward compatibility could still be kept while still permitting this to compile:

```cs
class Base
{
    public virtual void Foo<T>(T? value) where T : class { }
}
 
class Derived : Base
{
    public override void Foo<T>(T? value) { }
}
```

However, I worry it jumps too far to say that it's a bug that the compiler's current behavior is to reject this program. By rejecting this program, it enables a library to add a matching `where T : struct` overload without breaking the code of library consumers. We may even want to allow that the current behavior may have been deliberately chosen for this advantage.

Either way, `where T : struct` and `where T : class` do not perform disambiguation, in the current state of things. There is no ambiguity when they are absent.

standard/classes.md
jnm2: Without the type parameter constraints, the compiler _will_ know which base method is being overridden (the `where T : struct` one).

```suggestion
> Without the type parameter constraint `where T : class`, the base method with the reference-typed type parameter cannot be overridden. *end example*
```

standard/interfaces.md
```diff
@@ -660,8 +660,35 @@ It is not possible to access an explicit interface member implementation through
...
+> }
+>
+> class C : I
+> {
+>     void I.Foo<T>(T? value) where T : class { }
+>     void I.Foo<T>(T? value) where T : struct { }
+> }
+> ```
+>
+> Without the type parameter constraints in the implementing methods, the compiler won’t know which method is being implemented. *end example*
```
jnm2: Without the type parameter constraints, the compiler _will_ know which base method is being overridden (the `where T : struct` one).

```suggestion
> Without the type parameter constraint `where T : class`, the base method with the reference-typed type parameter cannot be overridden. *end example*
```

jnm2 (dismissed):

Nigel-Ecma:
All comments/suggestions resolved, only the final PR changes to the 2 files remain.
Taking off Draft and inviting final review.
[@jnm2 & @Nigel-Ecma]

jskeet (approved):
This is a lot clearer than before - thanks!

BillWagner (approved):
This new version LGTM.

gafter (approved):

jnm2 (approved):

PR #673, opened 2022-11-27
Add support for readonly members in structs

RexJaeschke:


KalleOlaviNiemitalo:
This will also need to update the defensive copying rule like in <https://github.com/dotnet/csharpstandard/issues/894>.

*Tracked as <https://github.com/dotnet/csharpstandard/issues/928>*

standard/structs.md
```diff
@@ -502,6 +505,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
 
 > *Note*: This access restriction means that constructors in structs containing automatically implemented properties often need an explicit constructor initializer where they would not otherwise need one, to satisfy the requirement of all fields being definitely assigned before any function member is invoked or the constructor returns. *end note*
 
+A *property_declaration* ([§14.7.1](classes.md#1471-general)) for an instance property in a *struct_declaration* may contain the *property_modifier* `readonly`. However, a static property shall not contain that modifier.
```
KalleOlaviNiemitalo: Does _"However, a static property shall not contain that modifier"_ also cover the case where the `readonly` modifier is on an accessor of the static property?

```
struct S {
    static int P {
        readonly get => 0;
        set {}
    }
}
```

standard/structs.md
```diff
@@ -502,6 +505,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
...
+
+### §cands-diffs-methods Methods
+
+A *method_declaration* ([§14.6.1](classes.md#1461-general)) for an instance method in a *struct_declaration* may contain the *method_modifier* `readonly`. However, a static method shall not contain that modifier.
+
+It is a compile-time error to attempt to modify the state of an instance struct variable via a readonly method declared in that struct.
+
+Although a readonly method may call a sibling, non-readonly method, or property or indexer get accessor, doing so results in the creation of an implicit copy of `this` as a defensive measure.
+
+A readonly method may call a sibling property or indexer set accessor that is readonly. If a sibling member’s accessor is not explicitly or implicitly readonly, a compile-error occurs.
```
KalleOlaviNiemitalo, line 531: Does "sibling" specifically mean an instance member of `this`?  The term is not used with that meaning anywhere else in the specification.  It is not clear that it excludes static members, or instance members of the same type but used on a different instance.

standard/structs.md
```diff
@@ -502,6 +505,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
...
+
+It is a compile-time error for an automatically implemented property in a `readonly` struct to have a `set` accessor.
+
+An automatically implemented property declared inside a `readonly` struct need not have a `readonly` modifier, as its `get` accessor is implicitly assumed to be readonly.
+
+It is a compile-time error for a property to have a `readonly` modifier on both the `get` and the `set` accessors.
+
+It is a compile-time error to have a `readonly` modifier on a property itself as well as on either of its `get` and `set` accessors.
+
+If the `get` accessor has a `readonly` modifier, the `set` shall exist and shall not have that modifier.
```
KalleOlaviNiemitalo: Should disallow this as well:

```csharp
struct S {
    int P { // error CS8664: 'S.P': 'readonly' can only be used on accessors if the property or indexer has both a get and a set accessor
        readonly set { }
    }
}
```

Perhaps:

> If an accessor has a `readonly` modifier, the property shall also have an accessor that does not have that modifier.

KalleOlaviNiemitalo: Or merge with the earlier rule

> It is a compile-time error for a property to have a `readonly` modifier on both the `get` and the `set` accessors.

resulting in:

> It is a compile-time error for a property to have a `readonly` modifier on all of its accessors.  *Note:* To correct the error, move the modifier from the accessors to the property itself.  *end note*

standard/structs.md
```diff
@@ -502,6 +505,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
...
+
+It is a compile-time error to have a `readonly` modifier on both the `get` and the `set` accessors.
+
+It is a compile-time error to have a `readonly` modifier on an indexer itself as well as on either of its `get` or `set` accessors.
+
+If the `get` accessor has a `readonly` modifier, the `set` shall exist and shall not have that modifier.
+
+### §cands-diffs-events Events
+
+An *event_declaration* ([§14.8.1](classes.md#1481-general)) for an instance, non-field-like event in a *struct_declaration* may contain the *event_modifier* `readonly`. However, a static event shall not contain that modifier.
```
KalleOlaviNiemitalo: OK; accessors of an event cannot have the `readonly` modifier, but the grammar does not allow any other modifiers on them either, so it does not need to be restated here.

standard/structs.md
```diff
@@ -502,6 +505,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
...
+
+If the `get` accessor has a `readonly` modifier, the `set` shall exist and shall not have that modifier.
+
+### §cands-diffs-methods Methods
+
+A *method_declaration* ([§14.6.1](classes.md#1461-general)) for an instance method in a *struct_declaration* may contain the *method_modifier* `readonly`. However, a static method shall not contain that modifier.
+
+It is a compile-time error to attempt to modify the state of an instance struct variable via a readonly method declared in that struct.
+
+Although a readonly method may call a sibling, non-readonly method, or property or indexer get accessor, doing so results in the creation of an implicit copy of `this` as a defensive measure.
```
KalleOlaviNiemitalo, line 529: OK, demo:

```csharp
struct S {
    int P => 0; // not readonly
    
    int this[string s] => 0; // not readonly

    readonly void M() {
        _ = P; // warning CS8656: Call to non-readonly member 'S.P.get' from a 'readonly' member results in an implicit copy of 'this'.
        _ = this[""]; // warning CS8656: Call to non-readonly member 'S.this[string].get' from a 'readonly' member results in an implicit copy of 'this'.
        N(); // warning CS8656: Call to non-readonly member 'S.N()' from a 'readonly' member results in an implicit copy of 'this'.
    }

    void N() {}
}
```

The standard does not require these warnings and I think that's OK.

standard/structs.md
KalleOlaviNiemitalo, line 531: Is there already a rule against `C.N` here:

```csharp
struct S {
    public int P {
        set {} // not readonly
    }
    
    public int this[string s] {
        set {} // not readonly
    }

    readonly void M() {
        this.P = 0; // error CS1604: Cannot assign to 'this.P' because it is read-only
        this[""] = 0; // error CS1604: Cannot assign to 'this[""]' because it is read-only
    }
}

class C {
    static void N(in S that) {
        that.P = 0; // error CS8332: Cannot assign to a member of variable 'in S' or use it as the right hand side of a ref assignment because it is a readonly variable
        that[""] = 0; // error CS8332: Cannot assign to a member of variable 'in S' or use it as the right hand side of a ref assignment because it is a readonly variable
    }
}
```

Not clear to me why these are separate diagnostic codes.  Can the C# standard cover both `S.M` and `C.N` with the same rule?

standard/structs.md
KalleOlaviNiemitalo: Another demo, also OK:

```csharp
struct S {
    private static int r;
    
    public ref int P {
        get => ref r; // not readonly
    }
    
    readonly void M() {
        this.P = 0; // warning CS8656: Call to non-readonly member 'S.P.get' from a 'readonly' member results in an implicit copy of 'this'.
    }
}
```

`this.P` is at the left side of the assignment, but this calls the `get` accessor and so the wording is OK.

BillWagner:
rebased on 09/26/2023, various links also fixed.

standard/classes.md
```diff
@@ -1984,7 +1985,7 @@ Grammar notes:
...
+A *method_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and one of the permitted kinds of declared accessibility ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `static` ([§15.6.3](classes.md#1563-static-and-instance-methods)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods)), `override` ([§15.6.5](classes.md#1565-override-methods)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods)), `extern` ([§15.6.8](classes.md#1568-external-methods)), `async` ([§15.15](classes.md#1515-async-functions)), and `readonly` modifiers.  However, it is a compile-time error for the *method_modifier* `readonly` to be used in a *method_declaration* that is not contained directly by a *struct_declaration* (§cands-diffs-methods).
```
jskeet: I think this is strictly correct due to specifying "the *method_modifier* `readonly`" - but it's still confusing, as at first glance this *looks* like it violates the rule... It's a *method_declaration* including `readonly`, after all. But it's not a "*method_modifier* `readonly`".

```csharp
class Test
{
    private int x;
    public ref readonly int GetXRef() => ref x;
}
````

It's only when it's part of a *method_modifier* that it's prohibited in a class. So this is only valid when `Test` is a struct, for example:

```csharp
struct Test
{
    private static int x;
    public readonly ref readonly int GetXRef() => ref x;
}
```

I'm not saying that's a good example of when `readonly` is usable, but it does highlight that `readonly` can occur twice in a single declaration with different meanings, which is a bit confusing. Perhaps that deserves a note as well?

standard/structs.md
```diff
@@ -154,6 +153,7 @@ Structs differ from classes in several important ways:
 - Instance field declarations for a struct are not permitted to include variable initializers ([§16.4.8](structs.md#1648-field-initializers)).
 - A struct is not permitted to declare a parameterless instance constructor ([§16.4.9](structs.md#1649-constructors)).
 - A struct is not permitted to declare a finalizer.
+- Some kinds of members in structs are permitted to have the modifier `readonly` while that is not generally permitted for those same member kinds in classes.
```
jskeet: While I can appreciate the desire not to repeat things, I suspect it's worth being more specific about which members can have `readonly` in structs. The "some" here feels really woolly.

standard/structs.md
```diff
@@ -502,6 +504,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
 
 > *Note*: This access restriction means that constructors in structs containing automatically implemented properties often need an explicit constructor initializer where they would not otherwise need one, to satisfy the requirement of all fields being definitely assigned before any function member is invoked or the constructor returns. *end note*
 
+A *property_declaration* ([§15.7.1](classes.md#1571-general)) for an instance property in a *struct_declaration* may contain the *property_modifier* `readonly`. However, a static property shall not contain that modifier.
+
+It is a compile-time error to attempt to modify the state of an instance struct variable via a readonly property declared in that struct.
```
jskeet, line 505: "attempt" feels more human than spec-like to me. What *rule* are we stating here, and can we do so more precisely?

jskeet (commented):
Overall it feels like we may be missing a brief section explaining the purpose of readonly members on structs.

I'd also expect there to be a change somewhere to what happens when member access via a `readonly` field (in either a class or a struct) invokes a readonly member... shouldn't we be removing a defensive copying step there? Apologies if I've missed it... I look forward to it being pointed out :)

standard/classes.md
```diff
@@ -3155,7 +3157,7 @@ ref_property_body
...
 
-There are two kinds of *property_declaration*:
+A *property_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and any one of the permitted kinds of declared accessibility ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `static` ([§15.7.2](classes.md#1572-static-and-instance-properties)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods), [§15.7.6](classes.md#1576-virtual-sealed-override-and-abstract-accessors)), `override` ([§15.6.5](classes.md#1565-override-methods), [§15.7.6](classes.md#1576-virtual-sealed-override-and-abstract-accessors)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods), [§15.7.6](classes.md#1576-virtual-sealed-override-and-abstract-accessors)), `extern` ([§15.6.8](classes.md#1568-external-methods)), and `readonly` modifiers. However, it is a compile-time error for the *property_modifier* `readonly` to be used in a *property_declaration* that is not contained directly by a *struct_declaration* (([§16.4.11](structs.md#16411-automatically-implemented-properties)).
```
BillWagner: We should specify that the `readonly` modifier can appear on an individual accessor (typically `get`) as well. The most common forms would be:

```csharp
public readonly int P => p; // readonly property

public int P2
{
    readonly get { return p; }
    set { p = value; }
}
```

standard/classes.md
BillWagner: Maybe a note is correct here. In the first example, `readonly` modifies the `int`. In the second, `readonly` modifies the *method_declaration*.

standard/classes.md
```diff
@@ -3246,6 +3249,7 @@ For a ref-valued property the *ref_get_accessor_declaration* consists optional a
 The use of *accessor_modifier*s is governed by the following restrictions:
 
 - An *accessor_modifier* shall not be used in an interface or in an explicit interface member implementation.
+- It is a compile-time error for the *accessor_modifier* `readonly` to be used in a *property_declaration* or *indexer_declaration* that is not contained directly by a *struct_declaration* (§16.4.11, §cands-diffs-indexers).
```
BillWagner: I'd include accessors on this restriction. Either in this bullet, or as a separate bullet item.

standard/classes.md
BillWagner: ~~Also, in this and several related bullets, I'm curious why the adverb "directly" is needed? ~~

Kalle is right. Ignore this comment.

standard/structs.md
```diff
@@ -502,6 +504,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
...
+
+If the `get` accessor has a `readonly` modifier, the `set` shall exist and shall not have that modifier.
+
+### §cands-diffs-methods Methods
+
+A *method_declaration* ([§15.6.1](classes.md#1561-general)) for an instance method in a *struct_declaration* may contain the *method_modifier* `readonly`. However, a static method shall not contain that modifier.
+
+It is a compile-time error to attempt to modify the state of an instance struct variable via a readonly method declared in that struct.
+
+Although a readonly method may call a sibling, non-readonly method, or property or indexer get accessor, doing so results in the creation of an implicit copy of `this` as a defensive measure.
```
BillWagner, line 529: We haven't defined "sibling" anywhere in the spec. I suggest replacing it with "may call a ... **declared in the same `struct` declaration**" (emphasis mine for suggested text.

BillWagner (commented):
I gave this a second read today, and added more comments.

standard/classes.md
KalleOlaviNiemitalo: >  I'm curious why the adverb "directly" is needed?

```csharp
public struct S {
    public class C {
        public readonly int P => 0;
    }
}
```

The *property_declaration* of P is contained in the *struct_declaration* of S, but not directly, and `readonly` should not be allowed.

standard/classes.md
KalleOlaviNiemitalo: The restriction in this bullet is only for accessors; it applies to *accessor_modifier*, rather than *property_modifier* or *indexer_modifier*.

standard/structs.md
KalleOlaviNiemitalo: The wording "declared in the same `struct` declaration" would incorrectly cause a copy here:

```csharp
struct S {
    void M(in S s) {}
    readonly void N(S s) {
        s.M(this); // must not make an implicit copy of `this`
    }
}
```

standard/structs.md
Nigel-Ecma: Agree in both cases, these warnings are up to a compiler. The first example in particular does however raise another potential issue – see #1053.

standard/structs.md
Nigel-Ecma: The separate diagnostic codes (and text) are not mandated, that is a compiler choice – these messages are just (in the view of the compiler writer at least, YMMV) better than “error: LHS is read-only”

Nigel-Ecma:
> Overall it feels like we may be missing a brief section explaining the purpose of readonly members on structs.
> 
> I'd also expect there to be a change somewhere to what happens when member access via a `readonly` field (in either a class or a struct) invokes a readonly member... shouldn't we be removing a defensive copying step there? Apologies if I've missed it... I look forward to it being pointed out :)

See #1053

standard/classes.md
Nigel-Ecma: I suggest a slightly more positive rewording:

```suggestion
A *method_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and one of the permitted kinds of declared accessibility ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `static` ([§15.6.3](classes.md#1563-static-and-instance-methods)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods)), `override` ([§15.6.5](classes.md#1565-override-methods)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods)), `extern` ([§15.6.8](classes.md#1568-external-methods)) and `async` ([§15.15](classes.md#1515-async-functions)). Additionally a  *method_declaration* that is contained directly by a *struct_declaration* may include the `readonly` modifier (§cands-diffs-methods).
```

standard/classes.md
Nigel-Ecma: Same suggestion as for methods above:

```suggestion
A *property_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and any one of the permitted kinds of declared accessibility ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `static` ([§15.7.2](classes.md#1572-static-and-instance-properties)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods), [§15.7.6](classes.md#1576-virtual-sealed-override-and-abstract-accessors)), `override` ([§15.6.5](classes.md#1565-override-methods), [§15.7.6](classes.md#1576-virtual-sealed-override-and-abstract-accessors)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods), [§15.7.6](classes.md#1576-virtual-sealed-override-and-abstract-accessors)) and `extern` ([§15.6.8](classes.md#1568-external-methods)). Additionally a *property_declaration* that is contained directly by a *struct_declaration* may include the `readonly` modifier (§cands-diffs-methods).
```

standard/classes.md
Nigel-Ecma: I suggest:

```suggestion
- The *accessor_modifier* `readonly` may only be used in a *property_declaration* or *indexer_declaration* that is contained directly by a *struct_declaration* (§16.4.11, §cands-diffs-indexers).
```

standard/classes.md
```diff
@@ -3897,7 +3902,7 @@ remove_accessor_declaration
...
+An *event_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and any one of the permitted kinds of declared accessibility ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `static` ([§15.6.3](classes.md#1563-static-and-instance-methods), [§15.8.4](classes.md#1584-static-and-instance-events)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods), [§15.8.5](classes.md#1585-virtual-sealed-override-and-abstract-accessors)), `override` ([§15.6.5](classes.md#1565-override-methods), [§15.8.5](classes.md#1585-virtual-sealed-override-and-abstract-accessors)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods), [§15.8.5](classes.md#1585-virtual-sealed-override-and-abstract-accessors)), `extern` ([§15.6.8](classes.md#1568-external-methods)), and `readonly` modifiers. However, it is a compile-time error for the *event_modifier* `readonly` to be used in an *event_declaration* that is not contained directly by a *struct_declaration* (§cands-diffs-events).
```
Nigel-Ecma: Following my previous suggestions:

```suggestion
An *event_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and any one of the permitted kinds of declared accessibility ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `static` ([§15.6.3](classes.md#1563-static-and-instance-methods), [§15.8.4](classes.md#1584-static-and-instance-events)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods), [§15.8.5](classes.md#1585-virtual-sealed-override-and-abstract-accessors)), `override` ([§15.6.5](classes.md#1565-override-methods), [§15.8.5](classes.md#1585-virtual-sealed-override-and-abstract-accessors)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods), [§15.8.5](classes.md#1585-virtual-sealed-override-and-abstract-accessors)) and `extern` ([§15.6.8](classes.md#1568-external-methods)) modifiers. Additionally an *event_declaration* that is contained directly by a *struct_declaration* may include the `readonly` modifier (§cands-diffs-methods).
```

standard/classes.md
```diff
@@ -3878,6 +3882,7 @@ event_modifier
     | 'override'
     | 'abstract'
     | 'extern'
+    | 'readonly'
```
Nigel-Ecma: Should we add a comment like for unsafe code?

```suggestion
    | 'readonly'        // direct struct members only
```

standard/classes.md
```diff
@@ -3220,6 +3222,7 @@ accessor_modifier
     | 'internal' 'protected'
     | 'protected' 'private'
     | 'private' 'protected'
+    | 'readonly'
```
Nigel-Ecma: Should we add a comment like for unsafe code?

```suggestion
    | 'readonly'        // direct struct members only
```

standard/classes.md
```diff
@@ -3135,6 +3136,7 @@ property_modifier
     | 'override'
     | 'abstract'
     | 'extern'
+    | 'readonly'
```
Nigel-Ecma: Should we add a comment like for unsafe code?

```suggestion
    | 'readonly'        // direct struct members only
```

standard/classes.md
```diff
@@ -1946,6 +1946,7 @@ ref_method_modifier
...
```
Nigel-Ecma: Should we add a comment like for unsafe code?

```suggestion
    | 'readonly'        // direct struct members only
```

standard/classes.md
```diff
@@ -4158,6 +4163,7 @@ indexer_modifier
...
```
Nigel-Ecma: Should we add a comment like for unsafe code?

```suggestion
    | 'readonly'        // direct struct members only
```

standard/classes.md
```diff
@@ -4179,7 +4185,7 @@ ref_indexer_body
 
 *unsafe_modifier* ([§23.2](unsafe-code.md#232-unsafe-contexts)) is only available in unsafe code ([§23](unsafe-code.md#23-unsafe-code)).
 
-There are two kinds of *indexer_declaration*:
+An *indexer_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and any one of the permitted kinds of declared accessibility ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods)), `override` ([§15.6.5](classes.md#1565-override-methods)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods)), `extern` ([§15.6.8](classes.md#1568-external-methods)), and `readonly` modifiers. However, it is a compile-time error for the *indexer_modifier* `readonly` to be used in an *indexer_declaration* that is not contained directly by a *struct_declaration* (§cands-diffs-indexers).
```
Nigel-Ecma: As above:

```suggestion
An *indexer_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and any one of the permitted kinds of declared accessibility ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods)), `override` ([§15.6.5](classes.md#1565-override-methods)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods)) and `extern` ([§15.6.8](classes.md#1568-external-methods)) modifiers. Additionally an *indexer_declaration* that is contained directly by a *struct_declaration* may include the `readonly` modifier (§cands-diffs-methods).
```

standard/structs.md
Nigel-Ecma: Agree with @jskeet and additionally we have some repetition here which I think should be factored out as it is for classes: under classes modifiers are generally described in subclauses under *§15.3 Class members*.

That same pattern can be repeated here with the current content of *§16.3 Struct members* becoming *§16.3.1 General* and adding *§16.3.2 The readonly modifier*.

In §16.3.2 the rules for readonly members can be given (not proposed text):

- They cannot mutate the state of the struct instance
- They may mutate the static state of the struct
- And the (subjectively) ugly one, a readonly member may call a mutating member (of the same instance only?) but any mutations made must not be observable after the call. This could in terms of the compiler strategy of “defensive copies”, see #1053, or not.

This would provide a single place for defining what a readonly member. The additional per-member differences could follow in §16.3 subclauses or remain as they are in §16.4 subclauses.

standard/structs.md
Nigel-Ecma: Agree, assuming CS8664 is the intended semantics and not a Roslyn quirk.

IIRC conformance allows *additional* members other than `get` & `set` so “all of its accessors” would be invalid in such a case. I'll offer the following:

```suggestion
If the `get` or `set` accessor has a `readonly` modifier then a sibling `set` or `get` accessor shall exist and shall not have that modifier.
```

Could use “corresponding” rather than “sibling”

standard/structs.md
Nigel-Ecma: Another thing that could be defined in §16.3.x as per suggestion on line 509

Nigel-Ecma (dismissed):
I think we need to pull out bits into a common ”readonly members” section so we have a single place to define what being a readonly member means.

Though I kept it in my suggestion changes I think “is contained directly by a *struct_declaration*” is a bit unwieldly, can we not use something like “is a member of a *struct_declaration*” – I don't think membership in the general sense ("member of the judo club") is automatically read to include grandchildren. But as I say I left it as it…

standard/classes.md
KalleOlaviNiemitalo: IIUC the comments on `unsafe` and pointer grammar are there because implementations are not required to support unsafe code.  That reason does not apply to `readonly`.

standard/structs.md
KalleOlaviNiemitalo: Assuming the implementation allows a property to have a nonstandard `increment` accessor in addition to the standard `get` and `set`:

```csharp
struct S {
    private int i;
    public int I {
        readonly get => i;
        increment => ++i;
    }
}
```

I think "all of its accessors" would be a good rule here.  If the developer wants all accessors to be readonly, then they should make the whole property readonly, regardless of whether the accessors are standard or nonstandard.  And if the developer wants the `get` accessor to be readonly but wants some nonstandard accessor not to be readonly, then the standard should not require them to make the whole property readonly, as that change would also affect the nonstandard accessor (or require contortions in the specification of the extension).

standard/classes.md
Nigel-Ecma: @KalleOlaviNiemitalo – yes & no. Comments are used in various places in the grammar to provide additional details and there are no hard & fast rules about them. In some cases there are different productions for differences small (e.g. *method_modifier*/*ref_method_modifier*) and sometimes not but with added text giving the rules and sometimes a comment in the production as well (e.g. *event_modifier* here).

To comment or not to comment here, that is the question ;-)

Nigel-Ecma:
It’s the next day and I’m commenting on my own introductory comment on my review as an important issue probably doesn't come across clearly in the review itself:

> I think we need to pull out bits into a common ”readonly members” section so we have a single place to define what being a read-only member means.

Unless I'm missing something the Standard is missing a clear explanation of semantics of read-only members, and those semantics some might find “surprising”.

It is reasonable to assume that a read-only member reports on the current state of an object and does not modify it. C# does the latter but not necessarily the former...

If a read-only member attempts to modify the state a reasonable assumption is that an error would be produced, the member is read-only after all.

C# chooses a different approach: it makes a *copy* of the object, mutates the state of that copy, and then discards the copy. The C# choice could be “surprising”, something Roslyn seems to acknowledge by issuing a warning, but I don’t think the Standard is clear as it could be on the semantics as it is preoccupied with the “how” a compiler might achieve this result, the so called “defensive copies”.

Such an description doesn’t need to be long and complicated, it doesn’t need to justify C#’s choice, it just needs to be clear.

Refs: https://github.com/dotnet/csharpstandard/pull/673#discussion_r1510548259  & #1053

jskeet:
Example of something that might be hard to understand:

```csharp
#nullable disable
string x = null;
string y = "";
#nullable enable
Console.WriteLine(x.ToString());
Console.WriteLine(y.ToString());
```

standard/structs.md
Nigel-Ecma, line 505: From [MS Learn](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/readonly-instance-members):

> Readonly can be applied to property accessors to indicate that this will not be mutated in the accessor. The following examples have readonly setters because those accessors modify the state of member field, but do not modify the value of that member field.
> ```
> public readonly int Prop1
> {
>     get
>     {
>         return this._store["Prop1"];
>     }
>     set
>     {
>         this._store["Prop1"] = value;
>     }
> }
> ```

This example demonstrates a subtly that might easily be missed from the statement on line 589:

> It is a compile-time error to attempt to modify the state of an instance struct variable via a readonly property declared in that struct.

An example like (maybe exactly) this one should be included in the Standard with explanatory text to cover:

- The “indirect” state of a readonly only struct can be modified; and
- That the somewhat oxymoronic `read-only set` makes sense in C#

Unless there are such explanations already here somewhere…

standard/classes.md
BillWagner: I think we also need something here that order matters for these modifiers:

This is legal:

```csharp
public class C
{
    private int field;
    
    public ref readonly int M() 
    {
        return ref field;
    }
}
```

This is not:

```csharp
public class C
{
    private int field;
    
    public readonly ref int M()
    {
        return ref field;
    }
}
```

However, (changing the field to `static` for ref safety, both are legal for a `struct`:

```csharp
public struct C {
    private static int field;
    
    public ref readonly int M() {
        return ref field;
    }
}
```

And

```csharp
public struct C {
    private static int field;
    
    public readonly ref int M() {
        return ref field;
    }
}
```

standard/structs.md
```diff
@@ -502,6 +504,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
...
 
+A *property_declaration* ([§15.7.1](classes.md#1571-general)) for an instance property in a *struct_declaration* may contain the *property_modifier* `readonly`. However, a static property shall not contain that modifier.
+
+It is a compile-time error to attempt to modify the state of an instance struct variable via a readonly property declared in that struct.
+
+It is a compile-time error for an automatically implemented property having a `readonly` modifier, to also have a `set` accessor.
+
+It is a compile-time error for an automatically implemented property in a `readonly` struct to have a `set` accessor.
+
+An automatically implemented property declared inside a `readonly` struct need not have a `readonly` modifier, as its `get` accessor is implicitly assumed to be readonly.
```
huoyaoyuan, line 511: Roslyn also treats auto property getter of non-readonly structs as implicitly readonly. Should this be intended in spec?

KalleOlaviNiemitalo: This behaviour of Roslyn avoids copying the non-`readonly` `struct` value when reading an automatically implemented property of an `in` parameter:

```csharp
struct S {
    int I { get; set; }
    int J => 42;

    static int M(in S s) {
        int i = s.I; // does not copy s
        int j = s.J; // defensive copy, as in ((S)s).J
        return i + j;
    }
}
```

However, I don't think a conforming C# program can detect whether this copying happens, because the `get` accessor of `S.I` does not contain any user code that could mutate the original instance via an alias and check whether the mutation also affected `this`.  (For C# implementations that target the CLR, it can be still detected by disassembling the IL code.)  Therefore, I think the C# standard does not have to specify it.

standard/classes.md
jskeet: A note sounds like the way forward to me, particularly given your comment above from December 2024.

standard/classes.md
jskeet: Possibly "is permitted only" to match the following line?

standard/structs.md
```diff
@@ -502,6 +504,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
...
+
+It is a compile-time error to have a `readonly` modifier on a property itself as well as on either of its `get` and `set` accessors.
+
+If the `get` accessor has a `readonly` modifier, the `set` shall exist and shall not have that modifier.
+
+### §cands-diffs-methods Methods
+
+A *method_declaration* ([§15.6.1](classes.md#1561-general)) for an instance method in a *struct_declaration* may contain the *method_modifier* `readonly`. However, a static method shall not contain that modifier.
+
+It is a compile-time error to attempt to modify the state of an instance struct variable via a readonly method declared in that struct.
```
jskeet, line 527: Another case of "attempt".

standard/structs.md
```diff
@@ -502,6 +504,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
...
+
+A readonly method may call a sibling property or indexer set accessor that is readonly. If a sibling member’s accessor is not explicitly or implicitly readonly, a compile-error occurs.
+
+All *method_declaration*s of a partial method shall have a `readonly` modifier, or none of them shall have it.
+
+### §cands-diffs-indexers Indexers
+
+An *indexer_declaration* ([§15.9](classes.md#159-indexers)) for an instance indexer in a *struct_declaration* may contain the *indexer_modifier* `readonly`.
+
+It is a compile-time error to attempt to modify the state of an instance struct variable via a readonly indexer declared in that struct.
```
jskeet, line 539: Another case of "attempt"

standard/structs.md
```diff
@@ -502,6 +504,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
...
+
+It is a compile-time error to have a `readonly` modifier on both the `get` and the `set` accessors.
+
+It is a compile-time error to have a `readonly` modifier on an indexer itself as well as on either of its `get` or `set` accessors.
+
+If the `get` accessor has a `readonly` modifier, the `set` shall exist and shall not have that modifier.
```
jskeet:
```suggestion
If the `get` accessor has a `readonly` modifier, the `set` accessor shall exist and shall not have that modifier.
```

standard/structs.md
```diff
@@ -502,6 +504,50 @@ Automatically implemented properties ([§15.7.4](classes.md#1574-automatically-i
...
+
+It is a compile-time error to attempt to modify the state of an instance struct variable via a readonly method declared in that struct.
+
+Although a readonly method may call a sibling, non-readonly method, or property or indexer get accessor, doing so results in the creation of an implicit copy of `this` as a defensive measure.
+
+A readonly method may call a sibling property or indexer set accessor that is readonly. If a sibling member’s accessor is not explicitly or implicitly readonly, a compile-error occurs.
+
+All *method_declaration*s of a partial method shall have a `readonly` modifier, or none of them shall have it.
+
+### §cands-diffs-indexers Indexers
```
jskeet, line 535: So just to check, that means an indexer *can* have a `readonly set` (so long as it doesn't have a `readonly get` as well)?

I guess if it's modifying the state of something referenced via the struct, that's reasonable. Odd though.

jskeet (commented):
I suspect we can end up merging this in the meeting, but I'd like us to fix up as many issues as possible *during* the meeting to avoid creating a huge raft of follow-on issues. This feels like a PR with "lots of nits" rather than either "mostly clean" or "a few big changes needed".

BillWagner:
rebased on dotnet/draft-v8

standard/classes.md
BillWagner: I think this is a great addition. Otherwise, I was thinking of suggesting two different productions for the modifiers to make it more clear that `readonly` was only allowed on direct members of a `struct` type. That felt a bit like overkill though.

standard/classes.md
BillWagner: I think we should either:

1. Add the comment in all productions (line 3285, line 3371, *line 4031*, line 4302).
2. Make two productions for the modifiers: one for *<type-of-declaration>_class_modifier* and one for *<type-of-declaration>_struct_modifier* which adds `readonly` as an allowed modifier.

To discuss at the meeting.

standard/classes.md
```diff
@@ -4043,7 +4048,7 @@ remove_accessor_declaration
...
+An *event_declaration* may include a set of *attributes* ([§22](attributes.md#22-attributes)) and any one of the permitted kinds of declared accessibility ([§15.3.6](classes.md#1536-access-modifiers)), the `new` ([§15.3.5](classes.md#1535-the-new-modifier)), `static` ([§15.6.3](classes.md#1563-static-and-instance-methods), [§15.8.4](classes.md#1584-static-and-instance-events)), `virtual` ([§15.6.4](classes.md#1564-virtual-methods), [§15.8.5](classes.md#1585-virtual-sealed-override-and-abstract-accessors)), `override` ([§15.6.5](classes.md#1565-override-methods), [§15.8.5](classes.md#1585-virtual-sealed-override-and-abstract-accessors)), `sealed` ([§15.6.6](classes.md#1566-sealed-methods)), `abstract` ([§15.6.7](classes.md#1567-abstract-methods), [§15.8.5](classes.md#1585-virtual-sealed-override-and-abstract-accessors)), `extern` ([§15.6.8](classes.md#1568-external-methods)), and `readonly` modifiers. However, it is a compile-time error for the *event_modifier* `readonly` to be used in an *event_declaration* that is not contained directly by a *struct_declaration* (§cands-diffs-events).
```
KalleOlaviNiemitalo: By the way, an `extern` event cannot work in .NET because its `add` and `remove` accessors cannot be explicitly declared and it is then not possible to decorate them with DllImportAttribute.  This however is not a defect of the C# standard because DllImportAttribute is not specified there.

standard/structs.md
BillWagner: I like the idea of an example.  The "read only set that changes the state of a member" is a clear extension of what C# already permits:

```csharp
public struct S
{
    private List<int> sequence;

    public IList<int> Sequence
    {
        get
        {
            sequence ??= new List<int>();
            return sequence;
        }
    }
}
```

Callers can modify the returned sequence. The property above can't be marked `readonly` until a future version of C# allows field initializers in struct (10, if I recall correctly).

standard/structs.md
BillWagner: I think the standard can remain quiet on this behavior. A conforming compiler shouldn't be *required* to add the readonly modifier in this instance.

standard/structs.md
BillWagner: Well, the grammar currently only allows `get` and `set` accessors. (See 15.7.3). Given that, I'd prefer the simpler "all accessors", but I'm not too concerned one way or the other

standard/structs.md
BillWagner: I made a note in #1053: I think we should defer this until we spec out `ref readonly` parameter in C# 12.

standard/structs.md
BillWagner: Yes. That's exactly why. @jnm2 has a proposal to allow this in more places: https://github.com/dotnet/csharplang/blob/main/proposals/readonly-setter-calls-on-non-variables.md  (Note: not currently in the working set)

standard/structs.md
BillWagner: The resolution here should follow the same pattern chosen for the similar restriction on line 515 regarding properties.

BillWagner (commented):
This is a great example PR for our new process.

I didn't find any new issues in this review. The additional comments should provide a clear path to completion, or a couple possible paths.

I see either of these as possible:

1. Review and build the list in the meeting. Merge, and create issues to address identified issues.
2. Leave open at the meeting. Assign one person for edits, and a second for review and merge prior to the next meeting. Add one or more comments that note the expected plan for each open discussion offline. Edit and merge offline before the June meeting.

standard/classes.md
Nigel-Ecma: @jskeet Good idea:

```suggestion
- The *accessor_modifier* `readonly` is permitted only in a *property_declaration* or *indexer_declaration* that is contained directly by a *struct_declaration* (§16.4.11, §cands-diffs-indexers).
```

standard/structs.md
Nigel-Ecma: While I wrote a short while ago in #1053 that I wasn’t intending to get into the defensive copy issue at this time re-reading comments like the above makes me think we have little choice.

Last year I suggested the [common aspects of readonly need to be defined in single place](https://github.com/dotnet/csharpstandard/pull/673#discussion_r1510548259) and that including covering that readonly members may call mutating members but any mutations must not be visible afterwards – i.e. specify *what* not *how*. Do this and we should find wording things like the above easier.

And I think we'll need to define *sibling*, at least for this topic if not for the wider Standard.

standard/classes.md
jnm2: `readonly ref readonly` also means a third, distinct, thing. The first `readonly` is saying something about the implicit `this` argument, while the `ref` and the second `readonly` are saying something about the returned value.

jnm2:
Could there be value in describing what a "readonly struct member" is, and apply that to each instance member type (method, property, indexer, event) without repeating the whole definition?

standard/classes.md
BillWagner: Issue:  Add a note to explain the order.

standard/classes.md
BillWagner: Resolution: Go with the comment.

standard/classes.md
jskeet: Meeting of 2025-05-14 - we think we probably need to add this, yes - we can't see where else it's specified.

RexJaeschke:
On the 2025-05-14 call: After 20 minutes, and accepting/tweaking some changes, we decided this PR is not yet ready to be merged.

**Action**: Jon and Bill will review the remaining issues offline. For each remaining comment, they will add a new comment suggesting the disposition.

standard/structs.md
jskeet:
```suggestion
- Event declarations, property declarations, property accessors, indexer declarations, and method declarations are permitted to have the modifier `readonly` while that is not generally permitted for those same member kinds in classes.
```

jskeet: This would be addressed by my suggestion below, assuming I haven't missed anything.

standard/classes.md
jskeet: My review 2025-06-02: I think this is still work to do. It may be simple enough to add within this PR, or we could create an issue - I don't mind which.

standard/structs.md
jskeet: My review 2025-06-02: I think this is still work to do, although I'd be fine with creating an issue for it and merging this.

standard/structs.md
jskeet: My review 2025-06-02: I'm happy to merge without a change here.

standard/structs.md
jskeet: My review 2025-06-02: We should do something here, and I don't think we need to worry about non-standard accessors. I'd be fine with the "It is a compile-time error for a property to have a readonly modifier on all of its accessors. Note: To correct the error, move the modifier from the accessors to the property itself. end note" suggestion.

standard/structs.md
jskeet: My review 2025-06-02: Add an issue rather than trying to fix this before merging? (Maybe a single issue for line 509 and this.)

standard/structs.md
jskeet: My review 2025-06-02: Happy to defer this.

standard/structs.md
jskeet: My review 2025-06-02: I suggest we file an issue for sibling, but not block this PR on it.

standard/structs.md
jskeet: (Probably to be included in the same issue as for line 509 etc.)

jskeet (commented):
@BillWagner I've looked through this and commented or thumbs-upped each one that sounded potentially blocking. Please could you:

- Commit suggestions that you're happy with
- Comment on my suggestions for deferring with an issue
- Reply on anything else that doesn't come under one of those

standard/structs.md
BillWagner: I made an edit that follows Jon's suggestion.

standard/structs.md
BillWagner: The same fix was applied here and for properties.

standard/structs.md
BillWagner: Logged as #1339

BillWagner: Comment added on #1339

standard/structs.md
BillWagner: See #1339

standard/structs.md
BillWagner: See #1339

standard/structs.md
BillWagner: See #1339

standard/structs.md
BillWagner: See #1339

standard/classes.md
BillWagner: I added and updated this paragraph and the struct diffs section in this commit: https://github.com/dotnet/csharpstandard/pull/673/commits/fa8542a7347c8a79214ffd502895ff7bc4e8570a

BillWagner:
@jskeet 

> I've looked through this and commented or thumbs-upped each one that sounded potentially blocking. Please could you:
> 
> * Commit suggestions that you're happy with
> * Comment on my suggestions for deferring with an issue
> * Reply on anything else that doesn't come under one of those

I made three commits today (June 3)

1. Accept suggestions where we had consensus.
2. Make edits where comments suggested a solution, but there wasn't a suggestion.
3. Clarify language and headers around `readonly` being allowed in `struct` member declarations, but not in `class member` declarations.

Other open issues are described in #1339 

I think this is ready to merge.

jskeet (approved):
Agreed, I think this is ready to merge, too. Will do so.

jskeet:
Test failure:

```text
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::  Mismatched errors: Expected CS1593, CS1661, CS1678, CS8030, CS1688, CS1661, CS1676, CS1643, CS0126, CS0029, CS1662, CS1670, CS0029, CS1662; Was CS1593, CS1661, CS1678, CS8030, CS1688, CS1676, CS1643, CS0126, CS0029, CS1662, CS1670, CS0029, CS1662
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::  Details of actual errors:
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 5: CS1593: Delegate 'Class1.D' does not take 0 arguments
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 6: CS1661: Cannot convert anonymous method to type 'Class1.D' because the parameter types do not match the delegate parameter types
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 6: CS1678: Parameter 1 is declared as type 'long' but should be 'int'
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 9: CS8030: Anonymous function converted to a void returning delegate cannot return a value
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 12: CS1688: Cannot convert anonymous method block without a parameter list to delegate type 'Class1.E' because it has one or more out parameters
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 14: CS1676: Parameter 1 must be declared with the 'out' keyword
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 17: CS1643: Not all code paths return a value in anonymous method of type 'Class1.P'
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 18: CS0126: An object of a type convertible to 'int' is required
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 20: CS0029: Cannot implicitly convert type 'string' to 'int'
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 20: CS1662: Cannot convert anonymous method to intended delegate type because some of the return types in the block are not implicitly convertible to the delegate return type
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 25: CS1670: params is not valid in this context
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 32: CS0029: Cannot implicitly convert type 'string' to 'int'
❌Example tester-ExampleTester::file=tools/conversions.md:771-809,line=771::    Line 32: CS1662: Cannot convert anonymous method to intended delegate type because some of the return types in the block are not implicitly convertible to the delegate return type
```

jskeet:
But conversions.md isn't changed by this PR, so I'll merge this and we'll need to fix it separately.

jnm2:
@jskeet That is the error caused by <https://github.com/dotnet/csharpstandard/pull/1334#pullrequestreview-2886364073> and fixed by <https://github.com/dotnet/csharpstandard/pull/1336>.

PR #677, opened 2022-12-01
Fix the test runner YAML

jskeet:
Will self-merge if the action runs successfully.


jskeet:
(Ah, can't merge without review - that's fine.)

BillWagner (approved):
Thanks @jskeet 

I'll :shipit: now.

PR #678, opened 2022-12-01
Remove references to pointers within the standard where appropriate.

### LINKED ISSUE #676: Separate out unsafe content into the unsafe chapter
We just merged PR #236 

But there is one small part of it that mentions unsafe that can be moved into the unsafe chapter.

There are only 2 or 3 other mentions of unsafe elsewhere outside the unsafe chapter.  Those could be moved as well.

jskeet:
- They're still present in the grammar, and in explanatory paragraphs after relevant grammar sections.
- There are references in Annexes B and D, but that feels reasonable.

Fixes #676


standard/classes.md
```diff
@@ -442,8 +442,6 @@ The value type constraint specifies that a type argument used for the type param
 
 > *Note*: The `System.Nullable<T>` type specifies the non-nullable value type constraint for `T`. Thus, recursively constructed types of the forms `T??` and `Nullable<Nullable<T>>` are prohibited. *end note*
 
-Pointer types are never allowed to be type arguments and are not considered to satisfy either the reference type or value type constraints.
```
jskeet, line 445: Existing text in unsafe code:

> A *pointer_type* cannot be used as a type argument ([§8.4](types.md#84-constructed-types)), and type inference ([§11.6.3](expressions.md#1163-type-inference)) fails on generic method calls that would have inferred a type argument to be a pointer type.

If type inference fails, it can't satisfy any constraints. On the other hand, we *might* need some more care here - failing to satisfy constraints might mean that type inference goes on to pick a different option instead of failing. Thoughts?

standard/classes.md
```diff
@@ -567,7 +565,6 @@ The ***dynamic erasure*** of a type `C` is type `Cₓ` constructed as follows:
 - If `C` is a nested type `Outer.Inner` then `Cₓ` is a nested type `Outerₓ.Innerₓ`.
 - If `C` `Cₓ`is a constructed type `G<A¹, ..., Aⁿ>` with type arguments `A¹, ..., Aⁿ` then `Cₓ` is the constructed type `G<A¹ₓ, ..., Aⁿₓ>`.
 - If `C` is an array type `E[]` then `Cₓ` is the array type `Eₓ[]`.
-- If `C` is a pointer type `E*` then `Cₓ` is the pointer type `Eₓ*`.
```
jskeet, line 570: New text in unsafe code:

> The *dynamic erasure* of a type `E*` is the pointer type with referent type of the dynamic erasure of `E`.

standard/classes.md
```diff
@@ -2961,7 +2958,7 @@ class Customer
 
 ### 14.6.10 Extension methods
 
-When the first parameter of a method includes the `this` modifier, that method is said to be an ***extension method***. Extension methods shall only be declared in non-generic, non-nested static classes. The first parameter of an extension method may have no modifiers other than `this`, and the parameter type may not be a pointer type.
```
jskeet, line 2964: New text in unsafe code:

> A *pointer_type* cannot be used as the type of the first parameter in an extension method ([§14.6.10](classes.md#14610-extension-methods)).

standard/expressions.md
```diff
@@ -2580,7 +2580,7 @@ class __Anonymous1
 }
 ```
 
-where each «Tx» is the type of the corresponding expression «ex». The expression used in a *member_declarator* shall have a type. Thus, it is a compile-time error for an expression in a *member_declarator* to be `null` or an anonymous function. It is also a compile-time error for the expression to have a pointer type ([§22.3](unsafe-code.md#223-pointer-types)).
```
jskeet, line 2583: New text in unsafe code:

> An expression with a pointer type cannot be used to provide the value in a *member_declarator* within an *anonymous_object_creation_expression* ([§11.7.15.7](expressions.md#117157-anonymous-object-creation-expressions)).

standard/types.md
```diff
@@ -481,7 +481,7 @@ type_argument
     ;
 ```
 
-A *type_argument* shall not be a pointer type ([§22](unsafe-code.md#22-unsafe-code)). Each type argument shall satisfy any constraints on the corresponding type parameter ([§14.2.5](classes.md#1425-type-parameter-constraints)).
```
jskeet, line 484: Existing text in unsafe code:

> A *pointer_type* cannot be used as a type argument ([§8.4](types.md#84-constructed-types))

standard/types.md
```diff
@@ -650,4 +650,3 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
 - `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`.
 - Any *enum_type*.
 - Any user-defined *struct_type* that is not a constructed type and contains fields of *unmanaged_type*s only.
-- In unsafe code ([§22.2](unsafe-code.md#222-unsafe-contexts)), any *pointer_type* ([§22.3](unsafe-code.md#223-pointer-types)).
```
jskeet, line 653: New text in unsafe code:

> Pointer types themselves are unmanaged types, so a pointer type may be used as the referent type for another pointer type.

standard/variables.md
```diff
@@ -162,7 +162,6 @@ The default value of a variable depends on the type of the variable and is deter
 
 - For a variable of a *value_type*, the default value is the same as the value computed by the *value_type*’s default constructor ([§8.3.3](types.md#833-default-constructors)).
 - For a variable of a *reference_type*, the default value is `null`.
-- In an unsafe context, for a variable of a *pointer_type*, the default value is `null`.
```
jskeet, line 165: New text in unsafe code:

> The default value ([§9.3](variables.md#93-default-values)) for any pointer type is `null`.

standard/classes.md
```diff
@@ -140,7 +140,7 @@ A *primary_expression* ([§11.7](expressions.md#117-primary-expressions)) is per
 
 In any other context, it is a compile-time error to reference a static class.
 
-> *Note*: For example, it is an error for a static class to be used as a base class, a constituent type ([§14.3.7](classes.md#1437-constituent-types)) of a member, a generic type argument, or a type parameter constraint. Likewise, a static class cannot be used in an array type, a pointer type, a new expression, a cast expression, an is expression, an as expression, a `sizeof` expression, or a default value expression. *end note*
```
jskeet, line 143: Existing text in unsafe code:

> the referent type of a pointer shall be an *unmanaged_type*

That excludes all reference types, including all classes, so we don't need to highlight static classes.

jskeet (commented):
This "review" is intended to provide justification for each removal, ensuring that we're not losing information.

standard/classes.md
BillWagner: I think we're OK, but I'll defer to others. My thinking is that a pointer type has an explicit conversion to the integral types, but only has implicit conversions to `void*`. That's also a pointer type, so any option for type inference would be a pointer type, and would fail.

BillWagner (approved):
Thanks @jskeet 

This LGTM.

Nigel-Ecma (commented):
The changes all look OK to me.

While I can see the value in adding the information to the Unsafe Code clause, I don't see the value of removing it from where it is now.

If this is the direction we wish to go the PR can be approved.

RexJaeschke:
TG2 call of 2023-02-01: Approved with change.

PR #680, opened 2022-12-05
Define term "derived interface"

RexJaeschke:
The C# spec defines the terms ***base class***, ***derived class***, and ***base interface***, but *not* “derived interface.” However, it uses the term “derived interface” or “derived [type] member” a few times.
 
As I complete the spec for the V8 feature, “Default member implementations in interfaces,” I will add more uses of the term “derived interface.”


jskeet (approved):
LGTM, but would appreciate others reviewing before we merge.

MadsTorgersen (approved):
Looks great!

BillWagner (approved):
This LGTM @RexJaeschke 

Let's :shiopit:

Nigel-Ecma:
I will disagree in ~30min (sorry on a call)

Nigel-Ecma (changes requested):
I'm going to recommend this is made a draft and left until we have looked at the default member implementations.

C# (originally?) is a single inheritance language which adopted the CLI model that interfaces no **not** inherit but rather the "base" interface list is a list of "adopting classes must also implement".

IIRC even the use of the word "inherit" in §18.1 "An interface may inherit from multiple base interfaces" was debated…

In the Annotated Standard there are annotations explaining that interface "inheritance" is not the same as class inheritance.

Now default member implementation _may_ to change how we describe this going forward, but I think we need to look at it and make sure the Standard is really clear on what "derived" means in this case.

On a minor point the current text defines every interface as a derived one – as it only *may* have base interfaces and surely to be derived you need 1 or more.

BillWagner:
This PR targets "draft-v7", but its milestone is set for V8. We should resolve that.

@RexJaeschke @jskeet

RexJaeschke:
@BillWagner  Initially, this was intended to change draft-v7, and I probably had that as the milestone. However, Nigel proposed we defer this until we deal with “Default member implementations in interfaces” in v8, so I changed the milestone to v8.

What are our options w.r.t GitHub? Can we change the base from v7 to v8, or do I need to close this PR and create one against v8 instead?

BillWagner:
I changed the base branch from `draft-v7` to `draft-v8`.

BillWagner:
rebased on the latest draft v8 on 09/25/2023

jskeet:
Let's reconsider in the light of #681.

gafter:
Inheritance is a property of the enclosing types. One doesn't "inherit members".

jnm2:
I've thought of "inherits" or "derives from" as meaning to create a new type and base it on an existing type (or types) in such a way that the defining aspects of the base type (or types) are present in the new type.

A class would not be able to inherit from an interface, because interfaceness is a defining aspect of the base type, and a class does not have interfaceness. But a class could inherit from a class, and an interface could inherit from an interface.

jnm2:
I also strongly believe that the spec is not made _better_ by avoiding common parlance, when the common parlance is perfectly specific. The de facto term for `interface IB : IA` has been "inheritance" and "IB derives from IA."

Nigel-Ecma:
A quarter of a century ago – which in computer science terms is eons! – the type models of the CLI & C# were co-developed and it is in that history that the term *derived interface* comes from, it is worth looking back. The last CLI Standard came out in 2012, the last C# for v7 in 2023, both pre-date the inclusion of implementation in interfaces. From the CLI Standard:

> **I.8.9.11 Interface type derivation**
> 
> Interface types can require the implementation of one or more other interfaces. Any type that implements support for an interface type shall also implement support for any required interfaces specified by that interface. This is different from object type inheritance in two ways:
> 
> - Object types form a single inheritance tree; interface types do not.
> - Object type inheritance specifies how implementations are inherited; required interfaces do not, since interfaces do not define implementation. Required interfaces specify additional contracts that an implementing object type shall support.
> 
> To highlight the last difference, consider an interface, `IFoo`, that has a single method. An interface, `IBar`, which derives from it, is requiring that any object type that supports `IBar` also support `IFoo`. It does not say anything about which methods `IBar` itself will have.

The difference here between object inheritance and interface derivation is the model that C# v7 describes. Up until this point the use of “inheritance” for both was probably unfortunate, even if common parlance; now of course interfaces can contain implementations and further an interface can provide an implementation for a different interface, i.e. provide a default implementation for a method from one of the interfaces it derives from through an explicit implementation.

The challenge for C# v8 and onwards is to change the description of interface derivation, or whatever it is now to be called, to cover implementations in interfaces – it is still **not** the same as class inheritance in C#, which is a single-inheritance system, and so calling it something other than “inheritance” or always qualifying it as “interface inheritance” might be required. There is no CLI Standard to fall back on for this specification.

When it is known that v8 has the new model correctly specified, if it doesn’t already, then defining (and renaming?) *derived interface* should fall out… (Famous last words!)

jskeet:
This PR is now over 3 years old. The last comment talked about "when it is known that v8 has the new model correctly specified" which may mean we have new information at this point. Let's at least see whether we can figure out the next steps here.

jskeet:
Closing this PR, as what we think we need at the moment is an issue for someone to think about the model. This should be pre-C#-9, but doesn't need to block C# 8. That work should *consider* `static abstract` and `static virtual`, even though those won't come up for a while. @RexJaeschke to create the new issue.

PR #686, opened 2022-12-15
Clarify the introduction

kevinlu11:
Clarified the introduction while preserving all important and unique info.

### DIFF
--- a/standard/documentation-comments.md
+++ b/standard/documentation-comments.md
@@ -4,11 +4,12 @@
 
 ## D.1 General
 
-C\# provides a mechanism for programmers to document their code using a comment syntax that contains XML text. In source code files, comments having a certain form can be used to direct a tool to produce XML from those comments and the source code elements, which they precede. Comments using such syntax are called ***documentation comments***. They must immediately precede a user-defined type (such as a class, delegate, or interface) or a member (such as a field, event, property, or method). The XML generation tool is called the ***documentation generator***. (This generator could be, but need not be, the C\# compiler itself.) The output produced by the documentation generator is called the ***documentation file***. A documentation file is used as input to a ***documentation viewer***; a tool intended to produce some sort of visual display of type information and its associated documentation.
 
-A conforming C\# compiler is not required to check the syntax of documentation comments; such comments are simply ordinary comments. A conforming compiler is permitted to do such checking, however.
+***Documentation comments*** are special comments used to annotate C# source code and generate ***documentation files*** therefor. In a given C# source code file, each documentation comment must directly precede the user-defined [type](https://learn.microsoft.com/en-us/dotnet/csharp/tour-of-csharp/#types-and-variables) or [member](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/members) it describes, using a special comment syntax containing XML. 
 
-This specification suggests a set of standard tags to be used in documentation comments, but use of these tags is not required, and other tags may be used if desired, as long as the rules of well-formed XML are followed. For C\# implementations targeting the CLI, it also provides information about the documentation generator and the format of the documentation file. No information is provided about the documentation viewer.
+In order to produce a documentation file, a C# source code file containing documentation comments is fed into a ***documentation generator***. The resulting XML documentation file can be viewed using a ***documentation viewer***. (Conveniently, the C# compiler itself can serve as the documentation generator.)
+
+This specification suggests a set of standard XML tags to be used in documentation comments, but these tags are optional, and other tags may be used as long as the XML is well-formed. This specification also covers, for C# implementations targeting the CLI, the documentation generator and the format of documentation files. Documentation viewers are not discussed.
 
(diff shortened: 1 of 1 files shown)


jskeet:
This looks *less* clear to me. What do you believe is clarified by this?

kevinlu11:
> This looks _less_ clear to me. What do you believe is clarified by this?

The first time I read this doc, I had to wade through what in my view was a verbose and barely comprehensible paragraph before I even knew what a "documentation comment" in C# was. Keep in mind that I am NOT a C# expert, and that all of my knowledge of C# documentation comments came from reading this doc several hours ago. Therefore, I don't suffer from the "curse of knowledge" in this domain; I am in touch with the beginner's perspective.

I replaced that verbose paragraph with a single sentence (the first sentence in my edit) that I thought would be crystal clear, even for a beginner like myself. What about my edit seems unclear to you?

kevinlu11:
(Well, okay, more like "half-paragraph", but still...)

jskeet:
It loses a lot of information - for example what is expected of a conformant compiler. The standard isn't designed to be a beginner's guide - while obviously we should make it as clear as possible, it's *expected* to be verbose and detailed, in order to be as precise as possible.

Every individual sentence in the existing standard seems to be reasonably clear to me. Would just breaking the paragraph up help a bit?

kevinlu11:
I thought all the info I removed was either mentioned later in the article, or clear from context. For example, I took the part about conforming compiler to be unnecessary if one knows that the comments are just ordinary comments, which should be clear from looking at their syntax (since each line starts with "//[...]")

kevinlu11:
Also, the docstrings in the other languages I've seen (Rust, Python, Clojure) are just ordinary comments, so I figured that documentation comments being ordinary comments was a universal pattern that didn't need to be mentioned.

jskeet:
> For example, I took the part about conforming compiler to be unnecessary if one knows that the comments are just ordinary comments, which should be clear from looking at their syntax (since each line starts with "//[...]")

No, it's definitely necessary - as the language certainly *could* require a conforming compiler to, for example, validate that all documentation comments are valid XML.

We'll see what other working group members think, but personally I would prefer not to merge this PR.

kevinlu11:
Fair enough, I don't know much about compilers, so it could well be that my edit was overzealous there. Would it be better if the part about conforming compilers were added back in, but the rest remained as I edited it?

I think my frustration during my first read-through mainly came from not being sure what documentation comments do; I had to furrow my brow and re-read a few times before I even understood that their purpose, in a nutshell, is to help generate documentation files. As you say, specifications should be detailed, but I think introductions should give an immediate idea of the most important points, and details should come later. This particular article does not seem aimed at such an expert audience that I think my edit was inappropriate, except *possibly* for the part about compilers where I am admittedly out of my depth.

Just my opinions; didn't mean to be rude (in case I came across that way).

jskeet:
We'll see what others think. Your edit removes definitions of other terms, too. It may be that what's really needed isn't a change to the existing text so much as the addition of an introductory paragraph.

I would say that the specification *in general* isn't a good place to start to learn about a particular topic - it's about the details rather than the fundamentals.

RexJaeschke:
I agree with @jskeet. This text has remained unchanged from the first edition, published in 2002. I don't see the proposal as an improvement. A formal specification is *definitely* not the place to start learning about a subject.

BillWagner:
I agree that we should close this PR for the standard.

@klu22 I have a question for you, based on your earlier comments:  How did you arrive at this content rather than [the C# language reference on docs.microsoft.com](https://learn.microsoft.com/dotnet/csharp/language-reference/xmldoc/). That explanation is meant for developers learning C#. I'd like to know how you landed here so we can do some SEO work to direct traffic better to each set of articles.

kevinlu11:
> @klu22 I have a question for you, based on your earlier comments: How did you arrive at this content rather than [the C# language reference on docs.microsoft.com](https://learn.microsoft.com/dotnet/csharp/language-reference/xmldoc/). That explanation is meant for developers learning C#. I'd like to know how you landed here so we can do some SEO work to direct traffic better to each set of articles.

I'm glad you asked! When I search "C# Documentation Comments", both DuckDuckGo and Google return this content as the first result, followed by links to either third-party (non-Microsoft) sources or to other Microsoft links titled something like "XML documentation comments[....]". Ironically, the addition of "XML" in those titles led me to believe that non-XML documentation comments also exist, so I thought "Okay, I should read about documentation comments in general before reading about the special XML ones." Of course, I now know that all documentation comments use XML. 

In contrast to those pages' titles, this page's title begins simply with "Documentation Comments [...]" which led me to believe this was the place to start learning about them. Since it's the first search result, too, I figured it must be a popular reference--perfect for beginners!

(After realizing my confusion, I went to the tutorial page and removed the confusing "XML" from its title, as you saw: https://github.com/dotnet/docs/pull/33055/commits)

Here's a pic of my DuckDuckGo search results, pre-title-edit:
Capture7

PR #690, opened 2023-01-04
Mark default literals and field-targeted autoprop attributes as completed

jskeet:
(This assumes #262 is merged later today.)


jskeet:
(Added meeting-discuss so we can merge it later on, if everyone is happy.)

gafter (approved):
🐸 😷

PR #697, opened 2023-01-05
Example metadata: Add new, and revise existing, templates

RexJaeschke:
- Added 4 new templates *without using directives* to parallel those having them, so we can compile examples starting with certain preprocessor directives..
- Add a using directive for a new namespace to the existing templates


jskeet (approved):
Yup, this all looks fine to me.

PR #698, opened 2023-01-05
Example metadata: Add new, and revise existing, additional-files

RexJaeschke:
This PR incorrectly creates a file called Attr3Attribute, which should have been Attr3Attribute.cs instead. The correctly named *.cs version was also added. I couldn't find any way to remove the erroneous commit. Perhaps I'll have to come back and delete that file once this PR is merged.

### DIFF
--- a/tools/example-templates/additional-files/A.cs
+++ b/tools/example-templates/additional-files/A.cs
@@ -1,0 +1,19 @@
+class A
+{
+    //int x;
+    int y;
+    int z;
+    public virtual int X
+    {
+        get => 0;
+    }
+    public virtual int Y
+    {
+       set => y = value;
+    }
+    public virtual int Z
+    {
+        get => z;
+        set => z = value;
+    }
+}
--- a/tools/example-templates/additional-files/Attr1Attribute.cs
+++ b/tools/example-templates/additional-files/Attr1Attribute.cs
@@ -1,0 +1,5 @@
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+public class Attr1Attribute : Attribute
+{
+    public Attr1Attribute() {}
+}
--- a/tools/example-templates/additional-files/Attr1Attribute.cs
+++ b/tools/example-templates/additional-files/Attr1Attribute.cs
@@ -1,5 +1,4 @@
 [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
 public class Attr1Attribute : Attribute
 {
-    public Attr1Attribute() {}
 }
--- a/tools/example-templates/additional-files/Attr2Attribute.cs
+++ b/tools/example-templates/additional-files/Attr2Attribute.cs
@@ -1,0 +1,14 @@
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+public class Attr2Attribute : Attribute
+{
+//    private string name;
+    public Attr2Attribute(string name)
+    {
+//        this.name = name;
+    }
+//    public string Name
+//    {
+//        get { return name; }
+//    }
+}
+
--- a/tools/example-templates/additional-files/Attr3Attribute
+++ b/tools/example-templates/additional-files/Attr3Attribute
@@ -1,0 +1,5 @@
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+public class Attr3Attribute : Attribute
+{
+    public Attr3Attribute() {}
+}
--- a/tools/example-templates/additional-files/Attr3Attribute
+++ b/tools/example-templates/additional-files/Attr3Attribute
@@ -1,5 +1,0 @@
-[AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
-public class Attr3Attribute : Attribute
-{
-    public Attr3Attribute() {}
-}
--- a/tools/example-templates/additional-files/Attr3Attribute.cs
+++ b/tools/example-templates/additional-files/Attr3Attribute.cs
@@ -1,0 +1,5 @@
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+public class Attr3Attribute : Attribute
+{
+    public Attr3Attribute() {}
+}
--- a/tools/example-templates/additional-files/AuthorAttribute.cs
+++ b/tools/example-templates/additional-files/AuthorAttribute.cs
@@ -1,0 +1,14 @@
+using System;
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+public class AuthorAttribute : Attribute
+{
+    private string name;
+    public AuthorAttribute(string name)
+    {
+        this.name = name;
+    }
+    public string Name
+    {
+        get { return name; }
+    }
+}
--- a/tools/example-templates/additional-files/AuthorAttribute.cs
+++ b/tools/example-templates/additional-files/AuthorAttribute.cs
@@ -2,13 +2,6 @@ using System;
 [AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
 public class AuthorAttribute : Attribute
 {
-    private string name;
-    public AuthorAttribute(string name)
-    {
-        this.name = name;
-    }
-    public string Name
-    {
-        get { return name; }
-    }
+     public string Name { get; }
+     public AuthorAttribute(string name) => Name = name;
 }
--- a/tools/example-templates/additional-files/S_Boxing.cs
+++ b/tools/example-templates/additional-files/S_Boxing.cs
@@ -1,5 +1,5 @@
 public class S_Boxing
 {
-	public S value;
-	public S_Boxing(object o) {}
+    public S value;
+    public S_Boxing(object o) {}
 }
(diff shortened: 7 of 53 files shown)


tools/example-templates/additional-files/A.cs
```diff
@@ -0,0 +1,19 @@
+class A
```
jskeet, line 1: I suspect we'll want to change this to be more descriptive at some point, but it's fine for the moment.

tools/example-templates/additional-files/Attr1Attribute.cs
```diff
@@ -0,0 +1,5 @@
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+public class Attr1Attribute : Attribute
+{
+    public Attr1Attribute() {}
```
jskeet: I would expect the compiler to generate this constructor for us by default - I don't think we need it explicitly declared.

tools/example-templates/additional-files/Attr2Attribute.cs
```diff
@@ -0,0 +1,14 @@
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+public class Attr2Attribute : Attribute
```
jskeet, line 2: Why are various parts here commented out?

tools/example-templates/additional-files/Attr3Attribute
```diff
@@ -0,0 +1,5 @@
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+public class Attr3Attribute : Attribute
```
jskeet: This looks exactly like Attr1Attribute - do we need both to demonstrate two attributes applied to the same class?

jskeet: And in fact, this doesn't have a file extension - is that deliberate?

tools/example-templates/additional-files/Attr3Attribute.cs
```diff
@@ -0,0 +1,5 @@
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
```
jskeet, line 1: This one *does* have a file extension, but is still the same as Attr1...

tools/example-templates/additional-files/AuthorAttribute.cs
```diff
@@ -0,0 +1,14 @@
+using System;
+[AttributeUsage(AttributeTargets.Class, AllowMultiple = true)]
+public class AuthorAttribute : Attribute
```
jskeet, line 3: Can we use autoprops now for attributes? If so, it would be good to do so.

```csharp
public string Name { get; }
public AuthorAttribute(string name) => Name= name;
```

tools/example-templates/additional-files/S_Boxing.cs
```diff
@@ -0,0 +1,5 @@
+public class S_Boxing
```
jskeet, line 1: Nit: outdent the members here.

tools/example-templates/additional-files/SimpleAttribute.cs
```diff
@@ -0,0 +1,4 @@
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface)]
+public class SimpleAttribute : Attribute
```
jskeet, line 2: Again this is effectively the same as Attr1Attribute. We should consider whether we can simplify the standard.

jskeet (commented):
Mostly fine with these - once the comments have been addressed, we can merge this, but I'd like to consider creating subdirectories... we've got a lot of files that are only about attributes, and a lot that are only about namespaces. I suspect if we limit the number of files that are expected to be used across multiple files in the standard, that would help us maintain them.

tools/example-templates/additional-files/Attr1Attribute.cs
RexJaeschke: Agreed; removed

tools/example-templates/additional-files/Attr2Attribute.cs
RexJaeschke: removed

tools/example-templates/additional-files/Attr3Attribute
RexJaeschke: They are equivalent. Both are used in the example in which the declarations of the attributes types is irrelevant. I suggest leaving this as is.

See my first comment for this PR. I inadvertently created this file without an extension.

tools/example-templates/additional-files/Attr3Attribute.cs
RexJaeschke: See previous reply

tools/example-templates/additional-files/AuthorAttribute.cs
RexJaeschke: Agreed; changed in both the addition-file and in attributes.md where this type is declared.

tools/example-templates/additional-files/S_Boxing.cs
RexJaeschke: Done

RexJaeschke:
@jskeet ready for review

jskeet:
I can remove the accidentally-added file and push back to the same branch. Will do that now...

jskeet:
@BillWagner: Do you know enough about the GH actions and Node to fix the error we're getting? It looks like we probably need to update the action to use a more recent version of Node.

jskeet:
(Whoops - closure was accidental; simply clicked the wrong button.)

KalleOlaviNiemitalo:
Or lock down the version in the <code>[npm i](https://docs.npmjs.com/cli/v6/commands/npm-install) -g markdownlint-cli</code> command.

tools/example-templates/additional-files/Attr2Attribute.cs
jskeet: They don't appear to be, at the moment. Is there a commit missing?

jskeet (commented):
One more comment, as well as us needing to fix the linter. I'd still like to remove the duplication if possible, but that can come later.

jskeet:
I've just run the tests locally, and there are 9 errors - most about not being able to use preprocessor definitions after the first token in the file. We discussed this by email; would you like me to add extra templates for this?

(Alternatively, I could potentially change the template processing to just put all lines starting "#" in the sample above any existing lines...)

RexJaeschke:
@jskeet Something is wrong! I thought I had them all working on my local copy. Please send me the list of failures, and I'll investigate.

jskeet:
I've attached the log of running tests. It's possible that there's something I've messed up, of course... but my guess is that it's the interaction of multiple PRs.

[test-failures.txt](https://github.com/dotnet/csharpstandard/files/10382756/test-failures.txt)

RexJaeschke:
Mystery solved!

I downloaded to my system the complete code for my branch RexJaeschke-patch-15, the one used to create this PR. And when I ran the extractor/test tools, I got *exactly* the same output as you; good!

The errors/warnings fall into two categories:

A. Explicit using directives are now reported as duplicates, as the 4 original templates (as changed by PR #697) now contain all such directives.

B. As the new versions (as added by PR #697) of the 4 original templates *have no using directives*, these can now be used instead.

I predicted both of these in my mail to you starting Jan 3. Here is the relevant part of that thread and your responses:

From: Jon Skeet <jonskeet@google.com> 
Sent: Tuesday, January 3, 2023 9:25 AM
To: Rex Jaeschke <rex@rexjaeschke.com>
Subject: Re: Status of Rex's work re example metadata with some Q.s

Right, now back, responses inline...

On Thu, Dec 15, 2022 at 2:16 PM Rex Jaeschke <[rex@rexjaeschke.com](mailto:rex@rexjaeschke.com)> wrote:
Hello, Jon, on a very wet day here with 9 hours of freezing rain just about to end!
 
I am now running the example extraction and testing toolchain on a clone of the repo, and I’m working my way through the spec chapters, one at a time, looking at the examples that are *not* currently being extracted. I started with attributes.md.
 
1.	Now that the templates have using directives, I’m removing all such directives from examples, that are *not* needed for expository purposes. ***However, once the master templates have all been updated, the toolset will report warnings (“already seen this using directive before”) on many tests until I get done removing them from the remaining chapters’ examples.***
1.	***A consequence of 1. above is that some examples *no* longer compile. They have preprocessing directives in their first line(s), and the automatically prepended using directives precede them, which is not permitted!*** I don’t have a neat solution other than to suggest having a parallel set of the templates, one with the using directives and one without (for example, code-in-main and code-in-main-without-using. What do you think?

Jon> That would allow us to both remove the warnings and fix the errors. Yes, I think we should do that. I don't think we'll end up with a huge number of templates that way.

So, if you approve the PR, as is, I will then create a series of PRs that

1. Remove unnecessary using directives from examples
2. Change from template X to X-without-using for all those examples having #define/#undef directives upfront
3. Complete many formerly incomplete examples

OK?

jskeet:
Right... I think to avoid this getting further out of date, it's probably best for me to approve now, we can merge and then you can address the remaining issues as per item 1 and 2 of your email.

jskeet (approved):

jskeet:
(We still need to fix the linter issue though - I'll try to do that in a new commit in this PR.)

jskeet:
Hooray - the linter is fixed. Will squash and merge now.

PR #700, opened 2023-01-09
Add support for nullable references

RexJaeschke:
> ***Note carefully*** The [spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/nullable-reference-types-specification.md) in the V8 proposals folder states, “***This is a work in progress - several parts are missing or incomplete. [An updated version of this document](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/nullable-reference-types-specification.md) can be found in the C# 9 folder. ***”. Just what was implemented in V8 is not yet known. **Even though this proposal is aimed at the V8 spec, it is based on the V9 proposal from MS.**

Separately, in discussions with Bill, he thinks that TG2 should be considering two aspects of this proposal:

1. The general topic of static analysis
1. The fact that this proposal adds syntax via reference-type annotation to allow analysis across non-local boundaries

---
[See also list of diagnostics](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/nullable-warnings)


RexJaeschke:
Re "Query expressions," the MS spec states, "Additional work needed here."

Bill's initial feedback re this was:

There’s probably a lot here. There are two major questions that need to be addressed.

1. What type is represented by the elements in the result of a query? Is it nullable or non-nullable?
2. What is the null-state of the elements if the type is nullable?

For example the result of First() would be not null. If it were null, the method throws an exception. By contrast , FirstOrDefault() could return a null value. What other type information should flow through a query expression?

RexJaeschke:
Re "Type inference," there are numerous differences between the V8 and V9 MS Specs. Specifically,

- The V8 sections “Type inference for `var`” and “Type inference for `var?`” have been replaced by V9 section “nullable implicitly typed local variables” with quite different content.
- Section “Generic type inference” exists in both versions, but V9 added a bit and removed most content.

The following sections are identical in both versions:

- The first phase
- Exact, upper-bound and lower-bound inferences
- Fixing

The opening paragraph of “Fixing” is not spec text, and indicates that more work is needed here.

Issue: We need to spec exactly what’s in V8. Is anything in the V9 spec that is **NOT** for V8 only?

Issue: What’s missing re “Fixing?”

Issue: We currently have 12.6.3, “Type inference” with sections 12.6.3.2, “The first phase”, 12.6.3.10 “Lower-bound inferences”, 12.6.3.11, “Upper-bound inferences,” and 12.6.3.12 “Fixing,” that match some of the proposed-edit homes. It would be good to get some suggestions as to where to put the other new stuff.

Bill: I admit that I get lost here as well. When you get ready to pick this up again, I can go through the history of the dotnet/csharplan repo and try to find the differences. We may also experiment using LangVersion in our csproj.

RexJaeschke:
Reviewers: Please check my attempt to merge in to the section "Fixing" the new steps with the old. This was quite complicated.

BillWagner:
rebased on updated draft-v8 branch on 09/25/2023

BillWagner:
~~The rebase removed some sections added in draft V7. Adding them back in newer commits.~~

This wasn't quite true. Sections weren't removed, but some headers had their name edited.

KalleOlaviNiemitalo:
There is a bug report <https://github.com/dotnet/roslyn/issues/70958> about `T?` in explicit interface implementation where `T` is an unconstrained type parameter of the interface method being implemented.  The compiler behaviour is surprising but I have not checked whether it matches the specification in this pull request.

… Apparently this is not relevant to the PR after all, because [Unconstrained type parameter annotations](https://github.com/dotnet/csharplang/blob/6ac33c5f93c3fdd4d1addd68f2e8f99a79d0c191/proposals/csharp-9.0/unconstrained-type-parameter-annotations.md) are a C# 9 feature so the problematic case would be invalid in C# 8 anyhow.

jskeet:
Intention of adding the "meeting: discuss" label is that the working group will review this before the next meeting only in enough depth to gauge whether the *approach* is suitable, and to think about how we can make progress perhaps without trying to do it all in one PR.

standard/variables.md
```diff
@@ -1249,3 +1249,20 @@ A `new` expression that invokes a constructor obeys the same rules as a method i
 - Neither a `ref` local, nor a local of a `ref struct` type shall be in context at the point of a `yield return` statement or an `await` expression.
 - For a ref reassignment `e1 = ref e2`, the ref-safe-context of `e2` shall be at least as wide a context as the *ref-safe-context* of `e1`.
 - For a ref return statement `return ref e1`, the ref-safe-context of `e1` shall be the caller-context.
+
+### §Null-Tracking Null tracking
```
BillWagner, line 1253: I'd name this "Null state tracking" or "Null state analysis".

I'd also put this as the new 9.5, so it immediately follows the section on definite assignment.

standard/variables.md
```diff
@@ -1249,3 +1249,20 @@ A `new` expression that invokes a constructor obeys the same rules as a method i
...
+    : simple_name
+    | 'this'
+    | 'base'
+    | tracked_expression '.' identifier
+    ;
+```
+
+where *identifier* denotes a field or property.
+
+The null state for a tracked variable in unreachable code is “not null.” (This follows other decisions around unreachable code like considering all locals to be definitely assigned.)
```
BillWagner, line 1268: Open question:  Does this section need to be expanded, somewhat analogous to the section on definite assignment? It "feels" like it can be simpler, as the null state of an expression is a special case of definite assignment: "Was this variable definitely assigned to an expression that is known to be *not-null*? Or, was this variable tested to be *not-null*?".

But, it feels incomplete.

standard/types.md
```diff
@@ -705,3 +751,122 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+## §Types-And-Nullability Types and nullability
+
+### §Nullable-Types-General General
+
+C# has long supported the notion of a nullable value type; for example, the type `int?` is the nullable version of the value type `int`. (Value types and their nullable variants are described in [§8.3](types.md#83-value-types) and [§8.3.12](types.md#8312-nullable-value-types).) Almost all of the remaining text in this subclause pertains to the addition of support for nullable reference types, with only occasional mention of nullable value types.
+
+By definition, a reference type is nullable; that is, a variable of that type can either contain a reference to an object or be the value `null`, which indicates “no reference.” However, in many applications, the value of a reference should never be `null`, yet for the first 20-odd years of C#’s existence, there was no way to indicate this, or to have attempts to use such a reference in a `nonnull`-related context be diagnosed. Now that support for nullable reference types exists, such indications and diagnosis are possible by setting the nullable contexts (§Nullable-Contexts).
```
BillWagner, line 762: This para reads informative rather than normative.

I would support a Note that explains how NRT was designed to works (mostly) seamlessly with existing codebases. It does provide context around the chosen syntax, the API annotations and so on.

standard/types.md
```diff
@@ -705,3 +751,122 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
 - `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`.
 - Any *enum_type*.
 - Any user-defined *struct_type* that is not a constructed type and contains instance fields of *unmanaged_type*s only.
+- In unsafe code ([§23.2](unsafe-code.md#232-unsafe-contexts)), any *pointer_type* ([§23.3](unsafe-code.md#233-pointer-types)).
+
+## §Types-And-Nullability Types and nullability
+
+### §Nullable-Types-General General
```
BillWagner, line 758: I'd add a bit more in this clause. The general clause should provide the relationship between the concepts of *nullable reference type*, the *nullable annotation context*, and the *null state analysis*.

standard/types.md
```diff
@@ -705,3 +751,122 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+If a nullable context is not specified, its default state is disabled.
+
+Throughout this specification, all C# code that does not contain nullable directives, or about which no statement is made regarding the current nullable context state, shall be assumed as having been compiled with nullable annotation and nullable warning contexts enabled.
+
+#### §Nullable-Annotation-Context Nullable annotation context
+
+This context controls whether nullable annotations have any effect.
+
+When the nullable annotation context is disabled
```
BillWagner, line 788: I'd define this as "nullable oblivious". It will make the language smoother.

standard/types.md
```diff
@@ -705,3 +751,122 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+  1. The variable has been definitely assigned a non-`null` value.
+  1. The variable or expression has been checked against `null` before dereferencing it.
+
+Some warnings are generated regardless of the state of the nullable annotation context. (Examples include throwing a value that maybe `null`, and dereferencing a possibly null reference.) Other warnings are generated only if both the nullable warning context and the nullable annotation context are enabled. (Examples include converting a null literal or possible null value to a non-nullable type, possible null reference assignment, and possible null reference return.)
+
+### §Nullabilities-And-Null-States Nullabilities and null states
+
+A type can have one of three kinds of ***nullability***: “non-nullable”, “nullable,” or “oblivious,” and this is determined by the compiler using static flow analysis. For a reference type `T`, if the nullable annotation context (§Nullable-Annotation-Context) is enabled, `T` is non-nullable and `T?` is nullable. If the nullable annotation context is disabled, `T` is nullable and the `?` in `T?` is ignored. If analysis cannot determine the nullability, the nullability is oblivious.
```
BillWagner, line 846: I don't like the way the last sentence is phrased. I think of "nullable oblivious" as the expression isn't annotated. The expression doesn't declare whether it's nullable or non-nullable.

standard/expressions.md
```diff
@@ -232,12 +234,14 @@ User-defined operator declarations cannot modify the syntax, precedence, or asso
...
+- If the set of candidate user-defined operators is not empty, then this becomes the set of candidate operators for the operation. Otherwise, the predefined binary `operator «op»` implementations, including their lifted forms, become the set of candidate operators for the operation. The predefined implementations of a given operator are specified in the description of the operator. The predefined operators provided by an enum or delegate type are only included in this set when the binding-time type—or the underlying type if it is a nullable value type—of either operand is the enum or delegate type.
```
BillWagner, line 244: Note to us that we should sweep the entire spec to find "nullable" where historically we meant "nullable value type". If @RexJaeschke hasn't already done this, let's make an issue and handle that in a separate PR.

standard/expressions.md
```diff
@@ -841,9 +869,11 @@ A *lower-bound inference from* a type `U` *to* a type `V` is made as follows:
     - If it is invariant then an *exact inference* is made.
 - Otherwise, no inferences are made.
 
+If `V` is a nullable reference type, see [§12.6.3.10](expressions.md#126310-lower-bound-inferences).
```
BillWagner, line 872: I'm confused. Isn't this xref to this section?

standard/expressions.md
```diff
@@ -1326,6 +1381,15 @@ object o = (new int[3])[1];
 
 A *primary_expression* that consists of a *literal* ([§6.4.5](lexical-structure.md#645-literals)) is classified as a value.
 
+The null state (§Nullabilities-And-Null-States) of a `null` literal depends on the target type of the expression. If the target type is a type parameter constrained to a reference type, then the state is “maybe default.” Otherwise, it is “maybe null.”
+The null state of a `default` literal depends on the target type of the `default` literal. A `default` literal with target type `T` has the null state (§Nullabilities-And-Null-States) based on the properties of the type `T`, as follows:
+
+- If `T` is a non-nullable type then the `default` literal has the null state “not null”
```
BillWagner, line 1387: I don't think this is correct. 

```
string s = default;
```

Generates a warning because `default` doesn't have *not-null* state.

standard/expressions.md
```diff
@@ -31,6 +31,8 @@ A property access or indexer access is always reclassified as a value by perform
 
 An ***instance accessor*** is a property access on an instance, an event access on an instance, or an indexer access.
 
+The null state (§Nullabilities-And-Null-States) of an expression is derived from its form and type, and from the null state of variables involved in it. See each operator’s description for more information.
```
BillWagner, line 34: As I went through all the items in this clause, I'd like to discuss if we want a clause that defines the null-state for each expression and statement. 

Alternatively, we need to express that the null-state is always *oblivious* when nullable annotations and warnings are disabled.

standard/expressions.md
```diff
@@ -4393,6 +4686,8 @@ There are two forms of the `is` operator. One is the *is-type operator*, which h
 
 #### 12.12.12.1 The is-type operator
 
+*type* shall not be *nullable_reference_type* or a nullable type parameter.
```
BillWagner, line 4689: Commenting here that we need to update 12.12.12.2 on the pattern matching is to point out that something like this is valid:

```csharp
string? s = SomeMethod();
if (s is string str)
{
    // str is not null
}
```

but, checking if an expression is a nullable reference type is invalid.

standard/statements.md
```diff
@@ -313,6 +313,11 @@ Implicitly typed declarations contain the contextual keyword ([§6.4.4](lexical-
 - If there is no type named `var` in scope and the input matches *implicitly_typed_local_variable_declaration* then it is chosen;
 - Otherwise if a type named `var` is in scope then *implicitly_typed_local_variable_declaration* is not considered as a possible match.
 
+For reference types and type parameters that aren't constrained to be a value type, `var` infers an annotated type. For example:
```
BillWagner, line 316: Did we define "annotated type"? I thought the preferred term was "nullable type"

standard/statements.md
```diff
@@ -606,6 +615,8 @@ Local function bodies are always reachable. The endpoint of a local function dec
 
 If the type of the argument to a local function is `dynamic`, the function to be called shall be resolved at compile time, not runtime.
 
+A local function is treated like a method, except in regard to its captured variables. The initial state of a captured variable inside a local function is the intersection of the nullable state of the variable at all the uses of that local function. A use of a local function is either a call to that function, or where it is converted to a delegate.
```
BillWagner, line 618: We need to define how "intersection" is calculated here. May say the initial state of the capture variable is not-null if and only if the state of the captured variable is not-null at all uses of that local function?

BillWagner (commented):
I added some comments that are a mix of broad and specific notes.

I like the direction this takes, but I'd like to call out a few things.

1. Where we first introduce NRT, we should introduce the concept that it relies on a combination of variable annotations and static analysis. Furthermore, those annotations don't create new types, but inform the compiler to intent and determine warnings.
2. Let's discuss if null-state should be a section like definite assignment, rather than sprinkled in each statement and expression.

standard/variables.md
Nigel-Ecma: **[Disclaimer: I've yet to review this PR, Bill’s post just caught my eye]**

Yes, it feels incomplete, but maybe for different reasons to each of us ;-)

*tracked_expression* is not a production *per se* in the language – the tracked expressions are a subset of those defined by other productions, which are in the language, which also follow the grammar specified for *tracked_expression* (unless there are major grammar changes).

Is a variable reference such as `A<X>.B<Y>.c` tracked? If so it is not a *tracked_expression* by the above grammar.

*tracked_expression* is also left-recursive which we try to avoid, but that is simple to address.

I looked a bit further in this PR and line 1953 of `expression.md` refers to the static flow analysis as “the compiler’s” – I think it is probably “the language’s”… (the specification defines the language, a compiler (*et al*) implements the specification).

Nigel-Ecma:
> > _**Note carefully**_ The [spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/nullable-reference-types-specification.md) in the V8 proposals folder states, “***This is a work in progress - several parts are missing or incomplete. [An updated version of this document](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/nullable-reference-types-specification.md) can be found in the C# 9 folder. ***”. Just what was implemented in V8 is not yet known. **Even though this proposal is aimed at the V8 spec, it is based on the V9 proposal from MS.**
> 
> Separately, in discussions with Bill, he thinks that TG2 should be considering two aspects of this proposal:
> 
>     1. The general topic of static analysis
> 
>     2. The fact that this proposal adds syntax via reference-type annotation to allow analysis across non-local boundaries
> 
> 
> [Associated WorkItem - 209385](https://dev.azure.com/msft-skilling/Content/_workitems/edit/209385)

Serious question: does this need to go into v8 at all?

It is not unknown for compilers to introduce prototype implementations of features being considered for future versions of a language and the above comment makes me wonder whether Roslyn’s support in v8 should be seen as that, and hence not in the Standard, and the feature should go into v9.

standard/variables.md
Nigel-Ecma: **[A few days later…]**

It is **massively** incomplete, as it is in the source documents mentioned…

For example [latest source document mentioned](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/nullable-reference-types-specification.md) contains:

- “Describe null state transitions similar to definite assignment” as a placeholder, the whole flow analysis algorithm appears to be missing.
- “The null state of `E1 ? E2 : E3` is based on the null state of `E2` and `E3`…” – this is fine as far as it goes but as far as it goes does not cover the null state of a variable tested against `null` in `E1` when used in `E2` and `E3`.

So there does appear to a significant hole based on the currently considered source documents.

Now I don’t doubt that the group could devise an algorithm to fill this hole, but surely there is a spec for it?

Nigel-Ecma (changes requested):
I've either missed the obvious or a fundamental algorithm is AWOL…

standard/attributes.md
```diff
@@ -857,6 +858,276 @@ For invocations that occur within field or event initializers, the member name u
 
 For invocations that occur within declarations of instance constructors, static constructors, finalizers and operators the member name used is implementation-dependent.
 
+### §Code-Analysis-Attributes Code analysis-attributes
+
+#### §Code-Analysis-Attributes-General General
+
+Code compiled with both nullable contexts (§Nullable-Contexts) disabled is null oblivious (§Nullabilities-And-Null-States). That means any reference type variable may be null, but null checks aren't required. Once such code is made nullable-aware, those rules change. Reference type variables should never have the null value, and such variables must be checked against null before being dereferenced.
```
jskeet, line 865: This all sounds quite explanatory - it's useful text, but perhaps should be in a note? Personally I'm all for the spec being a little less formal in places - this may be something to talk about in the meeting, as I suspect we'll get more and more features specified in this way.

standard/attributes.md
```diff
@@ -857,6 +858,276 @@ For invocations that occur within field or event initializers, the member name u
...
+
+> *Example*: Consider the following property in which null is the default value, but clients can only set it to a non-null value.
+>
+> <!-- Example: {template:"standalone-lib", name:"DisallowNullAttribute"} -->
+> ```csharp
+> #nullable enable
+> public class X
+> {
+>     [DisallowNull]
+>     public string ReviewComment
```
jskeet, line 928: Shouldn't this be declared as `public string? ReviewComment`?

standard/attributes.md
```diff
@@ -857,6 +858,276 @@ For invocations that occur within field or event initializers, the member name u
...
+>     }
+>     private string _comment = "";
+> }
+> ```
+>
+> The get accessor could return the default value of `null`, so the compiler warns that it must be checked before access. Furthermore, it warns callers that, even though it could be null, callers shouldn't explicitly set it to null. *end example*
+
+#### §The-DoesNotReturn-Attribute The DoesNotReturn attribute
+
+Specifies that a given method never returns.
```
jskeet, line 942: I suspect we might want this to be in a different feature PR.

standard/attributes.md
```diff
@@ -857,6 +858,276 @@ For invocations that occur within field or event initializers, the member name u
...
+>         throw new InvalidOperationException();
+>     }
+>
+>     public void SetState(object containedField)
+>     {
+>         if (!isInitialized)
+>         {
+>             FailFast();
+>         }
+>         // unreachable code:
```
jskeet, line 962: It's not actually unreachable code, is it? (We only call isInitialized conditionally.)

standard/attributes.md
jskeet: (I haven't checked whether there's a corresponding change to reachability, but I can imagine all kinds of ways in which this could affect the standard even if we didn't have the NRT feature.)

standard/attributes.md
```diff
@@ -857,6 +858,276 @@ For invocations that occur within field or event initializers, the member name u
...
+
+Specifies that a given method never returns if the associated `bool` parameter has the specified value.
+
+> *Example*: Consider the following:
+>
+> <!-- Example: {template:"standalone-lib", name:"DoesNotReturnIfAttribute"} -->
+> ```csharp
+> public class X
+> {
+>     private void FailFastIf([DoesNotReturnIf(false)] bool isValid)
```
jskeet, line 983: As an aside, I think this method could do with being renamed. I'm used to "if" conditions happening if something is *true*, not false.

standard/attributes.md
```diff
@@ -857,6 +858,276 @@ For invocations that occur within field or event initializers, the member name u
...
+>     {
+>         if (!isValid)
+>         {
+>             throw new InvalidOperationException();
+>         }
+>     }
+>
+>     public void SetFieldState(object containedField)
+>     {
+>         FailFastIf(isInitialized);
```
jskeet, line 993: As an example of how the method name is weird, this reads as "fail fast if initialized" whereas we want to fail if the state is *not* initialized.

standard/attributes.md
```diff
@@ -857,6 +858,276 @@ For invocations that occur within field or event initializers, the member name u
...
+>     private bool isInitialized = false;
+>     private object _field;
+> }
+> ```
+>
+> *end example*
+
+#### §The-MaybeNull-Attribute The MaybeNull attribute
+
+Specifies that a non-nullable return value may be null.
```
jskeet, line 1007: Possibly `null`, or make the end of the example just null. (Basically they should be consistent.)

standard/attributes.md
```diff
@@ -857,6 +858,276 @@ For invocations that occur within field or event initializers, the member name u
...
+>     public Container(string message)
+>     {
+>         Helper();
+>         _optionalMessage = message;
+>     }
+>
+>     [MemberNotNull(nameof(_uniqueIdentifier))]
+>     private void Helper()
+>     {
+>         _uniqueIdentifier = DateTime.Now.Ticks.ToString();
```
jskeet, line 1060:
```suggestion
>         _uniqueIdentifier = DateTime.UtcNow.Ticks.ToString();
```

(It's not a great unique identifier anyway, but I'd generally want to discourage the use of DateTime.Now...)

standard/attributes.md
```diff
@@ -857,6 +858,276 @@ For invocations that occur within field or event initializers, the member name u
...
+
+#### §The-MemberNotNullWhen-Attribute The MemberNotNullWhen attribute
+
+Specifies that the listed member won't be `null` when the method returns the specified `bool` value.
+
+> *Example*: This attribute is like `MemberNotNull` (§The-MemberNotNull-Attribute) except that `MemberNotNullWhen` takes a `bool` argument. `MemberNotNullWhen` is intended for use in situations in which a helper method returns a `bool` indicating whether it initialized fields. *end example*
+
+#### §The-NotNull-Attribute The NotNull attribute
+
+Specifies that a nullable return value will never be `null`.
```
jskeet, line 1075: This description feels wrong to me - there's no return value involved.

standard/classes.md
```diff
@@ -434,11 +434,13 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
 
 The list of constraints given in a `where` clause can include any of the following components, in this order: a single primary constraint, one or more secondary constraints, and the constructor constraint, `new()`.
 
-A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, or the ***unmanaged type constraint*** `unmanaged`.
+A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***nullable reference type constraint*** `class?`, or the ***value type constraint*** `struct`.
```
jskeet, line 437: Is `unmanaged` no longer valid, or no longer a primary constraint, or should it be included here?

standard/classes.md
```diff
@@ -434,11 +434,13 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
+If the nullable annotation context (§Nullable-Annotation-Context) is disabled, the reference type constraint specifies that a type argument used for the type parameter shall be a reference type. All class types, interface types, delegate types, array types, and type parameters known to be a reference type (as defined below) satisfy this constraint. If the nullable annotation context is enabled, the type argument used for the type parameter shall be a non-nullable reference type (§Nullable-Types-General).
+
+If the nullable annotation context is enabled, the nullable reference type constraint specifies that a type argument used for the type parameter shall be a reference type. All class types, interface types, delegate types, array types, and type parameters known to be a reference type (as defined below) satisfy this constraint. If the nullable annotation context is disabled, the `?` shall be ignored, and a message to that effect shall be generated.
```
jskeet, line 443: Having the enabled/disabled for this paragraph be laid out in the opposite order to the paragraph above is confusing for me.

standard/classes.md
```diff
@@ -469,6 +471,12 @@ In either case, the constraint may involve any of the type parameters of the ass
...
+
+For a parameter type type constraint, if the nullable annotation context is enabled, and the constraint includes `?`, a type argument used for the type parameter shall be a value type or a non-nullable reference type. If the nullable annotation context is disabled, the `?` shall be ignored, and a message to that effect shall be generated. If the nullable annotation context is enabled, and the constraint does not include `?`, a type argument used for the type parameter shall be a non-nullable reference type. If the nullable annotation context is disabled, a type argument used for the type parameter shall be a nullable or non-nullable reference type.
```
jskeet, line 476: Should this be "For a type parameter type constraint"?

standard/classes.md
```diff
@@ -469,6 +471,12 @@ In either case, the constraint may involve any of the type parameters of the ass
...
+
+The nullability of a type argument or of a constraint does not impact whether the type satisfies the constraint. However, if the type argument does not satisfy the nullability requirements of the constraint, a warning shall be generated.
```
jskeet, line 478: I think this merits an example.

standard/classes.md
```diff
@@ -1719,6 +1729,8 @@ The initial value of a field, whether it be a static field or an instance field,
 >
 > *end example*
 
+The following applies only if the nullable annotation and nullable warning contexts are enabled. If a non-nullable instance field has no *variable_initializer* and the class has no instance constructors, or any of its instance constructors leave that field with a null value, a warning shall be generated. If a non-nullable static field has no *variable_initializer* and the class has no static constructor, or its static constructors leaves that field with a null value, a warning shall be generated.
```
jskeet, line 1732: It's unclear how much "the following" covers. Is it just this paragraph? (I suspect so, but we should reword to clarify.)

standard/delegates.md
```diff
@@ -293,7 +293,7 @@ Attempting to invoke a delegate instance whose value is `null` results in an exc
 >         cd1(-1);             // call M1
 >         D cd2 = new D(C.M2);
 >         cd2(-2);             // call M2
->         D cd3 = cd1 + cd2;
+>         D? cd3 = cd1 + cd2;
```
jskeet, line 296: Does this need to be nullable?

standard/expressions.md
```diff
@@ -147,7 +149,7 @@ The precedence of an operator is established by the definition of its associated
 >
 > |  **Subclause**      | **Category**                     | **Operators**                                          |
 > |  -----------------  | -------------------------------  | -------------------------------------------------------|
-> |  [§12.8](expressions.md#128-primary-expressions)              | Primary                          | `x.y` `x?.y` `f(x)` `a[x]` `a?[x]` `x++` `x--` `new` `typeof` `default` `checked` `unchecked` `delegate` `stackalloc`  |
+> |  [§12.8](expressions.md#128-primary-expressions)              | Primary                          | `x.y` `x?.y` `f(x)` `a[x]` `a?[x]` `x++` `x--` `new` `typeof` `default` `checked` `unchecked` `delegate` `stackalloc` `!` |
```
jskeet, line 152: This is tricky, as `!` is already present as a unary operator below. Perhaps it should be `x!`?

standard/expressions.md
```diff
@@ -756,6 +760,10 @@ During the process of inference each type parameter `Xᵢ` is either *fixed* to
 Type inference takes place in phases. Each phase will try to infer type arguments for more type variables based on the findings of the previous phase. The first phase makes some initial inferences of bounds, whereas the second phase fixes type variables to specific types and infers further bounds. The second phase may have to be repeated a number of times.
 
 > *Note*: Type inference is also used in other contexts including for conversion of method groups ([§12.6.3.14](expressions.md#126314-type-inference-for-conversion-of-method-groups)) and finding the best common type of a set of expressions ([§12.6.3.15](expressions.md#126315-finding-the-best-common-type-of-a-set-of-expressions)). *end note*
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+> *Note*: Generic type inference helps decide whether inferred reference types should be nullable or not. This is a best effort. It may yield warnings regarding nullability constraints, and may lead to nullable warnings when the inferred types of the selected overload are applied to the arguments. *end note*
```
jskeet, line 766: This feels like it needs work - "best effort" sounds like something that could vary very significantly between compiler implementations, which is far from ideal.

standard/expressions.md
```diff
@@ -768,6 +776,18 @@ For each of the method arguments `Eᵢ`:
 - Otherwise, if `Eᵢ` has a type `U` and the corresponding parameter is an input parameter ([§15.6.2.3.2](classes.md#156232-input-parameters)) then a *lower bound inference* ([§12.6.3.10](expressions.md#126310-lower-bound-inferences)) is made *from* `U` *to* `Tᵢ`.
 - Otherwise, no inference is made for this argument.
 
+> *Note*: Nullable reference types flow into the bounds from the initial expressions, as described below. In addition, `null` and `default` bounds carry through occurrences of `null` or `default` in the input expressions, which may cause an inferred type to be nullable, even when it otherwise wouldn't. This works even for nullable value types, which are enhanced to pick up "nullness" in the inference process. *end note*
```
jskeet, line 779: Note to self: I should probably steel myself to actually review this at some point. But the combination of generic type inference and nullable reference types fills me with terror.

standard/types.md
jskeet, line 846: Should this really be "a type" or "an expression"? Or "a type within an expression"? Or something else?

(It feels like a type itself doesn't change depending on the context, but how that type is *handled* changes due to context.)

standard/expressions.md
```diff
@@ -1746,12 +1867,12 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
...
 A *null_conditional_member_access* is a conditional version of *member_access* ([§12.8.7](expressions.md#1287-member-access)) and it is a binding time error if the result type is `void`. For a null conditional expression where the result type may be `void` see ([§12.8.10](expressions.md#12810-null-conditional-invocation-expression)).
 
-A *null_conditional_member_access* consists of a *primary_expression* followed by the two tokens “`?`” and “`.`”, followed by an *identifier* with an optional *type_argument_list*, followed by zero or more *dependent_access*es.
+A *null_conditional_member_access* consists of a *primary_expression* followed by the two tokens “`?`” and “`.`”, followed by an *identifier* with an optional *type_argument_list*, followed by zero or more *dependent_access*es, followed by zero or one *suppression*..
 
 ```ANTLR
 null_conditional_member_access
     : primary_expression '?' '.' identifier type_argument_list?
-      dependent_access*
+      dependent_access* suppression?
```
jskeet, line 1875: I find it curious that this isn't "null-suppression" or something similarly specific. Is the intention that the same syntax is used for other kinds of suppression in the future?

standard/expressions.md
```diff
@@ -1819,6 +1942,55 @@ A  *null_conditional_member_access* expression `E` is of the form `P?.A`. The me
 
 A *null_conditional_projection_initializer* is a restriction of *null_conditional_member_access* and has the same semantics. It only occurs as a projection initializer in an anonymous object creation expression ([§12.8.16.7](expressions.md#128167-anonymous-object-creation-expressions)).
 
+A `null_conditional_member_access` has the null state (§Nullabilities-And-Null-States) based on the expression type of *primary_expression*
+
+- If the type is a nullable value type then it has the null state “maybe null”
+- Else if the type is a nullable type parameter then it has the null state “maybe default”
```
```suggestion
- Otherwise if the type is a nullable type parameter then it has the null state “maybe default”
```

standard/expressions.md
```diff
@@ -1819,6 +1942,55 @@ A  *null_conditional_member_access* expression `E` is of the form `P?.A`. The me
...
+- Else it has the null state “maybe null”
```
```suggestion
- Otherwise it has the null state “maybe null”
```

standard/expressions.md
```diff
@@ -1819,6 +1942,55 @@ A  *null_conditional_member_access* expression `E` is of the form `P?.A`. The me
...
+null_forgiving_expression
+    : primary_expression suppression
+    ;
+
+suppression
+    : '!'
+    ;
+```
+
+The *primary_expression* in *null_forgiving_expression* shall not be a *null_forgiving_expression*.
```
jskeet, line 1965: But presumably it could be a parenthesized one - so `x!!` isn't valid, but `(x!)!` is valid? (Not saying that's a problem - just checking my understanding.)

standard/expressions.md
```diff
@@ -1819,6 +1942,55 @@ A  *null_conditional_member_access* expression `E` is of the form `P?.A`. The me
...
+
+This operator has no runtime effect; it evaluates to the result of its operand, and that result retains that operand’s classification.
+
+The null-forgiving operator is used to declare that an expression having a reference type isn't null.
```
jskeet, line 1969: I wonder whether this should be "should be considered non-null" or similar. (In particular, I'm nervous of the use of "declare" for something that isn't a declaration in the sense it's used everywhere else in the spec.)

standard/expressions.md
```diff
@@ -1848,6 +2020,42 @@ The result of evaluating an *invocation_expression* is classified as follows:
...
 - Otherwise, the *invocation_expression* invokes a returns-by-value method ([§15.6.1](classes.md#1561-general)) or returns-by-value delegate, and the result is a value, with an associated type of the return type of the method or delegate. If the invocation is of an instance method, and the receiver is of a class type `T`, the associated type is picked from the first declaration or override of the method found when starting with `T` and searching through its base classes.
 
+The following applies only if the nullable warning context is enabled. If *primary_expression* may be null, a warning shall be generated.
+
+If an *invocation_expression* invokes a member that is declared with one or more attributes (§Code-Analysis-Attributes) for special null behavior, the null state (§Nullabilities-And-Null-States) is determined by those attributes. Otherwise, the null state of the expression is the default null state for its type.
+
+The null state of an *invocation_expression* is not tracked by the compiler.
```
jskeet, line 2027: While I know what this means, and the example helps to clarify it, I don't think elsewhere we ever consider `GetText()` occurring in multiple places to be a single invocation expression.

standard/expressions.md
```diff
@@ -2180,6 +2416,12 @@ A *null_conditional_element_access* expression `E` is of the form `P?[A]B`; wher
 >
 > *end note*
 
+A `null_conditional_element_access` has the null state (§Nullabilities-And-Null-States) based on the expression type of *primary_no_array_creation_expression*
+
+- If the type is a nullable value type then it has the null state “maybe null”
+- Else if the type is a nullable type parameter then it has the null state “maybe default”
```
jskeet, line 2422: Again, I think we should use "otherwise" instead of "else".

standard/expressions.md
```diff
@@ -3494,6 +3752,15 @@ The term “correct grammar” above means only that the sequence of tokens shal
 <!-- markdownlint-enable MD028 -->
 > *Note*: From the disambiguation rule, it follows that, if `x` and `y` are identifiers, `(x)y`, `(x)(y)`, and `(x)(-y)` are *cast_expression*s, but `(x)-y` is not, even if `x` identifies a type. However, if `x` is a keyword that identifies a predefined type (such as `int`), then all four forms are *cast_expression*s (because such a keyword could not possibly be an expression by itself). *end note*
 
+If a cast expression `(T)E` invokes a user-defined conversion, then the null state (§Nullabilities-And-Null-States) of the expression is the default null state for the type of the user-defined conversion. Otherwise:
+
+- If `T` is a non-nullable value type then `T` has the null state “not null”
+- Else if `T` is a nullable value type then `T` has the null state “maybe null”
```
jskeet, line 3758: More else that should probably be otherwise.

standard/expressions.md
jskeet: I think having a single section of for null state, similar to a section for definite assignment, may well be more readable than scattering it throughout the standard. (Is that what you were suggesting, or have I misinterpreted?)

jskeet (commented):
I've looked through most of this, but I'm still not sure I can answer the question of whether this is the best structure. For someone who knows C# pre-8, having a single section about how NRTs are handled (obviously with changes elsewhere as necessary, but lumping *most* of the material together) might be better - but I find it hard to imagine myself in the "coming to C# from scratch" situation.

(We should think about how generics panned out, as they were initially an addendum, but then fully integrated...)

standard/types.md
```diff
@@ -705,3 +751,122 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+A warning might result if a variable of a non-nullable type has a potential `null` value assigned to it. However, oblivious and nullable variables can have `null` values freely assigned to them.
+
+Values of oblivious and non-nullable types can be dereferenced or assigned without generating a warning. Values of nullable types, however, are “null-yielding” and can cause warnings when dereferenced or assigned without proper null checking.
+
+Every expression has a ***null state*** that indicates whether that expression might potentially evaluate to `null`. The null state is either “not null,” “maybe null,” or “maybe default.” The null state is used to determine whether a warning should be given about null-unsafe conversions and dereferences. The ***default null state*** of a null-yielding variable is “maybe null” or “maybe default.” The default null state of a non-null-yielding variable is “not null.”
```
Nigel-Ecma, line 852: In `if(x == null) { x.m(); }` is the null state of `x` in `x.m()` “is null” and a warning that this will always throw issued?

Nigel-Ecma: Answering my own comment – no, after a test against `null` the current compiler warns it is `possibly null` on dereferencing

standard/expressions.md
Nigel-Ecma: “Else” -> “Otherwise” no brainer, committing this and previous one

standard/expressions.md
Nigel-Ecma: Apparently not, at least in the current implementation `(x!)!` elicits the same error as `x!!`

standard/expressions.md
Nigel-Ecma: I think you are probably correct :-(

standard/expressions.md
```diff
@@ -3494,6 +3752,15 @@ The term “correct grammar” above means only that the sequence of tokens shal
 <!-- markdownlint-enable MD028 -->
 > *Note*: From the disambiguation rule, it follows that, if `x` and `y` are identifiers, `(x)y`, `(x)(y)`, and `(x)(-y)` are *cast_expression*s, but `(x)-y` is not, even if `x` identifies a type. However, if `x` is a keyword that identifies a predefined type (such as `int`), then all four forms are *cast_expression*s (because such a keyword could not possibly be an expression by itself). *end note*
 
+If a cast expression `(T)E` invokes a user-defined conversion, then the null state (§Nullabilities-And-Null-States) of the expression is the default null state for the type of the user-defined conversion. Otherwise:
+
+- If `T` is a non-nullable value type then `T` has the null state “not null”
```
Nigel-Ecma, line 3757: Is something mixed up here? It is saying that `T` (a *type*), not the *expression*, has a null state.

standard/types.md
```diff
@@ -620,6 +644,28 @@ Since a type parameter can be instantiated with many different type arguments, t
...
+> ```csharp
+> #nullable enable
+> // The value t here has the state "maybe null". It's possible for T to be instantiated
+> // with string?, in which case null would be within the domain of valid values here. The 
+> // assumption though is the value provided here is within the valid values of T. Hence 
+> // if T is `string` then null will not be a value, just as we assume that null is not
+> // provided for a normal string parameter
+> void M<T>(T t)
+> {
+>     // There is no guarantee that default(T) is within the valid values for T hence the
```
Nigel-Ecma, line 661: Humpty Dumpty is that you? (Sorry ;-))

standard/expressions.md
```diff
@@ -5628,6 +5939,23 @@ query_continuation
...
 
+The null state (§Nullabilities-And-Null-States) of an *anonymous_method_expression* or *lambda_expression* is “not null.”
+
+An anonymous function is treated like a method, except in regard to its captured variables. The initial state of a captured variable inside an anonymous function is the intersection of the nullable state of the variable at all the uses of that anonymous function. The use of an anonymous function is the point at which it is defined in source.
```
Nigel-Ecma, line 5944: Is this bit meant to be in the previous clause? (§12.19 Anonymous function expressions)

standard/statements.md
Nigel-Ecma: Do not all three states need to be considered? This might give:

- If at all uses the null state of a captured variable is “not null” then the initial state is “not null”
- Otherwise if at any of the uses the null state is “maybe default” then the initial state is “maybe default”
- Otherwise the initial state is “maybe null”

But that is a *guess* at this time, particularly as the definition and propagation of the “maybe” states probably requires some clarification.

jskeet:
Marking this as a draft, as we're going to try to specify this in small PRs (leaving a "known incomplete" standard until we're done).

@BillWagner is going to start by trying to specify `#nullable`.

KalleOlaviNiemitalo:
Re query expressions in <https://github.com/dotnet/csharpstandard/pull/700#issuecomment-1377099339>

> For example the result of First() would be not null. If it were null, the method throws an exception.

If Enumerable.First\<T\>(this IEnumerable\<T\>) is implemented in C#, then it cannot detect at run time whether T is a nullable reference type, and cannot use that information to decide whether to throw instead of returning null.

I was looking at <https://github.com/dotnet/csharplang/issues/8383> and a similar case (tried at sharplab.io, not with a C#-8-only compiler):

```csharp
using System;

// Syntactically similar to IEnumerable<T> as far as query expressions
// are concerned, but does not have any standard semantics.
class Seq<T> {
}

// Query expression patterns for Seq<T>.
static class SeqExtensions
{
    // Using Func<int, int> to be clear that func does not affect nullability of the result.
    public static Seq<T>? Select<T>(this Seq<T> sequence, Func<int, int> func)
    {
        throw new NotImplementedException();
    }
}

static class C {
    static Seq<T> UsingQueryExpression<T>(Seq<T?> sequence) 
        where T : class
    {
        // no warnings
        return from item in sequence select item;
    }

    static Seq<T> UsingExtensionMethod<T>(Seq<T?> sequence)
        where T : class
    {
        // warning CS8603: Possible null reference return.
        // warning CS8619: Nullability of reference types in value of type 'IEnumerable<T?>' doesn't match target type 'IEnumerable<T>'.
        return sequence.Select(item => item);
            
        // Perhaps §12.20.3 (Query expression translation) should specify:
        return sequence.Select(item => item)!;
    }
}
```

jskeet:
Dropped the meeting discuss label as we'll be discussing smaller ones.

jskeet:
Closing as this has been superseded by a number of smaller PRs with a slightly different approach.

PR #704, opened 2023-01-10
Example metadata: basic-concepts.md

RexJaeschke:


jskeet (approved):
These are all fine, although I wouldn't expect it to matter if we didn't bother changing to without-using for most of these.

PR #707, opened 2023-01-11
 Example metadata: documentation-comments.md

RexJaeschke:


standard/documentation-comments.md
```diff
@@ -1015,33 +1012,15 @@ namespace Graphics
     /// </summary>
     public class Point
     {
-        /// <summary>
```
jskeet, line 1018: We either shouldn't change this, or we should remove the field documentation from the XML below - as it stands, the change invalidates the example.

RexJaeschke: Good catch, Jon! I changed the XML to reflect the removal of the 2 named fields.

jskeet (approved):
Right. This is at least consistent now - although it's less complete, in terms of not demonstrating the docs for a field. It doesn't have events either though, or conversions etc - so I guess it doesn't have to be absolutely comprehensive.

PR #710, opened 2023-01-11
Example metadata: lexical-structure.md

RexJaeschke:


standard/lexical-structure.md
```diff
@@ -1311,7 +1313,8 @@ Any remaining conditional sections are skipped and no tokens, except those for p
 >
 > Except for pre-processing directives, skipped source code is not subject to lexical analysis. For example, the following is valid despite the unterminated comment in the `#else` section:
 >
-> <!-- IncompleteExample: {template:"standalone-lib", name:"PreproInvalidSkippedSource", replaceEllipsis:true} -->
+> <!-- FIX: my thinking was to make CheckConsistency an extension method, but I could not get that to work. -->
```
jskeet, line 1316: It would have to be called as `this.CheckConsistency()`. We can definitely look at this later.

(We should consider using more conventional names of DEBUG and TRACE at that point, too.)

jskeet (approved):

PR #717, opened 2023-01-12
Example metadata: expressions.md

RexJaeschke:


jskeet (approved):

jskeet:
Aargh - have merged this thinking it was a single whitespace change (having followed a link from GitHub which only showed me a single commit). Will now review all the changes instead...

standard/expressions.md
```diff
@@ -4811,24 +4808,14 @@ It is possible for anonymous function delegates to share some captured variables
...
-> the three delegates capture the same instance of `x` but separate instances of `y`, and the output is:
->
-> ```console
-> 1 1
-> 2 1
-> 3 1
-> ```
->
-> *end example*
+> the three delegates capture the same instance of `x` but separate instances of `y`. *end example*
```
jskeet, line 4811: Why was this change made? I really think we want the output here - and for it to be tested.

jskeet (commented):
Just one additional comment having reviewed the rest of the PR.

standard/expressions.md
RexJaeschke: @jskeet As the example was written, while it captured the delegates in an array of 3 elements, it never actually executed those functions, so no output was produced. What I will do in a separate PR is to add a loop that executes the functions, so output is actually produced, and we can check that.

jskeet: Sounds good to me.

PR #718, opened 2023-01-13
Example metadata: expressions.md, revisited

RexJaeschke:
Restores text from an example that was deleted in PR #717, and adds text make the code produce output.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4791,7 +4791,7 @@ If a for-loop declares an iteration variable, that variable itself is considered
 
 It is possible for anonymous function delegates to share some captured variables yet have separate instances of others.
 
-> *Example*: For example, if `F` is changed to
+> *Example*: For example, if `F` is changed, as follows
 >
 > <!-- Example: {template:"standalone-console", name:"InstantiationOfLocalVariables6", IgnoredWarnings:["CS8321"], additionalFiles:["DvoidNoArgs.cs"], inferOutput: true} -->
 > ```csharp
@@ -4820,6 +4820,8 @@ It is possible for anonymous function delegates to share some captured variables
 > 2 1
 > 3 1
 > ```
+>
+> *end example*
 
 Separate anonymous functions can capture the same instance of an outer variable.
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4753,7 +4753,6 @@ If a for-loop declares an iteration variable, that variable itself is considered
 > *Example*: Thus, if the example is changed to capture the iteration variable itself:
 >
 > <!-- Example: {template:"standalone-console", name:"InstantiationOfLocalVariables5", inferOutput: true} -->
-> <!-- Maintenance Note: A version of this delegate type exists in additional-files as "DvoidNoArgs.cs". As such, certain changes to this type definition might need to be reflected in that file, in which case, *all* examples using that file should be tested. -->
 > ```csharp
 > delegate void D();
 >
@@ -4791,15 +4790,10 @@ If a for-loop declares an iteration variable, that variable itself is considered
 
 It is possible for anonymous function delegates to share some captured variables yet have separate instances of others.
 
-> *Example*: For example, if `F` is changed, as follows
+> *Example*: For example, if `F` is changed to
 >
-> <!-- Example: {template:"standalone-console", name:"InstantiationOfLocalVariables6", IgnoredWarnings:["CS8321"], additionalFiles:["DvoidNoArgs.cs"], inferOutput: true} -->
+> <!-- Example: {template:"code-in-partial-class", name:"InstantiationOfLocalVariables6", IgnoredWarnings:["CS8321"], additionalFiles:["Caller.cs"], inferOutput: true} -->
 > ```csharp
-> foreach (D d in F())
-> {
->     d();
-> }
->
 > static D[] F()
 > {
 >     D[] result = new D[3];
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4793,8 +4793,13 @@ It is possible for anonymous function delegates to share some captured variables
 
 > *Example*: For example, if `F` is changed to
 >
-> <!-- Example: {template:"standalone-console", name:"InstantiationOfLocalVariables6", IgnoredWarnings:["CS8321"], additionalFiles:["DvoidNoArgs.cs"]} -->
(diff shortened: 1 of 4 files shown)


standard/expressions.md
```diff
@@ -4808,7 +4813,13 @@ It is possible for anonymous function delegates to share some captured variables
...
 > ```
 >
-> the three delegates capture the same instance of `x` but separate instances of `y`. *end example*
+> the three delegates capture the same instance of `x` but separate instances of `y`, and the output is:
+>
+> ```console
+> 1 1
+> 2 1
+> 3 1
+> ```
```
jskeet, line 4816: We need the *end example* restored.

jskeet (commented):
Hmm. This is tricky, as it doesn't quite match the introductory text of "For example, if `F` is changed to" any more.

We could handle this in a completely separate template, but it's not ideal.

I definitely think the text should be restored, but I don't know what to do about the code for now...

standard/expressions.md
RexJaeschke: @jskeet Clearly, as suggested by the opening sentence, this `F` is different to that in the previous example. I see two possible solutions:

1. Change the into text from "For example, if `F` is changed to" to "For example, if `F` is changed, as follows".
2. Clone the whole of the previous example, replacing its `F` with the new version. As that would declare the delegate type `D` as well, the `additionalFiles:["DvoidNoArgs.cs"]` can be removed, and as that is the only place that file is added  (I think), it can be deleted from the addition-files folder.

Yes, the *end example* marker needs to restored too.

jskeet: Hmm. I tend to think that the spec text was basically ideal as it was before - and I've always wanted the example testing to avoid detracting from the spec text. I think option 1 is probably the best for now, but I'll keep thinking about a way to avoid having the loop in the testable example.

RexJaeschke: Option 1 implemented

RexJaeschke: @jskeet Here's an idea that I have working with a small bit of manual intervention. Here's my test code:

````
> <!-- Example: {template:"code-in-class-lib", name:"InstantiationOfLocalVariables6", IgnoredWarnings:["CS8321"], additionalFiles:["Caller.cs"], inferOutput: true} -->
> ```csharp
> static D[] F()
> {
>     D[] result = new D[3];
>     int x = 0;
>     for (int i = 0; i < 3; i++)
>     {
>         int y = 0;
>         result[i] = () => Console.WriteLine($"{++x} {++y}");
>     }
>     return result;
> }
> ```
>
> the three delegates capture the same instance of `x` but separate instances of `y`, and the output is:
>
> ```console
> 1 1
> 2 1
> 3 1
> ```
>
> *end example*
````

When extracted, the Library.cs file contains the following:

```csharp
<<using directives ...>>

class Class1
{
    static D[] F()
{
    D[] result = new D[3];
    int x = 0;
    for (int i = 0; i < 3; i++)
    {
        int y = 0;
        result[i] = () => Console.WriteLine($"{++x} {++y}");
    }
    return result;
}
}
```

**However, what if the extractor added `partial` to `class Class1` for template code-in-class-lib (and code-in-main)?**

Caller.cs contains the following:

```csharp
delegate void D();

partial class Class1
{
   static void Main()
   {
       foreach (D d in F())
       {
           d();
       }
   }
}
```

This allows us to **supplement** the class from the example by declaring the delegate and adding a `Main`. And because both files contribute to the same class, we don't need `public` accessibility in method `F`'s declaration.

This relies on this template emiting `partial`. Now for all other examples that **don't** need supplementing, we'd have just a single Library.cs or Program.cs file containing a partial class, but that's OK, right?

When I compile/execute this, I use

```console
csc Library.cs Caller.cs -t:exe
```

and the resulting executable is Caller.exe.

However, if we have template code-in-main also emit `partial`, and the example contains `Main` and the supplementary file has support stuff, I'd have something like

```console
csc Program.cs Support.cs -t:exe
```

and the resulting executable would be Program.exe. My concern here is that the logic the tester uses to execute the result might be different for these two scenarios.

jskeet: Yes, I think partial classes would be a great way of doing this. There shouldn't be any need to change the execution code.

I don't want the extractor to have to "understand" the code though. I think we can just have a template for "code in a partial class" and then use an additional file for this example that supplies the Main method. (All such examples would just need to use the same partial class name, which is easy enough.) Effectively the new template would be like code-in-class-lib but with a partial class and an executable output type (instead of library).

I'm happy to do this myself if that description is too hard to follow :)

RexJaeschke: @jskeet I'll take a look at implementing this, but probably not for a few days. In the meantime, I'm going back to the V8 spec for nullable references based on recent private feedback from Mads and Bill.

RexJaeschke: @jon I've created the new template code-in-partial-class, so we're good to go. However, on my first attempt I inadvertently added that template to additional-files, but I don't see how to remove that. Can you please delete the 3rd-to-last commit? I then put it in the correct place.

standard/expressions.md
```diff
@@ -4793,7 +4792,7 @@ It is possible for anonymous function delegates to share some captured variables
 
 > *Example*: For example, if `F` is changed to
 >
-> <!-- Example: {template:"standalone-console", name:"InstantiationOfLocalVariables6", IgnoredWarnings:["CS8321"], additionalFiles:["DvoidNoArgs.cs"]} -->
+> <!-- Example: {template:"code-in-partial-class", name:"InstantiationOfLocalVariables6", IgnoredWarnings:["CS8321"], additionalFiles:["Caller.cs"], inferOutput: true} -->
```
jskeet, line 4795: We *may* want to rename this to console-partial-class at some point to make it clear what the output type is. Although if I implement project properties in metadata, we may find we can reduce the number of templates significantly anyway. Not a problem for now.

jskeet, line 4795: I think you've forgotten to add Caller.cs...

jskeet (commented):
Looks like it'll be fine when we've got Caller.cs. I've removed the extraneous file.

standard/expressions.md
RexJaeschke: OK

RexJaeschke: @jskeet OK, done. Now we no longer need additional-file DvoidNoArgs.cs, but I don't see how to remove that.

jskeet: > Now we no longer need additional-file DvoidNoArgs.cs, but I don't see how to remove that.

There's a "delete file" option in GitHub in the three dots on the top right corner. I can do it if you can't find the option easily.

jskeet (approved):

jskeet (approved):

PR #721, opened 2023-01-14
Enable all undefined-behavior examples

RexJaeschke:
In example PointerTypes1, the act of dereferencing `px1` and `px2` in `Main` is undefined. As such, the program might even abort! And even if the program executes, there is no point attempting to write out the underling `int` values, as they can be bogus, and certainly not predictable. (In my testing, *each* time I ran this program with output, one of the two values changed.)

So, I removed expectedOutput, changed the template, added some comments, and replaced the writes with dereferences to two new local variables.

For example PointerConversions1, I made no changes; I just activated this example.

For both examples, I added an HTML comment: `> <!-- Note: the behavior of this example is undefined. -->`.

### DIFF
--- a/standard/unsafe-code.md
+++ b/standard/unsafe-code.md
@@ -174,7 +174,8 @@ A *pointer_type* may be used as the type of a volatile field ([§14.5.4](classes
 
 > *Note*: Although pointers can be passed as `ref` or `out` parameters, doing so can cause undefined behavior, since the pointer might well be set to point to a local variable that no longer exists when the called method returns, or the fixed object to which it used to point, is no longer fixed. For example:
 >
-> <!-- Undefined$Example: {template:"standalone-console", name:"PointerTypes1", replaceEllipsis:true, expectedOutput:["*px1 = 10","*px2 = 20"], expectedWarnings:["x","x"]} -->
+> <!-- Example: {template:"standalone-console-without-using", name:"PointerTypes1", replaceEllipsis:true} -->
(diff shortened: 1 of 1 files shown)


standard/unsafe-code.md
```diff
@@ -286,22 +287,23 @@ When one pointer type is converted to another, if the resulting pointer is not c
...
+>     int* pi = (int*)pv; // pretend a 16-bit char is a 32-bit int
+>     int i = *pi;        // read 32-bit int; undefined
+>     *pi = 123456;       // write 32-bit int; undefined
 > }
 > ```
 >
 > *end example*
 
-When a pointer type is converted to a pointer to `byte`, the result points to the lowest addressed `byte` of the variable. Successive increments of the result, up to the size of the variable, yield pointers to the remaining bytes of that variable.
+When a pointer type is converted to a pointer to `byte` (as with the cast to `void*` above), the result points to the lowest addressed `byte` of the variable. Successive increments of the result, up to the size of the variable, yield pointers to the remaining bytes of that variable.
```
jskeet: This appears to say that `void*` *is* a pointer to `byte`. (I'd also argue that we don't have a cast to `void*` in the code above - we're using an implicit conversion, but we're not using a cast.)

Worth discussing in next week's meeting?

RexJaeschke: @jskeet I'll remove that parenthesized phrase. Reading it now, I don't recall why I added it; it certainly isn't needed to resolve this issue.

jskeet (approved):

PR #724, opened 2023-01-16
Add ignoreOutput option for example metadata

jskeet:
This allows the output to be ignored for nondeterministic or implementation-defined behavior.


RexJaeschke (approved):
Once this is merged, I'll apply the new directive to the remaining two impdef (finalize) examples.

PR #731, opened 2023-02-03
fix internal references

BillWagner:
Over time, a few links in this clause were going to the wrong section. Update them to refer to the correct location.

I went by the text surrounding the link. These were obviously wrong.

### DIFF
--- a/standard/unsafe-code.md
+++ b/standard/unsafe-code.md
@@ -347,10 +347,10 @@ Mappings between pointers and integers are implementation-defined.
 
 Arrays of pointers can be constructed using *array_creation_expression* ([§11.7.15.5](expressions.md#117155-array-creation-expressions)) in an usafe context. Only some of the conversions that apply to other array types are allowed on pointer arrays:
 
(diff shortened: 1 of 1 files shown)


standard/unsafe-code.md
```diff
@@ -1025,7 +1025,7 @@ stackalloc_initializer
...
+A stack allocation initializer of the form stackalloc `T[E]` requires `T` to be an unmanaged type (§8.8) and `E` to be an expression implicitly convertible to type `int`. The construct allocates `E * sizeof(T)` bytes from the call stack and returns a pointer, of type `T*`, to the newly allocated block. If `E` is a negative value, then the behavior is undefined. If `E` is zero, then no allocation is made, and the pointer returned is implementation-defined. If there is not enough memory available to allocate a block of the given size, a `System.StackOverflowException` is thrown.
```
BillWagner, line 1028: The text link says "unmanaged types", but the target is "pointer types". From both the context and the description in this paragraph, I think "unmanaged types" is a better target. However, this link has been to "pointer types" since C# 4.0 (or older). This one might be better left as is.

RexJaeschke: I favor the link going to "unmanaged type," as that's what the text clearly says. Later, the para says "returns a pointer, of type `T*`, to ...", but until then no pointer is involved.

RexJaeschke: @BillWagner Unless I'm missing something, all your replacement section numbers appear to be hard-coded rather than being actual links of the form "[...](...)".

RexJaeschke (approved):

PR #734, opened 2023-02-06
Add a new template specifically for extern aliases, and use it to implement "missing" examples

jskeet:
In a couple of examples I've renamed the alias so we don't need quite as many - this feels reasonable to me, but please shout if you disagree.


RexJaeschke (approved):
LGTM, @jskeet. I didn't run any of the tests, but if they worked for you, I'm good to merge.

PR #740, opened 2023-02-21
Fix md formatting

RexJaeschke:
The uses of "get accessor" and "set accessor" are set inconsistently, mostly without backtick fences, but some with. This removes the backticks from those uses having them.


BillWagner (approved):
This LGTM @RexJaeschke 

I'll :shipit:

PR #746, opened 2023-03-06
Clarify method group conversion

### LINKED ISSUE #552: Method group conversion existence and compatibility is inconsistent
(Conversions) 10.8 starts:

> An implicit conversion exists from a method group ([§11.2](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/expressions.md#112-expression-classifications)) to a compatible delegate type ([§19.4](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/delegates.md#194-delegate-compatibility)).

But later we see:

> Note that the existence of an implicit conversion from E to D does not guarantee that the compile-time application of the conversion will succeed without error.

and

> Even if the conversion exists, a compile-time error occurs if the selected method M is not compatible ([§19.4](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/delegates.md#194-delegate-compatibility)) with the delegate type D.

The first sentence seems to at least *imply* that an implicit conversion *only* exists from a method group to a compatible delegate type, but the rest of the description appears to say that an implicit conversion exists from a method group to a delegate type if a single method within the group is applicable in its normal form, without considering compatibility.

Now I know the first sentence doesn't say that implicit conversions *only* exist from method groups to compatible delegate types, but I think that's a reasonable implication. I suspect we want to change the wording of the first paragraph to remove comatibility.

BillWagner:
Fixes #552

First, remove the bullet regarding a compile time error if the method group is not compatible with the delegate type. That's now handled by applicable function member.

Earlier, point out that the conversion from a method group to a delegate includes the restrictions from the section on Applicable Function member that there must be an identity conversion from the method return type to the delegate's return type


jskeet:
Will review tomorrow morning, UK time.

standard/conversions.md
```diff
@@ -887,13 +887,12 @@ Not every lambda expression can be converted to expression tree types. The conve
...
 
-The compile-time application of the conversion from a method group `E` to a delegate type `D` is described in the following. Note that the existence of an implicit conversion from `E` to `D` does not guarantee that the compile-time application of the conversion will succeed without error.
+The compile-time application of the conversion from a method group `E` to a delegate type `D` is described in the following. Note that the existence of an implicit conversion from `E` to `D` does not guarantee that the compile-time application of the conversion will succeed without error. Therefore, the rules for applicable function member (§11.6.4.2) require an identity conversion between the method return type and the delegate's return type.
```
jskeet, line 890: It may be worth highlighting that although the compatibility requirement means "An identity or implicit reference conversion exists from the return type of M to the return type of D", the requirement for an identity conversion is stronger than this. (It's easy to miss that it's *only* identity conversions, compared with "identity or implicit reference conversion".)

jskeet (approved):
It still feels weird to have this shadowy "exists but isn't useful conversion" but this change definitely improves things.

At some point it might be worth considering what would break if we expanded the earlier paragraph to include the return type restriction, so that basically the conversion would exist and work, or not exist at all. But I'm not suggesting we get into that right now :)

standard/conversions.md
BillWagner: Let's clarify that now.

Also, should I update the text in 11.6.4.2 to match? It correctly says "identity conversion" and doesn't include "implicit reference conversion".

```suggestion
The compile-time application of the conversion from a method group `E` to a delegate type `D` is described in the following. Note that the existence of an implicit conversion from `E` to `D` does not guarantee that the compile-time application of the conversion will succeed without error. Therefore, the rules for applicable function member (§11.6.4.2) require an identity or implicit reference conversion from the method return type to the delegate's return type.
```

jskeet: Hmm... possibly. I'd thought that *was* a deliberate difference, but it's not. For example, this code is fine:

```csharp
using System;

class Test
{
    static void Main()
    {
        Func<string, object> func = Method;
    }

    static string Method(string text)
    {
        return "";
    }
}
```

I think it would be useful to get an example of a case where the delegate conversion exists, but the implicit conversion from `E` to `D` fails at compile-time.

BillWagner: I was trying a few things this morning, and I haven't found one for an implicit *reference* conversion. Other implicit conversions (like from `int` to `long`) do fail: 

```csharp
sing System;

class Test
{
    static void Main()
    {
        Func<long, object> func = Method; // Error: No overload for 'Method' matches delegate 'Func<long, object>'
    }

    static int Method(string text)
    {
        return 2;
    }
}
```

jskeet: Right - and it's fine for that to fail, because the delegate type isn't compatible under the rules of [19.4](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/delegates.md#194-delegate-compatibility), so the conversion doesn't exist according to the section in this PR (the paragraph at 888).

jskeet: Basically, it's possible that over time we've removed the existence of these "present but inapplicable" conversions.

BillWagner: I think you're right. I did add a sentence in 11.6.4.2 to allow implicit reference conversions as well as identity conversions.

standard/expressions.md
```diff
@@ -993,7 +993,7 @@ Additional rules determine whether a method is applicable or not based on the co
   - If the method group results from a *simple_name*, an instance method is only applicable if `this` access is permitted [§11.7.12](expressions.md#11712-this-access).
 - When the method group results from a *member_access* which could be via either an instance or a type as described in [§11.7.6.2](expressions.md#11762-identical-simple-names-and-type-names), both instance and static methods are applicable.
 - A generic method whose type arguments (explicitly specified or inferred) do not all satisfy their constraints is not applicable.
-- In the context of a method group conversion, there must exist an identity conversion ([§10.2.2](conversions.md#1022-identity-conversion)) or an implicit reference conversion (§10.2.8) between the method return type and the delegate’s return type. Otherwise, the candidate method is not applicable.
+- In the context of a method group conversion, there must exist an identity conversion ([§10.2.2](conversions.md#1022-identity-conversion)) or an implicit reference conversion (§10.2.8) from the method return type to the delegate’s return type. Otherwise, the candidate method is not applicable.
```
jskeet: Definitely happy with this change. It may be worth moving this to a separate PR and get this merged while the rest is under discussion?

standard/conversions.md
jskeet: Let's discuss this at the next meeting unless we have any brainwaves between now and then. I've a feeling I'm forgetting something, but I don't know what yet...

standard/expressions.md
BillWagner: See #747

jskeet:
Ah - I hadn't intended to approve this one (or if I did, it was before some changes). I think we should still discuss it next meeting. I'll reopen the issue and label that, as a simple way of keeping track.

standard/conversions.md
Nigel-Ecma, line 890: See review comment.

```suggestion
The compile-time application of the conversion from a method group `E` to a delegate type `D` is described in the following.
```

standard/conversions.md
```diff
@@ -887,13 +887,12 @@ Not every lambda expression can be converted to expression tree types. The conve
...
 
 - A single method `M` is selected corresponding to a method invocation ([§11.7.8.2](expressions.md#11782-method-invocations)) of the form `E(A)`, with the following modifications:
   - The argument list `A` is a list of expressions, each classified as a variable and with the type and modifier (`ref` or `out`) of the corresponding parameter in the *formal_parameter_list* of `D` — excepting parameters of type `dynamic`, where the corresponding expression has the type `object` instead of `dynamic`.
   - The candidate methods considered are only those methods that are applicable in their normal form and do not omit any optional parameters ([§11.6.4.2](expressions.md#11642-applicable-function-member)). Thus, candidate methods are ignored if they are applicable only in their expanded form, or if one or more of their optional parameters do not have a corresponding parameter in `D`.
 - A conversion is considered to exist if the algorithm of [§11.7.8.2](expressions.md#11782-method-invocations) produces a single best method `M` having the same number of parameters as `D`.
```
Nigel-Ecma, line 895: See review comment

```suggestion
- A conversion is considered to exist if the algorithm of [§11.7.8.2](expressions.md#11782-method-invocations) produces a single best method `M` which is compatible [§19.4] with `D`.
```

Nigel-Ecma (commented):
The original text just seems strange, why have a note that says something incomplete about failure when later on it says what the requirements are (a) a single method must be found and (b) it must be compatible (which handles the allowed conversions) – though it handles the later sort of backwards

> *Even if the conversion exists, a compile-time error occurs if the selected method `M` is not compatible*

Surely we can just drop the note and say the single found method must be compatible?

Or am I missing something obvious?

I've suggested the two changes to write it as above. I *suspect* the bulleted list can be shortened further as it probably overlaps with delegate compatibility (§19.4) – but I have *not* looked into that so far.

PR #747, opened 2023-03-08
Add implicit reference conversions as allowed

BillWagner:
For a method group conversion, an implicit reference conversion (in addition to an identity conversion) is allowed.

See https://github.com/dotnet/csharpstandard/pull/746#discussion_r1129175646

While working on the updates for method group conversions to a delegate, Jon and I noticed that *implicit reference conversions* are allowed, in addition to *identity conversions*. 

We think this change is good, but we are concerned we may be missing something.


jskeet (approved):

BillWagner:
@jskeet Did I misinterpret your comment on #746 and merge the wrong PR? Which one did you want to wait for the next meeting?

jskeet:
@BillWagner: Yup - this bit feels like the "clearly correct" change whereas the other part feels more debatable to me.

PR #751, opened 2023-03-13
Update Method Group conversions per feedback

BillWagner:
I'd misinterpreted Jon's comment on #746 and merged before more reviews. Nigel had great suggestions, which are now included in this PR.

See https://github.com/dotnet/csharpstandard/pull/746#pullrequestreview-1334122657


jskeet (commented):
I'm fine with this if it works, but I would like to discuss it at the meeting - as I wonder whether we're missing some previous "exists but causes an error" case elsewhere. If we can't come up with any reason for that still being a possibility, I'd be thrilled to do this :)

Nigel-Ecma (approved):
LGTM :-)

Nigel-Ecma:
Agree, discuss at meeting. I think the old text was an imprecise and confusing way of covering what §19.4 covers. If there are missing case(s), and that is certainly possible, then any fix should surely be in §19.4 and not here?

BillWagner:
#499 has two examples that motivate this change.

jskeet (approved):

PR #752, opened 2023-03-13
Add an example that explains why an anonymous function body may be a *null_conditional_invocation_expression*.

### LINKED ISSUE #621: 14.6.1: Add example for null conditional invocation expression
14.6.1 includes:

> when recognising a method_body if both the null_conditional_invocation_expression and expression alternatives are applicable then the former shall be chosen.

11.16.1 includes the same text, and local functions will introduce it as well.

There's no motivation given for it; @gafter had an explanation around it being classified as "None" in a way which would make it invalid as an expression in some cases. (That's a very hand-wavy explanation - Neal will be able to express it better.)

We should provide motivation and an example (just informatively) in 14.6.1, and refer to it in 11.16.1 and local functions (#104).

gafter:
Fixes #621

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4443,12 +4443,12 @@ When recognising an *anonymous_function_body* if both the *null_conditional_invo
 
 > *Note*: The overlapping of, and priority between, alternatives here is solely for descriptive convenience; the grammar rules could be elaborated to remove the overlap. ANTLR, and other grammar systems, adopt the same convenience and so *anonymous_function_body* has the specified semantics automatically.
 
-> *Note*: When treated as an *expression*, a syntaxtic form such as `x?.M()` would be an error if the result type of `M` is `void` [§11.7.11](expressions.md#11711-null-conditional-element-access). But when treated as a *null_conditional_invocation_expression*, the result type is permitted to be `void`.
+> *Note*: When treated as an *expression*, a syntactic form such as `x?.M()` would be an error if the result type of `M` is `void` ([§11.7.11](expressions.md#11711-null-conditional-element-access)). But when treated as a *null_conditional_invocation_expression*, the result type is permitted to be `void`.
 
-> *Example*: In the following code, the body of the anonymous expression is a *null_conditional_invocation_expression*, and so the result type of `Array.Reverse` may be `void`:
+> *Example*: The result type of `Array.Reverse` is `void`.  In the following code, the body of the anonymous expression is a *null_conditional_invocation_expression*, so it is not an error.
 >
 > ```csharp
-> Action<Array> a = () => a?.Reverse();
+> Action<Array> a = x => x?.Reverse();
 > ```
 >
 > *end example*
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4443,6 +4443,16 @@ When recognising an *anonymous_function_body* if both the *null_conditional_invo
 
 > *Note*: The overlapping of, and priority between, alternatives here is solely for descriptive convenience; the grammar rules could be elaborated to remove the overlap. ANTLR, and other grammar systems, adopt the same convenience and so *anonymous_function_body* has the specified semantics automatically.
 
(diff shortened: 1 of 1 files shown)


standard/expressions.md
```diff
@@ -4443,6 +4443,16 @@ When recognising an *anonymous_function_body* if both the *null_conditional_invo
 
 > *Note*: The overlapping of, and priority between, alternatives here is solely for descriptive convenience; the grammar rules could be elaborated to remove the overlap. ANTLR, and other grammar systems, adopt the same convenience and so *anonymous_function_body* has the specified semantics automatically.
 
+> *Note*: When treated as an *expression*, a syntaxtic form such as `x?.M()` would be an error if the result type of `M` is `void` [§11.7.11](expressions.md#11711-null-conditional-element-access). But when treated as a *null_conditional_invocation_expression*, the result type is permitted to be `void`.
+
+> *Example*: In the following code, the body of the anonymous expression is a *null_conditional_invocation_expression*, and so the result type of `Array.Reverse` may be `void`:
```
Nigel-Ecma: The result type of `Array.Reverse` *is* `void`, the use of "may be" here is somewhat confusing.

standard/expressions.md
```diff
@@ -4443,6 +4443,16 @@ When recognising an *anonymous_function_body* if both the *null_conditional_invo
...
+>
+> ```csharp
+> Action<Array> a = () => a?.Reverse();
```
Nigel-Ecma: The first `a` is declared in this line, the second `a` is?

standard/expressions.md
```diff
@@ -4443,6 +4443,16 @@ When recognising an *anonymous_function_body* if both the *null_conditional_invo
 
 > *Note*: The overlapping of, and priority between, alternatives here is solely for descriptive convenience; the grammar rules could be elaborated to remove the overlap. ANTLR, and other grammar systems, adopt the same convenience and so *anonymous_function_body* has the specified semantics automatically.
 
+> *Note*: When treated as an *expression*, a syntaxtic form such as `x?.M()` would be an error if the result type of `M` is `void` [§11.7.11](expressions.md#11711-null-conditional-element-access). But when treated as a *null_conditional_invocation_expression*, the result type is permitted to be `void`.
```
Nigel-Ecma: Typo, I think the Standard normally would put the x-ref in parentheses, e.g. ([§11.7.11](expressions.md#11711-null-conditional-element-access))

Nigel-Ecma (changes requested):
Sorry but I just found this somewhat confusing.

standard/expressions.md
jskeet: And I assume that syntaxtic should be syntactic?

standard/expressions.md
```diff
@@ -4443,6 +4443,16 @@ When recognising an *anonymous_function_body* if both the *null_conditional_invo
```
BillWagner: This will disable the MD lint warning:

```suggestion
<!-- markdownlint-disable MD028 -->
```

standard/expressions.md
```diff
@@ -4443,6 +4443,16 @@ When recognising an *anonymous_function_body* if both the *null_conditional_invo
...
+> *Note*: When treated as an *expression*, a syntactic form such as `x?.M()` would be an error if the result type of `M` is `void` ([§11.7.11](expressions.md#11711-null-conditional-element-access)). But when treated as a *null_conditional_invocation_expression*, the result type is permitted to be `void`.
+
+> *Example*: The result type of `Array.Reverse` is `void`.  In the following code, the body of the anonymous expression is a *null_conditional_invocation_expression*, so it is not an error.
+>
+> ```csharp
+> Action<Array> a = x => x?.Reverse();
+> ```
+>
+> *end example*
+
```
BillWagner: And then re-enable it here:

```suggestion
<!-- markdownlint-enable MD028 -->
```

BillWagner (approved):
This looks great, following the updates per Nigel's feedback.

I added suggestions to address the build warnings.

BillWagner:
ping @Nigel-Ecma 

I believe Neal has addressed all your concerns. Is this ready to merge?

Nigel-Ecma (approved):

PR #755, opened 2023-03-15
Fix typos (missing white spaces)

pkulikov:


Nigel-Ecma (approved):
Thanks for spotting these!

Issue #744, opened 2023-03-01
Write something about spec compliance and Roslyn


jskeet:
(Known non-compliance etc.)

gafter:
Please see all of the documents in https://github.com/dotnet/roslyn/tree/main/docs/compilers/CSharp . Some of them are about compiler/spec deviations. Also the Microsoft-internal notebook mentioned by @333fred .

BillWagner:
closing as this should be done in the roslyn repo, not here.

Closed in favor of dotnet/roslyn#70043 to make sure we don't lose anything.

PR #756, opened 2023-03-17
Introduce a document for Roslyn inconsistencies

jskeet:
First pass at addressing #744.

I've included one simple example just for the sake of discussion.

Aspects we should discuss:

- The term "inconsistency" vs "compliance violation" etc
- The location and name of the new file
- How we keep it up-to-date with changes in the standard
- Linking to an existing doc if possible (with suitable wording changes to this doc)
- Format of the document
- Testing of the inconsistencies (can we use the existing test framework?)

### DIFF
--- a/roslyn-differences.md
+++ b/roslyn-differences.md
@@ -1,0 +1,41 @@
+# Inconsistencies between Roslyn and the C# standard
+
+As noted in the [README](README.md), although the C# compiler within [Roslyn](https://github.com/dotnet/roslyn) is the closest compiler we have to a reference implementation, there are areas where its behavior is inconsistent with the standard. This document lists those inconsistencies where they are known, organized by standard section.
+
+TODO: There was talk of a document in the Roslyn repo which mentioned this, but I can't find it.
+
+TODO: How do we keep section references in sync? Will we need to update the section renumbering code? What about quoted text? (Periodically just check it?)
+
+# Conversions
+
+## Implicit enum conversions from zero
+
+From [§10.2.4](standard/conversions.md#1024-implicit-enumeration-conversions):
+
+> An implicit enumeration conversion permits a *constant_expression* ([§11.21](standard/expressions.md#1121-constant-expressions)) with any integer type and the value zero to be converted to any *enum_type* and to any *nullable_value_type* whose underlying type is an *enum_type*.
+
+Roslyn performs implicit enumeration conversions from constant expressions with types of `float`, `double` and `decimal` as well:
+
+```csharp
+enum SampleEnum
+{
+    Zero = 0,
+    One = 1
+}
+
+class EnumConversionTest
+{
+    const float ConstFloat = 0f;
+    const double ConstDouble = 0d;
+    const decimal ConstDecimal = 0m;
+
+    static void PermittedConversions()
+    {
+        SampleEnum floatToEnum = ConstFloat;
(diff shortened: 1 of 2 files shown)


roslyn-differences.md
```diff
@@ -0,0 +1,41 @@
+# Inconsistencies between Roslyn and the C# standard
+
+As noted in the [README](README.md), although the C# compiler within [Roslyn](https://github.com/dotnet/roslyn) is the closest compiler we have to a reference implementation, there are areas where its behavior is inconsistent with the standard. This document lists those inconsistencies where they are known, organized by standard section.
+
+TODO: There was talk of a document in the Roslyn repo which mentioned this, but I can't find it.
```
BillWagner, line 5: Adding @jcouv @333fred who would know where such a document is, if it exists.

roslyn-differences.md
```diff
@@ -0,0 +1,41 @@
...
+
+TODO: How do we keep section references in sync? Will we need to update the section renumbering code? What about quoted text? (Periodically just check it?)
```
BillWagner, line 7: I could add this file to the list of files updated in the 2nd pass when the tool runs. It runs in 2 phases:

1. Determine all the new section numbers, map old section numbers to any updated changes.
2. Replace any changed section number links.

i like that, because I may also add this article to docs in the section near the standard.

BillWagner (commented):
I had a couple initial comments. This is a great start.

roslyn-differences.md
jcouv: I don't think we have such a document yet, but I think this document should live in the Roslyn repo instead.

333fred: There are a number of pages in our team onenote that talk about spec deviations. That isn't publicly available, and AFAIK most (if not all) of them are documented inline in the code where they occur (if the deviation is known, of course).

gafter:
Please see all of the documents in https://github.com/dotnet/roslyn/tree/main/docs/compilers/CSharp .  Some of them are about compiler/spec deviations.  Also the Microsoft-internal notebook mentioned by @333fred .

gafter:
Also, please see https://github.com/dotnet/csharpstandard/pull/601#discussion_r925709489 .  This is a whole category of cases in which the compiler intentionally permits things that are supposed to be an error.

roslyn-differences.md
jskeet: SGTM, thanks :)

roslyn-differences.md
jskeet: I'd definitely be happier linking to a doc in the Roslyn repo. I do think it's useful to have a doc that collects them in one place (for easy checking) rather than expecting users to hunt around.

If there isn't such a public doc at the moment, are folks happy for us to start it here and then request a transfer when we've collected the low-hanging fruit?

roslyn-differences.md
KalleOlaviNiemitalo: Have you considered assigning tags to sections like in C++ drafts?  (I don't know whether the tags are in the published standards, too.)  Would make links from external documents more resilient to renumbering.  <https://github.com/cplusplus/draft/blob/d544cfcaec7559fe3b40b81343293390d3727a2d/source/macros.tex#L90-L93>

jskeet:
Closing - we want this to be in the Roslyn repo.

PR #664, opened 2022-11-02
Discards, tuples and out variables

MadsTorgersen:
This is a very rough cut of discards, tuples and out variables, and is here for discussion/sharing purposes. It lacks examples and cross references throughout and doesn't touch on patterns at all. It's also missing treatment of additional declaration spaces at statement/expression level, as well as the syntactic short form `var (a, b)` in deconstruction.

The intent is to test out an alternative approach to #44, #63 and #596 along the lines suggested in #561, and to fill out several gaps in those PRs.

The approach differs in several ways:

- Discards are proper variables, only without names. Otherwise, the treatment of out vars cannot explain where the variable comes from that is passed to the out parameter.
- The entities on the left hand side of a deconstructing assignment are proper tuple expressions. Otherwise we cannot adequately account for when their elements are full-blown expressions.
- A new expression classification for `tuples` helps describe when, whether and how tuples should be evaluated.
- The variable-declaring capability of tuples and out variables is factored out into a new kind of expression, a "declaration expression" instead of adding new special cases to each of their productions.

I believe this approach is more correct now, and more resilient to future language additions. But there is still a ways to go to clean it up.

PR #757, opened 2023-03-19
Add a new chapter to define concepts related to pattern-matching.

gafter:
I want to add a new chapter because this will get a lot more complicated
later, for example when we add recursive patterns.

This is proposed to replace #61.


standard/expressions.md
```diff
@@ -6124,6 +6137,7 @@ Constant expressions are required in the contexts listed below and this is indic
 - `goto case` statements ([§12.10.4](statements.md#12104-the-goto-statement))
 - Dimension lengths in an array creation expression ([§11.7.15.5](expressions.md#117155-array-creation-expressions)) that includes an initializer.
 - Attributes ([§21](attributes.md#21-attributes))
+- In a *conatant_pattern* (§constant-pattern-new-clause)
```
```suggestion
- In a *constant_pattern* (§constant-pattern-new-clause)
```

BillWagner:
Leaving notes for discussion Wednesday on how to add a chapter and ensure our tools work:

The order of clauses is defined in [clauses.json](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/clauses.json) for the renumbering tool. The markdown converter relies on the section numbers added by the renumbering tool.

Three other tools all have arrays of filenames in the shell scripts that run the tool:

- [smarten](https://github.com/dotnet/csharpstandard/blob/draft-v7/tools/run-smarten.sh)
- [update grammar annex](https://github.com/dotnet/csharpstandard/blob/draft-v7/tools/update-grammar-annex.sh)
- [validate grammar](https://github.com/dotnet/csharpstandard/blob/draft-v7/tools/validate-grammar.sh)

I can take on the task to update those tools. I'll propose keying all the tools off the `clauses.json` file.

standard/expressions.md
```diff
@@ -3700,6 +3700,7 @@ relational_expression
     | relational_expression '<=' shift_expression
     | relational_expression '>=' shift_expression
     | relational_expression 'is' type
+    | relational_expression 'is' pattern
```
MadsTorgersen, line 3786: In the future there will be a type pattern, and these two forms of `is` operator will reunify. In anticipation of that I wonder if we should avoid splitting them out quite so deeply here and instead have a single production like

``` antlr
    | relational_expression 'is' pattern_or_type
```

standard/expressions.md
```diff
@@ -4033,7 +4034,11 @@ where `x` is an expression of a nullable value type, if operator overload resolu
...
 
-The `is` operator is used to check if the run-time type of an object is compatible with a given type. The check is performed at runtime. The result of the operation `E is T`, where `E` is an expression and `T` is a type other than `dynamic`, is a Boolean value indicating whether `E` is non-null and can successfully be converted to type `T` by a reference conversion, a boxing conversion, an unboxing conversion, a wrapping conversion, or an unwrapping conversion.
+There are two forms of the `is` operator.  One is the *is-type operator*, which has a type on the right-hand-side.  The other is the *is-pattern operator*, which has a pattern on the right-hand-side.
+
+#### 11.11.11.1 The is-type operator
+
+The *is-type operator* is used to check if the run-time type of an object is compatible with a given type. The check is performed at runtime. The result of the operation `E is T`, where `E` is an expression and `T` is a type other than `dynamic`, is a Boolean value indicating whether `E` is non-null and can successfully be converted to type `T` by a reference conversion, a boxing conversion, an unboxing conversion, a wrapping conversion, or an unwrapping conversion.
```
BillWagner, line 4160: Is "compatible" the right term here? I don't have a better term, but I'm concerned because only the listed conversions are in used.

standard/expressions.md
```diff
@@ -4033,7 +4034,11 @@ where `x` is an expression of a nullable value type, if operator overload resolu
 
 ### 11.11.11 The is operator
 
-The `is` operator is used to check if the run-time type of an object is compatible with a given type. The check is performed at runtime. The result of the operation `E is T`, where `E` is an expression and `T` is a type other than `dynamic`, is a Boolean value indicating whether `E` is non-null and can successfully be converted to type `T` by a reference conversion, a boxing conversion, an unboxing conversion, a wrapping conversion, or an unwrapping conversion.
+There are two forms of the `is` operator.  One is the *is-type operator*, which has a type on the right-hand-side.  The other is the *is-pattern operator*, which has a pattern on the right-hand-side.
+
+#### 11.11.11.1 The is-type operator
```
MadsTorgersen: Could we simply retcon the behavior of the is-type operator as a shorthand for `E is T _`? There seems to be an opportunity for much less duplication here.

standard/expressions.md
MadsTorgersen:
```suggestion
- In a *constant_pattern* (§constant-pattern-new-clause)
```

standard/lexical-structure.md
```diff
@@ -65,13 +65,14 @@ The productions for *simple_name* ([§11.7.4](expressions.md#1174-simple-names))
...
 
-then the *type_argument_list* is retained as part of the *simple_name*, *member_access*, or *pointer_member_access* and any other possible parse of the sequence of tokens is discarded. Otherwise, the *type_argument_list* is not considered part of the *simple_name*, *member_access*, or *pointer_member_access*, even if there is no other possible parse of the sequence of tokens.
+If the following token is among this list, or an identifier in such a context, then the *type_argument_list* is retained as part of the *simple_name*, *member_access* or  *pointer_member-access* and any other possible parse of the sequence of tokens is discarded.  Otherwise, the *type_argument_list* is not considered to be part of the *simple_name*, *member_access* or *pointer_member_access*, even if there is no other possible parse of the sequence of tokens. (These rules are not applied when parsing a *type_argument_list* in a *namespace_or_type_name* [§7.8](basic-concepts.md#78-namespace-and-type-names).)
```
MadsTorgersen, line 75: I have a bit of a hard time following the flow of this rule, as it first "sees if" certain conditions hold, and then later draws conclusions from that. Probably a bit of wordsmithing is all it takes.

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
+I would suggest placing this new chapter between the existing 12 (Statements) and 13 (namespaces)
```
MadsTorgersen: Another placement would be before the existing 11 (Expressions). We tend to introduce building blocks before we use them, and patterns are used by expressions and statements. 

Another option is to not make them a top level chapter but add them to the set of sections in the beginning of Expressions providing building blocks for the subsequent syntax-based sections. This is what I have proposed to do with "Deconstruction" in #644, and pattern matching in a later version of C# will make use of deconstruction, so having patterns follow deconstruction would fit with that.

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+
+# §patterns-new-clause Patterns and pattern matching
+
+## §patterns-new-clause-general General
+
+A ***pattern*** is a syntactic form that can be used with the `is` operator ([§11.11.11](expressions.md#111111-the-is-operator)) and in a *switch_statement* ([§12.8.3](statements.md#1283-the-switch-statement)) to express the shape of data against which incoming data is to be compared. A pattern is tested in the context of a switch expression or a *relational_expression* that is on the left-hand side of an `is` operator. We call this a ***pattern input value***.
```
MadsTorgersen:
```suggestion
A ***pattern*** is a syntactic form that can be used with the `is` operator ([§11.11.11](expressions.md#111111-the-is-operator)) and in a *switch_statement* ([§12.8.3](statements.md#1283-the-switch-statement)) to express the shape of data against which incoming data is to be compared. A pattern is tested against the *expression* of a switch statement, or against a *relational_expression* that is on the left-hand side of an `is` operator. We call this a ***pattern input value***.
```

standard/statements.md
```diff
@@ -640,29 +640,54 @@ switch_section
...
+case_guard
+    : 'when' expression
+    ;
 ```
 
-A *switch_statement* consists of the keyword `switch`, followed by a parenthesized expression (called the ***switch expression***), followed by a *switch_block*. The *switch_block* consists of zero or more *switch_section*s, enclosed in braces. Each *switch_section* consists of one or more *switch_label*s followed by a *statement_list* ([§12.3.2](statements.md#1232-statement-lists)).
+A *switch_statement* consists of the keyword `switch`, followed by a parenthesized expression (called the ***switch expression***), followed by a *switch_block*. The *switch_block* consists of zero or more *switch_section*s, enclosed in braces. Each *switch_section* consists of one or more *switch_label*s followed by a *statement_list* ([§12.3.2](statements.md#1232-statement-lists)). Each *switch_label* containing `case` has an associated pattern (§patterns-new-clause) against which the value of the switch expression is tested (XREF NEEDED). If *case-guard* is present, its expression shall implicitly convertible to the type `bool` and that expression is evaluated as an additional condition to be satisfied for the case to be considered satisfied.
```
BillWagner: nit, missing word:

```suggestion
A *switch_statement* consists of the keyword `switch`, followed by a parenthesized expression (called the ***switch expression***), followed by a *switch_block*. The *switch_block* consists of zero or more *switch_section*s, enclosed in braces. Each *switch_section* consists of one or more *switch_label*s followed by a *statement_list* ([§12.3.2](statements.md#1232-statement-lists)). Each *switch_label* containing `case` has an associated pattern (§patterns-new-clause) against which the value of the switch expression is tested (XREF NEEDED). If *case-guard* is present, its expression shall be implicitly convertible to the type `bool` and that expression is evaluated as an additional condition to be satisfied for the case to be considered satisfied.
```

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+pattern
+    : declaration_pattern
+    | constant_pattern
+    | var_pattern
+    ;
+```
+
+A *declaration_pattern* and a *var_pattern* can result in the declaration of a local variable.
+
+Each pattern form defines the set of types for input values that that the pattern may be applied to.  We say a pattern `P` is *applicable to* a type `T` if `T` is among the types whose values the pattern may match.  It is an error if a pattern `P` appears in a program to match a *pattern input value* of type `T` if `P` is not applicable to `T`.
```
MadsTorgersen:
```suggestion
Each pattern form defines the set of types for input values that the pattern may be applied to.  We say a pattern `P` is *applicable to* a type `T` if `T` is among the types whose values the pattern may match.  It is an error if a pattern `P` appears in a program to match a *pattern input value* of type `T` if `P` is not applicable to `T`.
```

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+
+Each pattern form defines the set of values for which the pattern *matches* the value.
+
+### §declaration-pattern-new-clause Declaration pattern
+
+A *declaration_pattern* is used to test that a value has a given type and if the test succeeds, to cast that value to that type.
```
MadsTorgersen: The term "cast" seems potentially suggestive of different behavior.

```suggestion
A *declaration_pattern* is used to test that a value has a given type and, if the test succeeds, provide the value in a variable of that type.
```

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+
+```ANTLR
+declaration_pattern
+    : type simple_designation
+    ;
+simple_designation
```
MadsTorgersen, line 33: Remind me why we have this chain of trivial productions - is it in anticipation of future additions?

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+    : single_variable_designation
+    ;
+single_variable_designation
+    : identifier
+    ;
+```
+
+The runtime type of the value is tested against the *type* in the pattern. If it is of that runtime type (or some subtype), the pattern *matches* that value. This pattern form never matches a `null` value.
```
MadsTorgersen, line 41: > If it is of that runtime type (or some subtype), the pattern *matches* that value.

This seems underspecified. Compare to the detail in the specification of the existing type-based `is` operator. In fact I'd suggest moving all that detail here and respecifying the `E is T` operator as simply a shorthand for `E is T _`.

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+
+Given a *pattern input value* *e*, if the *simple_designation* is the *identifier* `_`, it denotes a discard (§discards-new-clause) the value of *e* is not bound to anything. (Although a declared variable with the name `_` may be in scope at that point, that named variable is not seen in this context.) If *simple_designation* is any other identifier, a local variable ([§9.2.8](variables.md#928-local-variables)) of the given type named by the given identifier is introduced. That local variable is assigned the value of the *pattern input value* when the pattern *matches* the value.
```
MadsTorgersen: This may need a bit of tinkering once this PR meets up with the discards coming in from #664.

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+> ```
+>
+> with the slightly more concise
+>
+> ```csharp
+> if (expr is Type v) { /* code using v */ }
+> ```
+>
+> *end example*
+It is an error if *type* is a nullable value type.
```
MadsTorgersen, line 65: Missing blank line after `> *end example*`

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+
+> *Example*: The declaration pattern can be used to test values of nullable types: a value of type `Nullable<T>` (or a boxed `T`) matches a type pattern `T2 id` if the value is non-null and the type of `T2` is `T`, or some base type or interface of `T`. For example, in the code fragment
```
MadsTorgersen:
```suggestion
> *Example*: The declaration pattern can be used to test values of nullable types: a value of type `Nullable<T>` (or a boxed `T`) matches a type pattern `T2 id` if the value is non-null and `T2` is `T`, or some base type or interface of `T`. For example, in the code fragment
```

standard/variables.md
```diff
@@ -287,8 +287,49 @@ if ( «expr» ) «then_stmt» else «else_stmt»
...
+
+The definite-assignment state of *v* at the beginning of a case's guard clause is
+- If *v* is a pattern variable declared in the *switch_label*: "definitely assigned".
+- If the switch label containing that guard clause is not reachable (XREF): "definitely assigned".
+- Otherwise, the state of *v* is the same as the state of *v* after *expr*.
+
+The definite-assignment state of *v* on the control flow transfer to a reachable switch block statement list is
+- If the control transfer was due to a 'goto case' or 'goto default' statement, then the state of *v* is the same as the state at the beginning of that 'goto' statement.
+- If the control transfer was due to the `default` label of the switch, then the state of *v* is the same as the state of *v* after *expr*.
+- If the control transfer was due to an unreachable switch label, then the state of *v* is "definitely assigned".
```
BillWagner, line 329: This one confuses me. I think I need an example to understand the implications of v being definitely assigned due to an unreachable switch label.

standard/patterns.md
BillWagner, line 41: This should include boxing and unboxing conversions, right?

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+- `P` is a var pattern and the set of patterns `Q` is *exhaustive* for the type of the pattern input value, and either the pattern input value is not of a nullable type or some pattern in `Q` would match `null`.
+- `P` is a declaration pattern with type `T` and the set of patterns `Q` is *exhaustive* for the type `T` (XREF).
+
+## Pattern Exhaustiveness
+
+Informally, we say that a set of patterns is exhaustive for a type if some pattern in the set is applicable to every possible value of that type other than null.
+Here we define when a set of patterns is *exhaustive* for a type.
+
+A set of patterns `Q` is *exhaustive* for a type `T` if any of the following conditions hold:
+1. `T` is an integral or enum type, or a nullable version of one of those, and for every possible value of `T`'s underlying type, some pattern in `Q` would that value; or
```
MadsTorgersen:
```suggestion
1. `T` is an integral or enum type, or a nullable version of one of those, and for every possible value of `T`'s underlying type, some pattern in `Q` would match that value; or
```

standard/statements.md
MadsTorgersen:
```suggestion
A *switch_statement* consists of the keyword `switch`, followed by a parenthesized expression (called the ***switch expression***), followed by a *switch_block*. The *switch_block* consists of zero or more *switch_section*s, enclosed in braces. Each *switch_section* consists of one or more *switch_label*s followed by a *statement_list* ([§12.3.2](statements.md#1232-statement-lists)). Each *switch_label* containing `case` has an associated pattern (§patterns-new-clause) against which the value of the switch expression is tested (XREF NEEDED). If *case-guard* is present, its expression shall be implicitly convertible to the type `bool` and that expression is evaluated as an additional condition for the case to be considered satisfied.
```

BillWagner (commented):
I do like the idea of a new clause for patterns. I had a few comments and nits, but otherwise this is great.

standard/expressions.md
gafter: This is the existing paragraph.  Such introductory paragraphs are often informal and then clarified by the text that follows.  Where I used that pattern, I introduced the text with "Informally, ".  See a couple of examples in this PR.

standard/patterns.md
gafter: Before expressions makes sense to me.  I suggested after only because expressions seem like a more fundamental and important concept, and each uses the other.

The patterns "chapter" will get more complicated when we add more pattern concepts in the next couple of versions (and in the future if discriminated unions are added), so I wanted to break them out so as not to interrupt the flow of the expressions description.

standard/patterns.md
gafter: Yes, we have tuple-like designations in the future, and it becomes optional.

standard/expressions.md
gafter: They are different.

https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYZQAs6AsgBQCU+hxTAlgGbrVzrMDO6AcgFcANsICGAI2EBTADzMAdgBcAfPTwBfRkyJsOXXgJHipcxavQB9dVqRNbt5FgCC+W0A

standard/expressions.md
```diff
@@ -4033,7 +4034,11 @@ where `x` is an expression of a nullable value type, if operator overload resolu
...
+
+The *is-type operator* is used to check if the run-time type of an object is compatible with a given type. The check is performed at runtime. The result of the operation `E is T`, where `E` is an expression and `T` is a type other than `dynamic`, is a Boolean value indicating whether `E` is non-null and can successfully be converted to type `T` by a reference conversion, a boxing conversion, an unboxing conversion, a wrapping conversion, or an unwrapping conversion.
 
 The operation is evaluated as follows:
```
gafter, line 4162: E is no longer permitted to be a method group.  I don't know when that change happened.

standard/patterns.md
gafter: The is-type operator permits a nullable type on the right-hand-side.  The is-pattern operator does not.  But we can still unify most of it.

gafter: Yes.  We should unify with the description of the is-type operator.

standard/patterns.md
gafter, line 66:
```suggestion

It is an error if *type* is a nullable value type.
```

standard/variables.md
gafter:
```
bool b;
switch (1) {
  case 2 when b: // b is definitely assigned here.
    break;
}
```

standard/expressions.md
```diff
@@ -4071,6 +4076,14 @@ User defined conversions are not considered by the `is` operator.
 >
 > *end note*
 
+#### 11.11.11.2 The is-pattern operator
+
+The *is-pattern operator* is used to check if the value computed by an expression *matches* a given pattern (XREF TO DEF OF "PATTERN MATCHES"). The check is performed at runtime. The result of the is-pattern operator is true if the value matches the pattern; otherwise it is false.
+
+For an expression of the form `E is P`, where `E` is a relational expression of type `T` and `P` is a pattern, it is a compile-time error if any of the following hold:
+- `E` does not designate a value or does not have a type.
```
gafter, line 4204:
```suggestion

- `E` does not designate a value or does not have a type.
```

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+> if (v != null) { /* code using v */ }
+> ```
+>
+> with the slightly more concise
+>
+> ```csharp
+> if (expr is Type v) { /* code using v */ }
+> ```
+>
+> *end example*
```
gafter, line 63:
```suggestion
> *end example*

```

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+```ANTLR
+constant_pattern
+    : constant_expression
+    ;
+```
+
+A constant pattern `P` is *applicable to* a type `T` if there is an implicit conversion from the constant expression of `P` to the type `T`.
+
+For a constant pattern `P`, we say its *converted value* is
+- if the input expression's type is an integral type or an enum type, the pattern's constant value converted to that type; otherwise
```
gafter, line 90:
```suggestion

- if the input expression's type is an integral type or an enum type, the pattern's constant value converted to that type; otherwise
```

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+- if the input expression's type is the nullable version of an integral type or an enum type, the pattern's constant value converted to its underlying type; otherwise
+- the value of the pattern's constant value.
+
+Given a *pattern input value* *e* and a a constant pattern `P` with converted value *v*,
+- if *e* has integral type or enum type, or a nullable form of one of those, and *v* has integral type, the pattern `P` *matches* the value *e* if result of the expression `e == v` is `true`; otherwise
```
gafter, line 96:
```suggestion

- if *e* has integral type or enum type, or a nullable form of one of those, and *v* has integral type, the pattern `P` *matches* the value *e* if result of the expression `e == v` is `true`; otherwise
```

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+## Pattern Subsumption
+
+In a switch statement, it is an error if a case's pattern is *subsumed* by the preceding set of unguarded cases (XREF).
+Informally, this means that any input value would have been matched by one of the previous cases.
+Here we define when a set of patterns *subsumes* a given pattern.
+
+We say a pattern `P` *would match* a constant `K` if the specification for that pattern's runtime behavior is that `P` matches `K`.
+
+A set of patterns `Q` *subsumes* a pattern `P` if any of the following conditions hold:
+- `P` is a constant pattern and any of the patterns in the set `Q` would match `P`'s *converted value*
```
gafter, line 146:
```suggestion

- `P` is a constant pattern and any of the patterns in the set `Q` would match `P`'s *converted value*
```

standard/patterns.md
```diff
@@ -0,0 +1,172 @@
...
+- `P` is a var pattern and the set of patterns `Q` is *exhaustive* for the type of the pattern input value, and either the pattern input value is not of a nullable type or some pattern in `Q` would match `null`.
+- `P` is a declaration pattern with type `T` and the set of patterns `Q` is *exhaustive* for the type `T` (XREF).
+
+## Pattern Exhaustiveness
+
+Informally, we say that a set of patterns is exhaustive for a type if some pattern in the set is applicable to every possible value of that type other than null.
+Here we define when a set of patterns is *exhaustive* for a type.
+
+A set of patterns `Q` is *exhaustive* for a type `T` if any of the following conditions hold:
```
gafter, line 153:
```suggestion
A set of patterns `Q` is *exhaustive* for a type `T` if any of the following conditions hold:

```

standard/statements.md
```diff
@@ -830,18 +855,45 @@ When the governing type of a `switch` statement is `string` or a nullable value
...
 The *statement_list*s of a *switch_block* may contain declaration statements ([§12.6](statements.md#126-declaration-statements)). The scope of a local variable or constant declared in a switch block is the switch block.
 
-The statement list of a given switch section is reachable if the `switch` statement is reachable and at least one of the following is true:
-
-- The switch expression is a non-constant value.
-- The switch expression is a constant value that matches a `case` label in the switch section.
-- The switch expression is a constant value that doesn’t match any `case` label, and the switch section contains the `default` label.
-- A switch label of the switch section is referenced by a reachable `goto case` or `goto default` statement.
+A switch label is reachable if at least one of the following is true:
+- The switch expression is a constant value and either
```
gafter, line 869:
```suggestion

- The switch expression is a constant value and either
```

standard/statements.md
```diff
@@ -830,18 +855,45 @@ When the governing type of a `switch` statement is `string` or a nullable value
...
+  - the label is a `case` without a guard or with a guard whose value is not the constant false; or
+  - it is a `default` label and
+    - the set of patterns appearing among the cases of the switch statement that do not have guards or have guards whose value is the constant true, is not *exhaustive* (NEED XREF) for the switch controlling type; or
+    - the switch controlling type is a nullable type and the set of patterns appearing among the cases of the switch statement that do not have guards or have guards whose value is the constant true does not contain a pattern that would match the value `null`.
+- The switch label is referenced by a reachable `goto case` or `goto default` statement.
 
-The end point of a `switch` statement is reachable if at least one of the following is true:
+The statement list of a given switch section is reachable if the `switch` statement is reachable and the switch section contains a reachable switch label.
 
+The end point of a `switch` statement is reachable if the switch statement is reachable and at least one of the following is true:
```
gafter, line 881:
```suggestion
The end point of a `switch` statement is reachable if the switch statement is reachable and at least one of the following is true:

```

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+
+```ANTLR
+pattern
+    : declaration_pattern
+    | constant_pattern
+    | var_pattern
+    ;
+```
+
+A *declaration_pattern* and a *var_pattern* can result in the declaration of a local variable.
```
jskeet, line 19: *Can* result, or *always* result (but possibly without the variable being definitely assigned)?

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+
+Each pattern form defines the set of types for input values that the pattern may be applied to.  We say a pattern `P` is *applicable to* a type `T` if `T` is among the types whose values the pattern may match.  It is an error if a pattern `P` appears in a program to match a *pattern input value* of type `T` if `P` is not applicable to `T`.
```
jskeet, line 21: I think we may want an example here to clarify this. I believe I get it, but it could be confusing.

For example, consider:

```csharp
if (x is string s)
```

It's fine for the compile-time type of `x` to be `object`, even though a value whose execution type is `object` *won't* match. But the expression type being `object` is okay because the value *might* be a `string` reference. Whereas if the compile-time type of `x` is (say) `Stream` then this would be a compile-time error.

I think that sort of clarification can be added later - I don't want us to wait for perfection before merging this, so long as the text is accurate when read appropriately (and I think it is here). We might want to keep a list of this sort of thing somewhere, so we can come back afterwards. (It's inevitable when adding a huge chunk this PR does.)

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+    : single_variable_designation
+    ;
+single_variable_designation
+    : identifier
+    ;
+```
+
+The runtime type of the value is tested against the *type* in the pattern. If it is of that runtime type (or some subtype), the pattern *matches* that value. This pattern form never matches a `null` value.
+
+Given a *pattern input value* *e*, if the *simple_designation* is the *identifier* `_`, it denotes a discard (§discards-new-clause) the value of *e* is not bound to anything. (Although a declared variable with the name `_` may be in scope at that point, that named variable is not seen in this context.) If *simple_designation* is any other identifier, a local variable ([§9.2.8](variables.md#928-local-variables)) of the given type named by the given identifier is introduced. That local variable is assigned the value of the *pattern input value* when the pattern *matches* the value.
```
jskeet: I suspect the rendering of *pattern input value* *e* would be clearer without "pattern input value" in italics. Or we could try to reword, something like "Given a *pattern input value* of the form *e*" - although that implies there are other forms. Thoughts welcome.

standard/patterns.md
jskeet: Ah - discards, right?

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+
+Certain combinations of static type of the pattern input value and the given type are considered incompatible and result in a compile-time error. A value of static type `E` is said to be ***pattern compatible*** with the type `T` if there exists an identity conversion, an implicit reference conversion, a boxing conversion, an explicit reference conversion, or an unboxing conversion from `E` to `T`, or if either `E` or `T` is an open type ([§8.4.3](types.md#843-open-and-closed-types)).  A declaration pattern naming a type `T` is *applicable to* every type `E` for which `E` is *pattern compatible* with `T`.
```
jskeet, line 45: Can we reduce "an implicit reference conversion, a boxing conversion, an explicit reference conversion" to "a reference conversion, a boxing conversion, ..."?

jskeet: Nit: the "period double-space" before "A declaration pattern" makes me twitch. I'm happy to change those afterwards :)

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+
+A constant pattern `P` is *applicable to* a type `T` if there is an implicit conversion from the constant expression of `P` to the type `T`.
+
+For a constant pattern `P`, we say its *converted value* is
+
+- if the input expression's type is an integral type or an enum type, the pattern's constant value converted to that type; otherwise
+- if the input expression's type is the nullable version of an integral type or an enum type, the pattern's constant value converted to its underlying type; otherwise
+- the value of the pattern's constant value.
+
+Given a *pattern input value* *e* and a a constant pattern `P` with converted value *v*,
```
```suggestion
Given a *pattern input value* *e* and a constant pattern `P` with converted value *v*,
```

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+
+- if *e* has integral type or enum type, or a nullable form of one of those, and *v* has integral type, the pattern `P` *matches* the value *e* if result of the expression `e == v` is `true`; otherwise
+- the pattern `P` *matches* the value *e* if `object.Equals(e, v)` returns `true`.
+
+> *Example*:
```
jskeet, line 99: I think this example could do with a bit more explanation - particularly as unless you're attuned to it already, you might be wondering where the pattern is.

(Also, please put the opening brace on a new line.)

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+For a constant pattern `P`, we say its *converted value* is
+
+- if the input expression's type is an integral type or an enum type, the pattern's constant value converted to that type; otherwise
+- if the input expression's type is the nullable version of an integral type or an enum type, the pattern's constant value converted to its underlying type; otherwise
+- the value of the pattern's constant value.
+
+Given a *pattern input value* *e* and a a constant pattern `P` with converted value *v*,
+
+- if *e* has integral type or enum type, or a nullable form of one of those, and *v* has integral type, the pattern `P` *matches* the value *e* if result of the expression `e == v` is `true`; otherwise
+- the pattern `P` *matches* the value *e* if `object.Equals(e, v)` returns `true`.
```
jskeet, line 97: It looks like the compiler is allowed to generate code with the same effect as that, but without actually calling `object.Equals` (which will box for float/decimal/double of course). For example, I've just tried `d is double.NaN` and Roslyn compiles that to `double.IsNaN(d)`.
Do we need to mention that?

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+
+A set of patterns `Q` *subsumes* a pattern `P` if any of the following conditions hold:
+
+- `P` is a constant pattern and any of the patterns in the set `Q` would match `P`'s *converted value*
+- `P` is a var pattern and the set of patterns `Q` is *exhaustive* for the type of the pattern input value, and either the pattern input value is not of a nullable type or some pattern in `Q` would match `null`.
+- `P` is a declaration pattern with type `T` and the set of patterns `Q` is *exhaustive* for the type `T` (XREF).
+
+## Pattern Exhaustiveness
+
+Informally, we say that a set of patterns is exhaustive for a type if some pattern in the set is applicable to every possible value of that type other than null.
```
jskeet, line 152: There are two ways of reading "if some pattern in the set is applicable to every possible value of that type" - my initial reading is "one of the patterns in the set has the property that it matches every possible value of that type". Whereas I assume you really mean "if every possible value of that type is matched by at least one pattern in the set".

Let's add this to the "non-blocking, could do with tweaking" list - assuming I've read it correctly.

standard/statements.md
```diff
@@ -640,29 +640,54 @@ switch_section
...
+>
+> ```csharp
+> switch (shape)
+> {
+>     case var x:
+>         break;
+>     case var _: // error: pattern subsumed, as previous case always matches
+>         break;
+>     default:
+>         break;  // warning: unreachable,
```
jskeet, line 685: Incomplete comment?

standard/statements.md
```diff
@@ -830,18 +855,47 @@ When the governing type of a `switch` statement is `string` or a nullable value
...
-The statement list of a given switch section is reachable if the `switch` statement is reachable and at least one of the following is true:
+A switch label is reachable if at least one of the following is true:
+
+- The switch expression is a constant value and either
+  - the label is a `case` whose pattern *would match* (XREF to "would match" in patterns.md) that value, and label's guard is either absent or not a constant expression with the value false; or
+  - it is a `default` label, and no switch section contains a case label whose pattern would match that value, and whose guard is either absent or a constant expression with the value true.
+- The switch expression is not a constant value and either
+  - the label is a `case` without a guard or with a guard whose value is not the constant false; or
+  - it is a `default` label and
+    - the set of patterns appearing among the cases of the switch statement that do not have guards or have guards whose value is the constant true, is not *exhaustive* (NEED XREF) for the switch controlling type; or
```
jskeet, line 875: governing type here and below? (Instead of controlling type.)

standard/statements.md
```diff
@@ -830,18 +855,47 @@ When the governing type of a `switch` statement is `string` or a nullable value
...
 
 - The `switch` statement contains a reachable `break` statement that exits the `switch` statement.
-- The `switch` statement is reachable, the switch expression is a non-constant value, and no `default` label is present.
-- The `switch` statement is reachable, the switch expression is a constant value that doesn’t match any `case` label, and no `default` label is present.
+- No `default` label is present and either
+    - The switch expression is a non-constant value, and the set of patterns appearing among the cases of the switch statement that do not have guards or have guards whose value is the constant true, is not *exhaustive* (NEED XREF) for the switch governing type.
+    - The switch expression is a non-constant value of a nullable type, and no pattern appearing among the cases of the switch statement that do not have guards or have guards whose value is the constant true would match the value `null` (XREF to "would match" in patterns).
+    - The switch expression is a constant value and no `case` label without a guard or whose guard is the constant true would match that value.
+
+> *Example*: The following code shows a succinct use of the when clause:
```
jskeet, line 889:
```suggestion
> *Example*: The following code shows a succinct use of the `when` clause:
```

standard/statements.md
```diff
@@ -830,18 +855,47 @@ When the governing type of a `switch` statement is `string` or a nullable value
...
+>
+> ```csharp
+> static object CreateShape(string shapeDescription)
+> {
+>    switch (shapeDescription)
+>    {
+>         case "circle":
+>             return new Circle(2);
+>         …
+>         case var o when (o?.Trim().Length ?? 0) == 0:
```
jskeet, line 899: I'd normally write this as `case var o when string.IsNullOrEmpty(o?.Trim())`

Happy to discuss merits either way (with the context of this example).

jskeet: (I suppose actually I'd use `string.IsNullWhiteSpace` actually...)

jskeet (commented):
Various comments, but I like having patterns in their own section. Let's talk in the meeting about concrete steps going from here to merged.

(This is the sort of thing where I'm a terribly judge of whether it's actually "done" or not - there are any number of paragraphs that Neal could have omitted and I'd never have noticed...)

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+
+A *declaration_pattern* and a *var_pattern* can result in the declaration of a local variable.
+
+Each pattern form defines the set of types for input values that the pattern may be applied to.  We say a pattern `P` is *applicable to* a type `T` if `T` is among the types whose values the pattern may match.  It is an error if a pattern `P` appears in a program to match a *pattern input value* of type `T` if `P` is not applicable to `T`.
+
+Each pattern form defines the set of values for which the pattern *matches* the value.
+
+### §declaration-pattern-new-clause Declaration pattern
+
+A *declaration_pattern* is used to test that a value has a given type and, if the test succeeds, provide the value in a variable of that type.
```
Nigel-Ecma, line 27: From below its "a value has a type which is pattern compatible with a given type" – which is a bit long winded so some wordsmithing is required…

standard/patterns.md
Nigel-Ecma: Also should "subtype" be "pattern-compatilble type"?

standard/expressions.md
jskeet, line 3786: Meeting notes: there's ambiguity here. We should somehow reference how it's resolved. We *may* be able to merge these, but we'll need to talk about how nullable value types are okay for `type` but not `pattern`.

standard/expressions.md
gafter: Also "of if" below should be "or if".  Better sooner rather than later.

standard/expressions.md
gafter: We need to deal with the fact that the right-hand-side has to have a precedence tighter than expression, or at least as tight as relational_expression.  We may have to bifurcate the grammar for "pattern".

gafter: Also weird lookup: right-hand-side first as a type then as an expression.  This probably ought to be documented, but it isn't urgent and could also be swept under the rug.

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+    | constant_pattern
+    | var_pattern
+    ;
+```
+
+A *declaration_pattern* and a *var_pattern* can result in the declaration of a local variable.
+
+Each pattern form defines the set of types for input values that the pattern may be applied to.  We say a pattern `P` is *applicable to* a type `T` if `T` is among the types whose values the pattern may match.  It is an error if a pattern `P` appears in a program to match a *pattern input value* of type `T` if `P` is not applicable to `T`.
+
+Each pattern form defines the set of values for which the pattern *matches* the value.
```
gafter, line 23: Say this is a runtime concept.  (Not urgent, or even necessary)

standard/patterns.md
gafter: Mention this is a compile-time concept.

The simplest way to do that is change "it is an error" to "it is a compile-time error".

standard/patterns.md
gafter: No, this is a different, runtime concept.

(no action needed)

gafter: Should take its description from the is-type operator.

Best addressed sooner rather than later.

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+>
+> with the slightly more concise
+>
+> ```csharp
+> if (expr is Type v) { /* code using v */ }
+> ```
+>
+> *end example*
+
+It is an error if *type* is a nullable value type.
```
gafter, line 65: ... unless it is the type of an is-type expression?

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+constant_pattern
+    : constant_expression
+    ;
+```
+
+A constant pattern `P` is *applicable to* a type `T` if there is an implicit conversion from the constant expression of `P` to the type `T`.
+
+For a constant pattern `P`, we say its *converted value* is
+
+- if the input expression's type is an integral type or an enum type, the pattern's constant value converted to that type; otherwise
```
gafter, line 90: not called input expression...

standard/patterns.md
gafter: Please propose an edit.

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+>         default: throw new ArgumentException(…);
+>     }
+> }
+> ```
+>
+> *end example*
+
+### §var-pattern-new-clause Var pattern
+
+A *var_pattern* matches every value.  That is, a pattern-matching operation with a *var_pattern* always succeeds.
```
gafter, line 119: Matches in italics

jskeet:
- Bill to add a commit to fix the GH Actions
- Neal can address any trivial issues
- If we can get the new xrefs working before merging, let's try to get that in
- We can then merge, and then fix any remaining nits afterwards
- If we didn't get xrefs working before, do it now.

standard/patterns.md
```diff
@@ -0,0 +1,176 @@
...
+- `P` is a var pattern and the set of patterns `Q` is *exhaustive* for the type of the pattern input value, and either the pattern input value is not of a nullable type or some pattern in `Q` would match `null`.
+- `P` is a declaration pattern with type `T` and the set of patterns `Q` is *exhaustive* for the type `T` (XREF).
+
+## Pattern Exhaustiveness
+
+Informally, we say that a set of patterns is exhaustive for a type if some pattern in the set is applicable to every possible value of that type other than null.
+Here we define when a set of patterns is *exhaustive* for a type.
+
+A set of patterns `Q` is *exhaustive* for a type `T` if any of the following conditions hold:
+1. `T` is an integral or enum type, or a nullable version of one of those, and for every possible value of `T`'s underlying type, some pattern in `Q` would match that value; or
```
gafter, line 157: "underlying" is meant to apply to nullable types, not enum types.

BillWagner:
My last commit gets the section renumber tool to work correctly (once #664 is merged).

I'll make a fresh PR to get the other tools using the same config file for chapter sources

standard/patterns.md
```diff
@@ -0,0 +1,174 @@
...
+- if the input expression's type is an integral type or an enum type, the pattern's constant value converted to that type; otherwise
+- if the input expression's type is the nullable version of an integral type or an enum type, the pattern's constant value converted to its underlying type; otherwise
+- the value of the pattern's constant value.
+
+Given a *pattern input value* *e* and a constant pattern `P` with converted value *v*,
+
+- if *e* has integral type or enum type, or a nullable form of one of those, and *v* has integral type, the pattern `P` *matches* the value *e* if result of the expression `e == v` is `true`; otherwise
+- the pattern `P` *matches* the value *e* if `object.Equals(e, v)` returns `true`.
+
+> *Example*:
```
jskeet, line 99:
```suggestion
> *Example*: The `switch` statement in the following method uses five constant patterns in its case labels.
```

jskeet:
I'm going to rebase this, fix the conflicts, and merge it, so that we've got it in the Word document draft. The remaining comments will still need to be addressed, but that can happen afterwards.

jskeet (approved):

BillWagner:
> I'm going to rebase this, fix the conflicts, and merge it, so that we've got it in the Word document draft. The remaining comments will still need to be addressed, but that can happen afterwards.

Good plan. Can you ping me when you merge it? (@ - mention is fine). I'll likely have some downstream links to fix once it's merged, and the sections have been renumbered.

jskeet:
@BillWagner: I've temporarily disabled the check for all conversations to be resolved (as we want to address those afterwards) but we need to get the renumbering working. Are you happy to pull this branch, fix that up and repush? I'm hoping the Word converter will work *after* the renumbering is done.

BillWagner:
> @BillWagner: I've temporarily disabled the check for all conversations to be resolved (as we want to address those afterwards) but we need to get the renumbering working. Are you happy to pull this branch, fix that up and repush? I'm hoping the Word converter will work _after_ the renumbering is done.

Yes. I can do it. My comment https://github.com/dotnet/csharpstandard/pull/757#issuecomment-1476571863 has my notes from earlier.

standard/expressions.md
gafter: A pattern can derive an expression (via `constant_pattern` and `constant_expression`), but in this context the expression must have precedence of `shift_expression` or tighter (or there would be a grammar ambiguity).  We should open an issue to reflect the fact that work to specify that fact still needs to be done someday.  Alternately, we could just say that in a note in the spec and leave the grammar munging as an exercise to the reader.

standard/lexical-structure.md
gafter: I agree.  I don't think there is any urgency to address this immediately.

standard/patterns.md
gafter: Maybe, maybe not.  I suggest creating an issue to examine this in detail at some later time.  I suspect no action is needed.

standard/expressions.md
gafter: I think the meaning of `compatible` in this context is clear, as it is explained in the rest of the paragraph.  If it isn't clear, we could add  the parenthetical phrase "in the sense defined below" or something of the sort after the word "compatible".  I think this can be left to be dealt with later.

standard/variables.md
gafter: I don't think this is a common enough case to deserve an example in the text, though I admit it would help the reader make sense of the text.  This can be left to be dealt with later.

standard/patterns.md
gafter: This section should refer to the specification of the is-type operator for the meaning of when it is true or false.  This is probably better done sooner rather than later.

standard/expressions.md
gafter: Two changes should be made sooner rather than later:
1. We should document the error `error CS0837: The first operand of an 'is' or 'as' operator may not be a lambda expression, anonymous method, or method group.` and update this section accordingly.
2. Change the words "of if" below to "or if".

standard/patterns.md
gafter: Right.

standard/patterns.md
gafter: Agreed, that expanding upon this, perhaps with a sentence or two just as you wrote here, is worth doing (later)

standard/patterns.md
gafter: I'm open to seeing the wording reconsidered.  Also, the type matching is more than just subtype.  I believe it should delegate the specification to the `is-type` operator (e.g. include boxing and unboxing).  Not urgent.

standard/patterns.md
gafter: > Can we reduce "an implicit reference conversion, a boxing conversion, an explicit reference conversion" to "a reference conversion, a boxing conversion, ..."?

We don't define the phrase "reference conversion".  However, we could say "implicit or explicit reference conversion".  However, this is super low priority.

> Nit: the "period double-space" before "A declaration pattern" makes me twitch. I'm happy to change those afterwards :)

One of us will end up twitching.

standard/patterns.md
gafter: Agreed.

standard/statements.md
gafter: Just remove the comma.  Or add an explanation "all possible values already handled".

Low priority.

standard/statements.md
gafter: Yes, please change.

standard/statements.md
gafter: Yes, should change.  Low priority, but also trivial to do.

standard/statements.md
gafter: It's not wrong, but I agree that would be better guidance.  Low priority.

standard/patterns.md
gafter: This would be better done sooner rather than later.  Two things:
1. The specification about the value being of the given type is actually a bit more complex and should delegate to the spec of the is-type operator.  This is a runtime type test.
2. The *pattern compatible* concept is to give warnings based on compile-time types that are impossible to match.  If that isn't clear, it should be made clear.

standard/expressions.md
gafter: Agreed, and also mention that the precedence of an expression, if the pattern is a constant pattern, is restricted to `shift_expression` or tighter.

gafter: Though honestly it could be done later too.

standard/patterns.md
gafter: Change "input expression" to "pattern input value".

standard/patterns.md
gafter: Generally, no.

standard/patterns.md
gafter: This minor typographical change can be deferred.

standard/patterns.md
gafter: Should be done sooner rather than later.

standard/patterns.md
gafter: Sooner, please.

standard/patterns.md
gafter: change "underlying" to "nullable underlying".

standard/expressions.md
BillWagner: Adding the note in #934

standard/lexical-structure.md
BillWagner: Logged #935 to address later.

standard/patterns.md
BillWagner: Logged in #936

standard/expressions.md
BillWagner: I raised it originally, so I'm happy leaving as is, unless we get feedback outside the committee.

standard/variables.md
BillWagner: Logged #937 to discuss later.

standard/expressions.md
BillWagner: Added to #934

standard/patterns.md
BillWagner: "compile time" added in #934 

#938 logs the request for an example.

standard/patterns.md
BillWagner: I added the change to #934 (because it was even easier than logging the low priority issue).

I checked and removed the remaining instances of "period double-space" exist in any clause, except where the two spaces was used at the end of a line to force a line break.

standard/statements.md
BillWagner: Made the edit in #934. More efficient than logging a new issue.

standard/patterns.md
BillWagner: Added a sentence, and moved the opening brace.

standard/patterns.md
BillWagner: Logged in #939

standard/statements.md
BillWagner: Added to #934

standard/statements.md
BillWagner: added to #934

standard/statements.md
BillWagner: trivial, so I made the edit.

standard/expressions.md
BillWagner: Added to #935

standard/patterns.md
BillWagner: Edited. Check for any additional occurrence of "input expression" where "pattern input value" should be used. None found.

standard/patterns.md
BillWagner: Added in #934

standard/patterns.md
BillWagner: added in #934

standard/patterns.md
BillWagner: Addressed in #934

Added text that the runtime test must produced the same results as the is-type operator. Earlier edits should have already clarified the *pattern compatible* concept as a compile-time operation.

standard/patterns.md
BillWagner: Already addressed in earlier edits in #934

standard/patterns.md
BillWagner: Done in #934

BillWagner: Addressed in #934 

reworded to use the rules in `is type`, with the added rule that the type can't be a nullable value type.

PR #759, opened 2023-03-22
Minor tweaks to #line directive and associated attributes

RexJaeschke:
Bill, the only normative change is the addition of the implementation-defined statement w.r.t the max value permitted for a line number. Prior to this edit, that was (implicitly) unspecified; now there is a requirement for an implementation to document it, which the current implementation does.

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -808,7 +808,7 @@ One exception is query expressions. These are considered syntactic expansions, a
 If more than one caller-info attribute is specified on a given parameter, they are recognized in the following order: `CallerLineNumber`, `CallerFilePath`, `CallerMemberName`. Consider the following parameter declaration:
 
 ```csharp
-[CallerMemberName][CallerFilePath][CallerLineNumber]object p = ...
+[CallerMemberName, CallerFilePath, CallerLineNumber] object p = ...
 ```
 
 `CallerLineNumber` takes precedence, and the other two attributes are ignored. If `CallerLineNumber` were omitted, `CallerFilePath` would take precedence, and `CallerMemberName` would be ignored. The lexical ordering of these attributes is irrelevant.
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -805,11 +805,17 @@ Caller information is only substituted when a function is explicitly invoked in 
 
 One exception is query expressions. These are considered syntactic expansions, and if the calls they expand to omit optional parameters with caller-info attributes, caller information will be substituted. The location used is the location of the query clause which the call was generated from.
 
-If more than one caller-info attribute is specified on a given parameter, they are preferred in the following order: `CallerLineNumber`, `CallerFilePath`, `CallerMemberName`.
+If more than one caller-info attribute is specified on a given parameter, they are recognized in the following order: `CallerLineNumber`, `CallerFilePath`, `CallerMemberName`. Consider the following parameter declaration:
+
+```csharp
+[CallerMemberName][CallerFilePath][CallerLineNumber]object p = ...
+```
+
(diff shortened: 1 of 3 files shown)


standard/attributes.md
```diff
@@ -805,11 +805,17 @@ Caller information is only substituted when a function is explicitly invoked in
 
 One exception is query expressions. These are considered syntactic expansions, and if the calls they expand to omit optional parameters with caller-info attributes, caller information will be substituted. The location used is the location of the query clause which the call was generated from.
 
-If more than one caller-info attribute is specified on a given parameter, they are preferred in the following order: `CallerLineNumber`, `CallerFilePath`, `CallerMemberName`.
+If more than one caller-info attribute is specified on a given parameter, they are recognized in the following order: `CallerLineNumber`, `CallerFilePath`, `CallerMemberName`. Consider the following parameter declaration:
+
+```csharp
+[CallerMemberName][CallerFilePath][CallerLineNumber]object p = ...
```
jskeet: Nit: can we have a space after the final `]`? And I'd personally write this as:

```csharp
[CallerMemberName, CallerFilePath, CallerLineNumber] object p = ...
```

... but I'm happy to stick to separating them entirely if you think it's clearer.

standard/portability-issues.md
```diff
@@ -24,20 +24,21 @@ The behavior is undefined in the following circumstances:
...
-1. The mappings between pointers and integers ([§22.5.1](unsafe-code.md#2251-general)).
-1. The effect of applying the unary `*` operator to a `null` pointer ([§22.6.2](unsafe-code.md#2262-pointer-indirection)).
-1. The behavior when pointer arithmetic overflows the domain of the pointer type ([§22.6.6](unsafe-code.md#2266-pointer-increment-and-decrement), [§22.6.7](unsafe-code.md#2267-pointer-arithmetic)).
-1. The result of the `sizeof` operator for non-pre-defined value types ([§22.6.9](unsafe-code.md#2269-the-sizeof-operator)).
-1. The behavior of the `fixed` statement if the array expression is `null` or if the array has zero elements ([§22.7](unsafe-code.md#227-the-fixed-statement)).
-1. The behavior of the `fixed` statement if the string expression is `null` ([§22.7](unsafe-code.md#227-the-fixed-statement)).
-1. The value returned when a stack allocation of size zero is made ([§22.9](unsafe-code.md#229-stack-allocation)).
+2. The maximum value allowed for `Decimal_Digit+` in `PP_Line_Indicator` ([§6.5.8](lexical-structure.md#658-line-directives)).
```
jskeet: Was the whole of this file change just adding this line, and then updating the numeric values? I'd personally be tempted to keep them all at 1 so that we can add new values anywhere without any other changes. Markdown just sorts it out.

BillWagner: I'd agree with Jon here. Over time, diffs are a lot easier to review if we let markdown sort it out.

BillWagner (approved):
This also looks good to me. I agree with Jon that we should let MD manage the numbers in lists.

standard/attributes.md
RexJaeschke: OK; will fix both of these.

standard/portability-issues.md
RexJaeschke: Ok, will change.

jskeet (approved):

RexJaeschke:
FYI, Re the Imp-Def list numbering, all items in the other 3 numbered list in that annex had number 1, but for some unknown reason, this list's items were numbered 1, 2, 3, ..., so when I added a new 2nd item, the web-based editor simply gave it the number 2! C'est la vie!

PR #761, opened 2023-03-28
limit runs to the dotnet org

BillWagner:
This action fails in forks, so don't run it in forks.


BillWagner:
@RexJaeschke This will fix the GitHub action failures on your fork that you emailed me about.

jskeet (approved):

PR #762, opened 2023-04-04
Update lexical-structure.md

johndward01:
Fixed the grammar for the first sentence of the **Pre-processing directives**  section

### DIFF
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -1004,7 +1004,7 @@ right_shift_assignment
 
 ### 6.5.1 General
 
-The pre-processing directives provide the ability to skip conditionally sections of compilation units, to report error and warning conditions, and to delineate distinct regions of source code.
+Pre-processing directives provide the ability to skip conditional sections of compilation units, report error and warning conditions, and delineate distinct regions of source code.
 
 > *Note*: The term “pre-processing directives” is used only for consistency with the C and C++ programming languages. In C#, there is no separate pre-processing step; pre-processing directives are processed as part of the lexical analysis phase. *end note*
 


johndward01:
@dotnet-policy-service agree

standard/lexical-structure.md
```diff
@@ -1004,7 +1004,7 @@ right_shift_assignment
 
 ### 6.5.1 General
 
-The pre-processing directives provide the ability to skip conditionally sections of compilation units, to report error and warning conditions, and to delineate distinct regions of source code.
+Pre-processing directives provide the ability to skip conditional sections of compilation units, report error and warning conditions, and delineate distinct regions of source code.
```
Nigel-Ecma, line 1007: I suspect this is two words in the wrong order rather than a misspelling:

```suggestion
The pre-processing directives provide the ability to conditionally skip sections of compilation units, to report error and warning conditions, and to delineate distinct regions of source code.
```

which I think make more sense and is what the text used to be – I don’t know offhand when the two words were swapped around.

As to dropping the The/to/to I'll leave that to a grammarian, though the text has been that way for a long time.

jskeet: That was my initial reaction when looking at this on a mobile earlier. I personally prefer "to conditionally skip" over either of the other options, despite being a split infinitive.

I'm fine with dropping "The" at the start of the sentence though.

Nigel-Ecma: FYI: Checked back and the two words were swapped around between v2 & v3.

johndward01: Now that I read it aloud, the suggestion made by Nigel-Ecma sounds better.

RexJaeschke (approved):

PR #764, opened 2023-04-05
Fix long lines in all but one case

jskeet:
The following line in lexical_structure.md (and then grammar.md) is 81 characters long. It looks okay in the Word doc and would be hard to fix without a line break that reduces readability:

```text
    | '.' Decimal_Digit Decorated_Decimal_Digit* Exponent_Part? Real_Type_Suffix?
```

I'm tempted to leave that and possibly hard-code warning suppression for that line.

(We *could* up the line limit from 80 to 81, but I'd prefer not to do that.)


RexJaeschke (approved):

jskeet:
Thanks @RexJaeschke - in this particular case I'd like @BillWagner to have a look too before I merge.

BillWagner (approved):
This also LGTM @jskeet @RexJaeschke 

Let's :shipit:

PR #781, opened 2023-04-20
Update expressions.md

hyryel:
add missing back-quote around a variable


hyryel:
@dotnet-policy-service agree

Nigel-Ecma (approved):
Thanks for spotting this.

PR #784, opened 2023-05-02
Fix typo

pkulikov:
By the way, what does this sentence mean? Is it somehow related to the `dynamic` type or "dynamic" is used with another meaning here?

### DIFF
--- a/standard/statements.md
+++ b/standard/statements.md
@@ -533,7 +533,7 @@ Local function bodies are always reachable. The endpoint of a local function dec
 >
 > In other words, the location of a local function declaration doesn’t affect the reachability of any statements in the containing function. *end example*
 
-If the argument to a local function is dynamic, the function to be called must be resolved at compile time, not runtime.
+If the argument to a local function is dynamic, the function to be called must be resolved at compile time, not run time.
 
 ## 13.7 Expression statements
 


KalleOlaviNiemitalo:
Yes I think it means the `dynamic` type.

```csharp
using System;

public class B {
    #if false
    protected static void Mem(int i) {
        Console.WriteLine("B.Mem(int)");
    }
    #endif
        
    protected static void Mem(long l) {
        Console.WriteLine("B.Mem(long)");
    }
}

public class D : B {
    static readonly dynamic d = 1;
    
    static void Main() {
        // Resolves to the Loc(int) local function
        // at compile time.  Does an implicit conversion
        // from dynamic to int and calls that.
        Loc(d);
       
        // Does tbe same as above.
        int i = d;
        Loc(i);
        
        // Resolves to some accessible Mem at run time.
        // If class B is defined in a different assembly
        // and a new version is installed that adds Mem(int),
        // then this starts choosing that instead of
        // Mem(long), even if Mem(int) did not exist at
        // compile time.
        Mem(d);
        
        void Loc(int i) {}
    }
}
```

Nigel-Ecma (changes requested):
Grammatically “runtime”, “run-time” and ”run time” are interchangeable and consistency of choice is suggested.

The Standard primarily uses “run-time”, “runtime” is the second choice (I would have expected these two to be the other way around), and “run time” never seems to appear.

On that basis this PR should be closed as “no change required”. Another PR could be raised to address the lack of consistency.

[Tagged "Request changes" as Github seems to lack a “Close” option.]

Nigel-Ecma:
@pkulikov – regarding “dynamic” @KalleOlaviNiemitalo is correct, it means dynamic type. As a non American-English speaker I might have added a few words if I was writing it, but to me its very American-English ;-) and not worth changing.

Nigel-Ecma:
Ah, I see I should have just commented and not reviewed, hitting the Close button…

Nigel-Ecma:
…but it hasn't closed, I've confused Github again :-)

jskeet:
Actually closing :)

pkulikov:
@jskeet @Nigel-Ecma why is this PR closed? Usually, opposite of "compile time" is "run time" (with space) not "runtime" (without space). Or spec really means CLR by "runtime"?

I'm sorry if my in-PR question mislead you.

[Update: I've noticed the motivation later.]

pkulikov:
> Grammatically “runtime”, “run-time” and ”run time” are interchangeable and consistency of choice is suggested.

I've got used that in .NET docs "runtime" refers to CLR. That is different from "run time", which is time when the code is executed (by runtime). So, I wouldn't say these two are interchangeable. And as many readers of Standard come from .NET docs (I assume), the language spec may be misleading and confusing to them (one example is me).

However, I see that the style guide has changed since some time ago and uses one word to mean different things:
https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/r/runtime

That's unfortunate.

pkulikov:
@KalleOlaviNiemitalo thank you for the example

PR #789, opened 2023-05-09
Constant expressions: mention local constants

### LINKED ISSUE #786: What is a `const` parameter and when a constant expression can reference a local variable?
The [Constant expressions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions#1223-constant-expressions) section has the following entry:
image

There are two questions to the highlighted list entry:

- What is "`const` parameter"?
- When can a constant expression reference a *local variable*?

For example, the following code results in a compiler error:

```csharp
const int a = 5;
const int b = a + 4;
int c = 4;
const int d = c + 1;  // Error CS0133 The expression being assigned to 'd' must be constant
```

Also, the list doesn't contain the item with references to local constants.

pkulikov:
Fixes #786 
@Nigel-Ecma please review

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -6283,11 +6283,11 @@ Only the following constructs are permitted in constant expressions:
 - Literals (including the `null` literal).
 - References to `const` members of class and struct types.
 - References to members of enumeration types.
-- References to `const` parameters or local variables
+- References to local constants.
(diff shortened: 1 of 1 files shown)


Nigel-Ecma (approved):
Thanks @pkulikov, this has been hiding in plain sight for too long!

PR #794, opened 2023-05-18
Update V8 admin status

RexJaeschke:


BillWagner (approved):
This LGTM @RexJaeschke 

Should we update the title of this PR? You've updated the V7 status, but the PR says V8

PR #796, opened 2023-05-20
Fix local variable usage terms in variables.md

siriak:
I think it was meant to be `impossible` there instead of `possible` because otherwise there's a conflict between the note and the following statement: `A variable shall be *definitely assigned* ([§9.4](variables.md#94-definite-assignment)) before its value can be obtained."`
Also, there's the following example which uses variable `a` in its own initializer, which should be possible according to the note.
```
int a = a + 2;
```

### DIFF
--- a/standard/variables.md
+++ b/standard/variables.md
@@ -135,7 +135,7 @@ The lifetime of a local variable is the portion of program execution during whic
 
 A local variable introduced by a *local_variable_declaration* or *declaration_expression* is not automatically initialized and thus has no default value. Such a local variable is considered initially unassigned.
 
-> *Note*: A *local_variable_declaration* that includes a *local_variable_initializer* is still initially unassigned. Execution of the declaration behaves exactly like an assignment to the variable ([§9.4.4.5](variables.md#9445-declaration-statements)). Using a variable before its *local_variable_initializer* has been executed; e.g., within the initializer expression itself or by using a *goto_statement* which bypasses the initializer; is a complie-time error:
+> *Note*: A *local_variable_declaration* that includes a *local_variable_initializer* is still initially unassigned. Execution of the declaration behaves exactly like an assignment to the variable ([§9.4.4.5](variables.md#9445-declaration-statements)). Using a variable before its *local_variable_initializer* has been executed; e.g., within the initializer expression itself or by using a *goto_statement* which bypasses the initializer; is a compile-time error:
 >
 > <!-- Example: {template:"code-in-main-without-using", name:"LocalVariables", expectedErrors:["CS0165"], expectedWarnings:["CS0162"]} -->
 > ```csharp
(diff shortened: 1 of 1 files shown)


BillWagner:
I believe the original text is correct, but could use a bit of wordsmithing.

What do others think of my suggestion?  @Nigel-Ecma @jskeet

standard/variables.md
```diff
@@ -135,7 +135,7 @@ The lifetime of a local variable is the portion of program execution during whic
 
 A local variable introduced by a *local_variable_declaration* or *declaration_expression* is not automatically initialized and thus has no default value. Such a local variable is considered initially unassigned.
 
-> *Note*: A *local_variable_declaration* that includes a *local_variable_initializer* is still initially unassigned. Execution of the declaration behaves exactly like an assignment to the variable ([§9.4.4.5](variables.md#9445-declaration-statements)). It is possible to use a variable without executing its *local_variable_initializer*; e.g., within the initializer expression itself or by using a *goto_statement* to bypass the initialization:
+> *Note*: A *local_variable_declaration* that includes a *local_variable_initializer* is still initially unassigned. Execution of the declaration behaves exactly like an assignment to the variable ([§9.4.4.5](variables.md#9445-declaration-statements)). It is impossible to use a variable without executing its *local_variable_initializer*; e.g., within the initializer expression itself or by using a *goto_statement* to bypass the initialization:
```
BillWagner: I think this captures the intent of the note better:

```suggestion
> *Note*: A *local_variable_declaration* that includes a *local_variable_initializer* is still initially unassigned. Execution of the declaration behaves exactly like an assignment to the variable ([§9.4.4.5](variables.md#9445-declaration-statements)). It is possible to use a variable without executing its *local_variable_initializer*; e.g., within the initializer expression itself or by using a *goto_statement* to bypass the initialization, however that emits a compiler error:
```

BillWagner:
closing and reopening for a fresh build

standard/variables.md
Nigel-Ecma: Rather than using *(im)possible* maybe just a definitive statement?

```suggestion
> *Note*: A *local_variable_declaration* that includes a *local_variable_initializer* is still initially unassigned. Execution of the declaration behaves exactly like an assignment to the variable ([§9.4.4.5](variables.md#9445-declaration-statements)). Using a variable before its *local_variable_initializer* has been executed; e.g., within the initializer expression itself or by using a *goto_statement* which bypasses the initializer; is a complie-time error:
```

Nigel-Ecma (commented):
A review as a comment wasn't allowed…

siriak:
Applied the suggested changes

BillWagner (approved):
Thanks @siriak 

I'll merge this as soon as the build finishes.

(I'm assuming that @Nigel-Ecma is happy with his own suggestion 😄 )

BillWagner:
The build issues are addressed in #818 

Once that merges, I'll merge this one.

BillWagner:
closing and reopening for a fresh CI build

Issue #768, opened 2023-04-06
Remaining work for discards, tuples and out variables


MadsTorgersen:
I'm using this issue to evacuate remaining unresolved conversations and pending improvements from PR #664, so that it can get merged without losing track of them.

- [ ] _More examples_: Several parts of the PR are short on examples, e.g.:
  - [ ] Deconstruction expressions (https://github.com/dotnet/csharpstandard/pull/664/files#r1145443903)
  - [ ] Deconstructing assignments
- [x] Annotation of examples for testing
- [x] _Tuple encoding_(https://github.com/dotnet/csharpstandard/pull/664/files#r1145298345, https://github.com/dotnet/csharpstandard/pull/664/files#r1145285224):
    > The compiler usually constructs the underlying `ValueTuple<>` chain. If the program constructs an n-tuple directly using `ValueTuple<>` then is the code potentially non-portable between implementations which declare a different number of `ValueTuple<>` sizes? Or does the compiler accept only flat `new ValueTuple…`?
    > This says an _n_-tuple in different implementations could be represented by a chain of _q_ _m_-tuples and a final _p_-tuple where _n_ = _q_ × (_m_-1) + _p_ and _m_ is implementation dependent number. Do we need to say anything about how one implementation handles tuple values created by another with a different _m_?
- [x] _Declaration spaces_ (https://github.com/dotnet/csharpstandard/pull/664/files#r1112356059):
    > Additional contexts that might not be covered by this yet:
    > 
    > * Pattern variables declared in a switch (case) label.  This one is subtle because it introduces a scope that encompasses just the switch section, nested within the switch block, but local variables declared in the statements of the switch section go in the scope of the switch block.
    > * A switch case guard (`when` clause).  [These go in just the switch block's scope](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGZwBsmALOgLIAUA8gEYBWApgMYAu6rAlIcQJJiI9AGcA7gEtO7ABbpaffANGj2AQzGt0cdBLmsAdjVoB7AK7cAbhoBO6AB68Qqte8Z3WGgNYBuN3d0TW10GH1DEzpLG3snF0Cg4k9vf0SiAF83LOFiKDJ0RjMzABtTGPQpI25nfCd0AF5dP0wAdnROOwtWFpycoA) (see previous bullet).
[...]

PR #798, opened 2023-06-02
State that the set of ValueTuple types provided is implementation-defined

jskeet:
Part of #768.

@Nigel-Ecma I'm not sure this fully addresses your concern of "Do we need to say anything about how one implementation handles tuple values created by another with a different m?" from #664, but I couldn't think of a simple way of stating that. I'm hoping you'll agree that saying that the set of types available is implementation-defined implies that you shouldn't expect to be able to use code compiled for one implementation on another implementation, but other suggestions are welcome :)

### DIFF
--- a/standard/types.md
+++ b/standard/types.md
@@ -389,7 +389,7 @@ An enumeration type is a distinct type with named constants. Every enumeration t
 
 ### 8.3.11 Tuple types
 
-A tuple type represents an ordered, fixed-length sequence of values with optional names and individual types. The number of elements in a tuple type is referred to as its ***arity***. A tuple type is written `(T1 I1, ..., Tn In)` with n ≥ 2, where the identifiers `I1...In` are optional ***tuple element names***. This syntax is shorthand for a type constructed with the types `T1...Tn` from `System.ValueTuple<...>`, which shall be a set of generic struct types capable of expressing tuple types of any arity greater than one.
+A tuple type represents an ordered, fixed-length sequence of values with optional names and individual types. The number of elements in a tuple type is referred to as its ***arity***. A tuple type is written `(T1 I1, ..., Tn In)` with n ≥ 2, where the identifiers `I1...In` are optional ***tuple element names***. This syntax is shorthand for a type constructed with the types `T1...Tn` from `System.ValueTuple<...>`, which shall be a set of generic struct types capable of expressing tuple types of any arity greater than one. The precise set of `System.ValueTuple<...>` types provided is implementation-defined.
 
 > *Note*: There does not need to exist a `System.ValueTuple<...>` declaration that directly matches the arity of any tuple type with a corresponding number of type parameters. Instead, larger tuples can be represented with a special overload `System.ValueTuple<..., TRest>` that in addition to tuple elements has a `Rest` field containing a nested tuple value of the remaining elements. Such nesting may be observable in various ways, e.g. via the presence of a `Rest` field. *end note*
 
--- a/standard/portability-issues.md
+++ b/standard/portability-issues.md
@@ -39,6 +39,7 @@ A conforming implementation is required to document its choice of behavior in ea
 1. The behavior of the `fixed` statement if the array expression is `null` or if the array has zero elements ([§23.7](unsafe-code.md#237-the-fixed-statement)).
 1. The behavior of the `fixed` statement if the string expression is `null` ([§23.7](unsafe-code.md#237-the-fixed-statement)).
 1. The value returned when a stack allocation of size zero is made ([§23.9](unsafe-code.md#239-stack-allocation)).
+1. The set of `System.ValueTuple<...>` types available for tuple types ([8.3.11](types.md#8311-tuple-types))
 
 ## B.4 Unspecified behavior
 


standard/types.md
```diff
@@ -389,7 +389,7 @@ An enumeration type is a distinct type with named constants. Every enumeration t
...
+A tuple type represents an ordered, fixed-length sequence of values with optional names and individual types. The number of elements in a tuple type is referred to as its ***arity***. A tuple type is written `(T1 I1, ..., Tn In)` with n ≥ 2, where the identifiers `I1...In` are optional ***tuple element names***. This syntax is shorthand for a type constructed with the types `T1...Tn` from `System.ValueTuple<...>`, which shall be a set of generic struct types capable of expressing tuple types of any arity greater than one. The precise set of `System.ValueTuple<...>` types provided is implementation-defined.
 
 > *Note*: There does not need to exist a `System.ValueTuple<...>` declaration that directly matches the arity of any tuple type with a corresponding number of type parameters. Instead, larger tuples can be represented with a special overload `System.ValueTuple<..., TRest>` that in addition to tuple elements has a `Rest` field containing a nested tuple value of the remaining elements. Such nesting may be observable in various ways, e.g. via the presence of a `Rest` field. *end note*
```
KalleOlaviNiemitalo, line 394: Reading this made me worry about possible conflicts in overloaded methods, if one method takes an (int, int, int, int, int, int, int, char, char) parameter and another overload takes (int, int, int, int, int, int, int, (char, char)) -- does the compiler translate both of those to ValueTuple\<int, int, int, int, int, int, int, ValueTuple\<char, char\>\>? But no, the latter becomes ValueTuple\<int, int, int, int, int, int, int, ValueTuple\<ValueTuple\<char, char\>\>\> instead.  Here, TRest is a ValueTuple\<T1\> with only one type argument, and does not correspond to any C# tuple type.  The wording _"a nested tuple value"_ is a bit misleading, then, as it could be taken to mean an instance of a tuple type.

Perhaps _"shall be a set of generic struct types capable of expressing tuple types of any arity greater than one_" should also be changed to require ValueTuple\<T1\>, at least in those implementations that translate large-arity tuple types to System.ValueTuple\<..., TRest>.  If the implementation does generate a dedicated generic type for tuple types of each arity, instead of using TRest, then I suppose it need not support ValueTuple\<T1\>, because:

* The non-generic `struct ValueTuple` and its `public static ValueTuple<T1> Create\<T1>(T1 item1)` method is not part of Annex C Standard library.
* `struct ValueTuple<T1>` is part of Annex C Standard library, but its omission can be excused because _"The precise set of `System.ValueTuple<...>` types provided is implementation-defined."_

Nigel-Ecma: @KalleOlaviNiemitalo – I see your point but rather than trying to improve the wording I would suggest a different tack.

The C# Standard does not specify a “wire format” for interoperability of binary data, or how arrays should be stored, and a myriad of other things. Why is it specifying how a tuple should be stored? Why even mention `ValueTuple`? Are these not implementation details of the kind the Standard doesn’t usually cover?

Rather than add a sentence to the end of the paragraph, and the following note, why not delete ”This syntax is shorthand for … arity greater than one” instead?

Nigel-Ecma (changes requested):
If the translation from language-level tuples to structures built from `ValueTuple` is to be included, as implementation-defined or not, then the issue raise by @KalleOlaviNiemitalo needs to be addressed.

Now while I might like the idea in my response of classing the whole translation as an implementation issue, and not discussing it past something like “*the underlying storage structure for tuples, like for arrays, is an implementation issue*”, I expect that is not going to happen. That leaves addressing the 1 “tuples” to address, somehow…

jskeet:
Actions from meeting 2023-07-12:

- @Nigel-Ecma to write a suggested paragraph that removes all reference to `System.ValueTuple`
- @jskeet to write a suggested paragraph that addresses the womple concern

jskeet:
Alternative suggestion from @gafter: we take the standard in the other direction, specifying exactly what set of types we have.

jskeet:
... and now we're considering whether womples are actually tuple types despite not having syntax.

New proposal:

- Specify the complete set of `ValueTuple<...>` types
- Specify that womples exist, but cannot be written using tuple type or tuple literal syntax (do this bit properly!)

Thing to check: whether `==` is handled "appropriately" (in the compiler) by `ValueTuple<T>` and `ValueTuple`. If it is, we should describe these as tuple types.

jskeet:
Nope, it looks like `ValueTuple` and `ValueTuple<T>` are *not* tuple types in terms of `==`:

```csharp
ValueTuple<int> x = new ValueTuple<int>(5);
ValueTuple<int> y = new ValueTuple<int>(5);
// Operator '==' cannot be applied to operands of type 'ValueTuple<int>' and 'ValueTuple<int>'
Console.WriteLine(x == y);

ValueTuple v1 = new ValueTuple();
ValueTuple v2 = new ValueTuple();
// Operator '==' cannot be applied to operands of type 'ValueTuple' and 'ValueTuple'
Console.WriteLine(v1 == v2);
```

(We are somewhat surprised by that.)

jskeet:
@jskeet to try to create a new PR that specifies all `System.ValueTuple` types, and rewrites the representation to only go one way (to System.ValueTuple without coming back out in terms of "tuples types" for the continuation part).

(This will be in a separate PR.)

jskeet:
Closing in favor of #876.

PR #811, opened 2023-06-12
Correct the used term

pkulikov:
The whole section is about and uses the term "ref-safe-context". Is that note exception (that is, "safe context" is something different) or this PR does the right thing?


BillWagner (approved):
Thanks @pkulikov 

This is correct, and I'll :shipit: now.

PR #815, opened 2023-06-16
Add the missing word

pkulikov:
Is that correct change, given other two examples:
image


BillWagner (approved):
Thanks @pkulikov 

Yes, this is correct. One instance of the previous iteration of terms snuck through. Great catch.

I'll :shipit: now.

Issue #806, opened 2023-06-07
Verify ref safety rules


BillWagner:
See https://github.com/dotnet/csharpstandard/pull/795/files#r1218623949

We believe we've managed to capture all the original rules correctly. @gafter made this [gist](https://gist.github.com/gafter/d6068f35deb2a558b08b478f9dba1b1f) which translates the speclet terms to those used in the standard.

Changes can be found in #795 

The edits made incorporating those rules are as follows:

1. Re-order local and parameter rules.
2. Change title of "method invocation" to "function invocation" because the same rules would apply to local function declarations.
3. "Rvalue" changes to "Value" for consistency with the standard's existing terms.
4. "Span constructor" and "default" rules move to structs.md, in the section of "safe context" for `ref struct` types. 

I did a quick scan, and none of the rules were removed, or had other semantic edits.

BillWagner:
Ping @jskeet 

I captured this discussion so I could resolve those conversations and merge the PR.

jskeet:
Thanks - I'll try get to this tomorrow or Friday.

jskeet:
Okay, actually trying to read through this today.  I'll keep notes here. I hadn't looked at the gist before - I expected a set of "things to check" rather than a separate draft spec, but I'll do what I can.

- 12.21.3: "It is a compile time error if the ref-safe-context" should be followed by a link; it's in code font.
- Gist: "We associate with each expression at compile-time the concept of what scope that expression is permitted to escape to"safe-context"" - in 16.4.12.1, this is " each expression whose type is a ref struct" rather than each expression. I think that's okay.
- 16.4.12.7 refers to "current-method"; should that be function-member?
- 16.4.12 general: I think it would be useful to define the three contexts in bold rather than in plain text
- The gist only has two safe-contexts: method and caller; the standard has 3 (declaration-block, function-member, and caller-context) - do we actually need all three?
- Checking the different contexts:
  - Parameters: fine
  - Locals: fine
  - Field references: fine
  - Operators: ref-safe-context looks okay; safe-context needs more work. (We talk about `?:` and `= ref`, but nothing else.)
  - Method invocation:
    - ref-safe-context: "The context of all argument expressions (including the receiver)" should probably refer to the safe-context of all argument expressions
    - safe-context: "calling-method" should be caller-context (I suspect)
[...]

PR #816, opened 2023-06-19
Minor changes to address issues found in #806

jskeet:
These were found comparing @gafter's gist with the current content. There are a couple of outstanding items which we can address separately.

### DIFF
--- a/standard/variables.md
+++ b/standard/variables.md
@@ -1217,7 +1217,7 @@ A value’s ref-safe-context is the nearest enclosing context.
 
 #### 9.7.2.7 Constructor invocations
 
-A `new` expression that invokes a constructor obeys the same rules as a method invocation ([§9.7.2.5](variables.md#9725-method-and-property-invocation)) that is considered to return the type being constructed.
+A `new` expression that invokes a constructor obeys the same rules as a method invocation ([§9.7.2.5](variables.md#9725-function-invocation)) that is considered to return the type being constructed.
 
 #### 9.7.2.8 Limitations on reference variables
 
--- a/standard/variables.md
+++ b/standard/variables.md
@@ -1184,7 +1184,7 @@ For a variable `c` resulting from a ref-returning function invocation, its ref-s
 - The caller-context.
 - The ref-safe-context of all `ref` and `out` argument expressions (excluding the receiver).
 - For each `in` parameter of the method, if there is a corresponding expression that is a variable, its ref-safe-context, otherwise the nearest enclosing context.
-- The safe-context ([§16.4.12](structs.md#16412-safe-context-constraint-for-ref-struct-types)) of all argument expressions (including the receiver).
+- The safe-context ([§16.4.12](structs.md#16412-safe-context-constraint)) of all argument expressions (including the receiver).
 
 > *Example*: the last bullet is necessary to handle code such as
 >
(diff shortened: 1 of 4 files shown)


standard/variables.md
```diff
@@ -1177,14 +1177,14 @@ For a variable designating a reference to a field, `e.F`:
 
 The conditional operator ([§12.18](expressions.md#1218-conditional-operator)), `c ? ref e1 : ref e2`, and reference assignment operator, `= ref e` ([§12.21.1](expressions.md#12211-general)) have reference variables as operands and yield a reference variable. For those operators, the ref-safe-context of the result is the narrowest context among the ref-safe-contexts of all `ref` operands.
 
-#### 9.7.2.5 Function invocation
+#### 9.7.2.5 Method and property invocation
```
BillWagner: "Method" might not be right here. I chose "Function" because these rules also apply when invoking a local function.

I was thinking that "function" would include methods, local functions, and property / indexer accessors. 

Thoughts on the correct term?

jskeet: @gafter is probably best placed to talk about this, as method was in the gist in #806.

Function might be okay, so long as we explicitly clarify what it means in this context. (Many years ago I tried to formalize what "function" meant, but ran out of steam.)

Nigel-Ecma: In this context it is the boundary crossing from a caller/invoker to a callee/invokee which is key, not what is being called/invoked *per se*. Generally the catch-all term for something that is called/invoked has become “function” over time – be they subroutines, procedures, lambdas, properties, methods *et al*; and I see no reason the Standard shouldn’t follow this so the original “Function invocation” would be the one to use here.

jskeet: I've reverted the title change, and added "local function" into line 1182.

jskeet (commented):
@BillWagner It would be good to get the uncontroversial bits of this in before next week's meeting - PTAL; I'm happy to remove any changes that we think still need more work.

standard/variables.md
```diff
@@ -1179,12 +1179,12 @@ The conditional operator ([§12.18](expressions.md#1218-conditional-operator)),
 
 #### 9.7.2.5 Function invocation
 
-For a variable `c` resulting from a ref-returning function invocation, `ref e1.M(e2, ...)`, its ref-safe-context is the narrowest of the following contexts:
+For a variable `c` resulting from a ref-returning method or local function `ref e1.M(e2, ...)`, or property invocation `ref e.P`, its ref-safe-context is the narrowest of the following contexts:
```
Nigel-Ecma: The previous function wording encompassed indexers, the revised wording does not. Intentional?

jskeet: Except it didn't really in terms of the syntax given, which is *only* valid for local functions and methods.

Maybe the simplest option would be to change it to just "from a ref-returning function invocation" and *not* provide the syntax... or change it to "such as `ref e1.M(e2, ...)`"?

Nigel-Ecma: Of the two I’d (slightly) favour the first, drop the code, as I don’t think the code adds much in this context.

Nigel-Ecma (changes requested):
One issue left to resolve, I’ve express a *slight* preference but either option suggested by @jskeet is fine. Then it’s good to go/

standard/variables.md
jskeet: Done in a final commit, PTAL to check I haven't done anything stupid.

Nigel-Ecma (approved):
All good

gafter (approved):

PR #819, opened 2023-06-21
clarify the ref-safe-context for `in` parameters

### LINKED ISSUE #296: ref-safe-to-escape rule for in parameter

BillWagner:
The ref safe context for an `in` parameter is different when the `in` argument is passed by reference vs. when a copy is made.

Fixes #296

### DIFF
--- a/standard/variables.md
+++ b/standard/variables.md
@@ -1183,8 +1183,8 @@ The conditional operator ([§12.18](expressions.md#1218-conditional-operator)), 
 For a variable `c` resulting from a ref-returning function invocation, its ref-safe-context is the narrowest of the following contexts:
 
 - The caller-context.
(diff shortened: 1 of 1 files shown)


jskeet (approved):
Still getting context back, but this looks right to me.

gafter (approved):

PR #824, opened 2023-06-27
Correct example and add annotation

RexJaeschke:
The original example used `Action<Array>`, however, method `Array.Reverse` is static and takes at least 1 argument, so at Bill's suggestion,  I uses `List<int>.Reverse` instead, and updated the narrative to match that.


jskeet (approved):

BillWagner (approved):
This LGTM as well.

I'll :shipit:

Issue #825, opened 2023-06-28
C# 7.3 method_body etc. grammar does not allow `=> ref`


KalleOlaviNiemitalo:
**Describe the bug**

In C# 7.3, a method/property/accessor/indexer body <code>=> ref <var>variable_reference</var>;</code> should be allowed and mean the same as <code>{ return ref <var>variable_reference</var>; }</code>.  Roslyn allows this, but the grammar in the C# 7 draft doesn't.

This issue concerns multiple clauses, listed under "Expected behavior" below.

**Example**

The following should be allowed, but `=> ref field` does not match the syntactic grammar.

```csharp
class C {
    int field;
    ref int ShortProperty => ref field;
    ref int LongProperty { get => ref field; }
    ref int this[int index] => ref field;
    ref int Method() => ref field;
    void Outer() {
        ref int Inner() => ref field;
    }
}
```

**Expected behavior**

In §15.7.1 (Classes / Properties / General), _property_body_  should allow `=>` `ref` _variable_reference_ `;`.

In §15.7.3 (Classes / Properties / Accessors), _accessor_body_ should allow `=>` `ref` _variable_reference_ `;`.

In §15.9 (Classes / Indexers), _indexer_body_ should allow `=>` `ref` _variable_reference_ `;`.

In §15.6.1 (Classes / Methods / General), _method_body_  should allow `=>` `ref` _variable_reference_ `;`.

In §13.6.4 (Statements / Declaration statements / Local function declarations), _local_function_body_ should allow `=>` `ref` _variable_reference_ `;`.

The standard should specify what the `=>` `ref` syntax means, perhaps by reference to §13.10.5 (Statements / Jump statements / The return statement), which specifies the meaning of `return` `ref`.

**Additional context**

In §12.19.1 (Expressions / Anonymous function expressions / General), _anonymous_function_body_ already allows `ref` _variable_reference_.  However, I haven't found anything that specifies what `ref` means in this context.

In §15.10.1 (Classes / Operators / General), _operator_body_ does not allow `=>` `ref` _variable_reference_ `;` but this is OK as operators cannot return by reference; _operator_declarator_ does not allow `ref` either.

In §15.11.1 (Classes / Instance constructors / General), _constructor_body_ does not allow `=>` `ref` _variable_reference_ `;` but this is OK as instance constructors cannot return values either.

[...]

Issue #826, opened 2023-06-28
C# 7.3 interface_method_declaration etc. grammar does not allow return by reference


KalleOlaviNiemitalo:
**Describe the bug**
In C# 7.3, a method, property, or indexer as a member of an interface should be able to return by (readonly) reference, but the syntactic grammar does not allow this.

§18.4.2, §18.4.3, §18.4.5.

**Example**

The following should be allowed, but `ref` does not match the syntactic grammar.

```csharp
interface I {
    ref int Method();
    ref int Property { get; }
    ref int this[int index] { get; }
}
```

**Expected behavior**

In §18.4.2, _interface_method_declaration_  should allow `ref` or `ref` `readonly` before _return_type_.

In §18.4.3, _interface_property_declaration_  should allow `ref` or `ref` `readonly` before _type_.  It is not necessary to explicitly forbid a `set` accessor on such a property, because the text already references §15.7.3, which states: "A ref-returning property shall not have a `set` accessor."

In §18.4.5, _interface_indexer_declaration_ should allow `ref` or `ref` `readonly` before _type_.  It is not necessary to explicitly forbid a `set` accessor on such an indexer, because the text already references §15.9, which states: "Aside from these differences, all rules defined in §15.7.3 and §15.7.4 apply to indexer accessors as well as to property accessors."

**Additional context**

In §18.4.4, _interface_event_declaration_  does not allow `ref`, but that is OK, because the _type_ is not the return type of any accessor of the event.

Operators cannot be members of interfaces, and they do not allow `ref` in classes and structs either.

jskeet:
@Nigel-Ecma: With your grammarian hat on, does this look like it's easy enough to fix?

Nigel-Ecma:
This will be covered by PR #837 along with issue #825

PR #837, opened 2023-07-06
Changes for ref-valued methods, properties, et al.

Nigel-Ecma:
This is a first draft of changes to address #825.

At this stage only ref-valued properties have been considered, other ref returning functions will follow later.

The grammar has been extended to be as prescriptive as reasonably possible as befits a language specification[1], this means that former statements such as it being illegal to have a `set` accessor for a ref-valued property are now covered by the grammar. However the usual verbose description of the grammar remains!

The term *ref-valued* rather than *ref-returning* is used, a term closer to the language of fields. To distingush the two kinds of properties, which do follow different rules, the corresponding and somewhat ugly term *non-ref-valued* is used (*non-ref-returning* would be just as ugly).

Please fire off brickbats, and maybe even bouquets, on the direction of this. TIA 

[1] The grammar used by a compiler may be less restrictive to support more descriptive error messages, that is of course an implementation issue!


standard/classes.md
```diff
@@ -3099,27 +3100,39 @@ property_body
...
+    : '{' ref_get_accessor_declaration '}'
+    | '=>' 'ref' variable_reference ';'
+    ;
 ```
 
 *unsafe_modifier* ([§23.2](unsafe-code.md#232-unsafe-contexts)) is only available in unsafe code ([§23](unsafe-code.md#23-unsafe-code)).
 
+There are two kinds of *property_declaration*:
+
+- The first declares a non-ref-valued property. It's value has type *type*. This kind of property may be readable and/or writeable.
```
KalleOlaviNiemitalo: "Its"

standard/classes.md
```diff
@@ -3099,27 +3100,39 @@ property_body
...
+- The second declares a ref-valued property. It’s value is a *variable_reference* ([§9.4](variables.md#94-definite-assignment)), that may be `readonly`, to a variable of type *type*. This kind of property is only readable.
```
KalleOlaviNiemitalo: "Its" again. Link to §9.5 Variable references rather than 9.4.

standard/classes.md
```diff
@@ -3099,27 +3100,39 @@ property_body
...
 
 The *type* of a property shall be at least as accessible as the property itself ([§7.5.5](basic-concepts.md#755-accessibility-constraints)).
 
 A *property_body* may either consist of an ***accessor body*** or an expression body. In an accessor body,  *accessor_declarations*, which shall be enclosed in “`{`” and “`}`” tokens, declare the accessors ([§15.7.3](classes.md#1573-accessors)) of the property. The accessors specify the executable statements associated with reading and writing the property.
 
-An expression body consisting of `=>` followed by an *expression* `E` and a semicolon is exactly equivalent to the statement body `{ get { return E; } }`, and can therefore only be used to specify read-only properties where the result of the get accessor is given by a single expression.
+In a *property_body* an expression body consisting of `=>` followed by an *expression* `E` and a semicolon is exactly equivalent to the statement body `{ get { return E; } }`, and can therefore only be used to specify read-only properties where the result of the get accessor is given by a single expression.
```
KalleOlaviNiemitalo, line 3169: Does "statement body" mean the same as "accessor body", which is defined above?

standard/classes.md
```diff
@@ -3099,27 +3100,39 @@ property_body
...
 
-*Note*: Even though the syntax for accessing a property is the same as that for a field, a property is not classified as a variable. Thus, it is not possible to pass a property as an `in`, `out`, or `ref` argument unless the property itself returns a reference ([§9.7](variables.md#97-reference-variables-and-returns)). *end note*
+A *ref_property_body* may either consist of an accessor body or an expression body. In an accessor body a  *get_accessor_declaration* declares the get accessor ([§15.7.3](classes.md#1573-accessors)) of the property. The accessor specifies the executable statements associated with reading the property.
+
+In a *ref_property_body* an expression body consisting of `=>` followed by `ref`, a *variable_reference* `V` and a semicolon is exactly equivalent to the statement body `{ get { return ref V; } }`.
+
+*Note*: Even though the syntax for accessing a property is the same as that for a field, a property is not classified as a variable. Thus, it is not possible to pass a property as an `in`, `out`, or `ref` argument unless the property is ref-valued and therefore returns a reference ([§9.7](variables.md#97-reference-variables-and-returns)). *end note*
```
KalleOlaviNiemitalo: This appears to be correct although I am a bit surprised that the special case for get-only auto property in constructor does not apply.

```csharp

class C {
    C() {
        P = 0;  // OK, assign to backing field
        P++; // OK
        F(out P); // Error CS0206
        ref int r = ref P; // Error CS0206
    }
    void F(out int i) { i = 0; }
    int P { get; }
}
```

standard/classes.md
```diff
@@ -3182,12 +3207,21 @@ The use of *accessor_modifier*s is governed by the following restrictions:
...
 
-A get accessor corresponds to a parameterless method with a return value of the property type. Except as the target of an assignment, when a property is referenced in an expression, the get accessor of the property is invoked to compute the value of the property ([§12.2.2](expressions.md#1222-values-of-expressions)). The body of a get accessor shall conform to the rules for value-returning methods described in [§15.6.11](classes.md#15611-method-body). In particular, all `return` statements in the body of a get accessor shall specify an expression that is implicitly convertible to the property type. Furthermore, the endpoint of a get accessor shall not be reachable.
+For `abstract` and `extern` ref-valued properties the *ref_accessor_body* is simply a semicolon. For the accessor of any other non-abstract, non-extern property, the *ref_accessor_body* is either:
+
+- a *block* that specifies the statements to be executed when the get accessor is invoked; or
+- an expression body, which consists of `=>` followed by `ref`, a *variable_reference* and a semicolon. The variable reference is evaluated when the get accessor is invoked.
```
KalleOlaviNiemitalo, line 3262: OK, this disallows:

```csharp
ref struct C {
    ref int R { get; } // error CS8145
}
```

but allows:

```csharp
ref struct C {
    ref int r;
    ref int R { get => ref r; }
}
```

standard/classes.md
```diff
@@ -3099,27 +3100,39 @@ property_body
...
 
 A *property_initializer* may only be given for an automatically implemented property ([§15.7.4](classes.md#1574-automatically-implemented-properties)), and causes the initialization of the underlying field of such properties with the value given by the *expression*.
 
-*Note*: Even though the syntax for accessing a property is the same as that for a field, a property is not classified as a variable. Thus, it is not possible to pass a property as an `in`, `out`, or `ref` argument unless the property itself returns a reference ([§9.7](variables.md#97-reference-variables-and-returns)). *end note*
+A *ref_property_body* may either consist of an accessor body or an expression body. In an accessor body a  *get_accessor_declaration* declares the get accessor ([§15.7.3](classes.md#1573-accessors)) of the property. The accessor specifies the executable statements associated with reading the property.
```
KalleOlaviNiemitalo: Unnecessary double space before *get_accessor_declaration*

standard/classes.md
```diff
@@ -3099,27 +3100,39 @@ property_body
...
+
+In a *ref_property_body* an expression body consisting of `=>` followed by `ref`, a *variable_reference* `V` and a semicolon is exactly equivalent to the statement body `{ get { return ref V; } }`.
```
KalleOlaviNiemitalo, line 3175: Again "statement body" vs "accessor body"

standard/classes.md
```diff
@@ -3182,12 +3207,21 @@ The use of *accessor_modifier*s is governed by the following restrictions:
...
+For `abstract` and `extern` ref-valued properties the *ref_accessor_body* is simply a semicolon. For the accessor of any other non-abstract, non-extern property, the *ref_accessor_body* is either:
+
+- a *block* that specifies the statements to be executed when the get accessor is invoked; or
+- an expression body, which consists of `=>` followed by `ref`, a *variable_reference* and a semicolon. The variable reference is evaluated when the get accessor is invoked.
+
+A get accessor corresponds to a parameterless method with a return value of the property type. Except as the target of an assignment, when a property is referenced in an expression, the get accessor of the property is invoked to compute the value of the property ([§12.2.2](expressions.md#1222-values-of-expressions)).
```
KalleOlaviNiemitalo: If ref-valued, the get accessor is invoked also as the target of an assignment.

KalleOlaviNiemitalo (changes requested):

standard/classes.md
Nigel-Ecma: Typo fixed, thanks.

standard/classes.md
Nigel-Ecma: Typo fixed, thanks.

standard/classes.md
Nigel-Ecma: Typo fixed, thanks.

standard/classes.md
Nigel-Ecma: This will confuse this PR a little but as you’ve mentioned it…

“accessor body” was introduced in v6 and used in both the specification of properties and indexers to refer to the statement/function based alternative of *property_body*/*indexer_body*; the term “expression body” being used for the `=>` alternative.

However as you note some uses of “statement body” occur rather and “accessor body”

Further in the text the term “accessor body” continued to be used in prose contexts to mean the grammar rule *accessor_body*…

The next version of this PR will use “statement body” and “*accessor_body*” to clean this up. This introduces a couple of changes to interfaces.md which wouldn’t have otherwise been impacted by this PR.

standard/classes.md
Nigel-Ecma: Addressed as per previous comment

standard/classes.md
Nigel-Ecma: Never cease to be surprised, in any walk of life, lose that and you come to greyness of acceptance ;-(

I’ll leave this conversation open so others can comment if they wish, but don’t expect this to change for v7.

standard/classes.md
Nigel-Ecma: See previous comment

standard/classes.md
Nigel-Ecma: Thanks, this case is now covered in the next version.

standard/classes.md
```diff
@@ -3182,12 +3209,23 @@ The use of *accessor_modifier*s is governed by the following restrictions:
...
+
+A get accessor for a non-ref-valued property corresponds to a parameterless method with a return value of the property type. Except as the target of an assignment, when such a property is referenced in an expression its get accessor is invoked to compute the value of the property ([§12.2.2](expressions.md#1222-values-of-expressions)).
+
+The body of a get accessor for a non-ref-valued property shall conform to the rules for value-returning methods described in [§15.6.11](classes.md#15611-method-body). In particular, all `return` statements in the body of a get accessor shall specify an expression that is implicitly convertible to the property type. Furthermore, the endpoint of a get accessor shall not be reachable.
+
+A get accessor for a ref-valued property corresponds to a parameterless method with a return value of a *variable_reference* to a variable of the property type. When such a property is referenced in an expression its get accessor is invoked to compute the *variable_reference* value of the property. That *variable reference*, like any other, is then used to read or, for non-readonly properties, write the referenced variable as required by the context.
```
KalleOlaviNiemitalo: "non-readonly properties" clashes a bit with this earlier text:

> - The second declares a ref-valued property. Its value is a *variable_reference* ([§9.4](variables.md#94-definite-assignment)), that may be `readonly`, to a variable of type *type*. This kind of property is only readable.

Earlier, the *variable_reference* could be readonly, not the ref-valued property.

standard/classes.md
```diff
@@ -3099,27 +3100,39 @@ property_body
...
 
 A *property_initializer* may only be given for an automatically implemented property ([§15.7.4](classes.md#1574-automatically-implemented-properties)), and causes the initialization of the underlying field of such properties with the value given by the *expression*.
 
-*Note*: Even though the syntax for accessing a property is the same as that for a field, a property is not classified as a variable. Thus, it is not possible to pass a property as an `in`, `out`, or `ref` argument unless the property itself returns a reference ([§9.7](variables.md#97-reference-variables-and-returns)). *end note*
+A *ref_property_body* may either consist of an statement body or an expression body. In an statement body a *get_accessor_declaration* declares the get accessor ([§15.7.3](classes.md#1573-accessors)) of the property. The accessor specifies the executable statements associated with reading the property.
```
KalleOlaviNiemitalo: <del>an</del> <ins>a</ins> statement body. This occurs in a few more places.

standard/classes.md
```diff
@@ -3099,27 +3100,39 @@ property_body
...
+    | '=>' 'ref' variable_reference ';'
+    ;
 ```
 
 *unsafe_modifier* ([§23.2](unsafe-code.md#232-unsafe-contexts)) is only available in unsafe code ([§23](unsafe-code.md#23-unsafe-code)).
 
+There are two kinds of *property_declaration*:
+
+- The first declares a non-ref-valued property. Its value has type *type*. This kind of property may be readable and/or writeable.
+- The second declares a ref-valued property. Its value is a *variable_reference* ([§9.4](variables.md#94-definite-assignment)), that may be `readonly`, to a variable of type *type*. This kind of property is only readable.
```
KalleOlaviNiemitalo: Wondering about array covariance – is the variable reference necessarily to a variable of type *type*?

```csharp
class C1 {
    static C2[] a2 = { new C2() };
#if false
    ref readonly C1 DirectR {
        get {
            return ref a2[0]; // error CS8151
        }
    }
    ref readonly C1 ConversionR {
        get {
            ref readonly C2 r2 = ref a2[0];
            ref readonly C1 r1 = ref r2; // error CS8151
            return ref r1;
        }
    }
#endif
    ref readonly C1 CovariantR {
        get {
            C1[] a1 = a2;
            return ref a1[0];
        }
    }
    
    static void Main() {
        C1 c1 = new C1();
        ref readonly C1 r = ref c1.CovariantR;
        System.Console.WriteLine(r);
    }
}

class C2 : C1 {}
```

This runs OK on .NET Framework 4.5.  However, if I remove `readonly` from `ref readonly C1 CovariantR`, then System.ArrayTypeMismatchException is thrown.

KalleOlaviNiemitalo (changes requested):

standard/classes.md
KalleOlaviNiemitalo: Planning to file a separate issue for ArrayTypeMismatchException.

KalleOlaviNiemitalo: Filed that as <https://github.com/dotnet/csharpstandard/issues/838>.

standard/classes.md
Nigel-Ecma: That was a bit sloppy, mea culpa. Fixed.

standard/classes.md
Nigel-Ecma: Good catch, thanks. All ref-valued properties are readonly, but the returned *variable_reference* may not be. Now says “non-readonly *variable_reference*s”.

standard/classes.md
Nigel-Ecma: Here be dragons!

I think this is consistent with other parts of the Standard and any change needs to ensure consistency. The definition as given is at *safe*, even if it could be less restrictive, so the issue can be punted till the next version.

But let’s see what others think…

KalleOlaviNiemitalo: So, would you say that the value of the `ref readonly C1 CovariantR` property is a `readonly` *variable_reference* to a variable of type C1?  Even though at run time, it refers to an element of an array whose type is C2[].

KalleOlaviNiemitalo: > The definition as given is at safe, even if it could be less restrictive, so the issue can be punted till the next version.

If "the issue" means <https://github.com/dotnet/csharpstandard/issues/838> — well yes, it is safe in the sense that the draft won't let a value of the wrong type to be written via a *variable_reference* to a covariant array element; but the .NET implementation throws the exception already when the ref is created, rather than later if/when the write is attempted.  Then the question is whether the draft allows such an early throw.

standard/classes.md
```diff
@@ -1966,31 +1988,47 @@ A declaration has a valid combination of modifiers if all of the following are t
 - If the declaration includes the `abstract` modifier, then the declaration does not include any of the following modifiers: `static`, `virtual`, `sealed`, or `extern`.
 - If the declaration includes the `private` modifier, then the declaration does not include any of the following modifiers: `virtual`, `override`, or `abstract`.
 - If the declaration includes the `sealed` modifier, then the declaration also includes the `override` modifier.
-- If the declaration includes the `partial` modifier, then it does not include any of the following modifiers: new, `public`, `protected`, `internal`, `private`, `virtual`, `sealed`, `override`, `abstract`, or `extern`.
+- If the declaration includes the `partial` modifier, then it does not include any of the following modifiers: `new`, `public`, `protected`, `internal`, `private`, `virtual`, `sealed`, `override`, `abstract`, or `extern`.
+
+Methods are classified according to what, if anything, they return:
+
+- If `ref` is present, the method is ***returns-by-ref*** and returns a *variable reference*, that is optionally read-only;
```
jskeet, line 2000: "and returns value" => "and returns a value"?

(It's a shame that "returns-no-value" isn't equivalent to "a bare return statement is okay" - due to async, where a Task-returning method is returns-by-value from the outside, but doesn't need a value to return on the inside.)

standard/classes.md
```diff
@@ -1966,31 +1988,47 @@ A declaration has a valid combination of modifiers if all of the following are t
...
+The *ref_return_type* of a returns-by-ref method declaration specifies the type of the variable referenced by the *variable_reference* returned by the method.
 
-The *return_type* of a method declaration specifies the type of the result, if any, returned by the method. A returns-no-value method does not return a value. A returns-by-ref method returns a *variable_reference* ([§9.5](variables.md#95-variable-references)), that is optionally read-only. A returns-by-value method returns a value. If the declaration includes the `partial` modifier, then *return_type* shall be `void` ([§15.6.9](classes.md#1569-partial-methods)). If the declaration includes the `async` modifier then *return_type* shall be `void` or the method returns-by-value and the return type is a *task type* ([§15.15.1](classes.md#15151-general)).
+A generic method is a method whose declaration includes a *type_parameter_list*. This specifies the type parameters for the method. The optional *type_parameter_constraints_clause*s specify the constraints for the type parameters.
```
jskeet, line 2008: Were there any semantic changes here that I should be checking, or was it just breaking a big paragraph into small ones?

standard/classes.md
```diff
@@ -1966,31 +1988,47 @@ A declaration has a valid combination of modifiers if all of the following are t
...
 
 For abstract and extern methods, the *method_body* consists simply of a semicolon. For partial methods the *method_body* may consist of either a semicolon, a block body or an expression body. For all other methods, the *method_body* is either a block body or an expression body.
 
 If the *method_body* consists of a semicolon, the declaration shall not include the `async` modifier.
 
+The *ref_method_body* of a returns-by-ref method is either a semicolon, a ***block body*** or an ***expression body***. A block body consists of a *block*, which specifies the statements to execute when the method is invoked. An expression body consists of `=>`, followed by `ref`, a *variable_reference*, and a semicolon, and denotes a single *variable_reference* to evlaute when the method is invoked.
```
```suggestion
The *ref_method_body* of a returns-by-ref method is either a semicolon, a ***block body*** or an ***expression body***. A block body consists of a *block*, which specifies the statements to execute when the method is invoked. An expression body consists of `=>`, followed by `ref`, a *variable_reference*, and a semicolon, and denotes a single *variable_reference* to evaluate when the method is invoked.
```

standard/classes.md
```diff
@@ -3099,27 +3138,39 @@ property_body
...
 
 The *type* of a property shall be at least as accessible as the property itself ([§7.5.5](basic-concepts.md#755-accessibility-constraints)).
 
-A *property_body* may either consist of an ***accessor body*** or an expression body. In an accessor body,  *accessor_declarations*, which shall be enclosed in “`{`” and “`}`” tokens, declare the accessors ([§15.7.3](classes.md#1573-accessors)) of the property. The accessors specify the executable statements associated with reading and writing the property.
+A *property_body* may either consist of a ***statement body*** or an ***expression body***. In a statement body,  *accessor_declarations*, which shall be enclosed in “`{`” and “`}`” tokens, declare the accessors ([§15.7.3](classes.md#1573-accessors)) of the property. The accessors specify the executable statements associated with reading and writing the property.
```
jskeet, line 3167: Can we make this "block body" instead of "statement body"? (That's what we did for methods a while ago.)

standard/classes.md
```diff
@@ -3099,27 +3138,39 @@ property_body
...
 
-*Note*: Even though the syntax for accessing a property is the same as that for a field, a property is not classified as a variable. Thus, it is not possible to pass a property as an `in`, `out`, or `ref` argument unless the property itself returns a reference ([§9.7](variables.md#97-reference-variables-and-returns)). *end note*
+A *ref_property_body* may either consist of a statement body or an expression body. In a statement body a *get_accessor_declaration* declares the get accessor ([§15.7.3](classes.md#1573-accessors)) of the property. The accessor specifies the executable statements associated with reading the property.
+
+In a *ref_property_body* an expression body consisting of `=>` followed by `ref`, a *variable_reference* `V` and a semicolon is exactly equivalent to the statement body `{ get { return ref V; } }`.
+
+> *Note*: Even though the syntax for accessing a property is the same as that for a field, a property is not classified as a variable. Thus, it is not possible to pass a property as an `in`, `out`, or `ref` argument unless the property is ref-valued and therefore returns a reference ([§9.7](variables.md#97-reference-variables-and-returns)). *end note*
```
jskeet: "Therefore returns a reference" sounds wrong to me - a string-valued non-ref property returns a reference, for example. Returns a reference variable?

standard/classes.md
```diff
@@ -3182,12 +3247,23 @@ The use of *accessor_modifier*s is governed by the following restrictions:
...
+
+A get accessor for a non-ref-valued property corresponds to a parameterless method with a return value of the property type. Except as the target of an assignment, when such a property is referenced in an expression its get accessor is invoked to compute the value of the property ([§12.2.2](expressions.md#1222-values-of-expressions)).
+
+The body of a get accessor for a non-ref-valued property shall conform to the rules for value-returning methods described in [§15.6.11](classes.md#15611-method-body). In particular, all `return` statements in the body of a get accessor shall specify an expression that is implicitly convertible to the property type. Furthermore, the endpoint of a get accessor shall not be reachable.
+
+A get accessor for a ref-valued property corresponds to a parameterless method with a return value of a *variable_reference* to a variable of the property type. When such a property is referenced in an expression its get accessor is invoked to compute the *variable_reference* value of the property. That *variable reference*, like any other, is then used to read or, for non-readonly *variable_reference*s, write the referenced variable as required by the context.
```
jskeet, line 3268: Would it be worth having an example of a ref-valued property appearing on the LHS of an assignment operator, to highlight that that still invokes the getter?

I'm not suggesting the code below *is* the example, but it demonstrates what I mean.

```csharp
class Program
{
    static int field;
    static ref int Property => ref field;

    static void Main()
    {
        field = 10;
        Console.WriteLine(Property); // Prints 10
        Property = 20; // This invokes the getter, then assigns to the resulting variable reference
        Console.WriteLine(field); // Prints 20
    }
}
```

jskeet (commented):
A few comments, but otherwise looks fine to me - which isn't to say I've verified it's complete; only that the changes here do make sense to me!

standard/classes.md
Nigel-Ecma: Fixed in next commit

standard/classes.md
Nigel-Ecma: Well some semantic statements are not required as they're now enforced by the grammar  – e.g. no need to say you can't have type constraints without type parameters. – and those grammar changes were introduced to cover refs better.

standard/classes.md
Nigel-Ecma: Fixed in next commit

standard/classes.md
Nigel-Ecma: Addressed but using “variable reference” rather and “reference variable” – which mean slightly different things! ;-)

standard/classes.md
Nigel-Ecma: @jskeet – The example is fine apart from the comment “then assigns to” – “then assigns using” would work (i.e. it doesn't assign *to* as the is `= ref` style assignment, but via/using).

Shall I add it?

jskeet: See what others think, but yes, I think we might as well...

Nigel-Ecma: @jskeet – It's in, or I'll forget as I'm about to add more of the missing bits to the PR, I used “via” in the end.

Leaving the conversation unresolved, if people prefer to remove it, or would like to change the wording, just comment.

standard/classes.md
KalleOlaviNiemitalo: How about "accessor list" or "accessor list body" — or would "list" imply comma-delimited?

Nigel-Ecma:
*Comment for last commit – somehow I managed to kill `vi`, rather than `git` as usual, my corpse collection expands! This means I have to remember what I typed…*

This is the last set of changes to address #825 & #826 and covers local functions, delegate and interfaces.

The grammar phrase `'ref' 'readonly'?` was getting used a lot and the prose around it sometimes tortuous, so the grammar rule `ref_kind : 'ref' | 'ref' 'readonly' ;` (longhand to make the two alternatives clear) has been introduced and the prose simplified by using it.

The grammars for local functions, delegates & interfaces have also been updated to follow those for methods *et al.* to keep things consistent.

Quite a few typos were found during the editing and these have been fixed, but don’t panic a similar number have probably been added so typo hunting is still required! ;-)

Please fire off brickbats & bouquets as usual, but further changes prior to the next meeting are unlikely.

Now everything is in theory done the “draft” status of this PR will be removed.

standard/classes.md
Nigel-Ecma: Marking as “resolved” here now #838 exists.

standard/classes.md
Nigel-Ecma: I think *block body* is used when the body is the grammar rule *block*, the *statement body* uses are for non-*block*s. While *statement body* is also somewhat of a misnomer it was presumably picked to contrast with *expression body* so I think is at least reasonable. While editing I just followed the pattern, though I may have cleaned up inconsistencies.

@jskeet – I suggest this is left for now, you add an issue to address it post v7, and you mark this converation as resolved.

standard/delegates.md
```diff
@@ -12,26 +12,33 @@ A *delegate_declaration* is a *type_declaration* ([§14.7](namespaces.md#147-typ
...
+    | unsafe_modifier   // unsafe code support
+    ;
+
+ref_delegate_modifier
     : 'new'
     | 'public'
     | 'protected'
     | 'internal'
     | 'private'
-    | unsafe_modifier   // unsafe code support
```
KalleOlaviNiemitalo, line 25: This should be allowed:

```csharp
unsafe delegate ref int D(int* p);
```

standard/classes.md
```diff
@@ -1926,15 +1945,17 @@ method_modifier
     | 'override'
     | 'abstract'
     | 'extern'
-    | 'async'
-    | unsafe_modifier   // unsafe code support
```
KalleOlaviNiemitalo, line 1930: This should be allowed:

```csharp
class C {
    unsafe ref int M(int* p) => ref *p;
}
```

standard/classes.md
```diff
@@ -1926,15 +1945,17 @@ method_modifier
     | 'override'
     | 'abstract'
     | 'extern'
-    | 'async'
```
KalleOlaviNiemitalo, line 1929: I wondered if `async` could work with `return ref` given a custom task builder type with SetResult(ref TResult), but apparently not.

```csharp
using System;

class C {
     int i;
     async RefTask<int> M() {
         return ref i; // error CS8149
     }
}

[AsyncMethodBuilder(typeof(RefTaskBuilder<>))]
struct RefTask<TResult> {}

struct RefTaskBuilder<TResult> {
    public RefTask<TResult> Task {
        get => new RefTask<TResult>();
    }
    public static RefTaskBuilder<TResult> Create() {
        return new RefTaskBuilder<TResult>();
    }
    public void SetResult(
        ref TResult result) // regardless of ref here
    {}
    public void SetException(Exception exception) {}
    public void Start<TStateMachine>(ref TStateMachine stateMachine)
        where TStateMachine : IAsyncStateMachine
    {}
    public void AwaitOnCompleted<TAwaiter,TStateMachine>(
        ref TAwaiter awaiter,
        ref TStateMachine stateMachine)
        where TAwaiter : INotifyCompletion
        where TStateMachine : IAsyncStateMachine
    {}
    public void AwaitUnsafeOnCompleted<TAwaiter,TStateMachine>(
        ref TAwaiter awaiter,
        ref TStateMachine stateMachine)
        where TAwaiter : INotifyCompletion
        where TStateMachine : IAsyncStateMachine
    {}
    public void SetStateMachine(
        IAsyncStateMachine stateMachine)
    {}
}
```

KalleOlaviNiemitalo (changes requested):

standard/classes.md
jskeet: Filed #877 - resolving this conversation now.

jskeet:
@Nigel-Ecma: Can you take another pass at the comments on this PR now? I'm happy for merge as-is with future work noted in another issue, if that simplifies things.

Nigel-Ecma:
@jskeet – Will do, hopefully later this week…

standard/delegates.md
Nigel-Ecma: Opened issue #886 to consider this while allowing this PR to proceed.

standard/classes.md
Nigel-Ecma: Opened issue #886 to consider this while allowing this PR to proceed.

Nigel-Ecma:
@jskeet – Reviewed remaining comments, pushed two to issue #886, all marked as resolved. If you're happy merge away!

standard/statements.md
```diff
@@ -446,23 +446,30 @@ A *local_function_declaration* declares a local function.
 
 ```ANTLR
 local_function_declaration
-    : local_function_header local_function_body
+    : local_function_modifier* return_type local_function_header local_function_body
+    | ref_kind ref_return_type local_function_header ref_local_function_body
```
KalleOlaviNiemitalo, line 450: Should allow `unsafe` here too, even if `async` is not allowed.  Added to <https://github.com/dotnet/csharpstandard/issues/886#issuecomment-1677046192>.

standard/classes.md
```diff
@@ -3099,27 +3142,39 @@ property_body
...
+
+ref_property_body
+    : '{' ref_get_accessor_declaration '}'
+    | '=>' 'ref' variable_reference ';'
+    ;
 ```
 
 *unsafe_modifier* ([§23.2](unsafe-code.md#232-unsafe-contexts)) is only available in unsafe code ([§23](unsafe-code.md#23-unsafe-code)).
 
+There are two kinds of *property_declaration*:
```
jskeet, line 3154: As an aside, it feels like this is really only true when considering things from a ref perspective. You could equally say "There are three kinds of property declaration: read-only, read-write or write-only" or split it by "manually-written code, or automatically implemented." I'm not suggesting a change here, and I'll resolve the comment myself, but we should potentially be on the look-out for better ways of talking about this sort of thing. (This isn't the only time this has come up.)

standard/classes.md
```diff
@@ -3445,7 +3546,9 @@ Properties can be used to delay initialization of a resource until the moment it
 
 ### 15.7.4 Automatically implemented properties
 
-An automatically implemented property (or auto-property for short), is a non-abstract, non-extern property with semicolon-only accessor bodies. Auto-properties shall have a get accessor and may optionally have a set accessor.
+*Note*: This clause applies to both properties ([§15.7](classes.md#157-properties)) and indexers ([§15.9](classes.md#159-indexers)). The clause is written in terms of properties, when reading for indexers substitute indexer/indexers for property/properties and consult the list of differences between properties and indexers given in [§15.9.2](classes.md#1592-indexer-and-property-differences). *end note*
```
jskeet: I hadn't seen this before - I don't think it's valid, as there's no such thing as an auto-implemented indexer (that I'm aware of).

I think we should remove the paragraph, unless I've missed something.

standard/classes.md
```diff
@@ -4050,24 +4160,32 @@ indexer_modifier
...
+
+ref_indexer_body
+    : '{' ref_get_accessor_declaration '}'
+    | '=>' 'ref' variable_reference ';'
+    ;
 ```
 
 *unsafe_modifier* ([§23.2](unsafe-code.md#232-unsafe-contexts)) is only available in unsafe code ([§23](unsafe-code.md#23-unsafe-code)).
 
+There are two kinds of *indexer_declarator*:
```
```suggestion
There are two kinds of *indexer_declaration*:
```

At least, I *think* that's what's meant. (While there are two types of indexer_declarator as well, that's about explicit interface implementation.)

standard/classes.md
```diff
@@ -4243,6 +4346,23 @@ When an indexer declaration includes an `extern` modifier, the indexer is said t
...
+- A set accessor of a property corresponds to a method with a single parameter named `value`, whereas a set accessor of an indexer corresponds to a method with the same formal parameter list as the indexer, plus an additional parameter named `value`.
+- It is a compile-time error for an indexer accessor to declare a local variable or local constant with the same name as an indexer parameter.
+- In an overriding property declaration, the inherited property is accessed using the syntax `base.P`, where `P` is the property name. In an overriding indexer declaration, the inherited indexer is accessed using the syntax `base[E]`, where `E` is a comma-separated list of expressions.
+- There is no concept of an “automatically implemented indexer”. It is an error to have a non-abstract, non-external indexer with semicolon accessors.
+
+Aside from these differences, all rules defined in [§15.7.3](classes.md#1573-accessors), [§15.7.4](classes.md#1574-automatically-implemented-properties), [§15.7.5](classes.md#1575-accessibility) and [§15.7.6](classes.md#1576-virtual-sealed-override-and-abstract-accessors) apply to indexer accessors as well as to property accessors.
```
jskeet: Again, I think we'll want to tweak this for automatic indexers - and probably add that to the list of differences above.

standard/statements.md
```diff
@@ -356,7 +356,7 @@ The value of a local variable is obtained in an expression using a *simple_name*
 
 The scope of a local variable declared in a *local_variable_declaration* is the block in which the declaration occurs. It is an error to refer to a local variable in a textual position that precedes the *local_variable_declarator* of the local variable. Within the scope of a local variable, it is a compile-time error to declare another local variable or constant with the same name.
 
-A local variable declaration that declares multiple variables is equivalent to multiple declarations of single variables with the same type and `ref`/`ref readonly` prefix. Furthermore, a variable initializer in a local variable declaration corresponds exactly to an assignment statement that is inserted immediately after the declaration.
+A local variable declaration that declares multiple variables is equivalent to multiple declarations of single variables with the same type and *ref_kind* prefix. Furthermore, a variable initializer in a local variable declaration corresponds exactly to an assignment statement that is inserted immediately after the declaration.
```
jskeet: Possibly remove "prefix" now? (It feels like it's obvious that it's a prefix from the grammar now. I very much approve of the introduction of *ref_kind*.)

jskeet (approved):
Just a very few comments now - sorry not to be able to merge immediately, but I *think* you'll agree they should be fixed first.

standard/classes.md
Nigel-Ecma: Fixed, thanks

standard/classes.md
Nigel-Ecma: It is just the clause list that is wrong, the restriction is the last bullet above. Fixed.

standard/classes.md
Nigel-Ecma: Thanks, the Note has just been copied'n'pasted into one two many clauses. Fixed

standard/statements.md
Nigel-Ecma: Removed

standard/statements.md
```diff
@@ -356,7 +356,7 @@ The value of a local variable is obtained in an expression using a *simple_name*
 
 The scope of a local variable declared in a *local_variable_declaration* is the block in which the declaration occurs. It is an error to refer to a local variable in a textual position that precedes the *local_variable_declarator* of the local variable. Within the scope of a local variable, it is a compile-time error to declare another local variable or constant with the same name.
 
-A local variable declaration that declares multiple variables is equivalent to multiple declarations of single variables with the same type and `ref`/`ref readonly` prefix. Furthermore, a variable initializer in a local variable declaration corresponds exactly to an assignment statement that is inserted immediately after the declaration.
+A local variable declaration that declares multiple variables is equivalent to multiple declarations of single variables with the same type and *ref_kind*. Furthermore, a variable initializer in a local variable declaration corresponds exactly to an assignment statement that is inserted immediately after the declaration.
```
KalleOlaviNiemitalo, line 359: It no longer corresponds exactly, because of ref safety.

```csharp
using System;
public class C {
    public void M() {
        Span<int> s1 = stackalloc int[1]; // OK
        
        Span<int> s2;
        s2 = stackalloc int[1]; // error CS8353
    }
}
```

Nigel-Ecma:
The raised error is:

CS8353: A result of a stackalloc expression of type 'Span<int>' cannot be used in this context because it may be exposed outside of the containing method

What logic is the compiler using to determine s2 may be exposed but not s1? Am I missing the glaringly obvious, its been a long day, or is something amiss here?

KalleOlaviNiemitalo:
Once a Span local has been initialised with stackalloc, it is apparently OK to assign other stackalloc expressions to it:

```csharp
using System;
public class C {
    public void M() {
        Span<int> s1 = stackalloc int[1]; // OK
        
        s1 = stackalloc int[1]; // OK
    }
}
```

I get the impression that the ref safety aspects of the local variable depend on whether and how it is initialised, and they then affect whether further assignments are considered safe.  But I haven't checked whether this is already specified adequately.

KalleOlaviNiemitalo:
9.7.2.1 (Local variable ref safe context):

> * If `v` is a reference variable, its ref-safe-context is the same as the ref-safe-context of its initializing expression.
> * Otherwise its ref-safe-context is the context in which it was declared.

So this is how the ref-safe-context depends on the initialiser.  However the term "reference variable" defined in 9.7.1 does not cover `Span<int> s1` as this declaration does not include the `ref` modifier.

KalleOlaviNiemitalo:
Likewise in 9.7.2.8 (Limitations on reference variables) there's:

> For a ref reassignment `ref e1 = ref e2`, the ref-safe-context of `e2` must be at least as wide a context as the ref-safe-context of `e1`.

which does not cover assignment of `Span<T>` as that is done with `=` rather than `= ref`.

jskeet (approved):

standard/statements.md
jskeet: I've filed #899 for this.

jskeet (approved):
This looks good to me; I've filed a new issue for the latest problem noted by @KalleOlaviNiemitalo, but that wasn't introduced in this PR, so I'll merge this in the interests of making progress.

jskeet:
Ah... can't merge immediately due to conflicts. @Nigel-Ecma are you happy to rebase, or would you like me to do so?

Nigel-Ecma:
@jskeet – I’ll take a look my tomorrow and if you haven’t done it in your today I’ll see if I can do it without my usual git-breaking skill…

KalleOlaviNiemitalo:
OK, the rule for how the initializer of a `Span<T>` local variable affects the ref safety of that variable is actually specified in §16.4.12.3 (Local variable safe context):

> * Otherwise if the variable’s declaration has an initializer then the variable’s safe-context is the same as the safe-context of that initializer.

That explains the CS8353 error.  It is not necessary to change §9.7.2 to cover `ref struct` variables in addition to reference variables.

It could be useful to add a note to §9.7.2.1 (Ref safe contexts / General):

> *Note:* Safe context for expressions of  `ref struct` type is defined in §16.4.12. *end note*

§16.4.12.1 (Safe context constraint / General) already has a link to §9.7.2:

> There are three different safe-context values, the same as the ref-safe-context values defined for reference variables (§9.7.2): **declaration-block**, **function-member**, and **caller-context**.

Filed this as <https://github.com/dotnet/csharpstandard/issues/901>.

jskeet:
> I’ll take a look my tomorrow and if you haven’t done it in your today I’ll see if I can do it without my usual git-breaking skill…

I've managed to fix everything, but I can't force push to your branch. I'll force push to *my* fork, create a PR from that and merge it.

Nigel-Ecma:
Rebased as PR #900, so can close this one.

PR #842, opened 2023-07-11
Minimal change to allow "in" parameters for operators

### LINKED ISSUE #835: C# 7.3 §15.10.1 operator with `in` parameter modifier
**Describe the bug**

C# 7 draft §15.10.1 (Classes / Operators / General) states:

> The parameter(s) of an operator shall have no modifiers.

However, the `in` parameter modifier should be allowed.  The C# compiler in .NET SDK 7.0.107 allows it when LangVersion is 7.3.

**Example**

This should be allowed:

```csharp
public struct S {
    public static S operator +(in S a, in S b) {
        return default(S);
    }
}
```

**Expected behavior**

The parameter(s) of an operator shall have no modifiers other than `in`.

(The `this`, `out`, `ref`, and `params` modifiers remain disallowed.)

**Additional context**

None.

jskeet:
Fixes #835.

I don't believe any grammar changes are required - the operator grammar uses `fixed_parameter` which already permits modifiers, so it's down to the operator *text* to reject the others (which it does).

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -4308,7 +4308,7 @@ For `extern` operators, the *operator_body* consists simply of a semicolon. For 
 The following rules apply to all operator declarations:
 
 - An operator declaration shall include both a `public` and a `static` modifier.
-- The parameter(s) of an operator shall have no modifiers.
+- The parameter(s) of an operator shall have no modifiers other than `in`.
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM.

Nigel-Ecma (approved):
This is a *sufficient* change, so I'll approve. However maybe a *Note* giving the purpose of using `in` – to avoid copying large values – might be a worthwhile addition?

jskeet:
I'll merge as-is, but create an issue for a note to be added post C# 7. At this point I'm trying to stick to the bare minimum for correctness in order to get C# 7 out :)

PR #846, opened 2023-07-12
Tweaks around task types

### LINKED ISSUE #844: C# 7.x 15.15.1 standard library task types
**Describe the bug**

In the C# 7 draft, 15.15.1 says

> (The Standard library types `System.Threading.Tasks.Task<T>` and types constructed from `System.Threading.Tasks.Task` are task types.)

However, there cannot be any types constructed from `System.Threading.Tasks.Task`, because it is not a generic type.

**Example**

None

**Expected behavior**

> (The Standard library type `System.Threading.Tasks.Task` and types constructed from `System.Threading.Tasks.Task<T>` are task types.)

**Additional context**

Could also mention ValueTask and ValueTask\<T\>, which are listed in C.3 (Standard Library Types not defined in ISO/IEC 23271).

C.3 is inconsistent on whether the type parameter of these is named `T` or `TResult`.  This name is not used when referencing the type in C#, but it is stored in the ECMA-335 GenericParam table and may be readable by reflection.

jskeet:
- Use `Task<TResult>` everywhere instead of `Task<T>`
- Fix typo around which type is generic and which isn't

Fixes #844.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -5205,7 +5205,7 @@ A method ([§15.6](classes.md#156-methods)) or anonymous function ([§12.19](exp
 
 It is a compile-time error for the formal parameter list of an async function to specify any `in`, `out`, or `ref` parameters, or any parameter of a `ref struct` type.
 
-The *return_type* of an async method shall be either `void` or a ***task type***. For an async method that returns a value, a task type shall be generic. For an async method that does not return a value, a task type shall not be generic. Such types are referred to in this specification as `«TaskType»<T>` and `«TaskType»`, respectively. (The Standard library types `System.Threading.Tasks.Task<T>` and types constructed from `System.Threading.Tasks.Task` are task types.)
+The *return_type* of an async method shall be either `void` or a ***task type***. For an async method that returns a value, a task type shall be generic. For an async method that does not return a value, a task type shall not be generic. Such types are referred to in this specification as `«TaskType»<T>` and `«TaskType»`, respectively. (The Standard library type `System.Threading.Tasks.Task` and types constructed from `System.Threading.Tasks.Task<TResult>` are task types.)
 A task type shall be a class or struct type that is associated with a ***task builder type*** via the attribute `System.Runtime.CompilerServices.AsyncMethodBuilderAttribute`. Such types are referred to in this specification as `«TaskBuilderType»<T>` and `«TaskBuilderType»`.
 
 An async method returning a task type is said to be ***task-returning***.
--- a/standard/standard-library.md
+++ b/standard/standard-library.md
@@ -445,7 +445,7 @@ namespace System.Threading
 
 The following types, including the members listed, must be defined in a conforming standard library. (These types might be defined in a future edition of ISO/IEC 23271.) It is expected that many of these types will have more members available than are listed.
 
-A conforming implementation may provide `Task.GetAwaiter()` and `Task<T>.GetAwaiter()` as extension methods.
+A conforming implementation may provide `Task.GetAwaiter()` and `Task<TResult>.GetAwaiter()` as extension methods.
 
 ```csharp
 namespace System


jskeet:
Note that custom task types are still expressed as `MyTask<T>` etc. I think that's okay.

standard/classes.md
```diff
@@ -5205,7 +5205,7 @@ A method ([§15.6](classes.md#156-methods)) or anonymous function ([§12.19](exp
 
 It is a compile-time error for the formal parameter list of an async function to specify any `in`, `out`, or `ref` parameters, or any parameter of a `ref struct` type.
 
-The *return_type* of an async method shall be either `void` or a ***task type***. For an async method that returns a value, a task type shall be generic. For an async method that does not return a value, a task type shall not be generic. Such types are referred to in this specification as `«TaskType»<T>` and `«TaskType»`, respectively. (The Standard library types `System.Threading.Tasks.Task<T>` and types constructed from `System.Threading.Tasks.Task` are task types.)
+The *return_type* of an async method shall be either `void` or a ***task type***. For an async method that returns a value, a task type shall be generic. For an async method that does not return a value, a task type shall not be generic. Such types are referred to in this specification as `«TaskType»<T>` and `«TaskType»`, respectively. (The Standard library types `System.Threading.Tasks.Task` and types constructed from `System.Threading.Tasks.Task<TResult>` are task types.)
```
KalleOlaviNiemitalo: The plural in "The Standard library types" feels a bit odd, but perhaps it's OK if it covers all of "`System.Threading.Tasks.Task` and types constructed from `System.Threading.Tasks.Task<TResult>`".

KalleOlaviNiemitalo: > For an async method that returns a value, a task type shall be generic.

Should this specify the same requirements as for task builder types:

> A task builder type can have at most one type parameter and cannot be nested in a generic type.

KalleOlaviNiemitalo: Not clear what "returns a value" means here.  Apparently not the same thing as in §15.6.1:

> A returns-by-value method returns a value. […] If the declaration includes the `async` modifier then *return_type* shall be `void` or the method returns-by-value and the return type is a *task type* ([§15.15.1](https://raw.githubusercontent.com/dotnet/csharpstandard/9b7d64c9ae222318dfdeee432bd3379cc714c91f/standard/classes.md#15151-general)).

i.e. if the *return_type* of an async method is a task type, then it is a returns-by-value method and thus "returns a value".

Also consider:

```csharp
using System;
using System.Threading.Tasks;

public class C {
    public async Task<int> M() {
        throw new Exception();
    }
}
```

so if the standard needs to distinguish between "an async method that returns a value" and "an async method that does not return a value", it cannot base that on what is in a `return` statement.

jskeet: I see what you mean - I think the singular would be better. Fixing.

jskeet: Hmm... I wonder whether it's okay for it to be non-generic, but nested in a generic type with a single parameter. I think this is worth considering in a separate issue.

jskeet: Again, let's keep that for a separate issue. If you're happy to raise an issue either for both this and the generic arity, that would be great - otherwise I can do so.

KalleOlaviNiemitalo: Compiler accepts:

```csharp
using System;

class C {
     async Outer<char>.CustomTask<int> M() => 1;
}

public class Outer<T1> {
    [AsyncMethodBuilder(typeof(CustomTaskBuilder<>))]
    public struct CustomTask<T2> {}
}

public struct CustomTaskBuilder<TResult> {
    public static CustomTaskBuilder<TResult> Create() {
        return new CustomTaskBuilder<TResult>();
    }
    public void Start<TStateMachine>(
        ref TStateMachine stateMachine)
        where TStateMachine : IAsyncStateMachine {}
    public void SetStateMachine(
        IAsyncStateMachine stateMachine) {}
    public void SetException(
        Exception exception) {}
    public void SetResult(TResult result) {}
    public void AwaitOnCompleted<TAwaiter,TStateMachine>(
        ref TAwaiter awaiter,
        ref TStateMachine stateMachine)
        where TAwaiter : INotifyCompletion
        where TStateMachine : IAsyncStateMachine {}
    public void AwaitUnsafeOnCompleted<TAwaiter,TStateMachine>(
        ref TAwaiter awaiter,
        ref TStateMachine stateMachine)
        where TAwaiter : INotifyCompletion
        where TStateMachine : IAsyncStateMachine {}
    public Outer<char>.CustomTask<TResult> Task => new Outer<char>.CustomTask<TResult>();
}
```

KalleOlaviNiemitalo: I'm not filing any more issues today, but can do it tomorrow.

- §C.5 misspelling TypeInializationException
  - filed <https://github.com/dotnet/csharpstandard/issues/852>
- §15.15.1 must special-case Task and Task\<TResult> because they lack AsyncMethodBuilderAttribute
  - filed <https://github.com/dotnet/csharpstandard/issues/854>
- §15.15.1 task type must not have multiple generic arguments (but may be nested in a generic type, unlike task builder)
  - first noted in <https://github.com/dotnet/csharpstandard/pull/846#discussion_r1260881867>
  - filed <https://github.com/dotnet/csharpstandard/issues/856>
- §15.15.1 mention AsyncVoidMethodBuilder, AsyncTaskMethodBuilder, AsyncTaskMethodBuilder\<TResult\>?
  - decided to leave unspecified, so that a C# implementation need not provide these types to applications and third-party compilers
- §15.15.1 "returns a value"
  - first noted in <https://github.com/dotnet/csharpstandard/pull/846#discussion_r1260859793>
  - filed <https://github.com/dotnet/csharpstandard/issues/857>
- §15.15.1 delegate or enum as task type (Roslyn accepts)
  - filed <https://github.com/dotnet/csharpstandard/issues/858>
  - filed <https://github.com/dotnet/csharpstandard/issues/859>

BillWagner (approved):
This LGTM @jskeet 

Once you're happy with all the conversations, let's :shipit:

jskeet:
I've rebased manually (as one of my other changes was to one line earlier) and I'll merge this now as "better than it was before"; when @KalleOlaviNiemitalo files new issues (and they're much appreciated btw) we can look at those separately.

standard/classes.md
jskeet: Resolving so I can merge this PR, but please do file bugs tomorrow :)

jskeet: (Again, resolving as this should be fixed in a different PR.)

PR #847, opened 2023-07-12
Fix name of async method builder attribute

### LINKED ISSUE #843: C# 7.x 15.15.1 AsyncMethodBuilderAttribute identifier
**Describe the bug**

C# 7 draft 15.15.1 says

> A task type shall be a class or struct type that is associated with a ***task builder type*** via the attribute `System.Runtime.CompilerServices.AsyncMethodBuilder`.

but the attribute type is AsyncMethodBuilderAttribute rather than AsyncMethodBuilder.

**Example**

None

**Expected behavior**

Reference the attribute type by its name.

**Additional context**

Makes it harder to find where in the standard the semantics of AsyncMethodBuilderAttribute are defined.

jskeet:
Fixes #843.


BillWagner (approved):
This LGTM @jskeet 

I'll :shipit: now.

PR #848, opened 2023-07-12
Unify used terms

pkulikov:
The [Expressions section](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/expressions) uses the terms "left operand" and "right operand". This PR removes the two occurrences of using "hand"


Nigel-Ecma (changes requested):
There are approx (commits may change the number) 73 uses of `(left|right) operand` and only 4 of `(left|right)[- ]?hand operand` (2 each in expressions.md & classes.md) so for consistency moving to the former is direction to go (and one of those uses is an echo of one already changed in this PR).

However there are 26 uses of `(left|right)[- ]?hand` overall (so 22 not followed by operand), should these be considered?

I think if you want to tackle this then to complete the job for operands the two in classes.md should be added to this PR. The other 22 can be left for latter consideration, or not!

So I'm marking this as "Request changes", but just 2 not 24!

pkulikov:
@Nigel-Ecma thank you, I've updated the two uses in *classes.md*

Nigel-Ecma (approved):
This now looks good.

Just 22 left to review… ;-)

jskeet (approved):

jskeet:
I'd be happy for this to be merged as-is, without trying to do the other 22. @BillWagner PTAL, and merge if you're happy too.

BillWagner (approved):
This LGTM as well.

I'll :shipit: now.

Issue #853, opened 2023-07-14
C# 7.x notes without "end note"


KalleOlaviNiemitalo:
**Describe the bug**

In the C# 7.x draft, the following single-line notes start with "*Note:*" but do not end with "*end note*".

§9.4.4.33 (Rules for variables in local functions):

* <https://github.com/dotnet/csharpstandard/blob/15fc576af042c775e37d829506f98e36eaac831f/standard/variables.md?plain=1#L926>

§9.7.2.6 (Ref safe contexts / Values):

* <https://github.com/dotnet/csharpstandard/blob/15fc576af042c775e37d829506f98e36eaac831f/standard/variables.md?plain=1#L1217>

§12.6.1 (Function members / General):

* <https://github.com/dotnet/csharpstandard/blob/15fc576af042c775e37d829506f98e36eaac831f/standard/expressions.md?plain=1#L425>

§12.8.1 (Primary expressions / General):

* <https://github.com/dotnet/csharpstandard/blob/15fc576af042c775e37d829506f98e36eaac831f/standard/expressions.md?plain=1#L1297>

§12.19.1 (Anonymous function expressions / General):

* <https://github.com/dotnet/csharpstandard/blob/15fc576af042c775e37d829506f98e36eaac831f/standard/expressions.md?plain=1#L4767>
* <https://github.com/dotnet/csharpstandard/blob/15fc576af042c775e37d829506f98e36eaac831f/standard/expressions.md?plain=1#L4827>
* <https://github.com/dotnet/csharpstandard/blob/15fc576af042c775e37d829506f98e36eaac831f/standard/expressions.md?plain=1#L4829>

§15.6.1 (Methods / General):

* <https://github.com/dotnet/csharpstandard/blob/15fc576af042c775e37d829506f98e36eaac831f/standard/classes.md?plain=1#L1956>

§15.6.2.1 (Method parameters / General):

* <https://github.com/dotnet/csharpstandard/blob/15fc576af042c775e37d829506f98e36eaac831f/standard/classes.md?plain=1#L2088>

§15.6.2.3 (Method parameters / Input parameters):

* <https://github.com/dotnet/csharpstandard/blob/15fc576af042c775e37d829506f98e36eaac831f/standard/classes.md?plain=1#L2102>

§15.6.2.4 (Method parameters / Reference parameters):

[...]

PR #862, opened 2023-07-14
Fix Note formatting

RexJaeschke:
@KalleOlaviNiemitalo, this addresses your Issue #853.

@jskeet. I think I broke the Word converter! At the end of 12.6.1, we have a very long "*Note*:" without an "*end note*", so I added one. However, the content of that Note is a multi-line table marked-up in HTML, which the Word converter needs to "recognize." The thing is that being inside a multi-line Note, I had to add a leading `>` to each table line, and my guess is the Word converter can't handle that.


BillWagner (approved):
These changes LGTM @RexJaeschke 

I won't merge until @jskeet returns and looks at the word converter concerns.

jskeet:
Will try to look at the Word converter issue later this week. (Sorry, I've got a mountain of docs to review at the moment - and then I'm away again at the end of the week...)

jskeet:
I've added a commit that just puts the comment *and the line before* in block quotes - that appears to fix things (at least locally!).

jskeet:
(I haven't reviewed the rest of the change, but if @BillWagner is happy, so am I...)

PR #869, opened 2023-07-22
Add support for static local functions

RexJaeschke:
Note 1:

In researching this V8 feature, I discovered that while V7 stated, “It is a compile-time error for a [non-static] local function to declare a parameter or local variable with the same name as one declared in the enclosing scope.”, based on the implementation’s behavior when compiling with LangVersion 8, that restriction was removed in V8 for non-static local functions, and also does not apply to static local functions either.

**I note this here to make sure this was the intent, as it was not so stated in the MS proposal for this feature.**

Note 2:

The spec for 13.6.4 contains a pointer to 9.4.4.33, “Rules for variables in local functions,” part of which talks about capturing variables, which, clearly, static local functions can’t do. As such, a knowledgeable reader of that referred-to reference might be expected to understand that it can’t apply to static local functions. As such, I did *not* say that explicitly in that section. 

**Reviewers might want to add an explicit note about that.**

---
[Associated WorkItem - 209387](https://dev.azure.com/msft-skilling/Content/_workitems/edit/209387)


BillWagner:
rebased on updated `draft-v8` branch on 09/25/2023.

standard/statements.md
```diff
@@ -561,9 +562,9 @@ Unless specified otherwise below, the semantics of all grammar elements is the s
...
+A *local_function_declaration* may include one `async` ([§15.15](classes.md#1515-async-functions)) modifier and one `unsafe` ([§23.1](unsafe-code.md#231-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a `«TaskType»` type ([§15.15.1](classes.md#15151-general)). The `unsafe` modifier uses the containing lexical scope. The `unsafe` modifier uses the containing lexical scope. The `async` modifier does not use the containing lexical scope. If the declaration includes the `static` modifier, the function is a ***static local function***; otherwise, it is a ***non-static local function***. It is a compile-time error for *type_parameter_list* or *formal_parameter_list* to contain *attributes*.
```
KalleOlaviNiemitalo: Doesn't need to describe `unsafe` twice.

KalleOlaviNiemitalo (changes requested):

standard/statements.md
KalleOlaviNiemitalo: The wording "uses the containing lexical scope" seems a bit vague too.  As if, when the local function has its own `unsafe` modifier, that causes the function to use something from the lexical scope.  Can this be clarified somehow… "Any `unsafe` modifier in the containing lexical scope also applies to the local function."

BillWagner: I agree that the "containing lexical scope" statements aren't as clear as they could be. (Not a C# 8 feature, but let's fix it anyway).  Here's my suggestion:

```suggestion
A *local_function_declaration* may include one `async` ([§15.15](classes.md#1515-async-functions)) modifier and one `unsafe` ([§23.1](unsafe-code.md#231-general)) modifier. If the declaration includes the `async` modifier then the return type shall be `void` or a `«TaskType»` type ([§15.15.1](classes.md#15151-general)). If the declaration includes the `static` modifier, the function is a ***static local function***; otherwise, it is a ***non-static local function***. It is a compile-time error for *type_parameter_list* or *formal_parameter_list* to contain *attributes*. If the local function is declared in an unsafe context (§23.2), the local function may include unsafe code, even if the local function declaration doesn't include the `unsafe` modifier.
```

BillWagner (changes requested):
I made a suggested change for the "enclosing scope" sections.

In the suggestion, I removed the sentence saying that `async` doesn't apply if the enclosing method is `async`. I don't think it's needed with the new wording.

BillWagner:
@RexJaeschke Should we mark this one ready for review?

standard/statements.md
```diff
@@ -604,6 +605,19 @@ Local function bodies are always reachable. The endpoint of a local function dec
 
 If the type of the argument to a local function is `dynamic`, the function to be called must be resolved at compile time, not runtime.
 
+A static local function
+
+- May not reference instance members from an implicit or explicit `this` or `base` reference.
```
```suggestion
- Shall not reference instance members from an implicit or explicit `this` or `base` reference.
```

standard/statements.md
```diff
@@ -604,6 +605,19 @@ Local function bodies are always reachable. The endpoint of a local function dec
...
+- May reference `static` members from the enclosing scope.
+- May reference `constant` definitions from the enclosing scope.
+- May use `nameof()` to reference locals, parameters, or `this` or `base` from the enclosing scope.
+- May capture state from an enclosing static local function, but may not capture state outside the enclosing static local function. The same rules apply to a lambda.
+- May not be invoked in an expression tree.
```
```suggestion
```

jskeet: Is it valid to invoke a non-static local function from an expressoin tree?

standard/statements.md
```diff
@@ -604,6 +605,19 @@ Local function bodies are always reachable. The endpoint of a local function dec
 
 If the type of the argument to a local function is `dynamic`, the function to be called must be resolved at compile time, not runtime.
 
+A static local function
+
+- May not reference instance members from an implicit or explicit `this` or `base` reference.
+- May reference `static` members from the enclosing scope.
+- May reference `constant` definitions from the enclosing scope.
+- May use `nameof()` to reference locals, parameters, or `this` or `base` from the enclosing scope.
+- May capture state from an enclosing static local function, but may not capture state outside the enclosing static local function. The same rules apply to a lambda.
```
jskeet, line 614:
```suggestion
- May capture state from an enclosing static local function, but may not capture state outside the enclosing static local function. The same rules apply to a lambda expression.
```

This final sentence feels weird though - are these rules for lambda expressions that are enumerated somewhere else? If so, let's link to them. It not, why not?

standard/statements.md
```diff
@@ -604,6 +605,19 @@ Local function bodies are always reachable. The endpoint of a local function dec
 
 If the type of the argument to a local function is `dynamic`, the function to be called must be resolved at compile time, not runtime.
 
+A static local function
+
+- May not reference instance members from an implicit or explicit `this` or `base` reference.
+- May reference `static` members from the enclosing scope.
+- May reference `constant` definitions from the enclosing scope.
```
```suggestion
- May reference constant definitions from the enclosing scope.
```

jskeet: (Maybe we should remove the backticks from around static as well?)

standard/statements.md
jskeet: Earlier we state:

> A non-static local function may capture variables from the enclosing scope while a static local function may not

Doesn't that contradict "May capture state from an enclosing static local function"?

jskeet (commented):
I do wonder whether we need a note about the purpose of making a local function static. At the moment we say that removing the modifier from valid code doesn't change the meaning - but we don't say why you'd want to keep it anyway.

standard/statements.md
```diff
@@ -561,9 +562,9 @@ Unless specified otherwise below, the semantics of all grammar elements is the s
...
 
-A local function is declared at block scope, and that function may capture variables from the enclosing scopes. It is a compile-time error if a captured variable is read by the body of the local function but is not definitely assigned before each call to the function. The compiler shall determine which variables are definitely assigned on return ([§9.4.4.33](variables.md#94433-rules-for-variables-in-local-functions)).
+A local function is declared at block scope, and that function may capture variables from the enclosing scopes. A non-static local function may capture variables from the enclosing scope while a static local function may not (so it has no access to enclosing locals, parameters, or `this`). It is a compile-time error if a captured variable is read by the body of the non-static local function but is not definitely assigned before each call to the function. The compiler shall determine which variables are definitely assigned on return ([§9.4.4.33](variables.md#94433-rules-for-variables-in-local-functions)).
```
BillWagner: I'd remove the second clause from the first sentence. It's repeated in the next sentence:

```suggestion
A local function is declared at block scope. A non-static local function may capture variables from the enclosing scope while a static local function may not (so it has no access to enclosing locals, parameters, or `this`). It is a compile-time error if a captured variable is read by the body of the non-static local function but is not definitely assigned before each call to the function. The compiler shall determine which variables are definitely assigned on return ([§9.4.4.33](variables.md#94433-rules-for-variables-in-local-functions)).
```

standard/statements.md
BillWagner: No, it's not. Local functions (static or not) aren't allowed in expression trees. We missed this in 7.3

standard/statements.md
BillWagner: This is incorrect.  A `static` local function can't capture state.

This should read that a non-static local function can capture state declared in an enclosing static local function:

```csharp
static void M()
{
    OuterM();
    static void OuterM()
    {
        var msg = "hello";
        Console.WriteLine("Hello, World!");

        InnerM();
        void InnerM() // not static
        {
            Console.WriteLine(msg);
        }
    }
}
```

BillWagner: yes.

BillWagner (changes requested):
I have a few comments to address, along with Jon's nits.

After that, this one is ready to merge. I don't think any rules on static local functions are missing.

Nigel-Ecma:
I'm not sure it's clear that a static local function can call another static local function (but not a non-static local function), and that includes a static local function defined in an enclosing local function.

@jskeet wrote:

> I do wonder whether we need a note about the purpose of making a local function static. At the moment we say that removing the modifier from valid code doesn't change the meaning - but we don't say why you'd want to keep it anyway.

A note might be worth it as using the keyword `static` to mean "limited scope non-capturing pure function" may not leap off the page ;-)

standard/statements.md
jskeet: Great - @BillWagner are you happy to propose a change for this?

BillWagner: I think we can simply remove this line:

```suggestion
```

standard/statements.md
```diff
@@ -604,6 +605,18 @@ Local function bodies are always reachable. The endpoint of a local function dec
 
 If the type of the argument to a local function is `dynamic`, the function to be called must be resolved at compile time, not runtime.
 
+A static local function
```
BillWagner: Move expression tree comment to apply to all local functions:


```suggestion
A local function shall not be invoked in an expression tree.

A static local function
```

standard/statements.md
```diff
@@ -604,6 +605,19 @@ Local function bodies are always reachable. The endpoint of a local function dec
 
 If the type of the argument to a local function is `dynamic`, the function to be called must be resolved at compile time, not runtime.
 
+A local function shall not be invoked in an expression tree.
```
```suggestion
A local function shall not be used in an expression tree.
```

standard/statements.md
```diff
@@ -561,9 +562,9 @@ Unless specified otherwise below, the semantics of all grammar elements is the s
...
 
-A local function is declared at block scope, and that function may capture variables from the enclosing scopes. It is a compile-time error if a captured variable is read by the body of the local function but is not definitely assigned before each call to the function. The compiler shall determine which variables are definitely assigned on return ([§9.4.4.33](variables.md#94433-rules-for-variables-in-local-functions)).
+A local function is declared at block scope. A non-static local function may capture variables from the enclosing scope while a static local function may not (so it has no access to enclosing locals, parameters, or `this`). It is a compile-time error if a captured variable is read by the body of the non-static local function but is not definitely assigned before each call to the function. The compiler shall determine which variables are definitely assigned on return ([§9.4.4.33](variables.md#94433-rules-for-variables-in-local-functions)).
```
```suggestion
A local function is declared at block scope. A non-static local function may capture variables from the enclosing scope while a static local function shall not (so it has no access to enclosing locals, parameters, non-static local functions, or `this`). It is a compile-time error if a captured variable is read by the body of a non-static local function but is not definitely assigned before each call to the function. The compiler shall determine which variables are definitely assigned on return ([§9.4.4.33](variables.md#94433-rules-for-variables-in-local-functions)).
```

standard/statements.md
```diff
@@ -604,6 +605,19 @@ Local function bodies are always reachable. The endpoint of a local function dec
...
 If the type of the argument to a local function is `dynamic`, the function to be called must be resolved at compile time, not runtime.
 
+A local function shall not be used in an expression tree.
+
+A static local function
+
+- Shall not reference instance members from an implicit or explicit `this` or `base` reference.
+- May reference `static` members from the enclosing scope.
+- May reference constant definitions from the enclosing scope.
+- May use `nameof()` to reference locals, parameters, or `this` or `base` from the enclosing scope.
```
```suggestion
- May use `nameof()` to reference locals, parameters, non-static local functions, `this` or `base` from the enclosing scope.
```

standard/statements.md
```diff
@@ -604,6 +605,19 @@ Local function bodies are always reachable. The endpoint of a local function dec
 
 If the type of the argument to a local function is `dynamic`, the function to be called must be resolved at compile time, not runtime.
 
+A local function shall not be used in an expression tree.
+
+A static local function
+
+- Shall not reference instance members from an implicit or explicit `this` or `base` reference.
```
```suggestion
- Shall not reference instance members from an implicit or explicit `this` or `base` reference.
- Shall not reference local variables, parameters, or non-static local functions from the enclosing scope.
```

standard/statements.md
```diff
@@ -604,6 +605,19 @@ Local function bodies are always reachable. The endpoint of a local function dec
...
+- May reference `static` members from the enclosing scope.
```
```suggestion
- May reference static members, type parameters, constant definitions and static local functions from the enclosing scope.
```

standard/statements.md
```diff
@@ -604,6 +605,19 @@ Local function bodies are always reachable. The endpoint of a local function dec
...
+- May reference constant definitions from the enclosing scope.
```
```suggestion
```

(This is now in 613.)

standard/statements.md
```diff
@@ -604,6 +605,19 @@ Local function bodies are always reachable. The endpoint of a local function dec
...
 If the type of the argument to a local function is `dynamic`, the function to be called must be resolved at compile time, not runtime.
 
+A local function shall not be used in an expression tree.
+
+A static local function
+
+- Shall not reference instance members from an implicit or explicit `this` or `base` reference.
+- Shall not reference local variables, parameters, or non-static local functions from the enclosing scope.
+- May reference static members, type parameters, constant definitions and static local functions from the enclosing scope.
+- May use `nameof()` to reference locals, parameters, non-static local functions, `this` or `base` from the enclosing scope.
```
```suggestion
- May reference static members, type parameters, constant definitions and static local functions from the enclosing scope.
- Shall not reference `this` or `base` nor instance members from an implicit `this` reference, nor local variables, parameters, or non-static local functions from the enclosing scope. However, all these are permitted in a `nameof()` expression.
```

standard/statements.md
```diff
@@ -604,6 +605,17 @@ Local function bodies are always reachable. The endpoint of a local function dec
 
 If the type of the argument to a local function is `dynamic`, the function to be called must be resolved at compile time, not runtime.
 
+A local function shall not be used in an expression tree.
+
+A static local function
+
+- May reference static members, type parameters, constant definitions and static local functions from the enclosing scope.
```
gafter, line 612:
```suggestion
- May reference static members, type parameters, constant definitions and static local functions from the enclosing scope;
```

standard/statements.md
```diff
@@ -604,6 +605,17 @@ Local function bodies are always reachable. The endpoint of a local function dec
...
+- Shall not reference `this` or `base` nor instance members from an implicit `this` reference, nor local variables, parameters, or non-static local functions from the enclosing scope. However, all these are permitted in a `nameof()` expression.
+
+Overload resolution of a call within a local function is not affected by whether the local function is static.
+
+Removing the `static` modifier from a local function in a valid program does not change the meaning of the program.
```
jskeet, line 617:
```suggestion
> *Note:* Removing the `static` modifier from a local function in a valid program does not change the meaning of the program. *end note*
```

```suggestion
```

jskeet (approved):

standard/statements.md
```diff
@@ -604,6 +605,15 @@ Local function bodies are always reachable. The endpoint of a local function dec
...
 If the type of the argument to a local function is `dynamic`, the function to be called must be resolved at compile time, not runtime.
 
+A local function shall not be used in an expression tree.
+
+A static local function
+
+- May reference static members, type parameters, constant definitions and static local functions from the enclosing scope.
+- Shall not reference `this` or `base` nor instance members from an implicit `this` reference, nor local variables, parameters, or non-static local functions from the enclosing scope. However, all these are permitted in a `nameof()` expression.
+
+Overload resolution of a call within a local function is not affected by whether the local function is static.
```
jskeet: I think this is correct:

```csharp
class Program
{
    public void Method(int x = 10)
    {
    }

    static void Method(string y = "")
    {
    }

    void M()
    {
        StaticLocal();
        NonStaticLocal();
        static void StaticLocal()
        {
            Method();
        }

        void NonStaticLocal()
        {
            Method();
        }
    }
}
```
Both calls to Method cause errors of "The call is ambiguous between the following methods or properties: 'Program.Method(int)' and 'Program.Method(string)'"

jskeet: Next attempt:

```csharp
class Program
{
    public void Method(string x)
    {
    }

    static void Method(object x)
    {
    }

    void M()
    {
        StaticLocal();
        NonStaticLocal();
        static void StaticLocal()
        {
            Method("");
        }

        void NonStaticLocal()
        {
            Method("");
        }
    }

    static void StaticM()
    {
        Method("");
    }
}
```

Here, calling `Method("")` from `StaticM` is fine, but not from `StaticLocal`.

```suggestion
```

Removing because we believe that the fact that it's currently true is a bug rather than a feature, and we may want to fix the compiler later.

jskeet: Interestingly, if you copy `StaticLocal` into `StaticM` (and call it from `StaticM`) then it's fine.

jskeet (approved):

BillWagner (approved):

MadsTorgersen (approved):

PR #871, opened 2023-07-26
Annotate Patterns Examples

RexJaeschke:


standard/patterns.md
```diff
@@ -160,13 +162,14 @@ A set of patterns `Q` is *exhaustive* for a type `T` if any of the following con
 
 > *Example*:
 >
+> <!-- Example: {template:"standalone-console-without-using", name:"PatternExhaustiveness1", replaceEllipsis:true, customEllipsisReplacements: [""], ignoredWarnings:["CS8321"]} -->
```
jskeet, line 165: Shouldn't we be expecting an error here?

jskeet: Or rather: fundamentally, the example doesn't fully show what it claims to, due to the "..."
It's not clear to me whether it's worth "testing" something that doesn't behave as documented.

BillWagner (approved):
I'm good with this.

My thinking on our testing is to validate what we can, while ensuring that the focus is still on the text of the standard, rather than creating a "testable standard".

To that end, let's leave this sample as is. We do verify there are no typos in the text, even though we're not actually testing subsumption.

jskeet (approved):

jskeet:
Right, I'm fine with that. I'll merge it now.

Issue #859, opened 2023-07-14
C# 7.x §15.15.1 interface, delegate, or enum as task type


KalleOlaviNiemitalo:
**Describe the bug**

In the C# 7.x draft, §15.15.1 (Async functions / General) says:

> A task type shall be a class or struct type that is associated with a ***task builder type*** via the attribute `System.Runtime.CompilerServices.AsyncMethodBuilderAttribute`. Such types are referred to in this specification as `«TaskBuilderType»<T>` and `«TaskBuilderType»`.

However, Roslyn allows a task type that is neither a class type nor a struct type, as long as it has the AsyncMethodBuilderAttribute.

**Example**

An interface should be allowed as a task type.

```csharp
using System;
using System.Runtime.CompilerServices;

class C {
    async ICustomTask M() {}
}

[AsyncMethodBuilder(typeof(CustomTaskBuilder))]
public interface ICustomTask {}

public struct CustomTaskBuilder {
    public static CustomTaskBuilder Create() => default;
    public void Start<TStateMachine>(
        ref TStateMachine stateMachine)
        where TStateMachine : IAsyncStateMachine {}
    public void SetStateMachine(
        IAsyncStateMachine stateMachine) {}
    public void SetException(
        Exception exception) {}
    public void SetResult() {}
    public void AwaitOnCompleted<TAwaiter,TStateMachine>(
        ref TAwaiter awaiter,
        ref TStateMachine stateMachine)
        where TAwaiter : INotifyCompletion
        where TStateMachine : IAsyncStateMachine {}
    public void AwaitUnsafeOnCompleted<TAwaiter,TStateMachine>(
        ref TAwaiter awaiter,
        ref TStateMachine stateMachine)
        where TAwaiter : INotifyCompletion
        where TStateMachine : IAsyncStateMachine {}
    public ICustomTask Task => default;
}
```

**Expected behavior**

In §15.15.1, also allow an interface, delegate, or enum type as a task type.  Say just "A task type shall be a type that" rather than "A task type shall be a class or struct type that" -- except <https://github.com/dotnet/csharpstandard/issues/854> will need more changes here.

**Additional context**

[...]

Issue #858, opened 2023-07-14
C# 7.x §C.3 AsyncMethodBuilderAttribute has AttributeTargets.Method


KalleOlaviNiemitalo:
**Describe the bug**

In the C# 7.x draft, §C.3 (Standard Library Types not defined in ISO/IEC 23271), AsyncMethodBuilderAttribute has `AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Interface | AttributeTargets.Delegate | AttributeTargets.Enum | AttributeTargets.Method`.  Of these, AttributeTargets.Method is not needed for C# 7.3 and instead corresponds to a C# 10 feature.

**Example**

This should not be allowed in a C# 7.3 implementation without extensions.

```csharp
using System.Runtime.CompilerServices;

public class C {
    [AsyncMethodBuilder(typeof(C))]
    void M() {}
}
```

**Expected behavior**

In §C.3, remove AttributeTargets.Method from AsyncMethodBuilderAttribute.

**Additional context**

The C# 7.0 [Async Task Types in C#](https://github.com/dotnet/csharplang/blob/05ed9dcfba84a038e02301a9b267516edfe14159/proposals/csharp-7.0/task-types.md) proposal does not mention AttributeTargets.

<https://github.com/dotnet/runtime/pull/50116> added AttributeTargets.Method to AsyncMethodBuilderAttribute, for the C# 10 [AsyncMethodBuilder override](https://github.com/dotnet/csharplang/blob/05ed9dcfba84a038e02301a9b267516edfe14159/proposals/csharp-10.0/async-method-builders.md) proposal <https://github.com/dotnet/csharplang/issues/1407>.

jskeet:
Closed by #875.

Issue #856, opened 2023-07-14
C# 7.x §15.15.1 task type must not have multiple type arguments


KalleOlaviNiemitalo:
**Describe the bug**

In the C# 7.x draft, §15.15.1 (Async functions / General) restricts the genericity of a *task type*:

> For an async method that returns a value, a task type shall be generic. For an async method that does not return a value, a task type shall not be generic.

This allows a constructed task type with more than one type argument, but Roslyn does not support that.

**Example**

The standard should not allow:

```csharp
using System;
using System.Runtime.CompilerServices;

class C {
    // «TaskType»<T> is CustomTask<int, int>.
    // «TaskBuilderType»<T> is not clear, but not outright disallowed either.
    async CustomTask<int, int> M() => 1; // error CS1983
}

[AsyncMethodBuilder(typeof(CustomTaskBuilder<>))]
public struct CustomTask<T1, T2> {}

public struct CustomTaskBuilder<TResult> {
    public static CustomTaskBuilder<TResult> Create() => default;
    public void Start<TStateMachine>(
        ref TStateMachine stateMachine)
        where TStateMachine : IAsyncStateMachine {}
    public void SetStateMachine(
        IAsyncStateMachine stateMachine) {}
    public void SetException(
        Exception exception) {}
    public void SetResult(TResult result) {}
    public void AwaitOnCompleted<TAwaiter,TStateMachine>(
        ref TAwaiter awaiter,
        ref TStateMachine stateMachine)
        where TAwaiter : INotifyCompletion
        where TStateMachine : IAsyncStateMachine {}
    public void AwaitUnsafeOnCompleted<TAwaiter,TStateMachine>(
        ref TAwaiter awaiter,
        ref TStateMachine stateMachine)
        where TAwaiter : INotifyCompletion
        where TStateMachine : IAsyncStateMachine {}
    public CustomTask<TResult, TResult> Task => default;
}
```

**Expected behavior**

Declare that the task type shall not have more than one type argument.

(Use "type argument" rather than "type parameter", because task types are constructed types.)

**Additional context**

Unlike a task builder type, a task type can be nested in a generic type.  Example:

```csharp
using System;
using System.Runtime.CompilerServices;

class C {
    // «TaskType»<T> is Outer<char>.CustomTask<int>.
    // «TaskBuilderType»<T> is CustomTaskBuilder<int>; the type argument in Outer<char> does not affect this.
    async Outer<char>.CustomTask<int> M() => 1;
}

[...]

PR #875, opened 2023-08-07
Clarify task types

### LINKED ISSUE #854: C# 7.x §15.15.1 must special-case Task and Task<TResult> because they lack AsyncMethodBuilderAttribute
**Describe the bug**

In the C# 7.x draft, §15.15.1 (Async functions / General) states:

> The *return_type* of an async method shall be either `void` or a ***task type***. For an async method that returns a value, a task type shall be generic. For an async method that does not return a value, a task type shall not be generic. Such types are referred to in this specification as `«TaskType»<T>` and `«TaskType»`, respectively. (The Standard library type `System.Threading.Tasks.Task` and types constructed from `System.Threading.Tasks.Task<TResult>` are task types.)
> A task type shall be a class or struct type that is associated with a ***task builder type*** via the attribute `System.Runtime.CompilerServices.AsyncMethodBuilderAttribute`. Such types are referred to in this specification as `«TaskBuilderType»<T>` and `«TaskBuilderType»`.

However, `System.Threading.Tasks.Task` and `System.Threading.Tasks.Task<TResult>` are not decorated with AsyncMethodBuilderAttribute. Not in §C.3 (Standard Library Types not defined in ISO/IEC 23271), and not in .NET Framework either.

**Example**

None.

**Expected behavior**

Special-case `System.Threading.Tasks.Task` and `System.Threading.Tasks.Task<TResult>` so that they need not be decorated with AsyncMethodBuilderAttribute.

**Additional context**

None.

jskeet:
- `Task` and `Task<T>` are classified as task types despite not specifying builder types
- Interfaces are allowed to be task types
- Enums and methods are prohibited from being decorated with AsyncMethodBuilderAttribute
- Task types must not be generic in more than one type parameter (including in terms of containing types)

Discussion required for all of this, but if merged, would fix #854, #856, #858 and #859.


jskeet:
cc @KalleOlaviNiemitalo for thoughts

@BillWagner I figure we can probably get this hammered into better shape between ourselves (I'm not entirely happy with the way the generic restriction is expressed) and then get broader review.

KalleOlaviNiemitalo:
> Task types must not be generic in more than one type parameter (including in terms of containing types)

That is not right.  Type arguments of types that contain the task type should be ignored; Outer\<char, int>.Inner\<long> should be allowed as a task type if it has the correct attribute.  But not as a task builder type.

jskeet:
> That is not right. Type arguments of types that contain the task type should be ignored; `Outer<char, int>.Inner<long>` should be allowed as a task type if it has the correct attribute. But not as a task builder type.

That would certainly be simpler to specify. Can I ask what your reasoning is here? I suspect it's more advanced than mine... and we should probably have a note to explain either way.

KalleOlaviNiemitalo:
Or is the idea that, if the task type is nested in a generic type, then the task builder type would not be able to have a Task property with the correct type, unless it hardcodes the type arguments of the containing type?  So that allowing such task types would not be useful in practice.  Which would then make the Roslyn behaviour an extension.

jskeet:
> Or is the idea that, if the task type is nested in a generic type, then the task builder type would not be able to have a Task property with the correct type, unless it hardcodes the type arguments of the containing type? So that allowing such task types would not be useful in practice. Which would then make the Roslyn behaviour an extension.

That was my thinking, I think - but I feel like I'm still wading through cotton wool trying to think about this. I'm definitely not feeling like we should be bound by exactly what Roslyn allows, but I'm open to useful justifications.

jskeet:
I've added another commit to basically copy the existing text from 15.15.2 (for task builders) and make it apply to task types.
This may be more restrictive than is strictly required (and Roslyn can allow a little more leeway), but I think it's a good starting point.

jskeet:
I need to go back and reapply changes from #846 to this...

jskeet:
Oops - no, I *don't* need to reapply changes from #846. It's all fine; @BillWagner does this look ready to merge?

BillWagner (approved):
This LGTM @jskeet 

Let's :shipit:

PR #876, opened 2023-08-07
Specify the mapping between tuple types and ValueTuple normatively.

jskeet:
Alternative to #798.


standard/types.md
```diff
@@ -389,9 +389,10 @@ An enumeration type is a distinct type with named constants. Every enumeration t
...
+This syntax is shorthand for a type constructed with the types `T1...Tn` from `System.ValueTuple<...>`, which shall be a set of generic struct types capable of directly expressing tuple types of any arity between two and seven inclusive.
+Tuple types with an arity of more than seven are constructed There does not need to exist a `System.ValueTuple<...>` declaration that directly matches the arity of any tuple type with a corresponding number of type parameters. Instead, larger tuples can be represented with a generic struct type `System.ValueTuple<T1 I1, ..., T7 I7, TRest>` that in addition to tuple elements has a `Rest` field containing a nested value of the remaining elements, using another `System.ValueTuple<...>` type. Such nesting may be observable in various ways, e.g. via the presence of a `Rest` field. Where only a single additional field is required, the generic struct type `System.ValueTuple<T1>` is used; this type is not considered a tuple type in itself. Where more than seven additional fields are required, `System.ValueTuple<T1 I1, ..., T7 I7, TRest>` is used recursively.
```
KalleOlaviNiemitalo: The first sentence of that line ends strangely.  "are constructed There does not"

KalleOlaviNiemitalo (changes requested):

standard/types.md
KalleOlaviNiemitalo: Remove the `I1` etc. from each `T1 I1` type argument of System.ValueTuple.

KalleOlaviNiemitalo (changes requested):

standard/types.md
jskeet: Done.

jskeet: Whoops, thanks. Abortive edit. Have pushed another commit to fix that and improve the wording a little more.

KalleOlaviNiemitalo:
The "shorthand" term feels a bit wrong because these types have identity conversions between them but are not the same type.  It's not the same situation as with `int` vs. `global::System.Int32`.  The distinction is explained in more detail later though so it could be ok.

Nigel-Ecma (approved):
Nice, clear, straightforward.

BillWagner (approved):
This is a clear concise explanation. Thanks @jskeet 

Let's :shipit:

jskeet:
Right, merging - we can go over it retrospectively, but I want to clear the decks as far as possible.

PR #884, opened 2023-08-09
Correction to the local function grammar

RexJaeschke:
I took this grammar from the MS proposal, which unfortunately contains an error that I only just spotted today. The parens are *required* tokens, but currently are interpreted as (redundant) ANTLR grouping parens. This correction makes the local function header match that for a method.


Nigel-Ecma (approved):
LGTM

PR #885, opened 2023-08-10
V7.3 stack allocation hack

Nigel-Ecma:
This is a replacement for PR #840 (as this bear of little brain was thwarted by Github in making the changes as comments/review on the PR, I expect there is a way to do it…)

The purpose, as identified by @RexJaeschke, is impose the C# 7.3 restrictions on `stackalloc` as the existing text applied to C# 8. However when I came to read the clause I found there was duplication of material, I'm guessing from moving stuff from the unsafe section, so I (foolishly?) took on cleaning it up, but just as much as necessary to be correct.

I have inserted markdown (HTML) comments around the bit which covers the C# 7.3, if we’re luck just deleting that in C# 8 should do – but someone should check when the time comes!

The discussion on #840 covered more concerns, such as C# 7.3 doesn't allow (pointless) casts. None of these are addressed as I don’t believe they are intended for C# 7.3, and C# 8 will allow them automatically without adding verbiage – but opinions may vary of that!

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -3091,16 +3091,25 @@ stackalloc_element_initializer
     : expression
     ;
 ```
+A *stackalloc_expression* is only permitted in two contexts:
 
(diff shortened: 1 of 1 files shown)


jskeet (commented):
This looks good to me, but it's so far out of my area of expertise that my opinion isn't worth much.

BillWagner (approved):
Thanks @Nigel-Ecma 

I like they way you've handled this.

I recommend merging this and closing the other PR.

PR #889, opened 2023-08-15
Add semantics for stackalloc used as a pointer initializer

Nigel-Ecma:
With the expansion/relocation of stackalloc to both Unsafe Code and Expressions the former was left without semantics allowing stackalloc to produce both Span<T> and T* in unsafe contexts. This PR addresses that. It is being left as “draft” for now so I can check it again in daylight before removing the draft status. Others casting their eyes over this would be appreciated – I think we know what the semantics are, it’s just describing them concisely and clearly…

### DIFF
--- a/standard/unsafe-code.md
+++ b/standard/unsafe-code.md
@@ -1049,6 +1049,10 @@ When the outermost containing struct variable of a fixed-size buffer member is a
 
 See [§12.8.21](expressions.md#12821-stack-allocation) for general information about the operator `stackalloc`. Here, the ability of that operator to result in a pointer is discussed.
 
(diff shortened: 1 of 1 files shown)


KalleOlaviNiemitalo:
About the pointer value:

- does it point to the beginning or to the end
- is it implementation-defined if `stackalloc T[0]`, like in the `Span<T>` case (it would be if it were somehow converted from the `Span<T>`)

KalleOlaviNiemitalo:
> so I can check it again in daylight before removing the draft status

🤭 for a moment, I misunderstood that the draft status somehow affects light/dark color scheme.

BillWagner:
I gave this a read, and everything looks correct to me. I'll do a formal approval once it's not in draft status.

Nigel-Ecma:
@KalleOlaviNiemitalo 

> About the pointer value:
> 
>     * does it point to the beginning or to the end

I take it as convention that memory allocators return pointers to the start/beginning/lowest byte address of allocated memory, but there is no harm in adding it – done.

>     * is it implementation-defined if `stackalloc T[0]`, like in the `Span<T>` case (it would be if it were somehow converted from the `Span<T>`)

The “In all other respects the semantics…” is there to cover this, so yes it is implementation defined.

BillWagner (approved):
This LGTM @Nigel-Ecma 

Let's :shipit:

I'll merge tomorrow, unless @jskeet finds anything that should be addressed.

jskeet (approved):
This looks fine to me, although it's an area I don't know much about (yet another one).

PR #892, opened 2023-08-15
Define SpanT

Nigel-Ecma:
Working on stackalloc turned up that Span<T> was used but what it is wasn't defined, and in the library (§C.3) the used members (indexer and the just used Length) not mentioned.

So here is a draft PR which squeezes what Span<T> is into stackalloc’s description, as it is the only place (currently) using it, and adds just the used members to Span<T>, and also ReadOnlySpan<T> as that was in §C.3 presumably as it is used in the example.

Is this sufficient? Should more members be included? Is defining it in stackalloc’s description OK? Etc.

Depending on feedback I'll remove draft status, edit, or kill.


jskeet:
This looks fine to me, thanks Nigel - I think it would be good for @BillWagner to review as well, but if he's happy with it too, I think we can merge as-is.

BillWagner:
I'm happy with this as written @Nigel-Ecma @jskeet 

When ready, let's merge it.

jskeet:
Will wait for @Nigel-Ecma to move it out of draft status, just in case he had any other tweaks to make :)

BillWagner (approved):
Thanks @Nigel-Ecma 

This LGTM.

PR #895, opened 2023-08-16
Fix samples

jskeet:
The addition of `ref` variables reduced the number of warnings of CS0219 - "The variable 'variable' is assigned but its value is never used"; the second sample was simply broken in terms of the ref readonly change.

### DIFF
--- a/standard/statements.md
+++ b/standard/statements.md
@@ -386,7 +386,7 @@ In an implicitly typed local variable declaration, the type of the local variabl
 
 > *Example*:
 >
-> <!-- Example: {template:"code-in-main", name:"LocalVariableDecls4", expectedWarnings:["CS0219","CS0219","CS0219"], additionalFiles:["Order.cs"]} -->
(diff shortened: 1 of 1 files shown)


jskeet:
(This demonstrates that we're not running the tests regularly though - we have a workflow, but it's not triggered automatically. @BillWagner do you remember why?)

BillWagner:
> (This demonstrates that we're not running the tests regularly though - we have a workflow, but it's not triggered automatically. @BillWagner do you remember why?)

IIRC, we didn't enable it because it wasn't yet ready. If you and @RexJaeschke think we've covered enough tests, I'll add an action to run it. I might need some guidance on detecting failures to block the build, but otherwise it will be straightforward.

BillWagner (approved):
This LGTM @jskeet 

Let's :shipit:

jskeet:
> If you and @RexJaeschke think we've covered enough tests, I'll add an action to run it. I might need some guidance on detecting failures to block the build, but otherwise it will be straightforward.

We've got an action already - I'm not sure why it's not triggering.

It should fail the action if any tests fail without anything else needing to be changed. (It uses an exit code of 1, IIRC.)

PR #897, opened 2023-08-16
Fix path matching for example test action

jskeet:
This now matches what we've got for markdownlint.yaml, which *is* being triggered.


BillWagner (approved):
I think that's right @jskeet 

If not, I'll track it down.

Let's :shipit:

PR #902, opened 2023-08-17
Update restrictions on parameters

### LINKED ISSUE #896: §9.7.2.9: clarity on capture of output or input parameters
**Describe the bug**

In the C# 7.x draft, §9.2.7 (Output parameters) says:

> Unlike value parameters an output parameter may not be captured (§9.7.2.9).

§9.2.8 (Input parameters) likewise says:

> Unlike value parameters an input parameter may not be captured (§9.7.2.9).

However, §9.7.2.9 (Limitations on reference variables) itself specifies restrictions that do not apply to all output or input parameters:

> * Neither a `ref` parameter, nor a `ref` local, nor a parameter or local of a `ref struct` type shall be captured by lambda expression or local function.
> * Neither a `ref` parameter nor a parameter of a `ref struct` type shall be an argument for an iterator method or an async method.

**Example**

```csharp
public class C {
    public void M(out int r) {
        r = 0;
        int Inner() => r;
    }
}
```

**Expected behavior**

In §9.7.2.9, make these restrictions on `ref` parameters also apply to output parameters and input parameters.

**Additional context**

ECMA-334 5th Edition §12.16.6.1 and ECMA-334 6th Edition §11.16.6.2 allow an anonymous function to capture a "value parameter", thus excluding `ref` and `out` parameters.  C# 5 and C# 6 do not have named local functions.

jskeet:
Fixes #896.

cc @KalleOlaviNiemitalo

I do wonder whether it would be useful to introduce a term which covers all of ref/out/in parameters, but that can come later.

### DIFF
--- a/standard/variables.md
+++ b/standard/variables.md
@@ -1224,8 +1224,8 @@ A `new` expression that invokes a constructor obeys the same rules as a method i
 
 #### 9.7.2.9 Limitations on reference variables
 
-- Neither a `ref` parameter, nor a `ref` local, nor a parameter or local of a `ref struct` type shall be captured by lambda expression or local function.
-- Neither a `ref` parameter nor a parameter of a `ref struct` type shall be an argument for an iterator method or an `async` method.
+- Neither a reference parameter, nor an output parameter, nor an input parameter, nor a `ref` local, nor a parameter or local of a `ref struct` type shall be captured by lambda expression or local function.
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @jskeet 

I'll give @Nigel-Ecma the chance to look, and I'll :shipit: tomorrow.

Nigel-Ecma:
@jskeet:

> I do wonder whether it would be useful to introduce a term which covers all of ref/out/in parameters, but that can come later.

*by-ref* parameters?

Nigel-Ecma (approved):
LGTM

PR #903, opened 2023-08-17
Add navigational note from ref-safe-context to safe-context

### LINKED ISSUE #901: §9.7.2.1: Reference §16.4.12 for clarity
In C# 7.x, it would be useful to add a note to §9.7.2.1 (Ref safe contexts / General):

> *Note:* Safe context for expressions of  `ref struct` type is defined in §16.4.12. *end note*

This would help a person reading the definition of **ref-safe-context** from §9.7.2 understand why the constraints are written as applying only to reference variables and not also to `ref struct` variables.

§16.4.12.1 (Safe context constraint / General) already has a link to §9.7.2:

> There are three different safe-context values, the same as the ref-safe-context values defined for reference variables (§9.7.2): **declaration-block**, **function-member**, and **caller-context**.

_Reformulated from <https://github.com/dotnet/csharpstandard/issues/837#issuecomment-1681946952>_

jskeet:
(And fix a bit of formatting.)

Fixes #901.

cc @KalleOlaviNiemitalo

### DIFF
--- a/standard/variables.md
+++ b/standard/variables.md
@@ -1052,6 +1052,8 @@ A ***reference return*** is the *variable_reference* returned from a returns-by-
 
 All reference variables obey safety rules that ensure the ref-safe-context of the reference variable is not greater than the ref-safe-context of its referent.
 
+> *Note*: The related notion of a *safe-context* is defined in ([§16.4.12](structs.md#16412-safe-context-constraint)), along with associated constraints. *end note*
+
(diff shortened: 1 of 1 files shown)


KalleOlaviNiemitalo:
I think I'd prefer having the note higher up, visually separating it from the examples.

jskeet:
> I think I'd prefer having the note higher up, visually separating it from the examples.

I've added a second commit which puts it pretty much at the start of the section. I'm not convinced there's a good place for it other than at the start or the end... we'll see what others think.

BillWagner (approved):
I like this @jskeet 

I like the note at the beginning. It is useful for readers to have the relationship in mind before reading the section.

Nigel-Ecma (approved):
LGTM

PR #834, opened 2023-07-02
Fix Editorial Nits

RexJaeschke:
@BillWagner, as I was working on the V8 spec for patterns, I had need to review the newly added chapter 7 (patterns), the first time I have done so since we agreed to break-off that text into its own chapter. Here are the editorial changes I made:

- Made section headings sentence-case rather than title-case.
- 11.2 contained text *prior* to its first child section, 11.2.1, which we don't allow, so I added a new section 11.2.1 General.
- I removed italics from references to terms that have already been defined.
- I added a bunch of xrefs, especially those placeholders flagged using `(XREF)`.
- Removed all occurrences of the the first-person pronoun "we" in newer text by rewriting.

PR #904, opened 2023-08-17
Add readonly modifier to standard library structs.

### LINKED ISSUE #893: Readonly structs in standard library
**Describe the bug**

Of the struct types listed in Annex C of the C# 7.x draft, only a few are specified as `readonly struct`.  In .NET however, most of them are defined as `readonly struct`.
Is this allowed, and is it an extension?

**Example**

Types that are `readonly struct` in .NET but plain `struct` in Annex C include:

- System.Int32
- System.Span\<T>
- System.Runtime.CompilerServices.TaskAwaiter\<TResult>

**Expected behavior**

Could be solved in any of these ways:

- make them `readonly struct` in the standard, too
- declare that an implementation can add `readonly` to standard library struct types
- as above, but also change every `readonly struct` in Annex C to plain `struct`

**Additional context**

Similar to <https://github.com/dotnet/csharpstandard/issues/870> on attributes.

jskeet:
The remaining non-readonly structs are:

- `Nullable<T>` (not sure why)
- `ValueTuple<...>` (as it's mutable)

Fixes #893.

@Nigel-Ecma and @RexJaeschke I'd particularly like your feedback on the note. Perhaps it shouldn't even be a note, but normative text?

### DIFF
--- a/standard/standard-library.md
+++ b/standard/standard-library.md
@@ -21,6 +21,9 @@ It is expected that a conforming C# implementation will supply a significantly 
 
 ## C.2 Standard Library Types defined in ISO/IEC 23271
 
+> *Note:* Some `struct` types below have the `readonly` modifier. This modifier was not available
+> when ISO/IEC 23271 was released, but is required for conforming implementations of this specification. *end note*
+
 ```csharp
 namespace System
 {
@@ -91,10 +94,10 @@ namespace System
         public AttributeTargets ValidOn { get; }
     }
 
-    public struct Boolean { }
-    public struct Byte { }
-    public struct Char { }
-    public struct Decimal { }
+    public readonly struct Boolean { }
+    public readonly struct Byte { }
+    public readonly struct Char { }
+    public readonly struct Decimal { }
     public abstract class Delegate { }
 
     public class DivideByZeroException : ArithmeticException
@@ -104,7 +107,7 @@ namespace System
         public DivideByZeroException(string message, Exception innerException);
     }
 
-    public struct Double { }
+    public readonly struct Double { }
 
     public abstract class Enum : ValueType
     {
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM.

standard/standard-library.md
```diff
@@ -21,6 +21,9 @@ It is expected that a conforming C# implementation will supply a significantly
 
 ## C.2 Standard Library Types defined in ISO/IEC 23271
 
+> *Note:* Some `struct` types below have the `readonly` modifier. This modifier was not available
+> when ISO/IEC 23271 was released, but is required for conforming implementations of this standard. *end note*
```
KalleOlaviNiemitalo: <https://github.com/dotnet/csharpstandard/pull/834> replaced "this standard" with "this specification" everywhere, except this one: <https://github.com/dotnet/csharpstandard/blob/87d585dbdc58f398b713f8a7a0123ee3a1d16c14/standard/expressions.md?plain=1#L1455>

Nigel-Ecma (dismissed):
Change “standard” to ”specification” as per @KalleOlaviNiemitalo (I don’t recall this change myself doing the diff for #900 correctness, it showed up a lot, I'm a “Standard” person myself)

Otherwise LGTM and I pre-approve.

jskeet:
Changed standard to specification. I'll "dismiss" Nigel's review given the pre-approval.

standard/standard-library.md
jskeet: Done.

KalleOlaviNiemitalo:
> * `Nullable<T>` (not sure why)

That was explained in <https://github.com/dotnet/corefx/pull/24997#issuecomment-346523578>: the Equals(object), GetHashCode(), and ToString() methods of Nullable\<T\> call the corresponding methods of T, and those may mutate the T instance stored within the Nullable\<T\>.  If Nullable\<T\> were changed to a readonly struct, then its methods would make defensive copies of T and the mutations would no longer affect the Nullable\<T\> instance.  That difference could break some applications.

Issue #695, opened 2023-01-05
Lock object for field-like event references "Lock statements" in definite assignment rules


KalleOlaviNiemitalo:
Normative text in [C# 6 §14.8.2 Field-like events](https://github.com/dotnet/csharpstandard/blob/V6-standard/standard/classes.md#1482-field-like-events) references §9.4.4.19 "Lock statements", which is part of [§9.4.4 "Precise rules for determining definite assignment"](https://github.com/dotnet/csharpstandard/blob/V6-standard/standard/variables.md#944-precise-rules-for-determining-definite-assignment):

> When compiling a field-like event, the compiler automatically creates storage to hold the delegate, and creates accessors for the event that add or remove event handlers to the delegate field. The addition and removal operations are thread safe, and may (but are not required to) be done while holding the lock ([§9.4.4.19](https://github.com/dotnet/csharpstandard/blob/V6-standard/standard/variables.md#94419-lock-statements)) in the containing object for an instance event, or the type `object` ([§11.7.15.7](https://github.com/dotnet/csharpstandard/blob/V6-standard/standard/expressions.md#117157-anonymous-object-creation-expressions)) for a static event.

The more appropriate reference would be [§12.13 "The lock statement"](https://github.com/dotnet/csharpstandard/blob/V6-standard/standard/statements.md#1213-the-lock-statement).

KalleOlaviNiemitalo:
This error is already in ECMA-335 5th edition, but not in 4th edition.

Issue #694, opened 2023-01-05
Lock object for static field-like event references "Anonymous object creation expressions"


KalleOlaviNiemitalo:
Normative text in [C# 6 §14.8.2 Field-like events](https://github.com/dotnet/csharpstandard/blob/V6-standard/standard/classes.md#1482-field-like-events) references §11.7.15.7 "Anonymous object creation expressions", which seems wrong:

> When compiling a field-like event, the compiler automatically creates storage to hold the delegate, and creates accessors for the event that add or remove event handlers to the delegate field. The addition and removal operations are thread safe, and may (but are not required to) be done while holding the lock ([§9.4.4.19](https://github.com/dotnet/csharpstandard/blob/V6-standard/standard/variables.md#94419-lock-statements)) in the containing object for an instance event, or the type `object` ([§11.7.15.7](https://github.com/dotnet/csharpstandard/blob/V6-standard/standard/expressions.md#117157-anonymous-object-creation-expressions)) for a static event.

Likewise, ECMA-334 5th edition (December 2017) §15.8.2 references §12.7.11.7 "Anonymous object creation expressions":

> When compiling a field-like event, the compiler automatically creates storage to hold the delegate, and creates accessors for the event that add or remove event handlers to the delegate field. The addition and removal operations are thread safe, and may (but are not required to) be done while holding the lock (§10.4.4.19) on the containing object for an instance event, or the type object (§12.7.11.7) for a static event.

In contrast, ECMA-334 4th edition (June 2006) §17.7.1 references §14.5.11 "The typeof operator":

> When compiling a field-like event, the compiler automatically creates storage to hold the delegate, and creates accessors for the event that add or remove event handlers to the delegate field. In order to be thread-safe:
>
> * The addition and removal operations on all instance events of a class shall be done while holding the lock (§15.12) on an object uniquely associated with the containing object (`this`). [_Note_: For example, the unique object may be the containing object itself, or a hidden instance field. _end note_]
> * The addition and removal operations on all static events of a class shall be done while holding the lock (§15.12) on an object uniquely associated with the containing class. [_Note_: For example, the unique object may be the type object (§14.5.11) for the containing class, or a hidden static field. _end note_]

So, C# 6 should instead say:

> the type object ([§11.7.16](https://github.com/dotnet/csharpstandard/blob/V6-standard/standard/expressions.md#11716-the-typeof-operator)) of the containing class for a static event.

[...]

PR #906, opened 2023-08-17
Update references to type objects

### LINKED ISSUE #905: Code font in "type `object`" when an instance of System.Type is meant
**Describe the bug**

In the C# 7.x draft, the following sentences are formatted with "type `object`", which gives the impression that the type System.Object is meant, but actually these mean instances of System.Type.

§12.8.17 (The typeof operator), describing `typeof(void)`:

> *Note*: This special type `object` is useful in class libraries that allow reflection onto methods in the language, where those methods wish to have a way to represent the return type of any method, including `void` methods, with an instance of `System.Type`. *end note*

§15.8.2 (Field-like events), previously noted in <https://github.com/dotnet/csharpstandard/issues/694>:

> The addition and removal operations are thread safe, and may (but are not required to) be done while holding the lock ([§9.4.4.19](variables.md#94419-lock-statements)) in the containing object for an instance event, or the type `object` (§12.8.16.7) for a static event.

**Example**

None.

**Expected behavior**

Format as "type object" with regular font, or even as "`Type` object".

**Additional context**

These came from <https://github.com/dotnet/csharpstandard/pull/403>.

jskeet:
Fixes #905.

(Also fixes a cross-link, which I believe should be to the typeof operator rather than anonymous object creation expressions...)

cc @KalleOlaviNiemitalo

We may need to do a little more - or less - in expressions.md, given that the previous paragraph does talk about "type object" - although it has `System.Type` earlier in the paragraph.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -3897,7 +3897,7 @@ Within the program text of the class or struct that contains the declaration of 
 >
 > *end example*
 
-When compiling a field-like event, the compiler automatically creates storage to hold the delegate, and creates accessors for the event that add or remove event handlers to the delegate field. The addition and removal operations are thread safe, and may (but are not required to) be done while holding the lock ([§9.4.4.19](variables.md#94419-lock-statements)) in the containing object for an instance event, or the type `object` ([§12.8.16.7](expressions.md#128167-anonymous-object-creation-expressions)) for a static event.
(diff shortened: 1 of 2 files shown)


KalleOlaviNiemitalo:
> (Also fixes a cross-link, which I believe should be to the typeof operator rather than anonymous object creation expressions...)

Fixes <https://github.com/dotnet/csharpstandard/issues/694>.

jskeet:
Nice, thanks - have added a commit to fix #695 at the same time. (And changed "holding the lock in" to "holding the lock on".)

KalleOlaviNiemitalo:
("`Type` object" would also be valid, based on §C.5 Library Type Abbreviations.)

jskeet:
> ("Type object" would also be valid, based on §C.5 Library Type Abbreviations.)

Yup, that's what I had at one point - but I wanted to be consistent with surrounding paragraphs which use `System.Type` quite a bit. We could update them all at some point, but arguably `System.Type` helps to be unambiguous (in a way that `Int32` doesn't need help with).

RexJaeschke (approved):

Nigel-Ecma (commented):
LGTM

Issue #899, opened 2023-08-17
Local variable initialization is not always equivalent to immediate assignment after declaration


jskeet:
From #837, in statements.md:

> A local variable declaration that declares multiple variables is equivalent to multiple declarations of single variables with the same type and *ref_kind*. Furthermore, a variable initializer in a local variable declaration corresponds exactly to an assignment statement that is inserted immediately after the declaration.

@KalleOlaviNiemitalo wrote:

> It no longer corresponds exactly, because of ref safety.
>
> ```csharp
> using System;
> public class C {
>     public void M() {
>         Span<int> s1 = stackalloc int[1]; // OK
>         Span<int> s2;
>         s2 = stackalloc int[1]; // error CS8353
>     }
> }

KalleOlaviNiemitalo:
@jskeet, do you intend this issue to also cover that many of the ref safety rules only apply to "reference variables" and not to variables of ref-like type?

KalleOlaviNiemitalo:
There's also this in 13.6.2 (Local variable declarations):

> However, it is a compile-time error to omit *local_variable_initializer* from a *local_variable_declarator* for a variable declared `ref` or `ref readonly`.

So you cannot transform a declaration with an initialiser, to a declaration without an initialiser, followed by an assignment.

jskeet:
I'd only expected it to cover the issue you brought up on #837. I'm not sure whether that answers your question or not, but I basically wanted to avoid that new issue from blocking the PR.

KalleOlaviNiemitalo:
@jskeet, it was a misunderstanding on my part.  Elaborated in <https://github.com/dotnet/csharpstandard/pull/837#issuecomment-1681946952>.

jskeet:
@KalleOlaviNiemitalo: Some aspects may have been a misunderstanding, but I still thing 13.6.2 needs clarification, as your example points out that the "exact correspondence" isn't the case. Or maybe *I've* misunderstood too :)

Nigel-Ecma:
The statement quoted at the start:

> A local variable declaration that declares multiple variables is equivalent to multiple declarations of single variables with the same type and *ref_kind*. Furthermore, a variable initializer in a local variable declaration corresponds exactly to an assignment statement that is inserted immediately after the declaration.

Needs to be updated to cope with:

1. `ref` & `ref readonly` locals: These must be initialised on declaration and so cannot be broken into declaration + assignment at all.
[...]

PR #909, opened 2023-08-18
Withdraw ”corresponds exactly” statement regarding initialization vs. declaration + assignment

Nigel-Ecma:
This addresses #899.

As explained there the statement “a variable initializer in a local variable declaration corresponds exactly to an assignment statement that is inserted immediately after the declaration” is not strictly true.

However describing the exceptions would at best make a simple statement rather more complicated.

This begs the question “What is the value of the statement?”

This PR answers that one way – by just deleting it.

Agree or disagree in your reviews (so yes, this is a “poll” which if passed automatically gets implemented).


jskeet (approved):
Yes, I think it's simplest to remove this.

BillWagner (approved):
Yes, this also LGTM.

PR #914, opened 2023-08-21
Clarify that only instance fields are relevant for unmanaged types

### LINKED ISSUE #891: §8.8: unmanaged type with static fields
**Describe the bug**

C# 7.x draft §8.8 (Unmanaged types) appears to forbid an unmanaged type from having static fields with a managed type.

> An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_parameter*, or a constructed type, and contains no fields whose type is not an *unmanaged_type*.

"contains no fields" might be understood as meaning both instance fields and static fields.

**Example**

Is this an unmanaged type?

```csharp
struct S {
    int i;
    static object obj;
}
```

**Expected behavior**

Static fields should not affect whether the type is an unmanaged type.

**Additional context**

Found while perusing <https://github.com/dotnet/csharpstandard/pull/888>.

jskeet:
Fixes #891

### DIFF
--- a/standard/types.md
+++ b/standard/types.md
@@ -696,8 +696,8 @@ unmanaged_type
     ;
 ```
 
-An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_parameter*, or a constructed type, and contains no fields whose type is not an *unmanaged_type*. In other words, an *unmanaged_type* is one of the following:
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @jskeet 


I'll :shipit:

PR #916, opened 2023-08-28
Update types.md

KamilStachera:
Remove wrongly placed "end example"


RexJaeschke (commented):
I disagree! This is an Example inside of a Note, and \*Example\* requires a terminating \*end example\*. The md-to-Word converter handles it correctly as it currently exists.

KamilStachera:
> 

Ah, yes, sorry, didn't notice the begging of example a little bit earlier, my bad

Issue #839, opened 2023-07-08
C# 7.x ref assignment to input parameter


KalleOlaviNiemitalo:
**Describe the bug**

In the C# 7.x draft, 12.21.3 could be a bit clearer about input parameters here:

> If the left operand is a writeable ref (i.e., it designates anything other than a `ref readonly` local or  `in` parameter), then the right operand shall be a writeable *variable_reference*. If the right operand variable is writeable, the left operand may be declared `ref` or `ref readonly`.

**Example**

This should be valid.

```csharp
class C {
    int j;
    void M(in int i) {
        i = ref j;
    }
}
```

**Expected behavior**

Say that the left operand may be declared `in` as an input parameter.

**Additional context**

Poking this, I met the ref-safe-to-escape rules

```csharp
class C {
    ref readonly int M(in int i, int j) {
        ////i = ref j; // error CS8374
        return ref i;
    }
}
```

The ref assignment here looks benign on its own, but it must be rejected because the method could then return that ref.  I have not checked whether the draft specifies an error here.

jskeet:
At the start of 12.21.3 we already have:

> The left operand shall be an expression that binds to a reference variable ([§9.7](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/variables.md#97-reference-variables-and-returns)), a reference parameter (other than this), an output parameter, or an input parameter.

The bit you've quoted is only meant to be restricting what's used for the right operand, e.g. to prevent:

```csharp
static void M(in int i)
{
    ref int j = ref i;
}
```

For the second part, the error occurs even without `return ref` (even in a `void` method) I believe it's prohibited already:

12.21.3: "It is a compile time error if the ref-safe-context (§9.7.2) of the left operand is wider than the ref-safe-context of the right operand."

9.7.2.2: "If p is a ref, or in parameter, its ref-safe-context is the caller-context. If p is an in parameter, it can’t be returned as a writable ref but can be returned as ref readonly." (So the ref-safe-context of `i` is caller-context.)

9.7.2.2: (After things that don't apply to `j`) "Otherwise, the parameter is a value parameter, and its ref-safe-context is the function-member." So the ref-safe-context of `j` is function-member.

caller-context is wider than function-member, so it's prohibited.

Proposal for the meeting: check my logic, then close the issue without further spec changes.

KalleOlaviNiemitalo:
So you're saying that this

[...]

PR #917, opened 2023-09-02
Address issue #839

Nigel-Ecma:
The original text appeared to restrict the left operand to variables declared `ref` or `ref readonly` when a writeable or read-only ref is allowed.


jskeet (approved):
Yes, this looks okay to me. Admittedly it feels a little odd to have the two "If this then that" approaching things from opposite sides - but it's possibly the simplest way of expressing it.

Issue #910, opened 2023-08-20
13.6.2: compile time error due to ref-safe-context of variable_reference in local_variable_initializer


KalleOlaviNiemitalo:
**Describe the bug**

In section 13.6.2 of the C# 7.x draft, the following paragraph specifies a "compile time error" that seems impossible to trigger.

> A *local_variable_initializer* for a variable declared `ref` or `ref readonly` shall be of the form “`ref` *variable_reference*”. It is a compile time error if the scope of the local variable is wider than the ref-safe-context of the *variable_reference* (§9.7.2).

**Example**

The paragraph applies to initializing a `ref` local variable, not to ref assignment:

```csharp
ref T local_variable = ref variable_initializer;
```

What would you put as *variable_initializer* to trigger the compile time error?

**Expected behavior**

Remove the sentence about "compile time error", or add an example of invalid code that would trigger it.

**Additional context**

Split off from <https://github.com/dotnet/csharpstandard/issues/899#issuecomment-1684461026>.

KalleOlaviNiemitalo:
The scope of a local variable:

* 13.3.1 (Blocks / General):

  > A block may contain declaration statements (§13.6). The scope of a local variable or constant declared in a block is the block.

* 13.6.2 (Local variable declarations):

  > The scope of a local variable declared in a *local_variable_declaration* is the block in which the declaration occurs. It is an error to refer to a local variable in a textual position that precedes the *local_variable_declarator* of the local variable. Within the scope of a local variable, it is a compile-time error to declare another local variable or constant with the same name.

Can the ref-safe-context of the initializing expression be smaller than the block in which the local variable is declared, without this violating some other constraint?

The ref-safe-context of *variable_reference*:

* 9.7.2.2 (Local variable ref safe context):

  > For a local variable `v`:
  >
  > * If `v` is a reference variable, its ref-safe-context is the same as the ref-safe-context of its initializing expression.
  > * Otherwise its ref-safe-context is the context in which it was declared.

KalleOlaviNiemitalo:
Ugh… from 9.7.2.1 (Ref safe contexts / General):

> * ***declaration-block***: The ref-safe-context of a *variable_reference* to a local variable (§9.2.9) extends from its declaration to the end of the scope in which it is declared, including into any nested *embedded-statements* in this range.

So in this code

```csharp
class C {
    void M() {
        int dummy;
        int target = 0;
        ref int r = ref target;
    }
}
```

[...]

PR #922, opened 2023-09-06
Update declaration-scope defn

Nigel-Ecma:
One way to address #910

### DIFF
--- a/standard/variables.md
+++ b/standard/variables.md
@@ -1060,7 +1060,7 @@ For any variable, the ***ref-safe-context*** of that variable is the context whe
 
 There are three ref-safe-contexts:
 
-- ***declaration-block***: The ref-safe-context of a *variable_reference* to a local variable ([§9.2.9](variables.md#929-local-variables)) is that local variable’s scope ([§13.6.2](statements.md#1362-local-variable-declarations)), including into any nested *embedded-statement*s in the scope.
+- ***declaration-block***: The ref-safe-context of a *variable_reference* to a local variable ([§9.2.9](variables.md#929-local-variables)) is that local variable’s scope ([§13.6.2](statements.md#1362-local-variable-declarations)), including any nested *embedded-statement*s in that scope.
 
   A *variable_reference* to a local variable is a valid referent for a reference variable only if the reference variable is declared within the ref-safe-context of that variable.
 
--- a/standard/variables.md
+++ b/standard/variables.md
@@ -1060,7 +1060,7 @@ For any variable, the ***ref-safe-context*** of that variable is the context whe
 
 There are three ref-safe-contexts:
 
-- ***declaration-block***: The ref-safe-context of a *variable_reference* to a local variable ([§9.2.9](variables.md#929-local-variables)) extends from its declaration to the end of the scope in which it is declared, including into any nested *embedded-statement*s in this range.
+- ***declaration-block***: The ref-safe-context of a *variable_reference* to a local variable ([§9.2.9](variables.md#929-local-variables)) is that local variable’s scope ([§13.6.2](statements.md#1362-local-variable-declarations)), including into any nested *embedded-statement*s in the scope.
 
   A *variable_reference* to a local variable is a valid referent for a reference variable only if the reference variable is declared within the ref-safe-context of that variable.
 
@@ -1160,7 +1160,7 @@ These values form a nesting relationship from narrowest (declaration-block) to w
 For a local variable `v`:
 
 - If `v` is a reference variable, its ref-safe-context is the same as the ref-safe-context of its initializing expression.
-- Otherwise its ref-safe-context is the context in which it was declared.
+- Otherwise its ref-safe-context is declaration-block.
 
 #### 9.7.2.3 Parameter ref safe context
 


KalleOlaviNiemitalo:
Does this make the "compile time error if the scope of the local variable is wider than the ref-safe-context of the *variable_reference*" in §13.6.2 trigger in some situation in which it legitimately should?

Nigel-Ecma:
@KalleOlaviNiemitalo – Do you mean in a situation like:

```
    static void DeclarationBlockScope()
    {
        int a = 42;
        ref int b = ref a;

        if(a == b)
        {
            int c = 24;
            // b = ref c; // CS8374  Cannot ref-assign 'c' to 'b' because 'c' has a narrower escape scope than 'b'.
            ref int d = ref a;
            ref int e = ref c;
            d = e * 2;
            Console.WriteLine($"a = {a}; b = {b}; c = {c}; d = {d}; e = {e}");
        }
    }
```

This correctly produces the error shown as the scope of `b` is wider than than the ref-safe-context of `c`.

KalleOlaviNiemitalo:
`b = ref c` in your example is a ref assignment expression, and it is disallowed by [§12.21.3 (Ref assignment)](https://github.com/dotnet/csharpstandard/blob/d538391e2615696c873282af2dec65ed6493fb8b/standard/expressions.md#12213-ref-assignment):

> It is a compile time error if the ref-safe-context ([§9.7.2](https://github.com/dotnet/csharpstandard/blob/d538391e2615696c873282af2dec65ed6493fb8b/standard/variables.md#972-ref-safe-contexts)) of the left operand is wider than the ref-safe-context of the right operand.

The text in [§13.6.2 (Local variable declarations)](https://github.com/dotnet/csharpstandard/blob/d538391e2615696c873282af2dec65ed6493fb8b/standard/statements.md#1362-local-variable-declarations) applies to ref initialization, not to ref assignment:

> A *local_variable_initializer* for a variable declared `ref` or `ref readonly` shall be of the form “`ref` *variable_reference*”. It is a compile time error if the scope of the local variable is wider than the ref-safe-context of the *variable_reference* ([§9.7.2](https://github.com/dotnet/csharpstandard/blob/d538391e2615696c873282af2dec65ed6493fb8b/standard/variables.md#972-ref-safe-contexts)).

So your example does not show why §13.6.2 should have this text.

KalleOlaviNiemitalo:
In other words, this PR may be useful by making the "compile time error" in §13.6.2 not trigger on code that should be allowed (although I have not checked whether changing the declaration scope causes other problems). But I question the need to specify the "compile time error" in §13.6.2 at all.  Instead, §13.6.2 could have a note saying that the ref-safe-context of the ref variable being initialized will not be wider than the ref-safe-context of the *variable_reference* in the *local_variable_initializer*, because [explanation with links to other sections].

Nigel-Ecma:
@KalleOlaviNiemitalo – Oops sorry, yes I was not thinking of that particular issue you raised in #910. I've added #923 to look at this later – post v7.x.

standard/variables.md
```diff
@@ -1060,7 +1060,7 @@ For any variable, the ***ref-safe-context*** of that variable is the context whe
 
 There are three ref-safe-contexts:
 
-- ***declaration-block***: The ref-safe-context of a *variable_reference* to a local variable ([§9.2.9](variables.md#929-local-variables)) extends from its declaration to the end of the scope in which it is declared, including into any nested *embedded-statement*s in this range.
+- ***declaration-block***: The ref-safe-context of a *variable_reference* to a local variable ([§9.2.9](variables.md#929-local-variables)) is that local variable’s scope ([§13.6.2](statements.md#1362-local-variable-declarations)), including into any nested *embedded-statement*s in the scope.
```
KalleOlaviNiemitalo: "into" used to be part of "extends from … into" but it should be removed now.

Nigel-Ecma: @KalleOlaviNiemitalo – "into" was added. The ", including into any nested *embedded-statement*s in the scope" is just referring redundantly (in both the original and this PR) to nesting (§7.7.1). Such redundant statements are made in various places where they are (subjectively!) deemed to be helpful. This PR keeps that redundancy simply because it was previously deemed helpful by somebody. Let’s see if others think it should now go.

KalleOlaviNiemitalo: I'd be happy with "including any nested *embedded-statement*s in the scope".  Just the "into" seems strange as I don't think "is … into" makes sense.

Nigel-Ecma: Small word changes: “any” removed as suggested by @KalleOlaviNiemitalo , “the scope” now “that scope”

jskeet (approved):
Right - I *think* I understand the problem being addressed, and I *think* I understand the proposed solution, and agree with it.

It's possible there are other problems, or other approaches to fix it, but I'm happy with this approach, basically.

BillWagner (approved):
I like this change. It's simpler language, and there are other rules that prevent creating a `ref` local variable that refers to a variable declared later in the same block. (use of uninitialized variable)

jskeet:
So, do we think this PR is ready to mark as ready, and then merge? (Just trying to tie up loose ends...)

Nigel-Ecma:
@jskeet – I thought it was already out of draft, should be now (provided I used a browser Github likes, which is probably the reason it didn't "stick" when I did it previously…)

jskeet:
Great. Given that Bill and I have both reviewed, I think this is fine to merge.

PR #924, opened 2023-09-08
Tighten up language around local function declarations

### LINKED ISSUE #913: Clarity for reuse of local function name
**Describe the bug**

In section 7.3 (Declarations) of the C# 7.x draft, there is wording that disallows some reuse of names in local variable declaration spaces:

> - Additional local variable declaration spaces may occur within member declarations, anonymous functions and local functions. Names are introduced into these declaration spaces through *declaration_expression*s and *declaration_statement*s. Local variable declaration spaces may be nested, but it is an error for a local variable declaration space and a nested local variable declaration space to contain elements with the same name. Thus, within a nested declaration space it is not possible to declare a local variable or constant with the same name as a parameter, type parameter, local variable or constant in an enclosing declaration space. It is possible for two declaration spaces to contain elements with the same name as long as neither declaration space contains the other. Local declaration spaces are created by the following constructs:

In the paragraph quoted above, the "it is not possible to declare" sentence lists various kinds of declarations but not local functions.  However, the *declaration_statement* grammar rule includes *local_function_declaration*, and I think that means the name of a local function is introduced into a local variable declaration space and then cannot be reused.  §13.6.1 (Declaration statements / General):

> ```ANTLR
> declaration_statement
>     : local_variable_declaration ';'
>     | local_constant_declaration ';'
>     | local_function_declaration
>     ;
> ```

**Example**

Make it clearer that this is not allowed:

```csharp
class ShadowDemo {
    void Method() {
        void SameName() { }

        {
            void SameName() { }
        }
    }
}
```

**Expected behavior**

Explicitly list local functions in the "it is not possible to declare" sentence quoted above.

In addition, local functions and their parameters and type parameters could be mentioned in the following sentences, or perhaps it would be better to replace the sentences with notes that reference the constraints on "local variable declaration space" in §7.3:

* §13.6.2 (Local variable declarations):

  > Within the scope of a local variable, it is a compile-time error to declare another local variable or constant with the same name.

* §13.6.3 (Local constant declarations):

  > Within the scope of a local constant, it is a compile-time error to declare another local variable or constant with the same name.

**Additional context**

Other similar constraints:

* §13.6.4 (Local function declarations) says "unique in its declared block scope", which might not include enclosing scopes:

  > The identifier of a *local_function_declaration* must be unique in its declared block scope.

* §13.6.4 (Local function declarations) does not mention type parameters of local functions here, and "the enclosing scope" seems too tight as multiple enclosing scopes should be considered:

  > It is a compile-time error for a local function to declare a parameter or local variable with the same name as one declared in the enclosing scope.

* §15.6.2.1 (Method parameters / General):

  > It is an error for the method declaration space and the local variable declaration space of a nested declaration space to contain elements with the same name.

jskeet:
Fixes #913.

cc @KalleOlaviNiemitalo 

This is an initial attempt, which uses "enclosing local variable declaration spaces" rather than "enclosing scopes". The purpose of that is to allow this:

```csharp
public class C
{
    private int field;

    public void M()
    {
        LM();
        
        void LM()
        {
            int field = this.field;
            Console.WriteLine(field); // Refers to the local variable
        }
    }
}
```

The mixture of declaration space and block is certainly awkward though, and I'd welcome alternative suggestions. I'd also welcome suggestions for which aspects are non-controversial so they could be merged first - I suspect the changes in 7, 13.6.2 and 13.6.3 are simple, and then we can haggle over the changes in 13.6.4.


BillWagner (approved):
All these changes look good to me. We should add @Nigel-Ecma for review regarding the language on "scope" and "declaration space"

Nigel-Ecma (approved):
LGTM

PR #925, opened 2023-09-08
Fix misdescribed example of input argument

### LINKED ISSUE #911: 12.6.2.3: Misdescribed example of input argument
**Describe the bug**

In section 12.6.2.3 (Run-time evaluation of argument lists) of the C# 7.x draft, the description of this example is wrong:

<https://github.com/dotnet/csharpstandard/blob/0ad8948fb078da1eec0ac0dda7df8f0e05215e92/standard/expressions.md?plain=1#L608-L620>

An unnamed `int` variable is created only in the `M1(i + 5)` call and not in the `M1(i)` call, unlike what the description says.  The C# code comments in the example are correct.

**Example**

Linked above.

**Expected behavior**

> <ins>In the `M1(i)` method call, `i` itself is passed as an input argument, because it is classified as a variable and has the same type `int` as the input parameter.</ins> In <del>each</del><ins>the `M1(i + 5)`</ins> method call, an unnamed `int` variable is created, initialized with the argument’s value, and then passed as an input argument. See §12.6.4.2 and §12.6.4.4.

**Additional context**

Split off from <https://github.com/dotnet/csharpstandard/issues/894#issuecomment-1681650853>.

jskeet:
Fixes #911 (as described there)

cc @KalleOlaviNiemitalo

@BillWagner: This looks pretty simple and correct to me. The wording was taken from the issue.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -615,7 +615,7 @@ During the run-time processing of a function member invocation ([§12.6.6](expre
      > M1(i + 5);     // transformed to a temporary input argument
      > ```
      >
-     > In each method call, an unnamed `int` variable is created, initialized with the argument’s value, and then passed as an input argument. See [§12.6.4.2](expressions.md#12642-applicable-function-member) and [§12.6.4.4](expressions.md#12644-better-parameter-passing-mode).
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @jskeet 

Let's :shipit:

*If* this were normative, I might quibble about the "same type" instead of the formal "there is an identity conversion", but in an example, I think it's fine.

Issue #324, opened 2021-05-22
Variables and Variable References


RexJaeschke:
This Issue *replaces* Issue #270, ‘Revising the Definition of “Variable Reference”.’

I’ve put each proposal in a separate Comment, so you can record a thumb-up/down for each.

## Defining two kinds of variable references

The key to my proposals below is to have two flavors of variable reference: modifiable and non-modifiable, so we can adequately spec the following V7 features:

- Input parameters
- Methods/delegates with ref returns
- local ref variables
- Conditional expressions having a ref result
- ref assignment

## Variable reference

10.5 Variable references says

> A *variable_reference* is an *expression* that is classified as a variable. A *variable_reference* denotes a storage location that can be accessed both to fetch the current value and to store a new value.
> ```ANTLR
> variable_reference
>     : expression
>     ;
> ```
> > *Note*: In C and C++, a *variable_reference* is known as an *lvalue*. *end note*

**Proposal 1**: We should use the term “variable reference” (rather than “variable”) in all contexts involving an expression that denotes a variable.

RexJaeschke:
## Use of “lvalue”

**Proposal 2**: As this is the *only* place in the spec that we use *lvalue*, rather than correcting the note (because it ignores the C89-invented distinction between modifiable and non-modifiable lvalues), we should delete it.

RexJaeschke:
## Classified as a variable

In the current V6 draft spec, we use the phrase “is [not] classified as a variable” numerous times, all in the context of an expression. Some are declarative statements; others are qualifiers of the form “If <some expression> is classified as a variable …”

**Proposal 3 (initial)**: Replace all these occurrences with “is [not] a *variable_reference*.” After all, that’s exactly what a variable reference is!

RexJaeschke:
## Variable

10 Variables|10.1 General says

> Variables represent storage locations. Every variable has a type that determines what values can be stored in the variable. C# is a type-safe language, and the C# compiler guarantees that values stored in variables are always of the appropriate type.
## Operators needing write access

**Proposal 4 (initial)**: Rewrite

12.7.10 Postfix increment and decrement operators

> The operand of a postfix increment or decrement operation shall be an expression classified as a variable, a property access, or an indexer access.

as 
> The operand of a postfix increment or decrement operation shall be a  *variable_reference*, a property access, or an indexer access.

Likewise, for 12.8.6 Prefix increment and decrement operators, and 12.18 Assignment operators|12.18.1 General.

[...]

PR #927, opened 2023-09-12
Specify temporary for `this` as an input parameter

### LINKED ISSUE #894: 12.6.6.1: Defensive copy of input parameter
**Describe the bug**

In the C# 7.x draft, 12.6.6.1 (Function member invocation / General) describes how a copy of a value-type instance "not classified as a variable" is made and used instead of the original.  This defensive copying should also apply to an input parameter unless its type is a `readonly struct`.

**Example**

```csharp
using System;
struct S {
    public int I;
    public void M() { I++; }
}
readonly struct R {
    public void M() {}
}
class C {
    static void Main() {
        N(new S(), new S(), new R());
    }
    static void N(S s1, in S s2, in R r) {
        s1.M(); // no copying
        Console.WriteLine(s1.I); // 1
        
        s2.M(); // copy required
        Console.WriteLine(s2.I); // 0

        r.M(); // no copying
    }
}
```

**Expected behavior**

In 12.6.6.1, declare:

> If `E` is not classified as a variable, <ins>or if `E` is an input parameter (15.6.2.3) and `V` is not a readonly struct type (16.2.2),</ins> then a temporary local variable of E’s type is created and the value of E is assigned to that variable.

**Additional context**

None.

---
[Associated WorkItem - 157261](https://dev.azure.com/msft-skilling/Content/_workitems/edit/157261)

BillWagner:
Fixes #894 

Issue #894 has several comments that describe different possible concerns. To help reviewers (and to keep my head around each different scenario). Each commit will fix one of the concerns.

The first commit fixes the concern in the description: 

If a method takes an `in` parameter of a struct type that is not a `readonly struct`, then the compiler creates a temporary to invoke a (non-readonly) member of that type. Note that for C# 8.0, this will need to be revisited again to note that a copy needn't be created to invoke a `readonly` member of a non-readonly struct.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1211,7 +1211,11 @@ The run-time processing of a function member invocation consists of the followin
   - `M` is invoked.
 - Otherwise, if the type of `E` is a value-type `V`, and `M` is declared or overridden in `V`:
   - `E` is evaluated. If this evaluation causes an exception, then no further steps are executed. For an instance constructor, this evaluation consists of allocating storage (typically from an execution stack) for the new object. In this case `E` is classified as a variable.
-  - If `E` is not classified as a variable, or if `E` is an input parameter (§15.6.2.3) and `V` is not a readonly struct type (§16.2.2), then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` is a true variable is it possible for the caller to observe the changes that `M` makes to `this`.
+  - If `E` is not classified as a variable, or if `V` is not a readonly struct type (§16.2.2), and `E` is one of:
+    - an input parameter (§15.6.2.3), or
+    - a `readonly` field (§15.5.3), or
+    - a `readonly` reference variable or return (§9.7),
+    then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` is a true variable is it possible for the caller to observe the changes that `M` makes to `this`.
   - The argument list is evaluated as described in [§12.6.2](expressions.md#1262-argument-lists).
   - `M` is invoked. The variable referenced by `E` becomes the variable referenced by `this`.
 - Otherwise:
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1216,7 +1216,7 @@ The run-time processing of a function member invocation consists of the followin
     - a `readonly` field (§15.5.3), or
     - a `readonly` reference variable or return (§9.7),
 
-  then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` is a true variable is it possible for the caller to observe the changes that `M` makes to `this`.
+  then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` can be written is it possible for the caller to observe the changes that `M` makes to `this`.
   - The argument list is evaluated as described in [§12.6.2](expressions.md#1262-argument-lists).
   - `M` is invoked. The variable referenced by `E` becomes the variable referenced by `this`.
 - Otherwise:
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1211,7 +1211,7 @@ The run-time processing of a function member invocation consists of the followin
   - `M` is invoked.
 - Otherwise, if the type of `E` is a value-type `V`, and `M` is declared or overridden in `V`:
   - `E` is evaluated. If this evaluation causes an exception, then no further steps are executed. For an instance constructor, this evaluation consists of allocating storage (typically from an execution stack) for the new object. In this case `E` is classified as a variable.
-  - If `E` is not classified as a variable, then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` is a true variable is it possible for the caller to observe the changes that `M` makes to `this`.
+  - If `E` is not classified as a variable, or if `E` is an input parameter (§15.6.2.3) and `V` is not a readonly struct type (§16.2.2), then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` is a true variable is it possible for the caller to observe the changes that `M` makes to `this`.
   - The argument list is evaluated as described in [§12.6.2](expressions.md#1262-argument-lists).
   - `M` is invoked. The variable referenced by `E` becomes the variable referenced by `this`.
 - Otherwise:
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1215,7 +1215,8 @@ The run-time processing of a function member invocation consists of the followin
     - an input parameter (§15.6.2.3), or
     - a `readonly` field (§15.5.3), or
     - a `readonly` reference variable or return (§9.7),
-    then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` is a true variable is it possible for the caller to observe the changes that `M` makes to `this`.
+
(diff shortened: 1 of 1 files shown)


BillWagner:
Regarding https://github.com/dotnet/csharpstandard/issues/894#issuecomment-1679958773 and https://github.com/dotnet/csharpstandard/issues/894#issuecomment-1680433790

I don't believe a spec change is needed to prohibit a defensive copy for a `readonly` struct as an *input_parameter* when a member is called. The fact that a defensive copy isn't required should be sufficient to allow compiler writers freedom of implementation.

BillWagner:
Regarding https://github.com/dotnet/csharpstandard/issues/894#issuecomment-1680455321 (what about `System.Enum.ToString()` and related methods?

I think this can be something we leave to implementations as an optimization. In the case of roslyn, they may avoid the defensive copy because the compiler "knows" that `System.Enum.ToString()` doesn't modify the instance. 

Without writing contrived code, you can't observe the lack of a defensive copy.

Question for the committee:  Should we add a note that a specific implementation may be able to avoid the defensive copy when the behavior of a specific method is well known?

BillWagner:
Regarding https://github.com/dotnet/csharpstandard/issues/894#issuecomment-1681573559 Is the `readonly struct` special case also required when the parameter is not classified as a variable?

I don't think any changes are needed. If `E` is not classified as a variable, a copy *must* be made, because `this` is classified as a `ref` parameter. (Starting in C# 11, `this` is classified as a [`scoped ref` parameter](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/low-level-struct-improvements.md#scoped-modifier), which makes the `ref` safety rules more consistent.) The result is that a temporary variable must be created for `E`.

BillWagner:
Regarding https://github.com/dotnet/csharpstandard/issues/894#issuecomment-1681573559 (other `readonly` ref variables

2nd commit updates the restriction to include `readonly` fields, `ref readonly` local variables and `ref readonly` returns.

BillWagner:
Regarding https://github.com/dotnet/csharpstandard/issues/894#issuecomment-1703734163 

>As another example, I think draft-v7 does not currently forbid this, even though it should:
>
> ```csharp
> class C {
>    void M(ref int i) {}
>     void N(in int i) {
>         M(ref i);
>     }
> }
> ```

I think the standard does prohibit this via §15.6.2.3 ("It is a compile-time error to modify the value of an input parameter.") A reasonable interpretation of passing a parameter as a `ref` parameter is that it is modified in the body of that method. For committee discussion. If a change is needed, we should prohibit passing an `in` parameter as a `ref` or `out` argument to another method, except as the receiver.

BillWagner:
The final two comments on #894 detail where we could take the strategy of defining a `*variable_reference*` and `*writable_variable_reference*` to distinguish.

I think the proposed changes fix the issue, and are smaller in scope.

BillWagner:
Tagging @cston to take a look at this PR and my notes regarding #894. I'd like you to make sure these changes to prohibit any optimizations for `in` parameters.

BillWagner:
Regarding https://github.com/dotnet/csharpstandard/issues/894#issuecomment-1716889949 

Example:

```csharp
struct S {
    public void M() {}
}
class C {
    void N(bool b, ref S r, in S i) {
        (b ? ref r : ref i).M();
    }
}
```

I believe this is covered, but it requires a look at 12.18:

> If `ref` is present:
...
> - The result is a *variable_reference*, which is writable if both *variable_references* are writable.

This example highlights why I'm trying to avoid limitations on implementers.  This doesn't require a defensive copy of the `ref` parameter. However, the semantics of an `in` parameter require the defensive copy when the result is the `in` parameter.

If a defensive copy were required regardless of the result, a compliant compiler would be forced to create that copy, even when it isn't necessary. The motivation for all the `ref` safety features is to improve performance by providing techniques to avoid both allocations and copies.

We could make a change to 12.18, but I think the description of `in` parameters already enforces the defensive copy there. We should also avoid mandating a copy when the result is a writable variable reference.

standard/expressions.md
```diff
@@ -1211,7 +1211,11 @@ The run-time processing of a function member invocation consists of the followin
   - `M` is invoked.
 - Otherwise, if the type of `E` is a value-type `V`, and `M` is declared or overridden in `V`:
   - `E` is evaluated. If this evaluation causes an exception, then no further steps are executed. For an instance constructor, this evaluation consists of allocating storage (typically from an execution stack) for the new object. In this case `E` is classified as a variable.
-  - If `E` is not classified as a variable, then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` is a true variable is it possible for the caller to observe the changes that `M` makes to `this`.
+  - If `E` is not classified as a variable, or if `V` is not a readonly struct type (§16.2.2), and `E` is one of:
+    - an input parameter (§15.6.2.3), or
+    - a `readonly` field (§15.5.3), or
+    - a `readonly` reference variable or return (§9.7),
```
cston, line 1217: Comparing with the list in https://github.com/dotnet/csharpstandard/issues/894#issuecomment-1703793806, does this set include get-only auto properties?

standard/expressions.md
```diff
@@ -1211,7 +1211,11 @@ The run-time processing of a function member invocation consists of the followin
...
+    then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` is a true variable is it possible for the caller to observe the changes that `M` makes to `this`.
```
cston: It looks like this line is merged with the previous bullet when the document is displayed. Is that intended?

standard/expressions.md
BillWagner: It doesn't. This text shouldn't need to, as it's concerned with when parameters must be copied. A get only auto prop doesn't have parameters.

The `ref readonly` part is already covered by other rules (you can't `ref` assign a `ref readyonly` return to a writable ref.

standard/expressions.md
```diff
@@ -1211,7 +1211,12 @@ The run-time processing of a function member invocation consists of the followin
   - `M` is invoked.
 - Otherwise, if the type of `E` is a value-type `V`, and `M` is declared or overridden in `V`:
   - `E` is evaluated. If this evaluation causes an exception, then no further steps are executed. For an instance constructor, this evaluation consists of allocating storage (typically from an execution stack) for the new object. In this case `E` is classified as a variable.
-  - If `E` is not classified as a variable, then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` is a true variable is it possible for the caller to observe the changes that `M` makes to `this`.
+  - If `E` is not classified as a variable, or if `V` is not a readonly struct type (§16.2.2), and `E` is one of:
+    - an input parameter (§15.6.2.3), or
+    - a `readonly` field (§15.5.3), or
```
AlekseyTs, line 1216: A readonly field is writable in some context, like in a constructor, for example, and, I think, won't be subject to the copy there. Instead of trying to explain all this here, we could introduce a notion of a readonly and writable variable as a classification for expression. That would cover in parameters, readonly fields in specific context, ref readonly references, etc. Then here we can simply say: "If `E` is not classified as a **writable** variable then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. ..." Without all the bullet points.

AlekseyTs (approved):

standard/expressions.md
jskeet: @BillWagner: I like this idea, but presumably you're not planning on doing that now? (Sounds like something to do soon after the C# 7 release though.)

standard/expressions.md
```diff
@@ -1211,7 +1211,12 @@ The run-time processing of a function member invocation consists of the followin
...
+    - a `readonly` reference variable or return (§9.7),
+
+  then a temporary local variable of `E`’s type is created and the value of `E` is assigned to that variable. `E` is then reclassified as a reference to that temporary local variable. The temporary variable is accessible as `this` within `M`, but not in any other way. Thus, only when `E` is a true variable is it possible for the caller to observe the changes that `M` makes to `this`.
```
jskeet: "true variable" has very faint sirens ringing for me ("no true Scotsman"?) but I think it's probably fine.

jskeet (approved):
A couple of comments, but generally okay.

standard/expressions.md
BillWagner: Rex has captured some of implications of introducing "writable" and "readonly" variables in #324.

I agree that this should be done very early in the C# 8 tasks.

standard/expressions.md
BillWagner: Language changed to "can be written to"

Nigel-Ecma:
> Regarding [#894 (comment)](https://github.com/dotnet/csharpstandard/issues/894#issuecomment-1680455321) (what about `System.Enum.ToString()` and related methods?
> 
> Question for the committee: Should we add a note that a specific implementation may be able to avoid the defensive copy when the behavior of a specific method is well known?

Can we not say that for *all* defensive copies? I.e. that a compiler can omit any of them if it knows the defense is not required? Indeed can we not specify that normatively as in “a defensive copy shall be made unless the compiler can determine it is not required”?

MadsTorgersen (approved):

PR #929, opened 2023-09-13
Prohibit access to "this" in local functions in structs

### LINKED ISSUE #926: Local function in struct accessing this (error CS1673)
**Describe the bug**

The standard should forbid a local function in a member of a `struct` type from accessing `this` or instance members of `this`.  I don't see where the C# 7.x draft does that.

**Example**

The following should not be allowed.  [SharpLab](https://sharplab.io/#v2:CYLg1APgzgLgTgVwMYwAQGVUG8CwAoVQ1ASwDs1iBufI1GogAQBZUBZACgEpt7bDMAclx4E+YhgHZUMABbEolQgHolqAKZw4AezioAwugCMANgkBmXnwC+l2raJk0AeWG5RY2pJKKV6zTv0jUwt3a0sbPAigA===)

```csharp
struct S {
    int i;
    
    void M() {
        S N() {
            return this;  // error CS1673
        }
        
        int O() {
            return i; // error CS1673
        }
    }
}
```

**Expected behavior**

In [§13.6.4 (Local function declarations)](https://github.com/dotnet/csharpstandard/blob/d538391e2615696c873282af2dec65ed6493fb8b/standard/statements.md#1364-local-function-declarations), declare that a local function in a member of a `struct` shall not access `this` or instance members of `this`.  Static members are okay though.

Add the constraint in or near this paragraph, which talks about variables captured by local functions: <https://github.com/dotnet/csharpstandard/blob/d538391e2615696c873282af2dec65ed6493fb8b/standard/statements.md#L518>

**Additional context**

This is a new problem in C# 7, as the equivalent thing for anonymous delegates and lambda expressions is disallowed in [§12.19.3 (Anonymous function bodies)](https://github.com/dotnet/csharpstandard/blob/d538391e2615696c873282af2dec65ed6493fb8b/standard/expressions.md#12193-anonymous-function-bodies): <https://github.com/dotnet/csharpstandard/blob/d538391e2615696c873282af2dec65ed6493fb8b/standard/expressions.md?plain=1#L4978>

jskeet:
Fixes #926

### DIFF
--- a/standard/statements.md
+++ b/standard/statements.md
@@ -519,7 +519,9 @@ A local function is declared at block scope, and that function may capture varia
 
 When the type of `this` is a struct type, it is a compile-time error for the body of a local function to access `this`. This is true whether the access is explicit (as in `this.x`) or implicit (as in `x` where `x` is an instance member of the struct). This rule only prohibits such access and does not affect whether member lookup results in a member of the struct.
 
-> *Note*: this rule mirrors the rule for anonymous functions in [§12.19.3](expressions.md#12193-anonymous-function-bodies). *end note*
+It is a compile-time error for the body of the local function to contain a `goto` statement, a `break` statement, or a `continue` statement whose target is outside the body of the local function.
+
+> *Note*: the above rules for `this` and `goto` mirror the rules for anonymous functions in [§12.19.3](expressions.md#12193-anonymous-function-bodies). *end note*
 
 A local function may be called from a lexical point prior to its declaration. However, it is a compile-time error for the function to be declared lexically prior to the declaration of a variable used in the local function ([§7.7](basic-concepts.md#77-scopes)).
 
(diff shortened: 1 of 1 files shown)


jskeet:
I've added a second commit which would also fix #930.

(Note that it doesn't talk about nested local functions - or indeed nested anonymous functions - as I believe that's already effectively covered in 13.10.4.)

standard/statements.md
```diff
@@ -517,6 +517,12 @@ A *local_function_declaration* may include one `async` ([§15.15](classes.md#151
 
 A local function is declared at block scope, and that function may capture variables from the enclosing scopes. It is a compile-time error if a captured variable is read by the body of the local function but is not definitely assigned before each call to the function. The compiler shall determine which variables are definitely assigned on return ([§9.4.4.33](variables.md#94433-rules-for-variables-in-local-functions)).
 
+When the type of `this` is a struct type, it is a compile-time error for the body of a local function to access `this`. This is true whether the access is explicit (as in `this.x`) or implicit (as in `x` where `x` is an instance member of the struct). This rule only prohibits such access and does not affect whether member lookup results in a member of the struct.
```
BillWagner, line 520: Do you think we need to explicitly state that assignment to `this` is also prohibited? I think assignment requires access, so we're good, but wanted your thoughts as well.

```csharp
public struct SThis
{
    public void M()
    {
        void LocalFunction()
        {
            this = new SThis(); // compiler error
        }
    }
}
```

BillWagner (approved):
This LGTM. I had one question, but I think this is fine as is.

standard/statements.md
jskeet: Nope, I think it's fine to treat that as covered by the proposed text. Thanks for checking though.

PR #934, opened 2023-09-15
Edits suggested for patterns

### LINKED ISSUE #774: Address outstanding comments on patterns
We merged #757 with several comments still open, in order to get the first draft to ECMA members.

This issue is basically a reminder that we should still go through them. Neal, feel free to just resolve any that have been addressed, or @ people for more details if they're insufficient. I'd expect that a single PR could address a bunch of the niggles that were left as comments.

BillWagner:
Fixes #774 

Respond to the comments from the committee on the first take at patterns.

First commit:  Smaller edits. Some of these are trivial, but making the edits was less time consuming than logging an issue.


standard/expressions.md
```diff
@@ -4378,16 +4380,16 @@ The tuple equality operator `x != y` is evaluated as follows:
...
 
 #### 12.12.12.1 The is-type operator
 
 The *is-type operator* is used to check if the run-time type of an object is compatible with a given type. The check is performed at runtime. The result of the operation `E is T`, where `E` is an expression and `T` is a type other than `dynamic`, is a Boolean value indicating whether `E` is non-null and can successfully be converted to type `T` by a reference conversion, a boxing conversion, an unboxing conversion, a wrapping conversion, or an unwrapping conversion.
 
 The operation is evaluated as follows:
 
-1. If `E` is an anonymous function, a compile-time error occurs
-1. If `E` is a method group or the `null` literal, of if the value of `E` is `null`, the result is `false`.
+1. If `E` is an anonymous function, lambda expression, or method group, a compile-time error occurs
```
jskeet: Isn't a lambda expression already an anonymous function? There may be some subtlety here that I'm missing. (It's possible that this should be either "anonymous method, lambda expression, or method group" or "anonymous function or method group". But this is definitely non-blocking.

standard/patterns.md
```diff
@@ -40,11 +40,13 @@ single_variable_designation
     ;
 ```
 
-The runtime type of the value is tested against the *type* in the pattern. If it is of that runtime type (or some subtype), the pattern *matches* that value. This pattern form never matches a `null` value.
+The runtime type of the value is tested against the *type* in the pattern using the same rules specified in the is-type operator (§12.12.12.1) returns true. If the test succeeds succeeds, the pattern *matches* that value. It is a compile-time error if the *type* is a nullable value type (§8.3.12). This pattern form never matches a `null` value.
+
+> *Note*: The is-type expression `e is T` and the declaration pattern `e is T _` are the same when `T` isn't a nullable type. *end note*
```
jskeet: "the same" to "equivalent" perhaps?

"isn't a nullable type" sounds a bit strange to me too - would "isn't a nullable value type" be better? (Aside from anything else, when we have nullable reference types, presumably it's *still* the same for reference types, whether the compiler thinks it's nullable or not.)

standard/patterns.md
```diff
@@ -158,7 +161,7 @@ The following rules define when a set of patterns is *exhaustive* for a type:
 
 A set of patterns `Q` is *exhaustive* for a type `T` if any of the following conditions hold:
 
-1. `T` is an integral or enum type, or a nullable version of one of those, and for every possible value of `T`’s underlying type, some pattern in `Q` would match that value; or
+1. `T` is an integral or enum type, or a nullable version of one of those, and for every possible value of `T`’s nullable underlying type, some pattern in `Q` would match that value; or
```
jskeet: "Nullable underlying type" sounds like a contradiction in terms to me. (The underlying type isn't nullable, T is.) Would "non-nullable underlying type (if `T` is a nullable type)" be better? Happy to keep this as it is if I'm missing something. (Currently thinking through a Covid brain fog, so may well not be clear...)

standard/statements.md
```diff
@@ -917,7 +917,7 @@ The end point of a `switch` statement is reachable if the switch statement is re
 >         case "circle":
 >             return new Circle(2);
 >         …
->         case var o when (o?.Trim().Length ?? 0) == 0:
+>         case var o when (string.IsNullOrWhiteSpace(o)) == true:
```
jskeet: Could this just be `case var o when string.IsNullableOrWhitespace(o)`? (No need for `== true` or the brackets?)

standard/variables.md
```diff
@@ -948,7 +948,7 @@ Delegate conversions have a control flow path to the local function body. Captur
 >     // OK. s is now definitely assigned.
 >     F1();
 >
->     // OK, F3 reads s2, which is definitely assigned in F2.
```
jskeet, line 951: It looks like this has removed two trailing spaces, but there are more to go?

jskeet (approved):
A few nits, but otherwise good to go, thanks!

standard/expressions.md
BillWagner: I thought so too. I'll leave this until our meeting because one of the comments was that this bullet "didn't cover lambda expressions". I don't know if that was a stale comment, or if one of had something different in mind.

standard/patterns.md
BillWagner: I updated to use the "equivalent" language.

IF I specify "nullable value type", we'll need to update for nullable reference types in C# 8.0. `e is string?` isn't valid, only `e is string`.

standard/patterns.md
BillWagner: It was just my typo. Updated.

standard/expressions.md
BillWagner: Making the suggestion that anonymous function includes lambda expression:

```suggestion
1. If `E` is an anonymous function or method group, a compile-time error occurs
```

standard/patterns.md
```diff
@@ -40,11 +40,13 @@ single_variable_designation
     ;
 ```
 
-The runtime type of the value is tested against the *type* in the pattern. If it is of that runtime type (or some subtype), the pattern *matches* that value. This pattern form never matches a `null` value.
+The runtime type of the value is tested against the *type* in the pattern using the same rules specified in the is-type operator (§12.12.12.1) returns true. If the test succeeds succeeds, the pattern *matches* that value. It is a compile-time error if the *type* is a nullable value type (§8.3.12). This pattern form never matches a `null` value.
```
```suggestion
The runtime type of the value is tested against the *type* in the pattern using the same rules specified in the is-type operator (§12.12.12.1). If the test succeeds, the pattern *matches* that value. It is a compile-time error if the *type* is a nullable value type (§8.3.12). This pattern form never matches a `null` value.
```

MadsTorgersen (approved):

PR #940, opened 2023-09-18
Revision of §13.6.2 *Local variable declarations* to fix errors and tidy up

Nigel-Ecma:
This clause has evolved over time as first implicitly typed, and now in v7 ref locals, have been introduced; the result ending up as something worthy of Heath Robinson or Professor Branestawm ;-) The result alas is also not entirely correct, the initial version was (almost), but subsequent kinds of local variables don't follow exactly the same rules...

As a simple example of what ended up going wrong is the assertion:

> A local variable declaration that declares multiple variables is equivalent to multiple declarations of single variables with the same type and *ref_kind*.

and its associated example. This was (almost) correct originally, but is not correct for implicitly typed or ref local declarations where the initialization cannot be split from the declaration. (It was “almost” as it doesn't allow for *for_initializer* or *resource_acqusition* cases where such a split is not possible either.)

The rework divides local variable declarations into three subgroups, there really are four but we can get away with three (whether we should can be debated). These groups are directly represented in the reworked grammar, which also replaces a lot of conditional English prose. The clause starts with the general rules and then has a section for each subgroup detailing just the rules that apply in that case.

All examples have been kept, though I'm not entirely persuaded by the value for the (now) first one.

The changes have been kept to the §13.6.2 clause; there are related issues in at least §7.7.1 *Scopes:General* and §13.14 *The using statement* – neither of these for example define the scope of a variable introduced in a *resource_acqusition*. Such issues are left for future (pre-v8) work!

I think some of the introduced errors this fixes are important enough to make v7, I expect we will debate that! :-), so I'll tag it that way for now.

This should address some of the more recent issues raised – the revision came from reading some of those, heading to this clause, and finding things needed fixing apart from those issues – but I haven't gone through to see which ones it addresses.

### DIFF
--- a/standard/statements.md
+++ b/standard/statements.md
@@ -313,7 +313,7 @@ Implicitly typed declarations contain the contextual keyword ([§6.4.4](lexical-
 
 Within a *local_variable_declaration* each variable is introduced by a *declarator*, which is one of *implicitly_typed_local_variable_declarator*, *explicitly_typed_local_variable_declarator* or *ref_local_variable_declarator* for impicitly typed, explicitly typed and ref local variables respectively. The declarator defines the name (*identifier*) and initial value, if any, of the introduced variable.
 
-If there are multiple declarators in a declaration then they are processed, including any initializing expressions, in order left to right ([§9.4.4.5](variables.md#945-declaration-statements)).
+If there are multiple declarators in a declaration then they are processed, including any initializing expressions, in order left to right ([§9.4.4.5](variables.md#9445-declaration-statements)).
 
 > *Note*: For a *local_variable_declaration* not occuring as a *for_initializer* ([§13.9.4](statements.md#1394-the-for-statement)) or *resource_acquisition* ([§13.14](statements.md#1314-the-using-statement)) this left to right order is equivalent to each declarator being within a separate *local_variable_declaration*. For example:
 >
@@ -339,7 +339,7 @@ If there are multiple declarators in a declaration then they are processed, incl
 >
 > *end note*
 
-The value of a local variable is obtained in an expression using a *simple_name* ([§12.8.4](expressions.md#1284-simple-names)). A local variable shall be definitely assigned ([§9.4](variables.md#94-definite-assignment)) at each location where its value is obtained. Each local variable introduced by a *local_variable_declaration* is *initially unassigned* ([§9.4.3](variables.md#99-initially-unassigned-variables)). If a declarator has an initializing expression then the introduced local variable is classified as *assigned* at the end of the declarator ([§9.4.4.5](variables.md#945-declaration-statements)).
+The value of a local variable is obtained in an expression using a *simple_name* ([§12.8.4](expressions.md#1284-simple-names)). A local variable shall be definitely assigned ([§9.4](variables.md#94-definite-assignment)) at each location where its value is obtained. Each local variable introduced by a *local_variable_declaration* is *initially unassigned* ([§9.4.3](variables.md#943-initially-unassigned-variables)). If a declarator has an initializing expression then the introduced local variable is classified as *assigned* at the end of the declarator ([§9.4.4.5](variables.md#9445-declaration-statements)).
 
 The scope of a local variable introduced by a *local_variable_declaration* is defined as follows ([§7.7](basic-concepts.md#77-scopes)):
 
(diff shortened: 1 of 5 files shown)


KalleOlaviNiemitalo:
This inspired me to play with `ref` declarations at sharplab.io, and I found an oddity, but it turned out to be a compiler bug already fixed as <https://github.com/dotnet/roslyn/issues/64259>.

BillWagner (commented):
I also haven't reviewed which current issues are fixed by this PR.

That said, I like the updated description of these features.

The grammar check passed, so the updated grammar didn't break anything in the grammar elsewhere. 

This PR should be updated to replace any of the removed grammar productions in text in the standard (for instance, *local_variable_declarator* occurs in a number of places in basic-concepts, other locations in statements, and one in variables.

I didn't check other removed or updated grammar elements.

standard/statements.md
```diff
@@ -368,21 +325,46 @@ A local variable declaration that declares multiple variables is equivalent to m
...
+
+The scope of a local variable introduced by a *local_variable_declaration* is defined as follows ([§7.7](basic-concepts.md#77-scopes)):
+
+- If the declaration occurs as a *for_initializer* then the scope is the *for_initializer*, *for_condition*, *for_iterator*, and *embedded_statement* ([§13.9.4](statements.md#1394-the-for-statement));
+- If the declaration occurs as a *resource_acquisition* then the scope is the outermost block of the semantically equivalent expansion of the *using_statement* ([§13.14](statements.md#1314-the-using-statement));
+- Otherwise the scope is the block in which the declaration occurs.
+
+It is an error to refer to a local variable by name in a textual position that precedes its declarator, or within any initializing expression within its declarator. Within the scope of a local variable, it is a compile-time error to declare another local variable, local function or constant with the same name.
```
jskeet, line 350: At some point it became okay to declare lambda expression parameters with existing names:

```csharp
static void Main()
{
    int x = 10;
    Action<int> action = x => Console.WriteLine(x);
    action(20);
    Console.WriteLine(x);
}
```

Is that post C# 7?

standard/statements.md
```diff
@@ -368,21 +325,46 @@ A local variable declaration that declares multiple variables is equivalent to m
...
+    | ref_kind 'var' ref_local_variable_declarator
+    ;
+
+implicitly_typed_local_variable_declarator
+    : identifier '=' expression
+    ;
+```
 
-In an implicitly typed local variable declaration, the type of the local variable being declared is taken to be the same as the type of the expression used to initialize the variable.
+An *implicity_typed_local_variable_declaration* introduces a single local variable, *identifier*. The *expression* or *variable_reference* must have a compile-time type, `T`. The first variant declares a variable with type `T` and an initial value of *expression*. The second variant declares a ref variable with type `ref T` and an initial value of `ref` *variable_reference*.
```
jskeet: Can we use "alternative" or something similar instead of "variant"? (Just because variant and variable are so similar-sounding.)

jskeet (approved):
Yes, I like this - just a couple of comments to discuss in the meeting.

standard/statements.md
KalleOlaviNiemitalo: Based on <https://github.com/dotnet/csharplang/issues/3906>, I think it's a C# 8 feature, but I haven't found the speclet for the change.

KalleOlaviNiemitalo: The feature is <https://github.com/dotnet/csharplang/issues/2777>

KalleOlaviNiemitalo: It's listed in the C# 8 feature tracker: <https://github.com/dotnet/csharpstandard/blob/1d44af8dbaf3e7a231b0276b4fe12bc699a3ab39/admin/v8-feature-tracker.md?plain=1#L23>

standard/statements.md
```suggestion
An *implicity_typed_local_variable_declaration* introduces a single local variable, *identifier*. The *expression* or *variable_reference* must have a compile-time type, `T`. The first alternative declares a variable with type `T` and an initial value of *expression*. The second alternative declares a ref variable with type `ref T` and an initial value of `ref` *variable_reference*.
```

MadsTorgersen (approved):

BillWagner (approved):
Ths LGTM with the most recent commit.

PR #942, opened 2023-09-19
Handle pattern variables in declaration spaces

MadsTorgersen:
Handle pattern variables and a few forgotten contexts for expression variables in declaration spaces.

This fixes the "Declaration spaces" bullet of #768.


standard/basic-concepts.md
```diff
@@ -82,12 +82,14 @@ There are several different types of declaration spaces, as described in the fol
...
   - Each *variable_initializer* in a field and property declaration introduces its own local variable declaration space, that is not nested within any other local variable declaration space.
   - The body of a function member, anonymous function, or local function, if any, creates a local variable declaration space that is considered to be nested within the function’s local variable declaration space.
   - Each *constructor_initializer* creates a local variable declaration space nested within the instance constructor declaration. The local variable declaration space for the constructor body is in turn nested within this local variable declaration space.
-  - Each *block*, *switch_block*, *iteration_statement* and *using_statement* creates a nested local variable declaration space.
+  - Each *block*, *switch_block*, *specific_catch_clause*, *iteration_statement* and *using_statement* creates a nested local variable declaration space.
```
Nigel-Ecma, line 89: Line 85 above now includes *exception_specifier* as introducing a declaration space, here *specific_catch_clause* is added. In the current grammar the former only occurs as part of the latter. Is the case here meant to be the *block* of a *specific_catch_clause*, so its space is nested inside that of the *exception_specifier*? Or something else? Off hand I'm not sure why introducing two nested spaces per *specific_catch_clause* makes sense…

Mistake? Lack of clarity?

standard/basic-concepts.md
```diff
@@ -82,12 +82,14 @@ There are several different types of declaration spaces, as described in the fol
...
   - Each *embedded_statement* that is not directly part of a *statement_list* creates a nested local variable declaration space.
+  - Each *switch_section* creates a nested local variable declaration space. However, variables declared directly within the *statement_list* of the *switch_section* (but not within a nested local variable declaration space inside the *statement_list*) are added directly to the local variable declaration space of the enclosing *switch_block*, instead of that of the *switch_section*.
```
Nigel-Ecma, line 91: I know you called this “subtle”, I might use another term… That said the description of this “feature” looks correct here, however I think an example or note would be particularly worthwhile here

Nigel-Ecma (commented):
A couple of comments that may result in changes.

standard/basic-concepts.md
jskeet: My guess is that it's for this:

```csharp
catch (Exception e) when (e is ArgumentException ae)
{
  // Use ae
}
```

We need the declaration space both for the filter (so that the expression can refer to e) and to make sure that `ae` is in scope in the block.

(I know that's a terrible filter, of course...)

standard/basic-concepts.md
jskeet: +1 for having a note, although I don't think that should block C# 7.

I note that in "The C# Programming Language 4th edition (Special Annotated Edition)" there was a suggestion that it's a pity the original C# designers hadn't redesigned switch/case more thoroughly...

jskeet (approved):
Looks fine to me, but I don't trust myself on it.

standard/basic-concepts.md
MadsTorgersen: *exception_specifier* is not listed as introducing a declaration space but as introducing a variable into one. *specific_catch_clause* introduces a declaration space that is shared between the *exception_specifier* and the filter (they don't each need to introduce their own, because the *exception_specifier* wouldn't have a way to observe variables introduced in the filter anyway), and within which the declaration space of the *block* is nested.

Make sense?

standard/basic-concepts.md
MadsTorgersen: I rewrote this particular bullet many times and did many tests to assure myself that it was correct. But straightforward it is not! 😄

I agree in principle on the note, but the problem is that I wouldn't know what to write in it!

standard/statements.md
```diff
@@ -1550,7 +1550,7 @@ A *try_statement* consists of the keyword `try` followed by a *block*, then zero
...
 
-When a `catch` clause specifies both a *class_type* and an *identifier*, an ***exception variable*** of the given name and type is declared. The exception variable corresponds to a local variable with a scope that extends over the `catch` block. During execution of the *exception_filter* and `catch` block, the exception variable represents the exception currently being handled. For purposes of definite assignment checking, the exception variable is considered definitely assigned in its entire scope.
+When a `catch` clause specifies both a *class_type* and an *identifier*, an ***exception variable*** of the given name and type is declared. The exception variable is introduced into the nearest enclosing declaration space, namely that of the *specific_catch_clause* ([§7.3](basic-concepts.md#73-declarations)). During execution of the *exception_filter* and `catch` block, the exception variable represents the exception currently being handled. For purposes of definite assignment checking, the exception variable is considered definitely assigned in its entire scope.
```
BillWagner: I'm confused by the wording here. When I read "nearest enclosing declaration space", I think that direction is "outward", meaning it would be the block containing the `try` and all associated `catch` blocks. The next phrase indicates the opposite: going "inward", to mean the *specific_catch_clause*. Maybe just shorten it as follows:

```suggestion
When a `catch` clause specifies both a *class_type* and an *identifier*, an ***exception variable*** of the given name and type is declared. The exception variable is introduced into the declaration space of the *specific_catch_clause* ([§7.3](basic-concepts.md#73-declarations)). During execution of the *exception_filter* and `catch` block, the exception variable represents the exception currently being handled. For purposes of definite assignment checking, the exception variable is considered definitely assigned in its entire scope.
```

BillWagner (commented):
I had only one additional comment.

This looks good.

standard/basic-concepts.md
Nigel-Ecma: Well in as much as the design makes sense ;-)

standard/basic-concepts.md
Nigel-Ecma: Agree this doesn't need to block 7.3, but maybe an issue for "pre 8" to add an example?

gafter (approved):

standard/basic-concepts.md
jskeet: @Nigel-Ecma: I'll add that now.

PR #941, opened 2023-09-19
Add the rules for relaxing ref safety restrictions in an `unsafe` context.

BillWagner:
I read through the grammar, and I don't believe any changes are necessary to allow pointer variables with the `ref`, `in`, or `out` modifiers. However, text regarding `out` and `ref` parameters has been modified to include `in` parameters. Text regarding returning pointer types now includes returning pointer types by `ref` or `readonly ref`.

Fixes #886

PR #943, opened 2023-09-20
allow `unsafe` and `ref` in grammar

### LINKED ISSUE #886: Investigate combinations of `unsafe` & `ref`
In PR #837 @KalleOlaviNiemitalo raised two comments regarding combinations of `unsafe` & `ref` that they believe should be allowed. They are being pushed to this issue to allow the PR to proceed.

> This should be allowed:
> ```
> unsafe delegate ref int D(int* p);
> ```
[link](https://github.com/dotnet/csharpstandard/pull/837#discussion_r1260615287)

> This should be allowed:
> ```
> class C {
>     unsafe ref int M(int* p) => ref *p;
> }
[link](https://github.com/dotnet/csharpstandard/pull/837#discussion_r1260615825)

---
[Associated WorkItem - 157262](https://dev.azure.com/msft-skilling/Content/_workitems/edit/157262)

BillWagner:
`unsafe` and ref returns are allowed for delegates, local functions, and methods. Update the grammar for that part.

Fixes #886

See https://github.com/dotnet/csharpstandard/pull/941#pullrequestreview-1634196517

### DIFF
--- a/standard/statements.md
+++ b/standard/statements.md
@@ -447,7 +447,7 @@ A *local_function_declaration* declares a local function.
 ```ANTLR
 local_function_declaration
     : local_function_modifier* return_type local_function_header local_function_body
-    | ref_kind ref_return_type local_function_header ref_local_function_body
+    | ref_local_function_modifier* ref_kind ref_return_type local_function_header ref_local_function_body
     ;
 
 local_function_header
@@ -456,8 +456,12 @@ local_function_header
     ;
 
 local_function_modifier
-    : 'async'
-    | unsafe_modifier   // unsafe code support
+    : ref_local_function_modifier
+    | 'async'
+    ;
+
+ref_local_function_modifier
+    : unsafe_modifier   // unsafe code support
     ;
 
 local_function_body
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -1930,7 +1930,6 @@ method_header
 method_modifier
     : ref_method_modifier
     | 'async'
-    | unsafe_modifier   // unsafe code support
     ;
 
 ref_method_modifier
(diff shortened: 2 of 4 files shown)


jskeet (approved):
I think this works, but @Nigel-Ecma may well have a nicer way of representing it. It would be nice to avoid all the branching - if we had a rule of "return_type or ref_kind + ref_return_type" then we could remove some duplication. We'd need to explicitly prohibit async ref local functions (rather than doing it in the grammar) but this wouldn't be the first time we did that sort of thing.

standard/statements.md
```diff
@@ -456,8 +456,12 @@ local_function_header
     ;
 
 local_function_modifier
-    : 'async'
-    | unsafe_modifier   // unsafe code support
+    : ref_local_function_modifier
+    | 'async'
+    ;
+
+ref_local_function_modifier
```
BillWagner, line 463: I chose to represent it this way rather than duplicate the *unsafe_modifier* in both nodes to make it easier to update when `static` local functions are added.

KalleOlaviNiemitalo:
Will the grammar summary be updated automatically after this is merged?

Nigel-Ecma:
> I think this works, but @Nigel-Ecma may well have a nicer way of representing it. It would be nice to avoid all the branching - if we had a rule of "return_type or ref_kind + ref_return_type" then we could remove some duplication. We'd need to explicitly prohibit async ref local functions (rather than doing it in the grammar) but this wouldn't be the first time we did that sort of thing.

I'll take a look but not before the meeting…

Nigel-Ecma:
> Will the grammar summary be updated automatically after this is merged?

Yes :-)

MadsTorgersen (approved):

PR #961, opened 2023-10-02
Change T&D entry format

RexJaeschke:
This reformats each term entry (except for "character", which has a bullet sublist) to match what I submitted in the V7 Word version for publication. (This means that we won't need to hand-edit this list in future final conversions to Word.)

Bill, this doesn't need TG2 review.


jskeet (approved):

BillWagner (approved):
Thanks @RexJaeschke 

This LGTM, and I'll :shipit:

Issue #939, opened 2023-09-16
Clarify text on pattern exhaustiveness


BillWagner:
See https://github.com/dotnet/csharpstandard/pull/757#discussion_r1326615608

In 11.4, a bit of wordsmithing could improve the text.

Deferred until after C# 7

jskeet:
Fixed in #963.

PR #963, opened 2023-10-03
Wordsmith pattern exhaustiveness

Nigel-Ecma:
Addresses #939

### DIFF
--- a/standard/patterns.md
+++ b/standard/patterns.md
@@ -156,7 +156,7 @@ A set of patterns `Q` *subsumes* a pattern `P` if any of the following condition
 
 ## 11.4 Pattern exhaustiveness
 
-Informally, a set of patterns is exhaustive for a type if some pattern in the set is applicable to every possible value of that type other than null.
+Informally, a set of patterns is exhaustive for a type if for every possible value of that type other than null some pattern in the set is applicable.
 The following rules define when a set of patterns is *exhaustive* for a type:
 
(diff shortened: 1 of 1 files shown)


standard/patterns.md
```diff
@@ -156,7 +156,7 @@ A set of patterns `Q` *subsumes* a pattern `P` if any of the following condition
 
 ## 11.4 Pattern exhaustiveness
 
-Informally, a set of patterns is exhaustive for a type if some pattern in the set is applicable to every possible value of that type other than null.
+Informally, a set of patterns is exhaustive for a type if for every possible value of that type other than null some pattern in the set is applicable.
```
```suggestion
Informally, a set of patterns is exhaustive for a type if, for every possible value of that type other than null, some pattern in the set is applicable.
```

Personally I think the commas help the reader pause in the right place - they're not *needed*, but I think they help. See what you think.

jskeet (approved):
Generally happy - one suggestion.

RexJaeschke (approved):

standard/patterns.md
Nigel-Ecma: I consulted a statistically significant sample size of one ;-) and they agreed that they not needed but if adding any they would add your second one and not the first. I'll admit to just following the comma count in the original! Let's see if @RexJaeschke proposes a change.

BillWagner (approved):
LGTM

PR #964, opened 2023-10-07
Fix editorial nits

RexJaeschke:
Bill, these are editorial fixes I noticed when reviewing the Word version of V7:

- Change all "must" in normative text to "shall," or rewrite, as needed.
- Change "Standard" to "specification."
- Plus a few minor nits.


standard/expressions.md
```diff
@@ -6610,7 +6610,7 @@ constant_expression
     ;
 ```
 
-A constant expression may be either a value type or a reference type. If a constant expression is a value type, it must be one of the following types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type. If a constant expression is a reference type, it must be the `string` type, a default value expression ([§12.8.20](expressions.md#12820-default-value-expressions)) for some reference type, or the value of the expression must be `null`.
+A constant expression may have either a value type or a reference type. If a constant expression has a value type, it shall be one of the following types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type. If a constant expression has a reference type, it shall be the `string` type, a default value expression ([§12.8.20](expressions.md#12820-default-value-expressions)) for some reference type, or the value of the expression shall be `null`.
```
jskeet: This one feels a little odd to me. I think I'd prefer "The type of a constant expression may be a value type or a reference type. If the type is a value type [...]"

standard/lexical-structure.md
```diff
@@ -124,7 +124,7 @@ If the following token is among this list, or an identifier in such a context, t
 >
 > *end example*
 
-A *relational_expression* ([§12.12.1](expressions.md#12121-general)) can have the form “*relational_expression* `is` *type*” or “*relational_expression* `is` *constant_pattern*,” either of which might be a valid parse of a qualified identifier. In this case, an attempt is made to bind it as a type ([§7.8.1](basic-concepts.md#781-general)); however, if that fails, it is bound as an expression, and the result must be a constant.
+A *relational_expression* ([§12.12.1](expressions.md#12121-general)) can have the form “*relational_expression* `is` *type*” or “*relational_expression* `is` *constant_pattern*,” either of which might be a valid parse of a qualified identifier. In this case, an attempt is made to bind it as a type ([§7.8.1](basic-concepts.md#781-general)); however, if that fails, it is bound as an expression, and the result is a constant.
```
jskeet: "the result is a constant" sounds like a classification, whereas we're stating a constraint on the code. (I think this part of me finding "shall be" generally more confusing than "must", so we may be able to ignore this as just a bit of Jon confusion, but I'd be interested to hear other views...)

jskeet (commented):
Mostly fine, just a couple of points to consider.

standard/expressions.md
Nigel-Ecma: Good suggestion for both the 2nd & 3rd sentences.

standard/lexical-structure.md
Nigel-Ecma: The language is wrong, e.g. “an attempt is made” is an operational description of the implementation of a specification rather than a specification. What it should say is that if both forms are valid (syntactically & semantically) parses of the input, i.e. there is an ambiguity, then the former (`is` *type*) shall be selected. There are similar situations elsewhere and the wording used might be suitable here, but no wordsmithing from me tonight – its too late!

BillWagner: Here's my first attempt at a wordsmithing.  Rex may have better ideas:

```suggestion
A *relational_expression* ([§12.12.1](expressions.md#12121-general)) can have the form “*relational_expression* `is` *type*” or “*relational_expression* `is` *constant_pattern*,” either of which might be a valid parse of a qualified identifier. When both forms result in correct code, the binding to a type ([§7.8.1](basic-concepts.md#781-general)) is preferred. If that fails, it is bound as an expression, and the result is a constant.
```

standard/expressions.md
```diff
@@ -1475,7 +1475,7 @@ In an *interpolation_minimum_width* the *constant_expression* shall have an impl
...
+The overall meaning of an *interpolated_string_expression*, including the above formatting and padding of interpolations, is defined by a conversion of the expression to a method invocation: if the type of the expression is `System.IFormattable` or `System.FormattableString` that method is `System.Runtime.CompilerServices.FormattableStringFactory.Create` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)) which returns a value of type `System.FormattableString`; otherwise the shall must be `string` and the method is `string.Format` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) which returns a value of type `string`.
```
BillWagner: typo in the replacement (original replaced "type" instead of "must"):

```suggestion
The overall meaning of an *interpolated_string_expression*, including the above formatting and padding of interpolations, is defined by a conversion of the expression to a method invocation: if the type of the expression is `System.IFormattable` or `System.FormattableString` that method is `System.Runtime.CompilerServices.FormattableStringFactory.Create` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)) which returns a value of type `System.FormattableString`; otherwise the type shall be `string` and the method is `string.Format` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)) which returns a value of type `string`.
```

standard/expressions.md
BillWagner: wordsmithing as a suggestion:

```suggestion
The type of a constant expression may be either a value type or a reference type. If a constant expression has a value type, it shall be one of the following types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type. If a constant expression has a reference type, it shall be the `string` type, a default value expression ([§12.8.20](expressions.md#12820-default-value-expressions)) for some reference type, or the value of the expression shall be `null`.
```

BillWagner (changes requested):
Hi @RexJaeschke 

I agree with Jon's and Nigel's comments, and offered suggestions.

One we resolve those, this is ready.

standard/lexical-structure.md
jskeet: Updated "bindint" to "binding" in the suggestion.

jskeet: I'm not sure about "correct code". It's more *valid* than *correct*, but I'm not sure whether that would be right either. Legal code? Or maybe talk about name lookups succeeding?

standard/expressions.md
Nigel-Ecma: The original text hails from the time when there was a null type, so “A constant expression may have either a value or a reference type…” made sense, but in modern C# `null` has no type… Also there is not really any need to call out value & reference types *per se* as a list of the valid types is provided.

This leads me to:

```suggestion
A constant expression shall either have the value `null` or one of the following types:
- `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool`, `string`;
- an enumeration type; or
- a default value expression ([§12.8.20](expressions.md#12820-default-value-expressions)) for a reference type.
```

standard/lexical-structure.md
Nigel-Ecma: First, I think this should probably be in $12.12.1 and not here – or in both, I've added #965 to address that and will suggest changes here for now.

It is not unusual for alternatives to overlap in language grammars, sometimes just because it produces a shorter/neater grammar, sometimes as some semantic information is required to resolve the syntactic choice, etc. For example in $12.19.1 we have:

> When recognising an *anonymous_function_body* if both the *null_conditional_invocation_expression* and *expression* alternatives are applicable then the former shall be chosen.

We can resolve the issue here by following the same pattern:

```suggestion
When recognising a *relational_expression* ([§12.12.1](expressions.md#12121-general)) if both the “*relational_expression* `is` *type*” and “*relational_expression* `is` *constant_pattern*” alternatives are applicable, and *type* resolves to an accessible type, then the “*relational_expression* `is` *type*” alternative shall be chosen.
```

I would refrain from explaining the overlap between the two alternatives further, the grammar does that, but if that is desired it should be added as a *Note*.

The bit “and the result is a constant” in the original is both confusing (is it referring to the *constant_pattern* or the result of the `is`?) and unneeded – any constness comes from the semantics of *constant_pattern* and not the `is` operator whose result is not a constant in general!)

Nigel-Ecma (changes requested):
I've added comments as responses in conversations. Some fixup required in issues this PR brought to light. (Note: I've only reviewed the conversations.)

standard/expressions.md
Nigel-Ecma: I think the revised text misses that a constant expression may have no type, i.e. when its value is `null`.

BillWagner (approved):

standard/expressions.md
```diff
@@ -6610,7 +6610,10 @@ constant_expression
     ;
 ```
 
-A constant expression may be either a value type or a reference type. If a constant expression is a value type, it must be one of the following types: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type. If a constant expression is a reference type, it must be the `string` type, a default value expression ([§12.8.20](expressions.md#12820-default-value-expressions)) for some reference type, or the value of the expression must be `null`.
+A constant expression shall either have the value `null` or one of the following types:
+- `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool`, `string`;
```
BillWagner, line 6615: Add blank line:

```suggestion

- `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool`, `string`;
```

jskeet (approved):

Nigel-Ecma (approved):

PR #969, opened 2023-10-25
Add additional example annotation support files

RexJaeschke:
Bill, when looking at example testing failures on the V8 Draft PRs, I discovered that a number of tests failed because supporting files had not been uploaded. This PR adds all such supporting files. Once you have merged this, I'll resume looking for any other testing failures on those Draft PRs.

Looking at the diffs below, it seems that 2 of the 10 files already exist. But as their proposed replacements appear to be identical, I presume that leaving them in this batch is OK.


BillWagner (approved):
This LGTM @RexJaeschke 

I'll :shipit: now.

Issue #933, opened 2023-09-15
[Tools] Single source of truth for clauses in the standard


BillWagner:
See https://github.com/dotnet/csharpstandard/pull/757#issuecomment-1476571863

We have multiple tools that process the entire standard. Some use the a list of files in a shell script, others use a clauses.json file. The word converter did other processing from the existing TOC.

It will save us time to have one source of truth for the order of various files that incorporate the standard.

I recommend setting doing this early in the C# 8 timeframe.

---
[Associated WorkItem - 161769](https://dev.azure.com/msft-skilling/Content/_workitems/edit/161769)

PR #971, opened 2023-10-26
Create word artifact when a PR is merged to the main branch

### LINKED ISSUE #745: Update scripts for deprecated commands
See https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/#examples

`::save-state` and `::set-output` will be deprecated in May. Our grammar validator script uses that. There may be others.

The link above shows the suggested replacement.

---
[Associated WorkItem - 161768](https://dev.azure.com/msft-skilling/Content/_workitems/edit/161768)

### LINKED ISSUE #953: Create word doc artifact when default branch updates
The word-converter runs on each PR commit, and when the *default* branch gets updated.

Ideally, we can use the [upload-artifact](https://github.com/actions/upload-artifact) action to save the output instead of running the converter by hand when needed.

Ideally, we can set a short lifetime for the artifact on PRs and commits. We can set a longer lifetime for when the default branch is updated. 

Even on the default branch, I don't think the lifetime should be infinite, as the ECMA repository is the correct location for downloading the approved standard. This functionality is for the committee to produce the doc for ECMA.

---
[Associated WorkItem - 161767](https://dev.azure.com/msft-skilling/Content/_workitems/edit/161767)

BillWagner:
Update the workflow for "update on merge" to also run the word converter, and upload the generated *standard.docx* file to the actions output.

While doing this work, remove the outdated "::set-output" command. And, remove all but one of the hardcoded lists of spec files.

For an example output for uploading the action artifact, see [here](https://github.com/BillWagner/csharpstandard/actions/runs/6655931948) for a test run.

Fixes #745 
Fixes #953 

Contributes to #933

### DIFF
--- a/tools/run-converter.sh
+++ b/tools/run-converter.sh
@@ -21,3 +21,13 @@ dotnet run --project $CONVERTER_PROJECT -- \
   $SPEC_DIRECTORY/*.md $TEMPLATE \
   -o $OUTPUT_DIRECTORY/standard.docx
 
+if [ "$?" -eq "0" ]
+then
+    # Success: Write key/value for GitHub action to read:
+    echo "status=success" >> $GITHUB_OUTPUT 
+else
+    # Failed: report the error to the GitHub action:
+    echo "status=failed" >> $GITHUB_OUTPUT 
+fi
+
+
--- a/tools/run-converter.sh
+++ b/tools/run-converter.sh
@@ -29,5 +29,3 @@ else
     # Failed: report the error to the GitHub action:
     echo "status=failed" >> $GITHUB_OUTPUT 
 fi
-
-
--- a/.github/workflows/update-on-merge.yaml
+++ b/.github/workflows/update-on-merge.yaml
@@ -68,6 +68,7 @@ jobs:
         body: 'renumber sections. Add grammar'
 
     - name: Run converter
+      id: run-converter
       run: |
         cd tools
         ./run-converter.sh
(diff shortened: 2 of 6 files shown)


tools/run-converter.sh
```diff
@@ -21,3 +21,11 @@ dotnet run --project $CONVERTER_PROJECT -- \
   $SPEC_DIRECTORY/*.md $TEMPLATE \
   -o $OUTPUT_DIRECTORY/standard.docx
 
+if [ "$?" -eq "0" ]
```
jskeet, line 24: Possibly just check whether GITHUB_OUTPUT is non-empty?

jskeet (approved):
Generally fine - one small suggestion.

tools/run-converter.sh
BillWagner: I was following the guidance here: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files

In all there examples, and the guidance uses `>>` and doesn't check for a non-empty environment variable. 

I agree to keep watching this, but I'll merge as is for now.

jskeet: > In all there examples, and the guidance uses >> and doesn't check for a non-empty environment variable.

I suspect that was because those examples are for code which is only expected to run as part of a GitHub action.

I run this shell script locally when testing the converter - what do we expect it to do in that case? (I have no idea what `>>` does if you don't tell it what to write to...)

BillWagner: Gotcha.  

I ran it locally, and that command fails with "ambiguous redirect".

I'll make a fresh PR that only sets those if the environment is set. (That's in multiple actions).

PR #973, opened 2023-10-26
test for GITHUB_OUTPUT

BillWagner:
When running locally, the environment variables set by GH actions aren't set. Check for their existence before writing to them.

### DIFF
--- a/tools/run-converter.sh
+++ b/tools/run-converter.sh
@@ -21,14 +21,7 @@ dotnet run --project $CONVERTER_PROJECT -- \
   $SPEC_DIRECTORY/*.md $TEMPLATE \
   -o $OUTPUT_DIRECTORY/standard.docx
 
-if [ -n "$GITHUB_OUTPUT" ]
-then 
-  if [ "$?" -eq "0" ]
-  then
-      # Success: Write key/value for GitHub action to read:
-      echo "status=success" >> $GITHUB_OUTPUT 
-  else
-      # Failed: report the error to the GitHub action:
-      echo "status=failed" >> $GITHUB_OUTPUT 
-  fi
+if [ -n "$GITHUB_OUTPUT" ] && [ "$?" -eq "0" ]
+then
+    echo "status=success" >> $GITHUB_OUTPUT 
 fi
--- a/tools/run-converter.sh
+++ b/tools/run-converter.sh
@@ -32,5 +32,10 @@ then
       echo "status=failed" >> $GITHUB_OUTPUT 
   fi
 else
+  if [ "$?" -eq "0" ]
+  then
       echo "Success generating word doc for standard."
+  else
+      echo "Failed generating word doc for standard."
+  fi
 fi
--- a/tools/run-converter.sh
+++ b/tools/run-converter.sh
@@ -21,7 +21,7 @@ dotnet run --project $CONVERTER_PROJECT -- \
   $SPEC_DIRECTORY/*.md $TEMPLATE \
   -o $OUTPUT_DIRECTORY/standard.docx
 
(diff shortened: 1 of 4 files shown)


tools/run-converter.sh
```diff
@@ -32,5 +32,10 @@ then
       echo "status=failed" >> $GITHUB_OUTPUT 
   fi
 else
+  if [ "$?" -eq "0" ]
```
jskeet: Not sure whether we actually need to do anything here - I think the console output is generally fairly obvious. (Not sure if that's true of the section renumbering tool.)

jskeet (commented):
I think the use of `set -e` means we'll never actually get into the "else" parts (so we'll never echo "status=failed" I suspect). It's probably worth either removing the `set -e`, or remove the "else" parts and trust that GitHub will note the error code. Personally I'd prefer the latter - I'm not keen on a script continuing on if part of it has already failed. (I should have spotted that aspect in the previous PR...)

tools/run-converter.sh
BillWagner: Likely true. I was using it to test and make sure the "else" branch was hit. I'll clean those up.

I think all are unnecessary.

tools/run-converter.sh
```diff
@@ -21,11 +21,7 @@ dotnet run --project $CONVERTER_PROJECT -- \
   $SPEC_DIRECTORY/*.md $TEMPLATE \
   -o $OUTPUT_DIRECTORY/standard.docx
 
-if [ "$?" -eq "0" ]
+if [ -n "$GITHUB_OUTPUT" ] && [ "$?" -eq "0" ]
```
jskeet: I don't think we'll ever see a non zero exit code due to set -e. But if we ever did, then when not running on GitHub this would fail because it would go into the else.

If I'm not making sense, I can raise a PR tomorrow - I'm currently only on a mobile and shouldn't be working...

BillWagner: Thanks Jon, or just add a suggestion when you have the time. I think we're just talking past each other and I'm missing your point.

BillWagner: OK, did a bit of reading. I think we should do one of two things. I prefer the former, but you do more shell programming than I do:

1. instead of `set -e`, use `trap ReportError ERR` instead. Then, the `ReportError` function can write "failed" to GITHUB_OUTPUT. I like this one better because our actions will reliably stop on the first error.
2. Remove the second check, as you point out. It will never be hit. This works *because* the YML files for our actions check for a success code, rather than a failure code. It feels a bit more brittle, but not by much.

What do you think?

jskeet: I prefer option 2, personally. I *believe* that GitHub actions will fail (and immediately stop) automatically if they run a command that fails - as if we'd written "failed" to GITHUB_OUTPUT.

To put it another way: I've never had to explicitly write "failed" to GITHUB_OUTPUT in any other context, and I've seen plenty of actions fail and stop appropriately :)

Option 2 is basically more general IMO - it means our scripts just work in a "normal" way without being aware of GitHub at all.

BillWagner:
Thanks @jskeet 

I updated based on our discussion. It should be ready to go.

tools/update-grammar-annex.sh
```diff
@@ -49,4 +49,8 @@ echo Insert EOF Stuff
 cat $GRAMMAR_PROJECT/grammar-eof-insert.md >>$OUTPUT_FILE
 
 # I think always success, but echo the success output anyway:
-echo "status=success" >> $GITHUB_OUTPUT 
\ No newline at end of file
+# I think always success, but echo the success output anyway:
```
jskeet: Remove duplicate line.

jskeet (approved):
One tiny nit, but otherwise yay :)

Issue #787, opened 2023-05-09
Possible typo in type inference


jskeet:
From https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#12632-the-first-phase, after an introduction which sets the scene with:

```csharp
Tₑ M<X₁...Xᵥ>(T₁ p₁ ... Tₓ pₓ)
```

we have this text:

> For each of the method arguments `Eᵢ`:
> 
> - If `Eᵢ` is an anonymous function, an *explicit parameter type inference* ([§12.6.3.8](expressions.md#12638-explicit-parameter-type-inferences)) is made *from* `Eᵢ` *to* `Tᵢ`
> - Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is a value parameter ([§15.6.2.2](classes.md#15622-value-parameters)) then a *lower-bound inference* ([§12.6.3.10](expressions.md#126310-lower-bound-inferences)) is made *from* `U` *to* `Tᵢ`.
 >- Otherwise, if `Eᵢ` has a type `U` and `xᵢ` is a reference ([§15.6.2.3](classes.md#15623-reference-parameters)) or output ([§15.6.2.4](classes.md#15624-output-parameters)) parameter then an *exact inference* ([§12.6.3.9](expressions.md#12639-exact-inferences)) is made *from* `U` *to* `Tᵢ`.
> - Otherwise, no inference is made for this argument.

Should the `xᵢ` in the second and third bullets be `pᵢ`?

I'll probably just create a PR for this, but it was easier to file an issue to start with...

Nigel-Ecma:
Well a parameter has the grammar:

>```
>attributes? parameter_modifier? type identifier default_argument?
>```

Matching `Tₑ M<X₁...Xᵥ>(T₁ p₁ ... Tₓ pₓ)` to this the `Tᵢ` are the *type* and `pᵢ` the *identifier*. Neither of these determine whether the parameter is value or reference parameter, that is down to the *parameter_modifier*. So it seems the writer is using `xᵢ` as the whole parameter spec but does not define it as such, so the `xᵢ` are neither `pᵢ` or `Tᵢ`…

We could try:

[...]

PR #975, opened 2023-10-30
Address typo in type inference

Nigel-Ecma:
Fix for issue #787

The PR uses “corresponding parameter” rather than the “ith parameter" from #787. I think this is better, avoids HTML tags, and matches the use of corresponding parameter elsewhere; but either choice is valid.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1,4 +1,4 @@
-# 12 Expressions
+/# 12 Expressions
 
 ## 12.1 General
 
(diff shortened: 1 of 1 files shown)


jskeet (commented):
I suspect it *has* to be corresponding parameter, and that we should change the "Ti" part to "the type of the corresponding parameter" or similar.

Otherwise, using named arguments in a way such that the order of the arguments doesn't match the order of the parameters breaks things... unless there's a notional reordering earlier on in the proces, of course.

BillWagner (approved):
This LGTM

Nigel-Ecma:
> I suspect it _has_ to be corresponding parameter, and that we should change the "Ti" part to "the type of the corresponding parameter" or similar.
> 
> Otherwise, using named arguments in a way such that the order of the arguments doesn't match the order of the parameters breaks things... unless there's a notional reordering earlier on in the proces, of course.

You were not meant to notice that ;-) I too realised it hasn't been updated for named arguments but choose just to fix the typo you found. We can always address named arguments later, or do you want to do this now?

jskeet:
Let's discuss in the meeting - if someone has a suggestion for a quick way to do it properly, let's do that; otherwise, this is better than it was before.

jskeet (approved):
Merging as better than what we've got.

PR #977, opened 2023-10-30
Combine merge tools

### LINKED ISSUE #933: [Tools] Single source of truth for clauses in the standard

BillWagner:
This PR combines the section renumbering tool with the grammar extraction tool. 

The grammar extractor runs after the section renumbering tool. Both tools use the same JSON input to determine the input files.

As part of this PR, update the grammar extraction algorithm to read the existing text from the grammar.md source, discarding the existing grammar productions. That way, any future updates made to the MD source will be automatically picked up by the tool.

This also removes the grammar extractor source from the repo and updates the workflows to run the renumbering tool, but not the grammar extractor when a PR is merged.

Fixes #933

This may be easier to review commit by commit.


tools/StandardAnchorTags/Properties/launchSettings.json
```diff
@@ -0,0 +1,9 @@
+{
```
jskeet: I suspect you didn't mean to add this file.

standard/grammar.md
```diff
@@ -10,15 +10,15 @@ This annex contains the grammar productions found in the specification, includin
 
 ```ANTLR
 
-// Source: §6.3.1 General
+// Source: [§6.3.1](lexical-structure.md#631-general) General
```
jskeet: Why are these links now here? I can't see a code change that would cause them... and I don't think we should have them, given that they don't render as links anyway.

tools/StandardAnchorTags/GenerateGrammar.cs
```diff
@@ -0,0 +1,112 @@
+﻿using System;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace StandardAnchorTags
```
jskeet: As we're adding a new file, maybe let's start with a file-scoped namespace? (I'm generally reluctant to retrofit as it messes up git blame, but I usually use it for new code.)

tools/StandardAnchorTags/GenerateGrammar.cs
```diff
@@ -0,0 +1,112 @@
...
+                    {
+                        throw new InvalidOperationException("Unexpected EOF; no closing grammar fence");
+                    }
+                    if (nextLine.Length < 3)   // Is it long enough to contain a closing fence?
+                    {
+                        await grammarStream.WriteLineAsync(nextLine);
+                    }
+                    else if (nextLine.Substring(0, 3) == "```")    // If line starts with ```
+                    {
+                        //			Console.WriteLine("------ End of a production");
```
jskeet: Remove comment?

jskeet (approved):
Looks good - a few comments, but generally seems like a very useful change :)

tools/StandardAnchorTags/GenerateGrammar.cs
BillWagner: Already removed in a more recent commit.

tools/StandardAnchorTags/GenerateGrammar.cs
BillWagner: I also plan to make an update once .NET 8 ships to make one PR that updates the code to C# 12. (List patterns in particular would be useful.)

standard/grammar.md
BillWagner: My first commit was broke. I'll run again and check it in. That will remove the links. The entire process will fail when the grammar extractor fails, it won't continue again.

(The reason it happened is that section renumber tool ran, but the grammar extractor failed. So, the renumbering tool updated all the text to be links, but the grammar extractor didn't remove them.)

PR #681, opened 2022-12-05
Adding support for default interface function members

RexJaeschke:
Re the [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/default-interface-methods.md): The main challenge when reading it was the (misleading) occurrence of the modifier `override` in numerous examples and narrative. As best as I can tell, this modifier is **not** actually permitted on any interface member. Instead, overriding is achieved via explicit implementation.

There is [one open question](https://github.com/dotnet/docs/issues/32871), which has to do with the topic "Base Interface Invocations," for which the decision "Decided on `base(N.I1<T>).M(s)`" was made. However, as far as I can tell, support for this was **not** added to the V8 compiler. Can someone please confirm (or refute) that.

BTW, I think the proposal is misnamed: it really applies to all interface function members (properties, indexers, and events as well), not just methods. (it also allows nested types.)

Note: No spec changes are needed to support entry point `Main` in an interface. *Any non-generic type* containing a `Main` method with `void`/`int` return and no/one `string[]` parameter list is already permitted, which not only allows `Main` in a struct or class, but now also in an interface.

As of V7, the descriptions of the various member kinds has been located in classes.md, with augmenting and/or overriding text in structs.md, with pointers from structs.md back into classes.md, and that continues. However, now that many members kinds can have implementations in interfaces as well, I've added an intro para to the start of most member sections in classes.md, which contains forward pointers to that member kind's occurrence in structs.md and interfaces.md, as appropriate.

Fixes #982 (which reinstates `abstract override`.)

PR #978, opened 2023-10-31
Add support for init accessors

RexJaeschke:
Notes:

1. New section "§init-accessors Init accessors" contains the place-holder link `([§xxx](plug in here link to v9 records grammar))`, which will need to be replaced by the corresponding grammar section number that results from the addition of the records feature (also in V9).


standard/classes.md
```diff
@@ -3194,14 +3198,16 @@ The differences between static and instance members are discussed further in [§
...
+*Note*: This subclause and its sibling subclauses apply to both properties ([§15.7](classes.md#157-properties)) and indexers ([§15.9](classes.md#159-indexers)). The clause is written in terms of properties, when reading for indexers substitute indexer/indexers for property/properties and consult the list of differences between properties and indexers given in [§15.9.2](classes.md#1592-indexer-and-property-differences). *end note*
 
 The *accessor_declarations* of a property specify the executable statements associated with writing and/or reading that property.
 
 ```ANTLR
 accessor_declarations
-    : get_accessor_declaration set_accessor_declaration?
-    | set_accessor_declaration get_accessor_declaration?
+    : get_accessor_declaration ( set_accessor_declaration | init_accessor_declaration)?
+    | ( set_accessor_declaration | init_accessor_declaration) get_accessor_declaration?
```
KalleOlaviNiemitalo: Asymmetric spacing within parentheses.

standard/classes.md
```diff
@@ -3548,11 +3592,197 @@ Properties can be used to delay initialization of a resource until the moment it
...
+>     }
+> }
+> ```
+>
+> *end example*
+
+At the point an init accessor is invoked, the instance is known to be in the construction phase. Hence an init accessor may take the following actions in addition to what a set accessor can do:
+
+1. Call other init accessors available through `this` or `base`
+1. Assign `readonly` fields declared on the same type through `this`
```
KalleOlaviNiemitalo, line 3846: Apparently, it can also make a writable reference to a `readonly` field, but I'm not sure whether that is a C# 9 feature.  However, unlike a constructor, it cannot assign to a `get`-only property.  [SharpLab](https://sharplab.io/#v2:C4LghgzgtgPgAgZgARwExIMJIN4FgBQSRSADgE4CWAbmMAKZJl1gAmA9gHYA2AnkhR2BIAHgG4CxJBOKJ+gpAE0cSAOZ1gopAF8C0orIFCAWjj2TV604XM2A9LaR0yZNmUwBlAAwBGAJyoQJABBRmZ2bj4AMwo6LhYkAGMwDg42IQAjBgBXCDp4yCQwUMikV1KsoRouLIYACjphBLoSIQFCxM4IYDIshOBXAEozGyImEsNGJABeYqRgAAsKCAA6MWGRuAB2JE9xaxsdffMBCiE8I5GkeyQAeQBpABokFjoIElOGJlZOXnWbMbkQjcMwBCyWqz2l0kf3M1ycLjcGC8qE8nkCAAUXCQnMA+GUBC9hE4kAByDDLBQkxLJVIZBiQCAUFQcPJzNhIAC0HP4rQgoVYpQiMMkYJWShmuykF2Ih0khy0QA==):

```csharp
public class C {
    private readonly int x;
    
    public int Y { get; }

    public int Z {
        get {
            // error CS0192: A readonly field cannot be used as a ref or out value (except in a constructor)
            ref int r = ref this.x;
            return 0;
        }
        init {
            // OK, despite readonly
            ref int r = ref this.x;
            
            // error CS0200: Property or indexer 'C.Y' cannot be assigned to -- it is read only
            this.Y = 0; 
        }
    }
}
```

KalleOlaviNiemitalo (commented):
Should this say something about referencing an `init` accessor in an expression tree (System.Linq.Expressions)?

standard/classes.md
```diff
@@ -3294,13 +3304,19 @@ A get accessor for a ref-valued property corresponds to a parameterless method w
...
+
+An init accessor corresponds to a method with a single value parameter of the property type and a `void` return type. The implicit parameter of an init accessor is always named `value`. Only during the construction phase of an object, and if an init accessor exists, when a property is referenced as the target of an assignment ([§12.21](expressions.md#1221-assignment-operators)), or as the operand of `++` or `–-` ([§12.8.15](expressions.md#12815-postfix-increment-and-decrement-operators), [§12.9.6](expressions.md#1296-prefix-increment-and-decrement-operators)), the init accessor is invoked with an argument that provides the new value ([§12.21.2](expressions.md#12212-simple-assignment)). The body of an init accessor shall conform to the rules for `void` methods described in [§15.6.11](classes.md#15611-method-body). In particular, return statements in the init accessor body are not permitted to specify an expression. Since an init accessor implicitly has a parameter named `value`, it is a compile-time error for a local variable or constant declaration in an init accessor to have that name.
```
KalleOlaviNiemitalo: In each file, please link from the first use of "the construction phase of an object" to the section where this term is defined.

standard/classes.md
```diff
@@ -3294,13 +3304,19 @@ A get accessor for a ref-valued property corresponds to a parameterless method w
...
 
-Based on the presence or absence of the get and set accessors, a property is classified as follows:
+Based on the presence or absence of get, set, and init accessors, a property is classified as follows:
 
 - A property that includes both a get accessor and a set accessor is said to be a ***read-write property***.
+- A property that includes both a get accessor and an init accessor is said to be a ***read-init property***. It is a compile-time error for a read-init property to be the target of an assignment except during the construction phase of an object.
 - A property that has only a get accessor is said to be a ***read-only property***. It is a compile-time error for a read-only property to be the target of an assignment.
 - A property that has only a set accessor is said to be a ***write-only property***. Except as the target of an assignment, it is a compile-time error to reference a write-only property in an expression.
+- A property that has only an init accessor is said to be an ***init-only property***. Except as the target of an assignment during the construction phase of an object, it is a compile-time error to reference an init-only property in an expression.
```
KalleOlaviNiemitalo, line 3502: Does `nameof` count as referencing a property in an expression?

KalleOlaviNiemitalo:
Expression tree seems to be working as expected.  [SharpLab](https://sharplab.io/#v2:C4LgTgrgdgPgAgJgIwFgBQcAMACOSB0AMgJZQCO+AogB4AOYApgM5PED2UTA3OunAMy4E2AMLYA3umzTcg0sGwBZCdgDmDYF2ylim7AF90htH0F4AbEOwARCVJmIE96ZLQz32GvWasOAHjgAVj8RAD5Q7AY6RhZ2KCRsAF5sAAoASiSIqAYAd1EVZWSE/S0AelLsAHkAaWcPcvKvGN8oAKQEEPDI6J844WSUgGMMxIjxbEH8QuwELRLscsiwMDYwUQBlAA5NwKc3D2kxQaTu71iOAhE2AFtaYgAbBnT8AEkoADc2AGsntJ59g54ACcQymfwWFTYEGAtGhTGwqABBiM6CAA==):

```csharp
using System;

public class C {
    public int M { get; init; }
}

public static class D {
    public static void Main()
    {
        Expression<Func<C>> expression1 = () => new C { M = 1 }; // OK
        ////Expression<Action<C>> expression2 = (c) => { c.M = 2; }; // error CS8852
        C c = expression1.Compile().Invoke();
        Console.WriteLine(c.M); // outputs 1
    }
}
```

KalleOlaviNiemitalo:
Please add `init` to the *interface_accessors* grammar rule.  Although an interface cannot be named in a `new` expression, the `init` accessor can be called via a type parameter constrained to that interface.  [SharpLab](https://sharplab.io/#v2:C4LghgzgtgPgAgZgAQEsB2wCmAnAZmAY0yQEkkBvAWACgk7UMkAFCpAc02AG4GVukAvjXoNgSYAAsUEANrox6ACaYAHgF1WHfuj48h1fTURI4AJiQBhJCFIVh9Y/OabOPHf30jHjSdLmMlVQ1ydk4kAF4APiQABjc0PgoBQRpDajgARgA2E3NMgHYaKloHbJMAFiQAWQAeABVIgAoASiQAdwkcYjrrUgAaJDRMNpa7EpE6HrFwweGkOtGQlhmMgZkAVg0ZnIEuezpPen2TMrhKqrB0FuPiifpaiybmvfGUgxoaIA):

```csharp
public interface I {
    int P { get; init; }
    int this[int index] { get; init; }
}

public class C : I {
    public int P { get; init; }
    public int this[int index] { get => 0; init {} }
}

static class Program
{
    static void M<T>() where T : I, new() {
        T t = new T() { P = 1, [5] = 6 };
    }
    
    static void Main()
    {
        M<C>();
    }
}
```

standard/classes.md
RexJaeschke: Fixed.

standard/classes.md
RexJaeschke: Good idea; fixed.

RexJaeschke:
Re @KalleOlaviNiemitalo's comment https://github.com/dotnet/csharpstandard/pull/978#issuecomment-1788486562:

This suggests the following changes to the grammar:

```ANTLR
interface_accessors
    : attributes? 'get' ';'
    | attributes? 'set' ';'
    | attributes? 'init' ';'
    | attributes? 'get' ';' attributes? ('set'|'init') ';'
    | attributes? ('set'|'init') ';' attributes? 'get' ';'
    ;
```

However, that will be superseded by changes to this same grammar by V8's PR #681, which adds support for default interface function members. We should revisit this once that PR has been merged.

standard/classes.md
RexJaeschke: I cloned the new para from the one preceding it (re a write-only property).

VS doesn't complain about `nameof(P)` where `P` is an init-only (or write-only) property.

According to [§12.8.23](expressions.md#12823-the-nameof-operator), "Nameof expressions,"

> A *nameof_expression* is a constant expression of type `string`, and has no effect at runtime. Specifically, its *named_entity* is not evaluated, and is ignored for the purposes of definite assignment analysis ([§9.4.4.22](variables.md#94422-general-rules-for-simple-expressions)).

This suggests that such a usage of the property name is not "a reference to a property."

standard/classes.md
```diff
@@ -3694,11 +3740,197 @@ Properties can be used to delay initialization of a resource until the moment it
 >
 > *end example*
 
+#### §init-accessors Init accessors
+
+An instance property containing an *init_accessor_declaration* is considered settable during the construction phase of the object, except when in a local function or lambda. The ***construction phase of an object*** includes the following:
+
+- During execution of an *object_initializer* ([§12.8.17.2.2](expressions.md#1281722-object-initializers))
```
KalleOlaviNiemitalo: This should exclude a *nested object initializer*, like `{ I = 1 }` in the following.

```csharp
class C {
    public int I { get; init; }
}

class D {
    public D(C c) { C = c; }
    public C C { get; }
}

class E {
    static void M() {
        new D(new C()) {
            C = { I = 1 } // error CS8852
        };
    }
}
```

standard/classes.md
```diff
@@ -3694,11 +3740,197 @@ Properties can be used to delay initialization of a resource until the moment it
 >
 > *end example*
 
+#### §init-accessors Init accessors
+
+An instance property containing an *init_accessor_declaration* is considered settable during the construction phase of the object, except when in a local function or lambda. The ***construction phase of an object*** includes the following:
```
KalleOlaviNiemitalo, line 3779: "except when in a local function or lambda" can be understood as incorrectly disallowing the following.

```csharp
class C {
    public int P { get; init; }
}

class D {
    System.Func<C> Method() {
        C Local() {
            return new C { P = 1 };
        }
        return Local;
    }
}
```

standard/classes.md
```diff
@@ -3694,11 +3740,197 @@ Properties can be used to delay initialization of a resource until the moment it
...
+When a property is specified as an automatically implemented property, a hidden backing field is automatically available for the property, and the accessors are implemented to read from and write to that backing field. The hidden backing field is inaccessible, it can be read and written only through the automatically implemented property accessors, even within the containing type. If the auto-property has no set or init accessor, the backing field is considered `readonly` ([§15.5.3](classes.md#1553-readonly-fields)). Just like a `readonly` field, a read-only auto-property may also be assigned to in the body of a constructor of the enclosing class. Such an assignment assigns directly to the read-only backing field of the property. If the auto-property has an init accessor, the backing field may be assigned to during the construction phase of an object (§init-accessors).
```
KalleOlaviNiemitalo, line 3967: The backing field should be considered `readonly` also when the auto-property has an init accessor.  Otherwise, this would lead to a `readonly struct` having a non-`readonly` field.

```csharp
readonly struct S {
    int I { get; init; }
}
```

BillWagner:
@jskeet @RexJaeschke 

This is the first C# 9 feature PR that I targeted for the updated V9 plan. Note that the base branch is now `v9-init-accessors`, a feature branch I created based on the updated `draft-v9`.

Note that there are failures at the moment:

- Grammar validation fails because the `init` keyword isn't expected.
- Test fixes are in #1451, enabling C# 9 in our test templates.
- Renumbering is because this feature depends on primary constructors in records.
- Word converter fails because of new sections and anchors.

Before moving forward in either of the two directions, I'd like @RexJaeschke to give it a look and see if I missed anything.

I see two procedural ways to move forward (with options):

- I can merge the PR into the feature branch. 
- I can push the changes directly to the feature branch, and add a comment pointing to this PR to preserve links to the existing comments.

Is there a preference?

In both plans, I could either squash the commits into one, or keep separate. 

In either case, I'd open a fresh PR from `v9-init-accessors` to `draft-v9`. That can stay in draft mode. That's where I'd add pointers back to this PR.

My first thought is to push a single squashed commit into the feature branch. Open a new PR, and point back to this one with a summary of the comments.

Thoughts?

Nigel-Ecma:
Hi @BillWagner

> - Grammar validation fails because the `init` keyword isn't expected.

I was curious so I took a peek. A failure is a failure so this doesn’t make any real difference to you, but the error isn’t down to the `init` keyword *per se*. The test that is failing is designed to fail – it is testing that right shift is not erroneously recognised – and the checker passes the test if the error message is what it expected. Antlr writes a long informative error message listing all the tokens it might have expected to find at the failing point of the code and that list now has `init` in it. Maybe I could look at supporting matching error messages (actually any stderr output) with a R.E. to avoid such false negatives… no promises! ;-)

BillWagner:
@Nigel-Ecma 

No need to make corrections right now. This PR shows how our tools help us as we move to C# 9:

- The grammar validator shows where we need updates to reflect the updated grammar in C# 9.
- The test examples runner validates that our samples don't use newer features while we're still working on older standards. See #1451 for updates in this branch to fix the failures in this PRs.
- The renumbering tool highlights that this PR / features depends on records. 

All told, this is good.

BillWagner:
Closing in favor of #1452

PR #980, opened 2023-11-01
Add support for top level statements

RexJaeschke:


standard/basic-concepts.md
```diff
@@ -47,6 +56,166 @@ If the effective entry point’s return type is `int`, the return value from the
...
+Async operations are allowed in top-level statements to the degree they are allowed in statements within a named async entry-point method. However, they are not required.
+
+The tokens that are generated in place of `«AsyncAndReturnType»` are determined based on operations used by the top-level statements, as follows:
+
+| **Top-level code contains** | **Generated entry-point signature**
+|   -----------------------   |   -------------------------------
+| No `await` or `return`      | `static void «Main»(string[] args)`
+| `return` only               | `static int «Main»(string[] args)`
+| `await` only                | `static async Task «Main»(string[] args)`
+| `await` and `return`        | `static async Task<int> «Main»(string[] args)`
```
KalleOlaviNiemitalo: How about `return;` without a return value?

```csharp
using System;

Console.WriteLine("Hah!");
return;
```

standard/basic-concepts.md
```diff
@@ -47,6 +56,166 @@ If the effective entry point’s return type is `int`, the return value from the
...
+partial class Program
+{
+    static «AsyncAndReturnType» «Main»(string[] args)
+    {
+        // top-level statements
+    }
+}
+```
+
+The class name `Program` shall be referenceable by name from within the application. However, the method name `«Main»` is used here for illustrative purposes only. The actual name generated by the implementation is unspecified, and cannot be referenced by name from within the application.
```
KalleOlaviNiemitalo: Should this say that the generated name is readable via CallerMemberNameAttribute?

```csharp
System.Console.WriteLine(WhoAmI()); // outputs "<Main>$"

System.Action action = () => System.Console.WriteLine(WhoAmI());
action(); // outputs "<Main>$" again, not the name of the method generated for the lambda

string WhoAmI(
    [System.Runtime.CompilerServices.CallerMemberName]string member = null)
{
    return member;
}
```

standard/basic-concepts.md
RexJaeschke: Thanks, @KalleOlaviNiemitalo. I added the text "with value" after `return` in column 1, rows 1, 2, and 4, to address this, so return without a value is equivalent to omitted return. In column 2, I also added `private` to the front of each signature (which is already implied in the 4 lots of implementation-generated code, as they have no explicit access modifier on the method).

standard/basic-concepts.md
RexJaeschke: @KalleOlaviNiemitalo, I confirm that the generated name can be found, but it's not clear to me how one might make use of that.

KalleOlaviNiemitalo: It might be best to add something about this to the specification of CallerMemberNameAttribute, to clarify that the name of the generated method is substituted even though it does not occur in the source code.  Because this doesn't work like in lambda expressions where the user-specified name is used even though the code is generated to a differently-named method.

```csharp
using System;

class C
{
    void M(int i)
    {
        // calls N(i, "M") even though the call is in a method named something like <M>b__0
        Action a = () => N(i);
    }
    
    static void N(int i, [CallerMemberName] string mem = null)
    {
    }
}
```

RexJaeschke: At the end of line 73, I added the following sentence:

> It is, however, available via the attribute `System.Runtime.CompilerServices.CallerMemberName`.

Also, in that attribute's section, I added

> (In the case of a function invocation from a top-level statement (§top-level-statements), the member name is that generated by the implementation.)

RexJaeschke: @KalleOlaviNiemitalo Do these changes resolve this conversation?

BillWagner:
Closed in favor of #1454

PR #983, opened 2023-11-02
Add support for record classes

RexJaeschke:
Notes:

A. I put the new subclause "With expression" prior to "Arithmetic operators", which once the V8 features "Indices and Ranges" and "Pattern matching" have been merged, should immediately follow "Range operator" and "Switch expression." Make sure these are all in the correct place. 12.4.2 Operator precedence and associativity will also need to be revised accordingly.

B. New subclause §rec-class-prtmem Printing members mentions a method `System.Runtime.CompilerServices.RuntimeHelpers.EnsureSufficientExecutionStack()`. It's quite possible that this is an implementation-specific detail, in which case, we'll need to find words to make it abstract, and possibly update the Portability annex accordingly.


BillWagner:
Closed in preference of #1458

Issue #750, opened 2023-03-13
Criteria for putting new unsafe-related text in chapters other than "Unsafe code" 


RexJaeschke:
***Note: Since this issue was created, chapter (and maybe section) numbers have changed!***

**[Although this is an issue for V9, Jon suggested we discuss the general principle soon. That is, how hard should we try to keep *all* unsafe-related text in the unsafe chapter? Are there reasonable exceptions for not doing so?]**

Support for unsafe mode is optional, and 1-2 years ago, we made the decision to push almost all the unsafe-related stuff into the unsafe chapter, §22, except for bits in the grammar, which we’ve flagged as “unsafe-mode only.”

I’ve nearly completed getting the [MS v9 proposal for function pointers](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/function-pointers.md) into shape for use by TG2. However, I have a situation for which I’m looking for guidance.

The addition of support for function pointers requires unsafe support, so most of the spec for that proposal will necessarily go in §22. However, this feature also impacts type inference, as described in six subsections of §11.6.3, “Type inference.”

I see two alternate approaches:

1. Following the current approach of putting as much as possible unsafe stuff into 22, I can add text to the corresponding sections in §11.6.3.* saying “This subclause is extended in unsafe code (§forward-pointer to §22.x.y).” And then describe those extensions in the new sections §22.x.y.

1. I can put the unsafe-related stuff in-line in §11.6.3.*, and somehow mark it as being unsafe-related.

Approach 1 is pure, but has the problem that some text in §22.x.y needs to be “merged” into specific places in lists in §11.6.3.*. For example, this would result in the following:

> §22.6.(x) Output type inferences [new section]

> **In §11.6.3.7, the following bullet is added between the second and third bullets:**

> - If `E` is an address-of method group and `T` is a function pointer type with parameter types `T1..Tk` and return type `Tb`, and overload resolution of `E` with the types `T1..Tk` yields a single method with return type `U`, then a *lower-bound inference* is made from `U` to `Tb`.

The reader of this new section, §22.6.x, will have to flip between this text and that in §11.6.3.7 to make sense of it. And then, we are at the mercy of this positional dependence of that list, which could easily get out of sync as §11.6.3.7 evolves.

Here’s a similar case:

> §22.6.(y) Lower-bound inferences [new section]

> **In §11.6.3.10, the following case is added to the third bullet:**

[...]

PR #984, opened 2023-11-04
Add support for function pointers

RexJaeschke:
There are two parts of this feature that are **not** covered by this PR, but that need to be considered:

**RESOLVED!** <s>## Part 1: Organizing the unsafe-related text
 
The following sections either need detailed text (which I have written, but it is *not* in this PR) to be added that is unsafe-specific, or will need forward pointers to the unsafe chapter, which would have the text added there:

12.6.3.4 Input types
12.6.3.5 Output types
12.6.3.7 Output type inferences
12.6.3.10 Lower-bound inferences
12.6.3.11 Upper-bound inferences
12.6.4.3 Better function member
12.6.4.5 Better conversion from expression
12.8.9 Invocation expressions|12.8.9.1 General

See Issue #750 for a discussion on dealing with unsafe situations outside of the unsafe chapter. The following two sections show the alternatives for one of these sections.

### Part 1 example - New unsafe text in Chapter 12

12.6.3.7 Output type inferences

> An *output type inference* is made *from* an expression `E` *to* a type T in the following way:
>
> - If `E` is an anonymous function with inferred return type `U` ([§12.6.3.13](expressions.md#126313-inferred-return-type)) and `T` is a delegate type or expression tree type with return type `Tₓ`, then a *lower-bound inference* ([§12.6.3.10](expressions.md#126310-lower-bound-inferences)) is made *from* `U` *to* `Tₓ`.
> - Otherwise, if `E` is a method group and `T` is a delegate type or expression tree type with parameter types `T₁...Tᵥ` and return type `Tₓ`, and overload resolution of `E` with the types `T₁...Tᵥ` yields a single method with return type `U`, then a *lower-bound inference* is made *from* `U` *to* `Tₓ`.
> - **[[IN UNSAFE MODE ONLY]] If `E` is an address-of method group and `T` is a function pointer type with parameter types `T1..Tk` and return type `Tb`, and overload resolution of `E` with the types `T1..Tk` yields a single method with return type `U`, then a *lower-bound inference* is made from `U` to `Tb`.**
> - Otherwise, if `E` is an expression with type `U`, then a *lower-bound inference* is made *from* `U` *to* `T`.
> - Otherwise, no inferences are made.

We might also want to add a backward pointer to this section from the unsafe chapter.

### Part 1 example - New unsafe text in Chapter 23

12.6.3.7 Output type inferences
…
This subclause is extended in unsafe code (§xx).

23.6.1.x Output type inferences

In [§12.6.3.7](expressions.md#12637-output-type-inferences), the following bullet is added between the second and third bullets:

- If `E` is an address-of method group and `T` is a function pointer type with parameter types `T1..Tk` and return type `Tb`, and overload resolution of `E` with the types `T1..Tk` yields a single method with return type `U`, then a *lower-bound inference* is made from `U` to `Tb`.

[The problem with this approach is that the reader has to flip between two chapters to understand this, and over time, paras might be added, removed, or rearranged back in 12.6.3.7, rendering the relative para instructions used here, incorrect. And we don't want to replicate all the bullets from 12.6.3.7 here.]</s>

## Part 2: The library type `System.Runtime.CompilerServices.UnmanagedCallersOnlyAttribute`

Spec'ing of this attribute has not yet been resolved. **However, I added this type to C.3, but TG2 might decide to remove it.** Following is a message thread between Fred Silberberg and me in March 2022:

### Comment 9: “UnmanagedCallersOnlyAttribute”

This part of the proposal looks like implementation detail. As such, I *don’t* mention it in my C# standard proposal. Is that reasonable?

The MS proposal doesn’t show any examples of initializing an unmanaged function pointer using `&`. I’m guessing it might go something like this:

```csharp
[UnmanagedCallersOnlyAttribute(CallConvs = new[] { typeof(CallConvCdecl) })]
static extern void Cfunction(int value);

delegate* unmanaged[Cdecl]<int, void> fp = &Cfunction;
```

Am I on the right track?

2022-03-24 Fred Silberberg responded: **No, this definitely cannot be removed.** It affects your Comment 6 [method signature and calling convention combination], and also affects the ability for methods to be called without taking a function pointer to them. It is very explicitly *not* an implementation detail.

2022-03-25 Rex replied: Thanks, [@333fred](https://github.com/333fred) for your quick response.
Re my Comment #9, I still haven't fully digested your response and the attribute in question, but I wanted to bring to your attention a statement from the C# Standard's Introduction:

> Ecma Technical Committee 39 (TC39) [later renamed to TC49] Task Group 2 (TG2) was formed in September 2000, to produce a standard for C#. Another Task Group, TG3, was also formed at that time to produce a standard for a library and execution environment called Common Language Infrastructure (CLI). (CLI is based on a subset of the .NET Framework.) Although Microsoft’s implementation of C# relies on CLI for library and run-time support, other implementations of C# need not, provided they support an alternate way of getting at the minimum CLI features required by this C# standard (see [Annex C](https://github.com/dotnet/csharpstandard/blob/draft-v6/standard/standard-library.md#annex-c-standard-library)).

My concern is, "How much of the machinery should be exposed in a standard that does not require .NET/CLI? That is, is this attribute part of the support needed by the minimum CLI features required by this C# standard?" My guess is that you will answer, "Yes, it's needed" and I'm OK with that.

2022-03-25 Fred replied: I will indeed. Since the presence of the attribute determines whether a method group can be converted to an unmanaged function pointer, it needs to be present in the spec. If a minimum cli didn't have the attribute, it would probably be fine technically: users of that cli would just never be able to directly convert a method group to an unmanaged function pointer.


standard/unsafe-code.md
```diff
@@ -156,106 +209,195 @@ For a given implementation, all pointer types shall have the same size and repre
...
 >     }
->     return null;
 > }
 > ```
 >
-> *end example*
+> Given that the function pointers in the array point to methods with no parameters, `&Log` takes the address of the `Log` method having no parameters. *end example*
 
-In an unsafe context, several constructs are available for operating on pointers:
+*unmanaged_calling_convention* supports a small number of predefined conventions (`Cdecl`, `Stdcall`, `Thiscall`, and `Fastcall`, all of which are contextual keywords), which may be used standalone or as an *identifier* in an *unmanaged_calling_convention* identifier list. Other implementation-defined conventions are permitted, and multiple conventions can be combined by using an *identifier* list, possibly containing one or more of these predefined conventions. Lookup and processing for identifiers in this list is done in an implementation-defined manner.
```
KalleOlaviNiemitalo, line 355: Are `@Stdcall` and `Stdc\u0061ll` required to mean the same thing as `Stdcall`, or is that too implementation-defined?

KalleOlaviNiemitalo:
Is there any definition of what the `Cdecl`, `Fastcall`, `Stdcall`, and `Thiscall` unmanaged calling conventions mean in C# — is it possible to write portable unmanaged code that uses these calling conventions and can interoperate with C#?  AFAIK, calling conventions with similar names are defined in

- Microsoft C++ compiler documentation
  - [/Gd, /Gr, /Gv, /Gz (Calling Convention)](https://learn.microsoft.com/en-us/cpp/build/reference/gd-gr-gv-gz-calling-convention?view=msvc-170)
  - [x64 calling convention](https://learn.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-170)
- ECMA-335 6th edition:
  - II.14.5 Method pointers, which references CallConv in II.15.3 Calling convention.  This has also `vararg`.  Metadata encoding in II.23.2 Blobs and signatures, and II.23.2.3 StandAloneMethodSig.
  - II.15.5.2 Platform invoke, and II.23.1.8 Flags for ImplMap [PInvokeAttributes].  This one has also `Platformapi`.

If nothing about semantics is being standardised then perhaps it would be better to leave the names entirely implementation-defined (so no contextual keywords) and add a *note* recommending these names in implementations that target ECMA-335.  Portable C# applications would then have to put `#if` around any use of `delegate* unmanaged[Cdecl]` as implementations would not be required to recognise the `Cdecl` identifier, even in code that won't be called at run time.

standard/unsafe-code.md
RexJaeschke: Regarding `Stdc\u0061ll`, [§6.4.2](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/lexical-structure.md#642-unicode-character-escape-sequences), Unicode character escape sequences, states:

> A Unicode escape sequence represents a Unicode code point. **Unicode escape sequences are processed in identifiers ([§6.4.3](lexical-structure.md#643-identifiers)), character literals ([§6.4.5.5](lexical-structure.md#6455-character-literals)), regular string literals ([§6.4.5.6](lexical-structure.md#6456-string-literals)), and interpolated regular string expressions ([§12.8.3](expressions.md#1283-interpolated-string-expressions)). A Unicode escape sequence is not processed in any other location (for example, to form an operator, punctuator, or keyword).**

This suggests that `Stdc\u0061ll` and `Stdcall` are equivalent. However, [§6.4.4](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/lexical-structure.md#644-keywords), Keywords, contains the following:

> Just as with keywords, contextual keywords can be used as ordinary identifiers by prefixing them with the `@` character.
>
> > *Note*: When used as contextual keywords, these identifiers cannot contain *Unicode_Escape_Sequence*s. *end note*

If that Note is true, and that info can't be deduced from normative text, the Note should be made normative. 

[§6.4.3](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/lexical-structure.md#643-identifiers) Identifiers, contains the grammar for `identifier`. See the detailed Note following it. After several readings of that, I am *not* completely convinced of the situation w.r.t `@`.

@Nigel-Ecma, as you were heavily involved in the rewrite of this grammar, what is your opinion?

KalleOlaviNiemitalo: The impression I get from the draft PR is that `delegate* unmanaged[Stdcall]<void>` uses a contextual keyword and shall be accepted by all implementations, but `delegate* unmanaged[@Stdcall]<void>` uses an identifier and does not need to be accepted; an implementation might even define that the identifier lookup depends on `using` directives.  But I'm not sure the specification should give this freedom to implementations.

If that interpretation is intended, then this text in §6.4.4 Keywords

> In certain cases the grammar is not enough to distinguish contextual keyword usage from identifiers. In all such cases it will be specified how to disambiguate between the two.

requires adding a disambiguation rule to §function-pointers, e.g.

> If the *unmanaged_calling_convention* is `Cdecl`, `Stdcall`, `Thiscall`, or `Fastcall`, then it is a contextual keyword.  Otherwise, it is an identifier.

But I feel it would be better to omit `Cdecl` etc. from the grammar and instead always parse *unmanaged_calling_convention* as a list of identifiers, and then specify that these four identifiers shall be accepted by all implementations.

RexJaeschke:
After a recent review of this PR, I revised the "Pointer types" section, to fix errors and omissions from my earlier proposal, to add clarification, and to tighten original wording.

With this proposal, pointer types go from a single section to being broken into a General section and 3 others: data, function, and void, and that has required distributing the original words across those 4 places. The important aspect is that given `void **p`, `p` is a data pointer to a void pointer; it is **not** itself a void pointer. Likewise, `delegate*<void>*` is a data pointer to a function pointer. And `int** p` declares a data pointer to a data pointer.

BillWagner:
Closing in favor of #1459

PR #985, opened 2023-11-06
Introduce term "identity convertible" for types that aren't the same type, but have an identity conversion

### LINKED ISSUE #730: Identity conversion and "Considered equivalent"
See https://github.com/dotnet/csharpstandard/pull/664/files#r1093246493

We use the same language that "dynamic and object are considered equivalent" in the section on identity conversions.

That is somewhat troubling because they aren't "equivalent". We should search for better wording.

---
[Associated WorkItem - 177627](https://dev.azure.com/msft-skilling/Content/_workitems/edit/177627)

BillWagner:
Fixes #730 

This landed slightly differently than originally discussed:

- the uses of "considered equivalent" were only in conversions where the concept of types that aren't the same have identity conversions.
- The uses of "the same type" often meant "the same type". However, not always. In places like overriding methods, partial method definitions, and so on, the types could be those that are runtime interchangeable (meaning any identity conversion).

So, the term "runtime interchangeable" is introduced in the clause on identity conversions. It refers to types that are not the same, but have identity conversions.

I also moved text from the tuple type to the identity conversions clause to define those conversions. 

Replace the term "the same type" with "the same type or runtime convertible" when that's what was meant.


standard/classes.md
```diff
@@ -233,7 +233,7 @@ The base class specified in a class declaration can be a constructed class type
 
 The direct base class of a class type shall be at least as accessible as the class type itself ([§7.5.5](basic-concepts.md#755-accessibility-constraints)). For example, it is a compile-time error for a public class to derive from a private or internal class.
 
-The direct base class of a class type shall not be any of the following types: `System.Array`, `System.Delegate`, `System.Enum`, or `System.ValueType`. Furthermore, a generic class declaration shall not use `System.Attribute` as a direct or indirect base class ([§22.2.1](attributes.md#2221-general)).
+The direct base class of a class type shall not be any of the following types: `System.Array`, `System.Delegate`, `System.Enum`, `System.ValueType` or the `dynamic` type. Furthermore, a generic class declaration shall not use `System.Attribute` as a direct or indirect base class ([§22.2.1](attributes.md#2221-general)).
```
BillWagner, line 236: This isn't strictly necessary with the addition in 8.7, but it's where both Jon and I looked first, so maybe others will find it helpful as well.

standard/conversions.md
```diff
@@ -71,19 +71,19 @@ The pre-defined implicit conversions always succeed and never cause exceptions t
...
 
 However, dynamic conversions ([§10.2.10](conversions.md#10210-implicit-dynamic-conversions) and [§10.3.8](conversions.md#1038-explicit-dynamic-conversions)) apply only to expressions of type `dynamic` ([§8.2.4](types.md#824-the-dynamic-type)).
 
 ### 10.2.2 Identity conversion
 
-An identity conversion converts from any type to the same type. One reason this conversion exists is so that a type T or an expression of type T can be said to be convertible to T itself.
+An identity conversion converts from any type to the same type. One reason this conversion exists is so that a type `T` or an expression of type `T` can be said to be convertible to `T` itself.
 
-In some cases there is an identity conversion between types that are not exactly the same, but are considered equivalent. Such identity conversions exist:
+Two types are *runtime interchangeable* when an identity conversion exists between two types that not exactly the same. The following such identity conversions exist:
```
BillWagner: my opinion is that "exactly" isn't need here. But, it was in the original text.

BillWagner (commented):
Adding comments for reviewers.

standard/expressions.md
```diff
@@ -245,7 +245,7 @@ An operation of the form `«op» x` or `x «op»`, where «op» is an overloada
 An operation of the form `x «op» y`, where «op» is an overloadable binary operator, `x` is an expression of type `X`, and `y` is an expression of type `Y`, is processed as follows:
 
 - The set of candidate user-defined operators provided by `X` and `Y` for the operation `operator «op»(x, y)` is determined. The set consists of the union of the candidate operators provided by `X` and the candidate operators provided by `Y`, each determined using the rules of [§12.4.6](expressions.md#1246-candidate-user-defined-operators). For the combined set, candidates are merged as follows:
-  - If `X` and `Y` are the same type, or if `X` and `Y` are derived from a common base type, then shared candidate operators only occur in the combined set once.
+  - If `X` and `Y` are the same type or runtime interchangeable, or if `X` and `Y` are derived from a common base type, then shared candidate operators only occur in the combined set once.
```
jskeet: "Common base type" implies "same" but I suspect should be runtime interchangeable. Not sure, but worth checking!

standard/types.md
```diff
@@ -396,7 +396,7 @@ There does not need to exist a `System.ValueTuple<...>` declaration that directl
 
 Element names within a tuple type shall be distinct. A tuple element name of the form `ItemX`, where `X` is any sequence of non-`0`-initiated decimal digits that could represent the position of a tuple element, is only permitted at the position denoted by `X`.
 
-The optional element names are not represented in the `ValueTuple<...>` types, and are not stored in the runtime representation of a tuple value. There is an identity conversion between all tuple types with the same arity and identity-convertible sequences of element types, as well as to and from the corresponding constructed `ValueTuple<...>` type.
+The optional element names are not represented in the `ValueTuple<...>` types, and are not stored in the runtime representation of a tuple value. Identity conversions (§10.2.2) exist between tuples with identity-convertible sequences of element types
```
```suggestion
The optional element names are not represented in the `ValueTuple<...>` types, and are not stored in the runtime representation of a tuple value. Identity conversions (§10.2.2) exist between tuples with identity-convertible sequences of element types.
```

(Just finishing the sentence with a period.)

jskeet: This loses the statement of an identity conversion between tuples and `ValueTuple<...>` though. Is that because you consider it to be covered two paragraphs above?

standard/types.md
```diff
@@ -546,7 +546,7 @@ A closed type is a type that is not an open type.
...
 
-Each closed constructed type has its own set of static variables, which are not shared with any other closed constructed types. Since an open type does not exist at run-time, there are no static variables associated with an open type. Two closed constructed types are the same type if they are constructed from the same unbound generic type, and their corresponding type arguments are the same type.
+Each closed constructed type has its own set of static variables, which are not shared with any other closed constructed types. Since an open type does not exist at run-time, there are no static variables associated with an open type. Two closed constructed types are runtime interchangeable (§10.2.2) if they are constructed from the same unbound generic type, and an identity conversion exists between their corresponding type arguments.
```
jskeet: Is there the possibility of runtime interchangeable unbound generic types? I can't think of any offhand, but I thought I'd mention it.

> and an identity conversion exists between their corresponding type arguments

Do we need to be more precise here about each corresponding pair of type arguments? (We're not really talking about a single identity conversion, but about N conversions for N type arguments in each constructed type.)

standard/types.md
```diff
@@ -666,12 +666,13 @@ The API surface provided by `Expression<TDelegate>` is implementation-specific b
 
 The type `dynamic` uses dynamic binding, as described in detail in [§12.3.2](expressions.md#1232-binding-time), as opposed to static binding which is used by all other types.
 
-`dynamic` is considered identical to `object` except in the following respects:
+The types `dynamic` is considered identical to `object` except in the following respects:
```
```suggestion
The type `dynamic` is considered identical to `object` except in the following respects:
```

standard/types.md
```diff
@@ -681,10 +682,13 @@ The type `dynamic` uses dynamic binding, as described in detail in [§12.3.2](ex
 
 Because of this equivalence, the following holds:
 
-- There is an implicit identity conversion between `object` and `dynamic`, and between constructed types that are the same when replacing `dynamic` with `object`.
+- There is an implicit identity conversion
+  - between `object` and `dynamic`
```
jskeet, line 686: I think we normally have ";" at the end of multi-bullet lists. @RexJaeschke to confirm precise punctuation here.

jskeet (commented):
Generally very much in favour of this - a few nits, including some probably-spurious ones.

standard/types.md
BillWagner: It does lose that information. I moved it to the section on Identity conversions, which is linked here. I thought that was the better place. (Similar to with the dynamic type, where there's text about identity conversions and refers to the section in Conversions on Identity Conversions

standard/types.md
KalleOlaviNiemitalo: "Each closed constructed type has its own set of static variables" -- so if you have a type

```csharp
class C<T> {
    public static int I;
}
```

then does that mean `C<(int a, int b)>.I` and `C<(int x, int y)>.I` are not the same variable?

standard/expressions.md
BillWagner: Well, no. "common base type" could be:

```csharp
public class B
{
    public static bool operator > (B left, B right)
    {
        // ... logic
    }
}

public class D1 : B
{
}

public class D2 : B
{
}
```

The types `D1` and `D2` are derived from a common base type. The operator is only added once.

standard/types.md
BillWagner: I re-organized this paragraph to address @KalleOlaviNiemitalo 's comment.

BillWagner: > Do we need to be more precise here about each corresponding pair of type arguments? (We're not really talking about a single identity conversion, but about N conversions for N type arguments in each constructed type.)

Very likely. I struggled with formal language here. Effectively, what we need is the conversion analog of "it's turtles all the way down". For each corresponding type argument, there must be an identity conversion between the corresponding types. Each of the type arguments may be other closed generic types or tuples, which requires another cycle down.

I'll take one more stab at it.

BillWagner:
Comment from the meeting: Where "the same type or runtime interchangeable" is used, replace with "runtime interchangeable" only. 

Instead of "runtime interchangeable", use "identity convertible" for this concept.

standard/conversions.md
```diff
@@ -71,19 +71,19 @@ The pre-defined implicit conversions always succeed and never cause exceptions t
...
 
-For the purposes of conversion, the types `object` and `dynamic` are considered equivalent.
+For the purposes of conversion, the types `object` and `dynamic` are *runtime interchangeable* (§10.2.2).
 
 However, dynamic conversions ([§10.2.10](conversions.md#10210-implicit-dynamic-conversions) and [§10.3.8](conversions.md#1038-explicit-dynamic-conversions)) apply only to expressions of type `dynamic` ([§8.2.4](types.md#824-the-dynamic-type)).
 
 ### 10.2.2 Identity conversion
 
-An identity conversion converts from any type to the same type. One reason this conversion exists is so that a type T or an expression of type T can be said to be convertible to T itself.
+An identity conversion converts from any type to the same type. One reason this conversion exists is so that a type `T` or an expression of type `T` can be said to be convertible to `T` itself.
```
BillWagner: change this to talk about interchangeable at runtime. (the same type or an equivalent type).

Add the identity conversions are always symetric: If there is a conversion from A to B, there is one from B to A

gafter:
```suggestion
An identity conversion converts from any type to the same type or a type that is equivalent at runtime. One reason this conversion exists is so that a type `T` or an expression of type `T` can be said to be convertible to `T` itself.
```

standard/conversions.md
gafter:
```suggestion
Two types are *identity convertible* when an identity conversion exists between two types. The following such identity conversions exist:
```

standard/conversions.md
```diff
@@ -71,19 +71,19 @@ The pre-defined implicit conversions always succeed and never cause exceptions t
...
 
-In some cases there is an identity conversion between types that are not exactly the same, but are considered equivalent. Such identity conversions exist:
+Two types are *runtime interchangeable* when an identity conversion exists between two types that not exactly the same. The following such identity conversions exist:
 
-- between `object` and `dynamic`.
-- between tuple types with the same arity, when an identity conversion exists between each pair of corresponding element types.
-- between types constructed from the same generic type where there exists an identity conversion between each corresponding type argument.
+- Between `object` and `dynamic`.
+- Between all tuple types with the same arity, and the corresponding constructed `ValueTuple<...>` type, when an identity conversion exists between each pair of corresponding element types. Element types may themselves by tuples, where an identity conversion exists between each pair of corresponding element types.
```
gafter:
```suggestion
- Between all tuple types with the same arity, and the corresponding constructed `ValueTuple<...>` type, when an identity conversion exists between each pair of corresponding element types.
```

gafter: Suggest: add an example with nesting and tuples.

BillWagner:
@jskeet @gafter @MadsTorgersen @RexJaeschke @Nigel-Ecma 

I made all the updates from our meeting. We can review again and hopefully merge next meeting.

standard/conversions.md
```diff
@@ -71,20 +71,48 @@ The pre-defined implicit conversions always succeed and never cause exceptions t
...
+> var t6 =((8, "eight"), 9);
+> t6 = t5;
+> t5 = t4;
+> ```
+>
+> The tuples `t1`, `t2` and `t3` all have two elements: an `int` followed by a `string`. Tuple element types may themselves by tuples, as in `t4`, `t5`, and `t6`. An identity conversion exists between each pair of corresponding element types, including nested tuples, therefore an identity conversion exists between the tuples `t4`, `t5`, and `t6`.
+>
+> *end example*
+
+All identity conversions are symmetric. If an identity conversion exists from `E₁` to `E₂`, then an identity conversion exists from `E₂` to `E₁`. Two types are *identity convertible* when an identity conversion exists between two types.
```
jskeet: E1 and E2 sound like expressions rather than types. Was that deliberate? If so, we probably need more clarity. Otherwise, I'd suggest just changing use T1 and T2.

standard/expressions.md
```diff
@@ -245,7 +245,7 @@ An operation of the form `«op» x` or `x «op»`, where «op» is an overloada
 An operation of the form `x «op» y`, where «op» is an overloadable binary operator, `x` is an expression of type `X`, and `y` is an expression of type `Y`, is processed as follows:
 
 - The set of candidate user-defined operators provided by `X` and `Y` for the operation `operator «op»(x, y)` is determined. The set consists of the union of the candidate operators provided by `X` and the candidate operators provided by `Y`, each determined using the rules of [§12.4.6](expressions.md#1246-candidate-user-defined-operators). For the combined set, candidates are merged as follows:
-  - If `X` and `Y` are the same type or runtime interchangeable, or if `X` and `Y` are derived from a common base type, then shared candidate operators only occur in the combined set once.
+  - If `X` and `Y` are the same type or identity convertible, or if `X` and `Y` are derived from a common base type, then shared candidate operators only occur in the combined set once.
```
jskeet: Do we still need "the same type" here? (Given that if they are the same type, they're identity convertible.)

standard/conversions.md
```diff
@@ -71,20 +71,48 @@ The pre-defined implicit conversions always succeed and never cause exceptions t
...
 
-An identity conversion converts from any type to the same type or a type that is equivalent at runtime. One reason this conversion exists is so that a type `T` or an expression of type `T` can be said to be convertible to `T` itself.
-
-Two types are *identity convertible* when an identity conversion exists between two types. The following such identity conversions exist:
+An identity conversion converts from any type to the same type or a type that is equivalent at runtime. One reason this conversion exists is so that a type `T` or an expression of type `T` can be said to be convertible to `T` itself. The following identity conversions exist:
 
+- Between `T` and `T`, for any type `T`.
 - Between `object` and `dynamic`.
 - Between all tuple types with the same arity, and the corresponding constructed `ValueTuple<...>` type, when an identity conversion exists between each pair of corresponding element types.
 - Between types constructed from the same generic type where there exists an identity conversion between each corresponding type argument.
```
jskeet, line 85: Should "same" still be here? I suspect it's okay because there are no (open) generic types with identity conversions, but I'm just checking. (I can imagine the potential in the future for this to change to "types constructed from identity convertible generic types where [...]")

standard/conversions.md
BillWagner: It should be types. Updated in next commit.

standard/expressions.md
BillWagner: Updated in next commit.

I also did a global search for "the same type" to make sure it wasn't used in a similar manner elsewhere. It was not.

standard/conversions.md
BillWagner: Should I add "...constructed from the same **closed** generic type where ..." (emphasis only in comment, not meant for a spec change).

standard/conversions.md
```diff
@@ -71,19 +71,47 @@ The pre-defined implicit conversions always succeed and never cause exceptions t
...
-- between types constructed from the same generic type where there exists an identity conversion between each corresponding type argument.
+> *Example*: The following illustrates the recursive nature of the third rule:
+>
+> <!-- Example: {template:"code-in-main-without-using", name:"IdentityTupleConversion"} -->
+> ```csharp
+> (int a , string b) t1 = (1, "two");
+> (int c, string d) t2 = (3, "four");
+> 
+> // Identity conversions exist between
+> // t1, t2, and t3.
```
```suggestion
> // the types of t1, t2, and t3.
```

standard/conversions.md
```diff
@@ -71,19 +71,47 @@ The pre-defined implicit conversions always succeed and never cause exceptions t
...
+> var t3 = (5, "six");
+> t3 = t2;
+> t2 = t1;
+> 
+> var t4 = (t1, 7);
+> var t5 = (t2, 8);
+> 
+> // Identity conversions exist between
+> // t4, t5, and t6.
```
```suggestion
> // the types of t4, t5, and t6.
```

standard/conversions.md
```diff
@@ -71,19 +71,47 @@ The pre-defined implicit conversions always succeed and never cause exceptions t
...
+> var t6 =((8, "eight"), 9);
+> t6 = t5;
+> t5 = t4;
+> ```
+>
+> The tuples `t1`, `t2` and `t3` all have two elements: an `int` followed by a `string`. Tuple element types may themselves by tuples, as in `t4`, `t5`, and `t6`. An identity conversion exists between each pair of corresponding element types, including nested tuples, therefore an identity conversion exists between the tuples `t4`, `t5`, and `t6`.
```
```suggestion
> The types of tuples `t1`, `t2` and `t3` all have two elements: an `int` followed by a `string`. Tuple element types may themselves by tuples, as in `t4`, `t5`, and `t6`. An identity conversion exists between each pair of corresponding element types, including nested tuples, therefore an identity conversion exists between the types of tuples `t4`, `t5`, and `t6`.
```

standard/conversions.md
```diff
@@ -772,7 +800,7 @@ Specifically, an anonymous function `F` is compatible with a delegate type `D`
 
 - If `F` contains an *anonymous_function_signature*, then `D` and `F` have the same number of parameters.
 - If `F` does not contain an *anonymous_function_signature*, then `D` may have zero or more parameters of any type, as long as no parameter of `D` has the out parameter modifier.
-- If `F` has an explicitly typed parameter list, each parameter in `D` has the same type and modifiers as the corresponding parameter in `F`.
+- If `F` has an explicitly typed parameter list, each parameter in `D` has the same modifiers as the corresponding parameter in `F` and an identity conversion exists between the corresponding parameter in `F` and `D`.
```
```suggestion
- If `F` has an explicitly typed parameter list, each parameter in `D` has the same modifiers as the corresponding parameter in `F` and an identity conversion exists between the corresponding parameter in `F`.
```

standard/types.md
```diff
@@ -681,10 +682,13 @@ The type `dynamic` uses dynamic binding, as described in detail in [§12.3.2](ex
...
+  - between constructed types that are the same when replacing `dynamic` with `object`
```
jskeet, line 687: I'll file an issue to say that this list shouldn't be needed here. We don't want to have to edit this section every time a new identity conversion between "things with identity convertible types" comes up.

jskeet (approved):

standard/types.md
jskeet: Raised #1034.

PR #988, opened 2023-11-12
Add support for static anonymous functions

RexJaeschke:


standard/expressions.md
```diff
@@ -4906,6 +4911,12 @@ anonymous_function_body
     ;
 ```
 
+If the modifier `static` is present, the anonymous function cannot capture state from the enclosing scope.
+
+A non-`static` local function or non-`static` anonymous function can capture state from an enclosing `static` anonymous function, but cannot capture state outside the enclosing static anonymous function.
+
+Removing the `static` modifier from an anonymous function in a valid program does not change the meaning of the program.
```
KalleOlaviNiemitalo, line 5062: Should `static` ensure that all delegate instances converted from the same anonymous function shall compare equal?  (That would not require any caching of delegate instances.)

```csharp
using System;
using System.Diagnostics;

class C {
    Action G() {
        return static () => {};
    }
    void M() {
        Debug.Assert(G() == G());
    }
}
```

I think `static` should **not** guarantee this equality.  IIRC Roslyn generates an instance method even for a static anonymous function because it is more efficient for the calling convention.  Then the delegate will refer to some instance, and will compare unequal if it refers to a different instance.  Requiring such an implementation to ensure that the same instance is used every time might cost additional interlocked operations for thread safety.

KalleOlaviNiemitalo: Maybe a note about this in the specification of delegate equality would be useful though.

BillWagner:
Closing in favor of #1461

PR #990, opened 2023-11-12
Add support for target typed new expressions

RexJaeschke:
The [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/target-typed-new.md) contains a "Miscellaneous" section, which contains a list of "consequences of the specification." These have *not* been added to this PR! Someone should verify that these are covered sufficiently by normative text, and even if they all are, it might be worth adding some Notes to that effect.


standard/conversions.md
```diff
@@ -352,6 +352,12 @@ An implicit conversion exists from a *default_literal* ([§12.8.20](expressions.
 
 While throw expressions do not have a type, they may be implicitly converted to any type.
 
+### §imp-obj-creation-conv Implicit object-creation conversions
+
+There is an implicit ***object-creation conversion*** from a *target_typed_new* expression ([§12.8.16.2](expressions.md#128162-object-creation-expressions)) to every type.
+
+Given a target type `T`, if `T` is an instance of `System.Nullable`, the type `T0` is `T`'s underlying type. Otherwise `T0` is `T`. The meaning of a *target_typed_new* expression that is converted to the type `T` is the same as the meaning of a corresponding *object_creation_expression* that specifies `T0` as the type.
```
KalleOlaviNiemitalo, line 387: System.Nullable is a type but not a generic type.  How is this worded elsewhere — `System.Nullable<>` or `System.Nullable<T>`?

standard/expressions.md
```diff
@@ -2296,6 +2301,10 @@ object_or_collection_initializer
 
 The *type* of an *object_creation_expression* shall be a *class_type*, a *value_type*, or a *type_parameter*. The *type* cannot be a *tuple_type* or an abstract or static *class_type*.
```
KalleOlaviNiemitalo, line 2448: Ok I suppose this disallows `dynamic` and pointer types as *type* even in target-typed new.

standard/expressions.md
```diff
@@ -2296,6 +2301,10 @@ object_or_collection_initializer
...
 
+If `type` can be inferred from usage, it can be omitted, as allowed by *target_typed_new*. It is a compile-time error to omit `type` if the type cannot be inferred. A *target_typed_new* expression has no type. However, there is an implicit object-creation conversion (§imp-obj-creation-conv) from a *target_typed_new* expression to every type. It is a compile-time error if a *target_typed_new* is used as an operand of a unary or binary operator, or if it is used where it is not subject to an object-creation conversion.
```
KalleOlaviNiemitalo, line 2450: Together with C# 9 [target-typed conditional expressions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/target-typed-conditional-expression), this will then allow

```csharp
class C {
    char M() {
        return new() ? new() : new();
    }
}
```

in which the target types will be inferred as

```csharp
class C {
    char M() {
        return new bool() ? new char() : new char();
    }
}
```

It might be clearest to merge that feature before this one, so that there is no doubt about why target-typed new is allowed in operands of the ternary operator.

BillWagner:
Closing in favor of #1469

PR #994, opened 2023-11-19
Add support for attributes and extern on local functions

RexJaeschke:
This PR will be applied *after* that for V8 PR #869, which adds support for static local functions, a feature to which this PR refers.

The grammar refers to rule *parameter_list*, which is added by V9 feature PR #983, for record classes.


BillWagner:
Closed in favor of #1466

PR #1003, opened 2023-11-22
Fix sole remaining example test failure

jskeet:
Generic attributes will be valid in the future (and Roslyn knows that), so the error code has changed.


BillWagner (approved):
LGTM. This is great progress

PR #1004, opened 2023-11-22
Xml doc bugfixes

### LINKED ISSUE #771: XML doc ID string format for explicit implementation of generic interface
**Describe the bug**
If a method of a type explicitly implements a method of a generic interface, then the name of the private method in ECMA-335 metadata includes the type arguments of the interface, with `<` and `>` around them. In the XML documentation file, Roslyn replaces the `<` and `>` with `{` and `}`, but the standard does not specify this replacement.

It should be specified in §D.4.2 _ID string format_ here: <https://github.com/dotnet/csharpstandard/blob/721ef2964efeb5b4b803e856a00094dc4fe21bb7/standard/documentation-comments.md?plain=1#L637>

The standard does specify the similar replacement of `<` and `>` with `{` and `}` in the argument-list part of the ID string: <https://github.com/dotnet/csharpstandard/blob/721ef2964efeb5b4b803e856a00094dc4fe21bb7/standard/documentation-comments.md?plain=1#L646>

**Example**

```csharp
namespace EiiDoc
{
    public interface IDemo<T>
    {
        void M(IDemo<T> arg);
    }

    public class Demo<T> : IDemo<int>
    {
        /// <summary>Explicit interface implementation</summary>
        void IDemo<int>.M(IDemo<int> arg) {}
    }
}
```

From .NET SDK 6.0.407, I get this metadata and IL:

```msil
.method private hidebysig newslot virtual final 
        instance void  'EiiDoc.IDemo<System.Int32>.M'(class EiiDoc.IDemo`1<int32> arg) cil managed
{
  .override  method instance void class EiiDoc.IDemo`1<int32>::M(class EiiDoc.IDemo`1<!0>)
  // Code size       2 (0x2)
  .maxstack  8
  IL_0000:  nop
  IL_0001:  ret
} // end of method Demo`1::'EiiDoc.IDemo<System.Int32>.M'
```

and this XML documentation file:

```XML
<?xml version="1.0"?>
<doc>
    <assembly>
        <name>EiiDoc</name>
    </assembly>
    <members>
        <member name="M:EiiDoc.Demo`1.EiiDoc#IDemo{System#Int32}#M(EiiDoc.IDemo{System.Int32})">
            <summary>Explicit interface implementation</summary>
        </member>
    </members>
</doc>
```

According to C# 6.0 §D.4.2 as written, the ID string would instead be ```M:EiiDoc.Demo`1.EiiDoc#IDemo<System#Int32>#M(EiiDoc.IDemo{System.Int32})```.

**Expected behavior**
Specify that the characters `.<>` are replaced with `#{}` in the member name. (Note that `.` is not replaced with `#` in the argument list.)

**Additional context**
As with C# 6.0 §D.4.1, this issue is only relevant to C# implementations targeting the CLI.  <https://github.com/dotnet/csharpstandard/blob/721ef2964efeb5b4b803e856a00094dc4fe21bb7/standard/documentation-comments.md?plain=1#L612-L614>

AFAIK, the standard does not specify how the compiler should construct the metadata name of the method that explicitly implements an interface method. However, software can discover the name by reflection, e.g. by calling Type.GetInterfaceMap(Type). If the metadata name differs from the documentation ID string in unspecified ways, that makes it more difficult to implement software that reads the documentation file and locates the documentation of individual methods.

---
[Associated WorkItem - 178982](https://dev.azure.com/msft-skilling/Content/_workitems/edit/178982)

### LINKED ISSUE #772: SZARRAY encoding in XML documentation ID string
**Describe the bug**

§D.4.2 _ID string format_ specifies that, when an array type is encoded in the ID string, the lower bound and size of each dimension is also encoded if known:

<https://github.com/dotnet/csharpstandard/blob/721ef2964efeb5b4b803e856a00094dc4fe21bb7/standard/documentation-comments.md?plain=1#L642>

However, one-dimensional zero-bound unspecified-size arrays (ELEMENT_TYPE_SZARRAY in ECMA-335) are actually encoded with just `[]` rather than `[0:]`.

Also, multidimensional arrays appear to have `0:` for each dimension. According to §16.1, C# array types have a lower bound of zero in every dimension. It might be worth adding a note about this to §D.4.2, especially because the System.Type and System.Reflection.ParameterInfo APIs do not provide methods for querying the bounds of an array type, although System.Array has methods for querying the bounds of an array instance.

<https://github.com/dotnet/csharpstandard/blob/721ef2964efeb5b4b803e856a00094dc4fe21bb7/standard/arrays.md?plain=1#L7>

(I suppose it might be possible to define a non-zero-bound array parameter in a different programming language and reference that with the C# XML documentation comments `cref` attribute, in which case the C# compiler might generate an ID string with the non-zero-bound syntax.)

**Example**

ArrayDoc.csproj:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>net6.0</TargetFramework>
    <LangVersion>6.0</LangVersion>
    <GenerateDocumentationFile>true</GenerateDocumentationFile>
  </PropertyGroup>

</Project>
```

ArrayDoc.cs:

```csharp
public abstract class ArrayDoc
{
    /// <summary>Some method</summary>
    protected abstract void M(int[] a, int[,] b);
}
```

Generated ArrayDoc.xml:

```xml
<?xml version="1.0"?>
<doc>
    <assembly>
        <name>ArrayDoc</name>
    </assembly>
    <members>
        <member name="M:ArrayDoc.M(System.Int32[],System.Int32[0:,0:])">
            <summary>Some method</summary>
        </member>
    </members>
</doc>
```

Note the ID string is not ```M:ArrayDoc.M(System.Int32[0:],System.Int32[0:,0:])```.

**Expected behavior**

Declare that a one-dimensional, zero-bound array type is encoded in the ID string as `[]` rather than `[0:]`.

Perhaps add a note saying that array types defined in C# always have a lower bound of zero, with a cross reference to §16.1.

**Additional context**

I vaguely recall that a one-dimensional, unknown-lower-bound array type might be encoded as `[*]`, but I don't remember where I saw that. It's not ECMA-335 §II.14.2, which uses `[...]` instead.

---
[Associated WorkItem - 178981](https://dev.azure.com/msft-skilling/Content/_workitems/edit/178981)

### LINKED ISSUE #915: Wrong XML in the example or misleading description of <param> tag
>The <param> tag is used to describe parameters. If such a tag is used, the documentation generator must verify that the specified parameter exists and that all parameters are described in documentation comments. If such verification fails, the documentation generator issues a warning.

This comment and D.3.9. example made me think that `name` attribute in `<param>` tag are mandatory, however, in D.5. section `<param>` tags do not have `name` attribute

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: db74e905-523c-dae9-f741-af5e52785a9c
* Version Independent ID: c9b7c7a1-0584-b6e6-95cd-5276788fdda0
* Content: [Documentation comments - C# language specification](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#d33-code)
* Content Source: [standard/documentation-comments.md](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/documentation-comments.md)
* Product: **dotnet-csharp**
* Technology: **csharp-spec**
* GitHub Login: @dotnet-bot
* Microsoft Alias: **wiwagn**

---
[Associated WorkItem - 178987](https://dev.azure.com/msft-skilling/Content/_workitems/edit/178987)

BillWagner:
Fixes three issues:

Fixes #771. The string IDs for methods that are explicit implementations of members of generic interfaces have the `<` and `>` characters replaced with `{` and `}`.
Fixes #772:  single dimension arrays don't specify the lower bound in their string ids. Instead of `[0]` its `[]`.
Fixes #915:  The example didn't include the "name=" attributes on parameters.

The fixes are in three distinct commits to make it easier to review.


standard/documentation-comments.md
```diff
@@ -1031,8 +1031,8 @@ namespace Graphics
         /// This constructor initializes the new Point to
         /// (<paramref name="xPosition"/>,<paramref name="yPosition"/>).
         /// </summary>
-        /// <param><c>xPosition</c> is the new Point's x-coordinate.</param>
-        /// <param><c>yPosition</c> is the new Point's y-coordinate.</param>
+        /// <param name="xPosition"><c>xPosition</c> is the new Point's x-coordinate.</param>
```
jskeet: At some point we should improve the docs to remove the "xPosition is..." bits. But not now :)

jskeet (approved):

standard/documentation-comments.md
BillWagner: I had wanted to do it anyway. Thanks for pointing that out.

jskeet (approved):
Hooray, thanks :)

KalleOlaviNiemitalo:
Now I wonder how the name of a type is formatted in these if it involves an `extern alias`.  Particularly if class C has conversion operators to both X::T and Y::T, where X and Y are extern aliases, and T is the same identifier in both but means different types.

standard/documentation-comments.md
BillWagner: I went ahead and just did it.

KalleOlaviNiemitalo:
The extern alias thing is a bit related to https://github.com/dotnet/docfx/issues/2620

standard/documentation-comments.md
```diff
@@ -691,12 +691,12 @@ The documentation generator observes the following rules when it generates the I
...
-    - Arguments that are arrays are represented as `[` *lowerbound* `:` *size* `,` … `,` *lowerbound* `:` *size* `]` where the number of commas is the rank less one, and the lower bounds and size of each dimension, if known, are represented in decimal. If a lower bound or size is not specified, it is omitted. If the lower bound and size for a particular dimension are omitted, the “`:`” is omitted as well. Jagged arrays are represented by one “`[]`” per level.
+    - Arguments that are multi-dimensional arrays are represented as `[` *lowerbound* `:` *size* `,` … `,` *lowerbound* `:` *size* `]` where the number of commas is the rank less one, and the lower bounds and size of each dimension, if known, are represented in decimal. If a lower bound or size is not specified, it is omitted. If the lower bound and size for a particular dimension are omitted, the “`:`” is omitted as well. Jagged arrays are represented by one “`[]`” per level. Single dimension arrays omit the lower bound when the lower bound is 0 (the default) (§17.1).
```
jskeet:
```suggestion
    - Arguments that are arrays are represented as `[` *lowerbound* `:` *size* `,` … `,` *lowerbound* `:` *size* `]` where the number of commas is the rank less one, and the lower bounds and size of each dimension, if known, are represented in decimal. If a lower bound or size is not specified, it is omitted. If the lower bound and size for a particular dimension are omitted, the “`:`” is omitted as well. Jagged arrays are represented by one “`[]`” per level. Single dimensional arrays omit the lower bound when the lower bound is 0 (the default) (§17.1).
```

MadsTorgersen (approved):

Nigel-Ecma (approved):

gafter (approved):

jskeet:
> Now I wonder how the name of a type is formatted in these if it involves an `extern alias`.

Meeting note: We're not sure either. Please feel free to experiment and either raise a PR or an issue with your findings.

PR #1005, opened 2023-11-22
Link sections on `throw` and handling exceptions

### LINKED ISSUE #776: Information about what happens when an exception is thrown is scattered between two sections.
There are two places in the standard that describe what happens when an exception is thrown:

- The second half of [13.10.6 The throw statement](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/statements#13106-the-throw-statement)
- [21.4 How exceptions are handled](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/exceptions#214-how-exceptions-are-handled)

Those two complement each other, as each contains the information that is not in another. However, they don't reference each other. So, if one reads one of them, most probably one would miss another one. 

Would be helpful to add the cross-links between these two parts of the standard.

---
[Associated WorkItem - 178980](https://dev.azure.com/msft-skilling/Content/_workitems/edit/178980)

BillWagner:
Fixes #776

Links are added in both directions between §21.4 and §13.10.6.


RexJaeschke (changes requested):
Bill, these links look like hard-coded numbers rather than md `[...](...)` links.

jskeet (commented):
Looks good to me other than Rex's comments about the form of the links.

standard/exceptions.md
```diff
@@ -25,7 +25,7 @@ The value of these properties can be specified in calls to the instance construc
...
+When an exception is thrown (§13.10.6), the system searches for the nearest catch clause that can handle the exception, as determined by the run-time type of the exception. First, the current method is searched for a lexically enclosing `try` statement, and the associated `catch` clauses of the `try` statement are considered in order. If that fails, the method that called the current method is searched for a lexically enclosing `try` statement that encloses the point of the call to the current method. This search continues until a `catch` clause is found that can handle the current exception, by naming an exception class that is of the same class, or a base class, of the run-time type of the exception being thrown. A `catch` clause that doesn’t name an exception class can handle any exception.
```
KalleOlaviNiemitalo: I'd prefer linking from here to [21.2 Causes of exceptions](https://github.com/dotnet/csharpstandard/blob/1b8e8c35345d9cbb3838de7cd40c23da1070c215/standard/exceptions.md#212-causes-of-exceptions), which already includes a link to [13.10.6 The throw statement](https://github.com/dotnet/csharpstandard/blob/1b8e8c35345d9cbb3838de7cd40c23da1070c215/standard/statements.md#13106-the-throw-statement).  This would make it clearer that the paragraph also applies to NullReferenceException etc. when thrown by the runtime without a `throw` statement.

BillWagner:
> Bill, these links look like hard-coded numbers rather than md `[...](...)` links.

@RexJaeschke The tool should fix those once we merge this PR.

standard/exceptions.md
BillWagner: Making the comment into a suggestion:

```suggestion
When an exception is thrown (§21.2), the system searches for the nearest catch clause that can handle the exception, as determined by the run-time type of the exception. First, the current method is searched for a lexically enclosing `try` statement, and the associated `catch` clauses of the `try` statement are considered in order. If that fails, the method that called the current method is searched for a lexically enclosing `try` statement that encloses the point of the call to the current method. This search continues until a `catch` clause is found that can handle the current exception, by naming an exception class that is of the same class, or a base class, of the run-time type of the exception being thrown. A `catch` clause that doesn’t name an exception class can handle any exception.
```

MadsTorgersen (approved):

Nigel-Ecma (approved):

standard/exceptions.md
gafter:
```suggestion
When an exception is thrown (§21.2), the system searches for the nearest catch clause that can handle the exception, as determined by the run-time type of the exception. First, the current method is searched for a lexically enclosing `try` statement, and the associated `catch` clauses of the `try` statement are considered in order. If that fails, the method that called the current method is searched for a lexically enclosing `try` statement that encloses the point of the call to the current method. This search continues until a `catch` clause is found that can handle the current exception, by naming an exception class that is of the same class, or a base class, of the run-time type of the exception being thrown. A `catch` clause that doesn’t name an exception class can handle any exception.
```

gafter (approved):

RexJaeschke (approved):

PR #1006, opened 2023-11-22
Wordsmithing return by ref

### LINKED ISSUE #805: clarify the *return-by-ref* expression evaluation
See https://github.com/dotnet/csharpstandard/pull/795/files#r1217459209

The language for *return-by-ref* implies a different evaluation. Could use some wordsmithing.

---
[Associated WorkItem - 178988](https://dev.azure.com/msft-skilling/Content/_workitems/edit/178988)

BillWagner:
Fixes #805

Clarify that return-by-ref expressions are evaluated the same as return-by-value expressions. The distinction is that for return-by-ref, the result must be a variable reference.


standard/statements.md
```diff
@@ -1523,7 +1523,7 @@ It is a compile-time error for a `return` statement to appear in a `finally` blo
 
 A `return` statement is executed as follows:
 
-- For a return-by-value, *expression* is evaluated and its value is converted to the effective return type of the containing function by an implicit conversion. The result of the conversion becomes the result value produced by the function. For a return-by-ref, a reference to the *variable_reference* designated by *expression* becomes the result produced by the function. That result is a variable. If the enclosing method’s return-by-ref includes `readonly`, the resulting variable is read-only.
+- For a return-by-value, *expression* is evaluated and its value is converted to the effective return type of the containing function by an implicit conversion. The result of the conversion becomes the result value produced by the function. For a return-by-ref, the *expression* is evaluated the same, but the result must be a reference to a *variable_reference* designated by *expression*. If the enclosing method’s return-by-ref includes `readonly`, the resulting variable is read-only.
```
```suggestion
- For a return-by-value, *expression* is evaluated and its value is converted to the effective return type of the containing function by an implicit conversion. The result of the conversion becomes the result value produced by the function. For a return-by-ref, the *expression* is evaluated, and the result shall be classified as a variable. If the enclosing method’s return-by-ref includes `readonly`, the resulting variable is read-only.
```

jskeet: Is "reference to a variable_reference" correct? And it's not clear how "designated by *expression*" comes into this... possibly remove?

jskeet (commented):
Getting there, but I think we'll need some more group discussion.

standard/statements.md
BillWagner, line 1526: Updated text from meeting:

```suggestion
- For a return-by-value, *expression* is evaluated and its value is converted to the effective return type of the containing function by an implicit conversion. The result of the conversion becomes the result value produced by the function. For a return-by-ref, the *expression* is evaluated in the same way as a return-by-value expression, but the result must designate a *variable_reference*. If the enclosing method’s return-by-ref includes `readonly`, the resulting variable is read-only.
```

gafter (approved):

MadsTorgersen (approved):

jskeet (approved):

PR #1008, opened 2023-11-23
Add note for the purpose of input parameters

### LINKED ISSUE #845: §15.10.1 Add note for purpose of in parameters in operators
#842 adds a fix so that operators can have `in` parameters. We might want to add a note to explain the purpose of this (to avoid copying large values) - but exactly where that should go (operators, or `in` parameters in general?) could take some discussion, so this is going to be post-v7.

jskeet:
Fixes #845.

I'm not at all convinced of the tone or precision of this - it's effectively a straw man for further discussion.

(Requesting review from Bill and Nigel just for preliminary reactions.)

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -2156,6 +2156,8 @@ Input parameters are not allowed on functions declared as an iterator ([§15.14]
 
 In a method that takes input parameters, it is possible for multiple names to represent the same storage location.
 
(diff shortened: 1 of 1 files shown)


BillWagner:
I think this looks reasonable. It's the combination of two factors for efficiency:  

1. `struct` types may be stored on the stack, minimizing heap allocations and GC work.
2. `in` parameters aren't copied, and can't be modified.

Adding @jaredpar for his thoughts.

jaredpar (approved):
Looks good to me

MadsTorgersen (approved):

Nigel-Ecma (approved):

gafter (approved):

PR #1011, opened 2023-11-24
Delete duplicated Note in §6.2.5

Nigel-Ecma:
A note was duplicated (historically it was an in para note in v5, a standalone note in v6, and now in v7 is both – I'd blame git ;-))


jskeet (approved):

RexJaeschke (approved):

jskeet:
Merging given two approvals.

PR #1012, opened 2023-11-24
Add example for unreachable code

### LINKED ISSUE #937: Add example for definite assignment due to unreachable switch label
See https://github.com/dotnet/csharpstandard/pull/757#discussion_r1326604587

In 9.4.4.7 we have:

> - If the control transfer was due to an unreachable switch label, then the state of *v* is "definitely assigned".

An example would make this rule more clear. But, it would break the flow of the text, and it is an uncommon case.

Deferred to discuss post 7.3

---
[Associated WorkItem - 178984](https://dev.azure.com/msft-skilling/Content/_workitems/edit/178984)

BillWagner:
Fixes #937

I admit this rule confused me, and I asked to add the example.

After reading the updated, I'm ambivalent about adding the example. I don't want to move it into the bullet list, as that breaks the flow even more.

Let's discuss if it's needed at the meeting.

### DIFF
--- a/standard/variables.md
+++ b/standard/variables.md
@@ -343,6 +343,19 @@ The definite-assignment state of *v* on the control flow transfer to a reachable
   - If *v* is a pattern variable declared in the *switch_label*: “definitely assigned”.
   - Otherwise, the state of *v* is the same as the stat of *v* after *expr*.
 
+> *Example*: The third rule eliminates the need for the compiler to warn if an unassigned variable is accessed in unreachable code. The state of *b* is "definitely assigned" in the unreachable switch label `case 2 when b`.
+>
+> ```csharp
+> bool b;
+> switch (1) 
+> {
+>     case 2 when b: // b is definitely assigned here.
+>     break;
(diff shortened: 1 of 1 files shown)


jskeet (approved):
Gosh, interesting!

standard/variables.md
```diff
@@ -343,6 +343,19 @@ The definite-assignment state of *v* on the control flow transfer to a reachable
   - If *v* is a pattern variable declared in the *switch_label*: “definitely assigned”.
   - Otherwise, the state of *v* is the same as the stat of *v* after *expr*.
 
+> *Example*: The third rule eliminates the need for the compiler to warn if an unassigned variable is accessed in unreachable code. The state of *b* is "definitely assigned" in the unreachable switch label `case 2 when b`.
```
KalleOlaviNiemitalo: Wouldn't that be an error rather than a warning?

jskeet (approved):

gafter (approved):

PR #1013, opened 2023-11-24
Add example of compile-time error

### LINKED ISSUE #938: Example clarifying compile-time type validation in patterns
See https://github.com/dotnet/csharpstandard/pull/757#discussion_r1326609450

Original comment on section 11.2.1

> I think we may want an example here to clarify this. I believe I get it, but it could be confusing.
> 
> For example, consider:
> 
> ```csharp
> if (x is string s)
> ```
> 
> It's fine for the compile-time type of `x` to be `object`, even though a value whose execution type is `object` *won't* match. But the expression type being `object` is okay because the value *might* be a `string` reference. Whereas if the compile-time type of `x` is (say) `Stream` then this would be a compile-time error.

Note that in #934 I added the clarifying change from "it is an error" to "it is a compile-time error" referenced in the same conversation.

The additional example is deferred to after C# 7

---
[Associated WorkItem - 178983](https://dev.azure.com/msft-skilling/Content/_workitems/edit/178983)

BillWagner:
Fixes #938

Add an example of a type pattern to demonstrate when a pattern generates a compile-time error.

### DIFF
--- a/standard/patterns.md
+++ b/standard/patterns.md
@@ -22,23 +22,23 @@ A *declaration_pattern* and a *var_pattern* can result in the declaration of a l
 
 Each pattern form defines the set of types for input values that the pattern may be applied to. A pattern `P` is *applicable to* a type `T` if `T` is among the types whose values the pattern may match. It is a compile-time error if a pattern `P` appears in a program to match a pattern input value ([§11.1](patterns.md#111-general)) of type `T` if `P` is not applicable to `T`.
 
-> *Example*: The following example generates a compile-time error because the compile-time type of `v` is `Stream`. A variable of type `Stream` can never be an expression of type `string`:
+> *Example*: The following example generates a compile-time error because the compile-time type of `v` is `Stream`. A variable of type `Stream` can never have a value that is reference compatible with `string`:
 >
 > ```csharp
 > Stream v = OpenDataFile(); // compile-time type of 'v' is 'Stream'
 > if (v is string) // compile-time error
 > {
->     /* code assuming v is a string*/ 
+>     // code assuming v is a string
 > }
 > ```
 >
-> However, the following doesn't generate a compile-time error because the compile-time type of `v` is `object`. A variable of type `object` could be an expression of type `string`:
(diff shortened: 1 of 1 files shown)


standard/patterns.md
```diff
@@ -22,6 +22,28 @@ A *declaration_pattern* and a *var_pattern* can result in the declaration of a l
 
 Each pattern form defines the set of types for input values that the pattern may be applied to. A pattern `P` is *applicable to* a type `T` if `T` is among the types whose values the pattern may match. It is a compile-time error if a pattern `P` appears in a program to match a pattern input value ([§11.1](patterns.md#111-general)) of type `T` if `P` is not applicable to `T`.
 
+> *Example*: The following example generates a compile-time error because the compile-time type of `v` is `Stream`. A variable of type `Stream` can never be an expression of type `string`:
+>
+> ```csharp
+> Stream v = OpenDataFile(); // compile-time type of 'v' is 'Stream'
+> if (v is string) // compile-time error
+> {
+>     /* code assuming v is a string*/
```
jskeet: Nit: space before `*/` - or just change it to a `//` comment.

standard/patterns.md
```diff
@@ -22,6 +22,28 @@ A *declaration_pattern* and a *var_pattern* can result in the declaration of a l
 
 Each pattern form defines the set of types for input values that the pattern may be applied to. A pattern `P` is *applicable to* a type `T` if `T` is among the types whose values the pattern may match. It is a compile-time error if a pattern `P` appears in a program to match a pattern input value ([§11.1](patterns.md#111-general)) of type `T` if `P` is not applicable to `T`.
 
+> *Example*: The following example generates a compile-time error because the compile-time type of `v` is `Stream`. A variable of type `Stream` can never be an expression of type `string`:
```
jskeet: I'm not sure about "expression of type `string`" here. How about:

> A variable of type `Stream` can never have a value which is a reference compatible with `string`

? Or some variation?

Nigel-Ecma (approved):

gafter (approved):

PR #1020, opened 2023-11-30
Add extra example in switch/case

jskeet:
Let's wait until next meeting before merging this, as I worry that it might muddy the waters a bit.


BillWagner (approved):
This LGTM @jskeet 

It matches our discussion from yesterday.

PR #1022, opened 2023-11-30
Fix overload resolution for non-returning async lambdas matching task-returning delegates

### LINKED ISSUE #625: Check preference for Func<Task> over Action
The following code prints `Func<Task>`:

```csharp
using System;
using System.Threading.Tasks;

class Program
{
    static void Main()
    {
        M(async () => {});
    }

    static void M(Action action)
    {
        Console.WriteLine("Action");
    }

    static void M(Func<Task> task)
    {
        Console.WriteLine("Func<Task>");
    }
}
```

So `Func<Task>` is a better conversion target than `Action` in this case. I'm not sure where that's specified at the moment.

jskeet:
Fixes #625 (which contains more details)

We can merge this when you're happy, Bill.


BillWagner (approved):
This LGTM. I still wonder what `S` was (or if it was a typo).

jskeet:
> This LGTM. I still wonder what `S` was (or if it was a typo).

Left over from the clause above, I suspect...

PR #1024, opened 2023-11-30
Fix codefence

333fred:
Fixed a codefence to be around the fully qualified type, instead of just the typename.


jskeet (approved):
Thanks!

PR #1025, opened 2023-11-30
Add example annotation to 3 new examples

RexJaeschke:
@BillWagner, I changed the 2 examples in patterns.md, so they don't need any non-standard supporting types/methods. I believe I have preserved your original intent.

BTW, should you want to add annotations to examples in the future, add the following line right before the line starting the source code, and then replace/delete options as you need (according to https://github.com/dotnet/csharpstandard/blob/standard-v7/tools/ExampleExtractor/ExtractorAndTesterUsersGuide.md):

`> <!-- Example: {template:"xxx", name:"xxx", replaceEllipsis:true, customEllipsisReplacements: ["xxx", "xxx"], expectedErrors:["CSxxx"], expectedWarnings:["CSxxx"], ignoredWarnings:["CSxxx"], inferOutput:true, expectedOutput:["xxx"], ignoreOutput:true, expectedException:"xxx", additionalFiles:["xxx.cs"]} -->`

The leading `> ` is only needed if the example is inside an Example or Note.


BillWagner (approved):
Thanks @RexJaeschke 

This looks good to me. I'll :shipit: now.

PR #873, opened 2023-08-04
Add support for new pattern kinds

RexJaeschke:
Prior to V8, we have a term “switch expression” defined as the *expression* inside parens in a `switch` statement.

Now, we’re introducing a switch as an expression, and we have a new grammar rule called *switch_expression*. The difference between “switch expression” (without a \*…\* fence or separating underscore) and \*switch_expression\* likely is too subtle. So, how to differentiate them?

Although C calls such a switch expression a “controlling expression” (and uses that term as well for `if`, `while`, `for`, and `do` statements), we already state, “The ***governing type*** of a `switch` statement is established by the switch expression.” 

**As such, I have replaced all occurrences of “switch expression” (which exist only in the statements chapter) with “switch’s governing expression.”**

Fixes #576.
Fixes #300.

PR #1026, opened 2023-12-07
Add support for pattern additions

RexJaeschke:
**This PR *might* be incomplete; read below.**

Regarding the [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/patterns3.md):

1. Type patterns: Grammar rule *primary_pattern* is defined as something different from *pattern*, but as best as I can tell, they are exactly the same thing, so I replaced all occurrences of *primary_pattern* with *pattern*. **Is that OK?** (Also, the use of *primary* suggests here are secondary/other forms, but I don't see what they are.)
1. Relational patterns: the proposed grammar shows *relational_expression* as the operand of each relational pattern, but the accompanying narrative goes on to constrain such relational expressions to constants only. So what not use *constant_expression" instead? And that's what I did. **Am I missing something?**
1. Relational patterns: The proposal did not include support for enums, but as they *are* supported, I included them in the spec.
1. Logical patterns: The proposal calls these *pattern combinators*; however, that term is *not* used elsewhere on MS's tutorial pages, so I used the more obvious (and elsewhere used) *logical patterns* instead.
1. Logical patterns: The fact that the rule *logical_pattern* always expands to *disjunctive_pattern* may seem unnecessary, but it allows the placement of *logical_pattern* in the rule *pattern* rather than pushing the (odd-looking) *disjunctive_pattern* up there instead. (We use this approach elsewhere in the grammar.)
1. The (substantial) second half of the proposal, "Open Issues with Proposed Changes," raises questions/issues. However, unlike all (or at least most) other MS proposals I've worked with, there is no follow-on discussion to say how these topics were actually addressed in the final implementation. Certainly, if any of the suggested approaches were taken, more words will need to be added to this PR to reflect that. **I'll need help from someone at MS to resolve this.**

***This PR was created based on the assumption that the edits proposed by PR #873 for V8-related pattern-matching features will be adopted, and no other non-trivial edits will be made to that part of V8. When that PR is finally merged, this PR might need adjustment to reflect any changes in the V8 proposal review/adoption.***


gafter:
Regarding the OP, I have the following comments:

(1) Re "Type patterns: Grammar rule *primary_pattern* is defined as something different from *pattern*, but as best as I can tell, they are exactly the same thing, so I replaced all occurrences of *primary_pattern* with *pattern*. **Is that OK?** (Also, the use of *primary* suggests here are secondary/other forms, but I don't see what they are.)"

Several kinds of `pattern`s are not `primary_pattern`s. Specifically: `disjunctive_pattern`, `conjunctive_pattern`, and `negated_pattern`.

(2) Re: "Relational patterns: the proposed grammar shows *relational_expression* as the operand of each relational pattern, but the accompanying narrative goes on to constrain such relational expressions to constants only. So what not use *constant_expression" instead? And that's what I did. **Am I missing something?**"

Yes, there is a reason. The grammar rule for `constant_expression` makes it equivalent to `expression` (see section 12.23). So it includes, for example, operators at a looser precedence level, such as `equality_expression ` which in this context it ought not do. Your change makes the following code ambiguous:

`x = a is < b == c;`

Because the left operand of the `== c` could be `a is < b` (correct) or it could be `b` (incorrect) due to your change.

(3) Re: "Relational patterns: The proposal did not include support for enums, but as they *are* supported, I included them in the spec."

Yes, good.

(4) Re: "Logical patterns: The proposal calls these *pattern combinators*; however, that term is *not* used elsewhere on MS's tutorial pages, so I used the more obvious (and elsewhere used) *logical patterns* instead."

Yes, good.

(5) Re: "Logical patterns: The fact that the rule *logical_pattern* always expands to *disjunctive_pattern* may seem unnecessary, but it allows the placement of *logical_pattern* in the rule *pattern* rather than pushing the (odd-looking) *disjunctive_pattern* up there instead. (We use this approach elsewhere in the grammar.)"

Okay, but I don't see a problem with using `disjunctive_pattern` as a right-hand-side in the rule for `pattern`.

(6) Re: "The (substantial) second half of the proposal, "Open Issues with Proposed Changes," raises questions/issues. However, unlike all (or at least most) other MS proposals I've worked with, there is no follow-on discussion to say how these topics were actually addressed in the final implementation. Certainly, if any of the suggested approaches were taken, more words will need to be added to this PR to reflect that. **I'll need help from someone at MS to resolve this.**"

We should open issues for each of the open issues (or one issue for all of them) to make sure we address them.

standard/patterns.md
```diff
@@ -2,7 +2,7 @@
 
 ## 11.1 General
 
-A ***pattern*** is a syntactic form that can be used with the `is` operator ([§12.12.12](expressions.md#121212-the-is-operator)) and in a *switch_statement* ([§13.8.3](statements.md#1383-the-switch-statement)) to express the shape of data against which incoming data is to be compared. A pattern is tested against the *expression* of a switch statement, or against a *relational_expression* that is on the left-hand side of an `is` operator, each of which is referred to as a ***pattern input value***.
+A ***pattern*** is a syntactic form that can be used with the `is` operator ([§12.12.12](expressions.md#121212-the-is-operator)) and in a *switch_statement* ([§13.8.3](statements.md#1383-the-switch-statement)) to express the shape of data against which incoming data is to be compared. A pattern is tested against the *expression* of a switch statement, or against a *relational_expression* that is on the left-hand side of an `is` operator, each of which is referred to as a ***pattern input value***. Patterns may be combined using Boolean logic.
```
gafter, line 5: I don't know what "Patterns may be combined using Boolean logic" means. Yes, there are pattern forms for disjunction, conjunction, and negation, but the "Boolean logic" sentence suggests something more general.

standard/patterns.md
```diff
@@ -140,6 +144,157 @@ Given a pattern input value ([§11.1](patterns.md#111-general)) *e*, if *designa
...
+    : '<'  constant_expression
+    | '<=' constant_expression
+    | '>'  constant_expression
+    | '>=' constant_expression
+    ;
+```
+
+Relational patterns support the relational operators `<`, `<=`, `>`, and `>=` on all of the built-in types that support such binary relational operators with both operands having the same type: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `nint`, `nuint`, and enums.
+
+It is a compile-time error if `constant_expression`is `double.NaN`, `float.NaN`, or `null_literal`.
```
gafter, line 198: Not exactly. It is not a syntactic constraint. It is a compile-time error if it evaluates to one of those values.

standard/patterns.md
```diff
@@ -140,6 +144,157 @@ Given a pattern input value ([§11.1](patterns.md#111-general)) *e*, if *designa
...
+> Too high
+> Unknown
+> Acceptable
+> ```
+>
+> *end example*
+
+§logical-pattern-new-clause Logical pattern
+
+A *logical_pattern* is used to negate a pattern input value ([§11.1](patterns.md#111-general)) or to combine that value with a pattern using a Boolean operator.
```
gafter, line 231: No, it doesn't negate the input value. It inverts (in the boolean sense) the result of the pattern-match. Similarly, the combination doesn't happen with the value, but with the result of pattern-match tests.

standard/patterns.md
```diff
@@ -140,6 +144,157 @@ Given a pattern input value ([§11.1](patterns.md#111-general)) *e*, if *designa
...
+    ;
+```
+
+`not`, `and`, and `or` are collectively called ***pattern operators***.
+
+A *negated_pattern* matches if the pattern being negated does not match, and vice versa. A *conjunctive_pattern* requires both patterns to match. A *disjunctive_pattern* requires either pattern to match. Unlike their language operator counterparts, `&&` and `||`, `and` and `or` are *not* short-circuiting operators.
+
+> *Note*: As indicated by the grammar, `not` has precedence over `and`, which has precedence over `or`. This can be explicitly indicated or overridden by using parentheses. *end note*
+
+When a *pattern* is used with `is`, any pattern operators in that *pattern* have higher precedence than their logical operator counterparts. Otherwise, those pattern operators have lower precedence.
```
gafter, line 260: I'm not sure what this means. An example would be helpful.

standard/patterns.md
```diff
@@ -12,9 +12,13 @@ A pattern may have one of the following forms:
 
 ```ANTLR
 pattern
-    : declaration_pattern
+    : '(' pattern ')'
```
gafter, line 15: It is worth a sentence to say what this means.

standard/patterns.md
Nigel-Ecma: If is an operation of many names: not, invert & negate, and in C# it is indeed called logical negation (§12.9.4).

gafter: I'm not concerned with which of those names we use. I'm pointing out that the meaning is not "to negate a pattern input value", but rather to negate the result of determining whether or not the input value matches the pattern.

BillWagner:
Closing in favor of #1460

PR #1028, opened 2023-12-11
Create general by-reference clause

### LINKED ISSUE #799: De-triplicate input, output, and reference parameters.
See https://github.com/dotnet/csharpstandard/pull/795/files#r1212255596

The description of value parameters, input parameters and reference parameters contain repeated text. Remove duplication.

---
[Associated WorkItem - 178979](https://dev.azure.com/msft-skilling/Content/_workitems/edit/178979)

BillWagner:
Fixes #799

There are several declarations that are duplicated, or triplicated in the description of `in`, `out`, and `ref` parameters.

This PR pushes those into a common section on "by-reference parameters". Each subsequent section describes the particular keyword.

The diffs are less helpful than I'd hoped, so I've annotated the changes to make reviews easier.


standard/classes.md
```diff
@@ -2140,35 +2140,31 @@ When a formal parameter is a value parameter, the corresponding argument in a me
...
 
-> *Note*: As with reference variables the referent of an input parameter can be changed using the ref assignment (`= ref`) operator, however the value stored in the referent itself cannot be changed. *end note*
+A parameter declared with one of the `in`, `ref` or `out` modifiers is a ***by-reference*** parameter. A by-reference parameter is a local reference variable ([§9.7](variables.md#97-reference-variables-and-returns)) that gets its initial referent from the corresponding argument supplied in the method invocation.
 
-When a formal parameter is an input parameter, the corresponding argument in a method invocation shall consist of either the keyword `in` followed by a *variable_reference* ([§9.2.8](variables.md#928-input-parameters)) of the same type as the formal parameter, or an *expression* for which an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from that argument expression to the type of the corresponding parameter. A variable shall be definitely assigned before it can be passed as an input parameter.
+> *Note*: The referent of a by-reference parameter can be changed using the ref assignment (`= ref`) operator.
```
BillWagner, line 2149: This note was only in the `in` and `ref` sections. However, it applies to `out` parameters as well.

standard/classes.md
```diff
@@ -2140,35 +2140,31 @@ When a formal parameter is a value parameter, the corresponding argument in a me
...
 
-It is a compile-time error to modify the value of an input parameter.
-
-Within a method, an input parameter is always considered definitely assigned.
+When a formal parameter is a by-reference parameter, the corresponding argument in a method invocation shall consist of the corresponding keyword, `in`, `ref`, or `out`, followed by a *variable_reference* ([§9.5](variables.md#95-variable-references)) of the same type as the formal parameter. However, when the formal parameter is an `in` parameter, the argument may be an *expression* for which an implicit conversion ([§10.2](conversions.md#102-implicit-conversions)) exists from that argument expression to the type of the corresponding parameter.
```
BillWagner, line 2151: This paragraph was almost identical in all sub-clauses. I propose this edit, where the exception for `in` is declared in the general clause. An alternative would be to keep the (almost) triplicate code in each of the three sub-clauses.

standard/classes.md
```diff
@@ -2140,35 +2140,31 @@ When a formal parameter is a value parameter, the corresponding argument in a me
...
 
-Input parameters are not allowed on functions declared as an iterator ([§15.14](classes.md#1514-iterators)) or async function ([§15.15](classes.md#1515-async-functions)).
+By-reference parameters are not allowed on functions declared as an iterator ([§15.14](classes.md#1514-iterators)) or async function ([§15.15](classes.md#1515-async-functions)).
 
-In a method that takes input parameters, it is possible for multiple names to represent the same storage location.
-
-> *Note*: The primary purpose of input parameters is for efficiency. When the type of a method parameter is a large struct (in terms of memory requirements), it is useful to be able to avoid copying the whole value of the argument when calling the method. Input parameters allow methods to refer to existing values in memory, while providing protection against unwanted changes to those values. *end note*
+In a method that takes multiple by-reference parameters, it is possible for multiple names to represent the same storage location.
```
BillWagner, line 2155: This was interesting. The original text disallowed `in` on iterators and async methods. For `ref`, only iterators were disallowed (incorrect: `ref` is disallowed for async). For `out` parameters, it was only disallowed for iterators and `partial` methods. The statement on `partial` methods is now in the section for `out`, and the other restrictions are here.

standard/classes.md
```diff
@@ -2229,20 +2225,16 @@ In a method that takes reference parameters, it is possible for multiple names t
...
-Within a method, just like a local variable, an output parameter is initially considered unassigned and shall be definitely assigned before its value is used.
+A parameter declared with an `out` modifier is an output parameter. Within a method, just like a local variable, an output parameter is initially considered unassigned and shall be definitely assigned before its value is used. Every output parameter of a method shall be definitely assigned before the method returns.
 
-Every output parameter of a method shall be definitely assigned before the method returns.
+A method declared as a partial method ([§15.6.9](classes.md#1569-partial-methods)) may not have output parameters.
 
-A method declared as a partial method ([§15.6.9](classes.md#1569-partial-methods)) or an iterator ([§15.14](classes.md#1514-iterators)) may not have output parameters.
-
-Output parameters are typically used in methods that produce multiple return values.
+> *Note: Output parameters are typically used in methods that produce multiple return values. *end note*
```
BillWagner, line 2234: This was normative in the existing draft. I don't think it's required to be normative, so I made this a note.

standard/classes.md
```diff
@@ -2140,35 +2140,31 @@ When a formal parameter is a value parameter, the corresponding argument in a me
...
+In a method that takes multiple by-reference parameters, it is possible for multiple names to represent the same storage location.
 
-#### 15.6.2.4 Reference parameters
+##### 15.6.2.3 Input parameters
 
-A parameter declared with a `ref` modifier is a reference parameter. A reference parameter is a local reference variable ([§9.7](variables.md#97-reference-variables-and-returns)) that gets its initial referent from the corresponding argument supplied in the method invocation.
+A parameter declared with an `in` modifier is a input parameter. The argument corresponding to an `in` parameter is either a variable existing at the point of the method invocation, or one created by the implementation ([§12.6.2.3](expressions.md#12623-run-time-evaluation-of-argument-lists)) in the method invocation. A variable shall be definitely assigned before it can be passed as an input parameter. Within a method, an input parameter is always considered definitely assigned.
```
BillWagner: The rules on definite assignments have the "2 out of 3" problem. The rules for `in` and `ref` are identical: The variable representing the argument must be definitely assigned before using it as an argument, and the parameter is considered definitely assigned in the entirety of the method. The rules for `out` are different. Here, I chose to keep the three distinct copies. I think it read better.

BillWagner (commented):
Adding comments for reviewers because the diffs aren't as useful as I'd hoped.

standard/classes.md
```suggestion
A parameter declared with an `in` modifier is a input parameter. The argument corresponding to an `in` parameter is either a variable existing at the point of the method invocation, or one created by the implementation ([§12.6.2.3](expressions.md#12623-run-time-evaluation-of-argument-lists)) in the method invocation. A variable shall be definitely assigned before it can be passed as an argument for an input parameter. Within a method, an input parameter is always considered definitely assigned.
```

See whether this works as a change - I think *something* is needed, to differentiate argument/parameter.

standard/classes.md
```diff
@@ -2140,35 +2140,31 @@ When a formal parameter is a value parameter, the corresponding argument in a me
...
+> *Note*: The primary purpose of input parameters is for efficiency. When the type of a method parameter is a large struct (in terms of memory requirements), it is useful to be able to avoid copying the whole value of the argument when calling the method. Input parameters allow methods to refer to existing values in memory, while providing protection against unwanted changes to those values. *end note*
 
-Within a method, a reference parameter is always considered definitely assigned.
+##### 15.6.2.4 Reference parameters
 
-A method declared as an iterator ([§15.14](classes.md#1514-iterators)) may not have reference parameters.
+A parameter declared with a `ref` modifier is a reference parameter. A variable shall be definitely assigned before it can be passed as a reference parameter. Within a method, a reference parameter is always considered definitely assigned.
```
```suggestion
A parameter declared with a `ref` modifier is a reference parameter. A variable shall be definitely assigned before it can be passed as an argument for a reference parameter. Within a method, a reference parameter is always considered definitely assigned.
```

(If we change my suggestion earlier, we should change this one too.)

standard/classes.md
```diff
@@ -2229,20 +2225,16 @@ In a method that takes reference parameters, it is possible for multiple names t
...
+##### 15.6.2.5 Output parameters
 
-Within a method, just like a local variable, an output parameter is initially considered unassigned and shall be definitely assigned before its value is used.
+A parameter declared with an `out` modifier is an output parameter. Within a method, just like a local variable, an output parameter is initially considered unassigned and shall be definitely assigned before its value is used. Every output parameter of a method shall be definitely assigned before the method returns.
 
-Every output parameter of a method shall be definitely assigned before the method returns.
+A method declared as a partial method ([§15.6.9](classes.md#1569-partial-methods)) may not have output parameters.
```
```suggestion
A method declared as a partial method ([§15.6.9](classes.md#1569-partial-methods)) shall not have output parameters.
```

(To avoid the "does may not mean it's optional" ambiguity. I'm tempted to look for every occurrence of "may not" in the spec...)

standard/classes.md
jskeet: Agreed, anything with "typically" shouldn't be normative.

jskeet (approved):
I've looked at the new text more than the change (as you noted the diffs weren't helpful). A couple of suggestions, but otherwise looks good as a reconciliation.

Would be good to get feedback from others as well though.

standard/classes.md
BillWagner: There are 16 occurrences. I'll add them to this PR.

standard/classes.md
BillWagner: I like it. I added it.

standard/classes.md
BillWagner: Update: three of the occurrences were refering to conditions that may or may not be true. Those remain.

standard/classes.md
```diff
@@ -2140,35 +2140,31 @@ When a formal parameter is a value parameter, the corresponding argument in a me
...
 
-#### 15.6.2.3 Input parameters
+#### §by-reference-parameters By-reference parameters
 
-A parameter declared with an `in` modifier is an input parameter. An input parameter is a local reference variable ([§9.7](variables.md#97-reference-variables-and-returns)) that gets its initial referent from the corresponding argument supplied in the method invocation. That argument is either a variable existing at the point of the method invocation, or one created by the implementation ([§12.6.2.3](expressions.md#12623-run-time-evaluation-of-argument-lists)) in the method invocation.
+##### §by-reference-general General
 
-> *Note*: As with reference variables the referent of an input parameter can be changed using the ref assignment (`= ref`) operator, however the value stored in the referent itself cannot be changed. *end note*
+A parameter declared with one of the `in`, `ref` or `out` modifiers is a ***by-reference*** parameter. A by-reference parameter is a local reference variable ([§9.7](variables.md#97-reference-variables-and-returns)) that gets its initial referent from the corresponding argument supplied in the method invocation.
```
jskeet: "that gets its" feels a little more colloquial than our normal spec language, but I think it's fine. We *could* change it to:

"A by-reference parameter is a local reference variable; the initial referent is obtained from the corresponding argument supplied in the method invocation." (Or similar.)

Separate note: is "method invocation" right, or should this be "member invocation"?

jskeet: (And do delegate invocations count as member invocations? We're back to wanting "function"...)

jskeet: After consideration: We're in 15.6, which is "Methods". Other function members will presumably refer here, and that's fine.

```suggestion
A parameter declared with one of the `in`, `ref` or `out` modifiers is a ***by-reference*** parameter. A by-reference parameter is a local reference variable ([§9.7](variables.md#97-reference-variables-and-returns)); the initial referent is obtained from the corresponding argument supplied in the method invocation.
```

jskeet (approved):

Issue #1031, opened 2024-01-09
Make AsyncMethodBuilderAttribute a "Reserved attribute"


RexJaeschke:
In V7, in [§15.15.1](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#15151-general), we introduced the new attribute type `AsyncMethodBuilderAttribute` and in the library annex [§C.3](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271), we showed its declaration.

It seems to me that this attribute needs to be recognized by the compiler, so it needs to have its own section in [§22.5, Reserved attributes](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/attributes.md#225-reserved-attributes). If members agree, I'll write up the PR to add that (small/simple) section.

jskeet:
We agree, the attribute should be there. @RexJaeschke will create a PR.

RexJaeschke:
Closed by https://github.com/dotnet/csharpstandard/pull/1035.

PR #1035, opened 2024-01-11
Make AsyncMethodBuilder a reserved attribute

RexJaeschke:
This addresses Issue #1031.

Jon/Bill, as this fix is trivial, let's merge it without further TG2 discussion.


jskeet (approved):
(Might "described" be better than "discussed"? I'm not that bothered.)

standard/attributes.md
```diff
@@ -768,6 +769,10 @@ If a program uses a type or member that is decorated with the `Obsolete` attribu
 >
 > *end example*
 
+### §AsyncMethodBuilderAttribute The AsyncMethodBuilder attribute
+
+This attribute is discussed in [§15.15.1](classes.md#15151-general).
```
BillWagner: Since @jskeet mentioned it, I agree with "described" is better. Note: I searched the text, and I found "discussed" was used 13 times, and "described" was used 138 times. We're well on the way to "described".

```suggestion
This attribute is described in [§15.15.1](classes.md#15151-general).
```

BillWagner (approved):
I approve this as well.

Since we're here, I made and applied a suggestion based on Jon's comment.

Issue #857, opened 2023-07-14
C# 7.x §15.15.1 meaning of "returns a value" in async function


KalleOlaviNiemitalo:
**Describe the bug**

In the C# 7.x draft, the phrase "returns a value" is used in §15.6.1 (Methods / General) and §15.15.1 (Async functions / General), but not with the same meaning.

In §15.6.1:

> If `ref` is present, the method is ***returns-by-ref***; otherwise, if *return_type* is `void`, the method is ***returns-no-value***; otherwise, the method is ***returns-by-value***.
>
> The *return_type* of a method declaration specifies the type of the result, if any, returned by the method. A returns-no-value method does not return a value. A returns-by-ref method returns a *variable_reference* ([§9.5](variables.md#95-variable-references)), that is optionally read-only. A returns-by-value method returns a value. If the declaration includes the `partial` modifier, then *return_type* shall be `void` ([§15.6.9](classes.md#1569-partial-methods)). If the declaration includes the `async` modifier then *return_type* shall be `void` or the method returns-by-value and the return type is a *task type* ([§15.15.1](classes.md#15151-general)).

i.e. if the *return_type* of an async method is a *task type*, then it is a returns-by-value method and thus "returns a value".

In §15.15.1:

> The *return_type* of an async method shall be either `void` or a ***task type***. For an async method that returns a value, a task type shall be generic. For an async method that does not return a value, a task type shall not be generic.

**Example**

The following should be allowed, but is C.M() a method that "returns a value"?

```csharp
using System.Threading.Tasks;

class C {
    async Task M() {}
}
```

According to §15.6.1, this method is *returns-by-value*, and thus "returns a value".

However, if this is "an async method that returns a value", then §15.15.1 requires that "a task type shall be generic", which System.Threading.Tasks.Task is not.

**Expected behavior**

In §15.15.1, use some other phrase than "returns a value".

**Additional context**

None.

---
[Associated WorkItem - 187393](https://dev.azure.com/msft-skilling/Content/_workitems/edit/187393)

KalleOlaviNiemitalo:
Could perhaps say "has an *effective return type* (15.6.11) other than `void`", but I did not check whether this would result in a circular definition.

jskeet:
Again, punting - and this one has probably been broken since C# 5. (I suspect the eventual fix will be a matter of just a few words, but choosing those words will take time...)

KalleOlaviNiemitalo:
[...]

PR #1043, opened 2024-02-08
async "does not return a value"

BillWagner:
See https://github.com/dotnet/csharpstandard/issues/857#issuecomment-1933830824

When fixing #857, the corresponding text for async methods that don't produce a value wasn't updated.

I suspect we can merge this with a single review, without waiting for the next meeting.


standard/classes.md
```diff
@@ -5319,7 +5319,7 @@ A method ([§15.6](classes.md#156-methods)) or anonymous function ([§12.19](exp
...
+The *return_type* of an async method shall be either `void` or a ***task type***. For an async method that produces a result value, a task type shall be generic. For an async method that does not produce a value, a task type shall not be generic. Such types are referred to in this specification as `«TaskType»<T>` and `«TaskType»`, respectively. The Standard library type `System.Threading.Tasks.Task` and types constructed from `System.Threading.Tasks.Task<TResult>` are task types, as well as a class, struct or interface type that is associated with a ***task builder type*** via the attribute `System.Runtime.CompilerServices.AsyncMethodBuilderAttribute`. Such types are referred to in this specification as `«TaskBuilderType»<T>` and `«TaskBuilderType»`. A task type can have at most one type parameter and cannot be nested in a generic type.
```
```suggestion
The *return_type* of an async method shall be either `void` or a ***task type***. For an async method that produces a result value, a task type shall be generic. For an async method that does not produce a result value, a task type shall not be generic. Such types are referred to in this specification as `«TaskType»<T>` and `«TaskType»`, respectively. The Standard library type `System.Threading.Tasks.Task` and types constructed from `System.Threading.Tasks.Task<TResult>` are task types, as well as a class, struct or interface type that is associated with a ***task builder type*** via the attribute `System.Runtime.CompilerServices.AsyncMethodBuilderAttribute`. Such types are referred to in this specification as `«TaskBuilderType»<T>` and `«TaskBuilderType»`. A task type can have at most one type parameter and cannot be nested in a generic type.
```

Just to be consistent with the previous sentence.

jskeet (approved):

PR #1045, opened 2024-02-09
Update dependencies on the word converter

### LINKED ISSUE #997: Markdown Converter outdated NuGet package
See notes in #996 

The *FSharp.Formatter* package is many versions of out date. But, moving to the latest version introduces both some source breaks and some behavior breaks. 

I don't know of reasons we *must* upgrade, but it's probably a good idea.

---
[Associated WorkItem - 199062](https://dev.azure.com/msft-skilling/Content/_workitems/edit/199062)

BillWagner:
Fixes #997 

This PR updates all the NuGet packages for the word converter, and the other tools.

The first commit updates every NuGet package to the "latest stable" except the FSharp.Formatting package.

After that, I updated the FSharp.Formatting package iteratively in successive commits to get it to the latest version. In most cases, I fixed issues as they appeared. Many were compilation errors because new arguments were added to methods, or namespaces changed.

v 16.0.1 fixed the issue Jon had noted regarding parsing emphasis blocks. In that update, I removed the block of code that worked around that issue.

That same update introduced some false positive unit test failures. The new version created text nodes with a single space character before changing styles. The appearance was the same. Even after updating the existing results these tests still failed. *This is the first area I'd like help fixing.* The validation code seems to view those nodes as empty, not having a single space. It also introduced one new bug.  If a block of emphasis text didn't end with at the end of a word, the parser didn't find the end of the emphasis block, but rather added the `*` character in the text and extended the emphasis text until the next end marker was found. I added a new test for this. It's the test resource for "emphasis-with-plural.md".

v 16.0.2 fixed a parsing error introduced in the previous update. However, the unit test still failed with the same false positive noted above. In addition, it introduced a new failure. Starting in this version, tables that included bold code were are incorrect.

The source is `**+x**`, with the "+x" also code font:  **`x`**. The output is italicized, with the additional `*`:  `*+x*`. I added a new test for this, and tried to update the XML output correctly.

Those failures remain through the additional updates until the current version.


BillWagner:
@jskeet 

This isn't ready to merge, but it's worth taking a look.

Any pointers on fixing the false positive unit tests is appreciated. I'll work on trying to fix the actual issue regarding the math operator tables.

BillWagner:
Marking this ready, as I now need help on getting the unit tests to pass.

tools/MarkdownConverter.Tests/markdown-lint.xml
```diff
@@ -20,7 +20,10 @@
       <t space="preserve">Note</t>
     </r>
     <r>
-      <t space="preserve">: Note 1. </t>
+      <t space="preserve">: Note 1.</t>
+    </r>
+    <r>
+      <t space="preserve"> </t>
```
BillWagner, line 26: These are the diffs in output that are causing the unit tests to fail. The updated XML does match the generated output, but for some reason, the difference parser is viewing the new nodes as "null text" rather than a single space.

tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
```diff
@@ -662,33 +682,6 @@ IEnumerable<OpenXmlElement> Span2Elements(MarkdownSpan md, bool nestedSpan = fal
         {
             IEnumerable<MarkdownSpan> spans = (md.IsStrong ? ((MarkdownSpan.Strong) md).body : ((MarkdownSpan.Emphasis) md).body);
 
-            // Workaround for https://github.com/tpetricek/FSharp.formatting/issues/389 - the markdown parser
```
BillWagner, line 665: This issue has been fixed, so this workaround has been removed.

tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
```diff
@@ -951,7 +944,9 @@ IEnumerable<OpenXmlElement> Literal2Elements(string literal, bool isNested, bool
         }
     }
 
-    IEnumerable<OpenXmlCompositeElement> GenerateCustomBlockElements(string customBlockId, MarkdownParagraph.InlineBlock block) => customBlockId switch
+    // Note for future PR: "function_members", "format_strings_1", and "format_strings_2" are the
+    // only special cases still used. The others can be safely removed in a future PR.
+    IEnumerable<OpenXmlCompositeElement> GenerateCustomBlockElements(string customBlockId, MarkdownParagraph.InlineHtmlBlock block) => customBlockId switch
```
BillWagner: @jskeet Let me know if you want me to remove the (now) unused custom blocks in this PR.

tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
```diff
@@ -424,6 +423,26 @@ IEnumerable<OpenXmlCompositeElement> Paragraph2Paragraphs(MarkdownParagraph md)
                 {
                     var mdcell = mdrow[icol];
                     var cell = new TableCell();
+
```
BillWagner, line 426: I freely admit that this is ugly.

But it works.

Any suggestions welcome.  I did try a pattern match and that was less readable.

tools/MarkdownConverter.Tests/MarkdownConverter.Tests.csproj
```diff
@@ -11,6 +11,11 @@
     <EmbeddedResource Include="*.xml" />
     <EmbeddedResource Include="*.md" />
   </ItemGroup>
+  
+  <ItemGroup>
```
jskeet, line 15: I suspect this block isn't necessary/desirable.

tools/MarkdownConverter.Tests/emphasis-with-plural.md
```diff
@@ -0,0 +1,6 @@
+# 1 Heading
+
+Rules:
+
+- If `x` is zero and the namespace declaration contains an *extern_alias_directive* or *using_alias_directive* that associates the name�`I` with an imported namespace or type, then the *namespace_or_type_name* refers to that namespace or type.
```
jskeet: I'm seeing odd non-ASCII characters here after "name". Are those deliberate?

tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
```diff
@@ -423,6 +423,27 @@ IEnumerable<OpenXmlCompositeElement> Paragraph2Paragraphs(MarkdownParagraph md)
...
+                    // This logic deals with the special case where
+                    // the cell is meant to cntain bold inline code text.
```
jskeet: Nit: cntain => contain

tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
```diff
@@ -423,6 +423,27 @@ IEnumerable<OpenXmlCompositeElement> Paragraph2Paragraphs(MarkdownParagraph md)
...
+                    // The formatter parses it as emphasis, but surrounded
+                    // by single asterisks. We'll detect that and rewrite
+                    // it as the bold we intended.
+                    var para = mdcell.FirstOrDefault() as MarkdownParagraph.Paragraph;
+                    var leadingAsterisk = para?.body.First() as MarkdownSpan.Literal;
+                    var span2 = para?.body.Skip(1).FirstOrDefault() as MarkdownSpan.Emphasis;
+                    var cellText = span2?.body.FirstOrDefault() as MarkdownSpan.InlineCode;
+                    var trailingAsterisk = span2?.body.Skip(1).FirstOrDefault() as MarkdownSpan.Literal;
+                    if (leadingAsterisk?.text == "*" &&  cellText != null && trailingAsterisk?.text == "*")
```
jskeet: Nit: double space before cellText

tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
```diff
@@ -923,6 +944,8 @@ IEnumerable<OpenXmlElement> Literal2Elements(string literal, bool isNested, bool
         }
     }
 
+    // Note for future PR: "function_members", "format_strings_1", and "format_strings_2" are the
```
jskeet: I don't understand this comment. What's happened to CreateMultiplicationTable etc? (Did we stop *using* those a while ago, but didn't remove the code?)

tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
jskeet: Looking into it now - I need to understand the comment a bit more first.

jskeet (approved):
I'm looking at the custom block bit now. While the workaround for the bold in tables is a bit complicated, my experience is that it's almost always really hard to make code like this any more readable.

tools/MarkdownConverter/Converter/MarkdownSourceConverter.cs
jskeet: Right - I see now that the tables were removed in https://github.com/dotnet/csharpstandard/pull/584.

I suggest leaving the code here for now, but I've filed https://github.com/dotnet/csharpstandard/issues/1046 to remove it in a separate PR. I'm happy to do that as soon as this one is merged.

BillWagner: Added a comment to make it easier to remember this

tools/MarkdownConverter.Tests/markdown-lint.xml
BillWagner: @jskeet 

In my final two commits, I fix the unit tests, by modifying the comparison tests. Only add a failed comparison when the expect node is null and the actual node is a single space.

PR #1047, opened 2024-02-12
Remove unused table generation methods

### LINKED ISSUE #1046: Remove legacy table handling in the Word converter
Since https://github.com/dotnet/csharpstandard/pull/584 most of the custom tables have gone away - but the code still remains. We should remove it.

jskeet:
This also removes helper methods that would only have been used by the test case.

Fixes #1046.


BillWagner (approved):
This LGTM @jskeet 

Let's :shipit:

Issue #962, opened 2023-10-02
New Language Features that Rely on Features NOT in the Current CLI Spec


RexJaeschke:
[After discussing this with Bill, I'm generalizing Kalle's comment for other future features/spec versions.]

As a [comment ](https://github.com/dotnet/csharpstandard/pull/681#issuecomment-1734482343) to PR #681, @KalleOlaviNiemitalo  wrote:

> Should the references to "C# implementations targeting the CLI" be changed somehow because the ECMA standard CLI does not support default interface members? Possible answers include:
> •	Remove them; C# 8 implementations cannot target "the CLI".
> •	Change them to say "targeting a CLI with extensions" or the like.
> •	Keep them; a C# implementation can still target the standard CLI, although the mapping may become complex or inefficient or have restrictions (like not supporting dynamic loading of assemblies that were not provided at build time).

FYI, in the V7 Introduction, we state:

> Although Microsoft’s implementation of C# relies on CLI for library and run-time support, other implementations of C# need not, provided they support an alternate way of getting at the minimum CLI features required by this C# standard (see Annex C).

RexJaeschke:
Adding the Pre-C# 8.0 milestone, as this issue impacts a V8 feature.

Nigel-Ecma:
I think we can leave the CLI references as is, I don’t think trying to find alternative wording or adding caveats is worth it. (In some sense features added to .NET since the last CLI edition and used by C# are coming in some future edition until the CLI is declared as done/abandoned.)

However what we are missing is a source for the “formal definition of these types and their members” (§C.1) for those types in §C.3 and we should consider how to address that as it leaves a hole in the C# Standard – “What are the semantics of these types?”

While looking at occurrences of CLI I noted the following:

1. In §22.4.3 there is an example starting “In an implementation of the CLI…” but the example does not say why it applies only to C# on the CLI. Is it worth stating this?
2. In $D.4.2 it says “The encoding of each argument is the same as a CLI signature, as follows:” – is there a word missing in that sentence?

jskeet:
Notes from the meeting October 4th:

- Try to describe the new features without referring to the CLI
- Effectively ignore the fact that "just a CLI" implementation (without any extensions) wouldn't support some of this
- Potentially get XML documentation for the non-CLI-standard-specified types
- Default interface implementation feature will affect this and #959; we should probably prioritize that feature (and working out the implications)

Nigel-Ecma:
> I think we can leave the CLI references as is

[...]

PR #1054, opened 2024-03-05
Tweak CLI wording

Nigel-Ecma:
Spun off from a comment in #962 so it doesn’t get lost – accept, reject or revise as you see fit!

Tweaks wording around CLI as per https://github.com/dotnet/csharpstandard/issues/962#issuecomment-1832553160

### DIFF
--- a/standard/introduction.md
+++ b/standard/introduction.md
@@ -2,7 +2,7 @@
 
 This specification is based on a submission from Hewlett-Packard, Intel, and Microsoft, that described a language called C#, which was developed within Microsoft. The principal inventors of this language were Anders Hejlsberg, Scott Wiltamuth, and Peter Golde. The first widely distributed implementation of C# was released by Microsoft in July 2000, as part of its .NET Framework initiative.
 
-Ecma Technical Committee 39 (TC39) \[later renamed to TC49\] Task Group 2 (TG2) was formed in September 2000, to produce a standard for C#. Another Task Group, TG3, was also formed at that time to produce a standard for a library and execution environment called Common Language Infrastructure (CLI). (CLI is based on a subset of the .NET Framework.) Although Microsoft’s implementation of C# relies on CLI for library and run-time support, other implementations of C# need not provided they support the features and API required by this C# Standard (see [Annex C](standard-library.md#annex-c-standard-library)).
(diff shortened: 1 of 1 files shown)


standard/introduction.md
```diff
@@ -2,7 +2,7 @@
...
+Ecma Technical Committee 39 (TC39) \[later renamed to TC49\] Task Group 2 (TG2) was formed in September 2000, to produce a standard for C#. Another Task Group, TG3, was also formed at that time to produce a standard for a library and execution environment called Common Language Infrastructure (CLI). (CLI is based on a subset of the .NET Framework.) Although Microsoft’s implementation of C# relies on CLI for library and run-time support, other implementations of C# need not provided they provide the features and API required by this C# standard (see [Annex C](standard-library.md#annex-c-standard-library)).
```
jskeet: I haven't had enough coffee yet today to suggest an alternative, but it would be good to avoid "provided they provide". Would "provided they support the features and API" work?

Nigel-Ecma: Well I must not have had enough coffee back in Nov when I wrote that ;-) The original had support in it, can't think why it was changed.

```suggestion
Ecma Technical Committee 39 (TC39) \[later renamed to TC49\] Task Group 2 (TG2) was formed in September 2000, to produce a standard for C#. Another Task Group, TG3, was also formed at that time to produce a standard for a library and execution environment called Common Language Infrastructure (CLI). (CLI is based on a subset of the .NET Framework.) Although Microsoft’s implementation of C# relies on CLI for library and run-time support, other implementations of C# need not provided they support the features and API required by this C# Standard (see [Annex C](standard-library.md#annex-c-standard-library)).
```


(I capitalised Standard as well).

@jskeet go ahead and accept the change (as I put it in) if you're happy.

standard/introduction.md
```diff
@@ -2,7 +2,7 @@
...
+Ecma Technical Committee 39 (TC39) \[later renamed to TC49\] Task Group 2 (TG2) was formed in September 2000, to produce a standard for C#. Another Task Group, TG3, was also formed at that time to produce a standard for a library and execution environment called Common Language Infrastructure (CLI). (CLI is based on a subset of the .NET Framework.) Although Microsoft’s implementation of C# relies on CLI for library and run-time support, other implementations of C# need not provided they support the features and API required by this C# Standard (see [Annex C](standard-library.md#annex-c-standard-library)).
```
BillWagner: nit: I think we need a comma here.

```suggestion
Ecma Technical Committee 39 (TC39) \[later renamed to TC49\] Task Group 2 (TG2) was formed in September 2000, to produce a standard for C#. Another Task Group, TG3, was also formed at that time to produce a standard for a library and execution environment called Common Language Infrastructure (CLI). (CLI is based on a subset of the .NET Framework.) Although Microsoft’s implementation of C# relies on CLI for library and run-time support, other implementations of C# need not, provided they support the features and API required by this C# Standard (see [Annex C](standard-library.md#annex-c-standard-library)).
```

Nigel-Ecma: @RexJaeschke, what do you think?

RexJaeschke: I'm OK with this.

RexJaeschke (approved):

jskeet (approved):

PR #1061, opened 2024-03-20
Minor nit: make sentence a new para

RexJaeschke:
We have a para following the final bullet in a list incorrectly being considered part of that bullet. It was a separate para in V5, but was (presumably accidentally) made incorrect in V6.


jskeet (approved):

BillWagner (approved):
This LGTM

PR #1063, opened 2024-03-27
Fix examples where feasible

jskeet:
Towards #646


BillWagner (approved):
This LGTM @jskeet 

Let's :shipit:

PR #1065, opened 2024-03-28
Fix escape sequence examples which were missing backslash

### LINKED ISSUE #1064: Examples for character literals.
### Type of issue

Typo

### Description

The below example in [6.4.5.5 Character literals](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/lexical-structure#6455-character-literals)

> string good = "x9Good text";
> string bad = "x9Bad text";

shouldn't be like the below.

> string good = "\x9Good text";
> string bad = "\x9Bad text";

### Page URL

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/types#838-the-decimal-type

### Content source URL

https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/types.md

jskeet:
Fixes #1064

(Might be worth porting back to standard-v7 as well?)

### DIFF
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -842,8 +842,8 @@ fragment Hexadecimal_Escape_Sequence
 >
 > <!-- Example: {template:"standalone-console-without-using", name:"CharacterLiterals", ignoredWarnings:["CS0219"]} -->
 > ```csharp
-> string good = "x9Good text";
-> string bad = "x9Bad text";
+> string good = "\x9Good text";
+> string bad = "\x9Bad text";
 > ```
 >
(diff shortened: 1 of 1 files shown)


KalleOlaviNiemitalo (approved):
I skimmed the results of `git grep -P -e "([\"'])"'.*x[0-9a-fA-F].*\1' 9bfbc254877745e33cb6db58dd29ad3c0ce4b919 -- "*.md"` and didn't find any other similar errors.

KalleOlaviNiemitalo:
The ECMA-334 5th edition PDF does not have this error, but the 6th edition does.

BillWagner (approved):
This LGTM as well.

I'm fine with back-porting this fix. I'll defer to @RexJaeschke on how that process would work.

PR #1077, opened 2024-04-12
Correct the spelling and format of accessor-related term usage

RexJaeschke:
For some spec versions now, we've used multiple ways of referring to the same kind of accessor. Using `set` as an example, those ways are:
1. set accessor [our colloquial term that is not defined in the spec]
1. `set` accessor [what is being used as a synonym for the colloquial term]
1. *set_accessor* [a reference to a non-existent grammar rule, but which is being used a synonym for the colloquial term]
 
This PR makes the following changes for set, get, add, remove, and event accessors:
1. Leaves all the many >>xxx accessor<< as they are.
1. Changes the few >>`xxx` accessor<< to >>xxx accessor<<.
1. Changes the few >>*xxx_accessor*<< to >>xxx accessor<<.
1. Changes the one occurrence of >>getter<< to >>get accessor<<. [The spec does not use the colloquial terms getter and setter.]

Once this has been merged, I'll review the V9 draft PR that adds support for init-only accessors, to make sure it is consistent with these changes.


standard/statements.md
```diff
@@ -1512,7 +1512,7 @@ It is a compile-time error to use a return-by-ref if *expression* is not a *vari
...
 
-A function member is said to ***compute a value*** if it is a method with a returns-by-value method ([§15.6.11](classes.md#15611-method-body)), a returns-by-value `get` accessor of a property or indexer, or a user-defined operator. Function members that are returns-no-value do not compute a value and are methods with the effective return type `void`, `set` accessors of properties and indexers, `add` and `remove` accessors of event, instance constructors, static constructors and finalizers. Function members that are returns-by-ref do not compute a value.
+A function member is said to ***compute a value*** if it is a method with a returns-by-value method ([§15.6.11](classes.md#15611-method-body)), a returns-by-value get accessor of a property or indexer, or a user-defined operator. Function members that are returns-no-value do not compute a value and are methods with the effective return type `void`, set accessors of properties and indexers, add and remove accessors of event, instance constructors, static constructors and finalizers. Function members that are returns-by-ref do not compute a value.
```
Nigel-Ecma:
```suggestion
A function member is said to ***compute a value*** if it is a method with a returns-by-value method ([§15.6.11](classes.md#15611-method-body)), a returns-by-value get accessor of a property or indexer, or a user-defined operator. Function members that are returns-no-value do not compute a value and are methods with the effective return type `void`, set accessors of properties and indexers, add and remove accessors of events, instance constructors, static constructors and finalizers. Function members that are returns-by-ref do not compute a value.
```

Nigel-Ecma (approved):
One old missing ‘s’ added as a suggestion.

LGTM

jskeet (approved):
LGTM once Nigel's suggestion is committed.

BillWagner (approved):
This LGTM as well, once Nigel's suggestion is applied.

PR #1082, opened 2024-04-14
Correct README.md

RexJaeschke:
Correct/bring up to date. Removed all .x from version numbers, as Ecma hasn't published any dot versions; besides, 7.0 might have confused readers with MS's V7.0, when Ecma V7 actually matches MS V7.3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ The branch `draft-v9` has Draft PRs and Issues for C# 9.
 
 The branch `draft-v8` has the evolving draft text for C# 8.
 
-### C# 7 draft (the most-recently published version)
+### C# 7 standard
 
 The branch `standard-v7` has the ECMA C# 7 standard text, in Markdown format. For the official standard, see the [ECMA site](https://www.ecma-international.org/publications-and-standards/standards/ecma-334/).
 
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ More broadly, *no* comments should be regarded as being part of the standard its
 
 ## Admin folder
 
-A home for adminstrative files.
+A home for administrative files.
 
 For now, it contains separate logs for past (V6, V7), present (V8), and future (V9) work going on to add new features.
 
--- a/README.md
+++ b/README.md
@@ -9,25 +9,25 @@ This project has adopted the code of conduct defined by the Contributor Covenant
 
 ## C# Language Specification
 
-### C# 9.0 draft
+### C# 9 draft
 
-The branch `draft-v9` has Draft PRs and Issues for C# 9.0.
+The branch `draft-v9` has Draft PRs and Issues for C# 9.
 
-### C# 8.0 draft
+### C# 8 draft
 
-The branch `draft-v8` has the evolving draft text for C# 8.0.
+The branch `draft-v8` has the evolving draft text for C# 8.
 
-### C# 7.0 draft
+### C# 7 draft (the most-recently published version)
 
-The branch `standard-v7` has the text for C# 7.0. It has been submitted for consideration as a formal standard to ECMA.
+The branch `standard-v7` has the ECMA C# 7 standard text, in Markdown format. For the official standard, see the [ECMA site](https://www.ecma-international.org/publications-and-standards/standards/ecma-334/).
 
-### C# 6.0 standard
+### C# 6 standard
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -39,15 +39,15 @@ This version is stored in this branch as a base markdown version to compare with
...
 
-Some automated test comments refer to error codes that are specific to the Microsoft C# compiler (e.g. "CS0509") to test that compilation fails as expected, where an example presents deliberately-invalid code. These error codes are not part of the standard, and should not be viewed as any kind of compliance check for other compilers.
+Some automated test comments refer to error codes that are specific to the Microsoft C# compiler (e.g., "CS0509") to test that compilation fails as expected, where an example presents deliberately-invalid code. These error codes are not part of the standard, and should not be viewed as any kind of compliance check for other compilers.
 
 More broadly, *no* comments should be regarded as being part of the standard itself.
 
 ## Admin folder
 
-A home for adminstrative files (such as [eventually] meeting agendas and minutes).
+A home for adminstrative files.
```
KalleOlaviNiemitalo: "admin<ins>i</ins>strative"

README.md
```diff
@@ -9,25 +9,25 @@ This project has adopted the code of conduct defined by the Contributor Covenant
...
+The branch `draft-v9` has Draft PRs and Issues for C# 9.
 
-### C# 8.0 draft
+### C# 8 draft
 
-The branch `draft-v8` has the evolving draft text for C# 8.0.
+The branch `draft-v8` has the evolving draft text for C# 8.
 
-### C# 7.0 draft
+### C# 7 draft (the most-recently published version)
```
KalleOlaviNiemitalo: Is it still a "draft" then?

KalleOlaviNiemitalo:
How about the tables at <https://ecma-international.org/publications-and-standards/standards/ecma-334/> and <https://ecma-international.org/task-groups/tc49-tg2/>?  Those still say "V7.0" (rather than "V7.3") in the "MICROSOFT VISUAL C# VERSION" column, and the latter doesn't mention ECMA-334:2023.  (Also, listing ISO/IEC 20619:2023 in the table on the ECMA-334 page seems a bit misleading as that standard better corresponds to ECMA-422.)

README.md
KalleOlaviNiemitalo: I guess it means the standard was already published by ECMA but the `standard-v7` branch in this repository does not have all the changes that went in the final standard.  So there won't be a `V7-standard` tag similar to the existing `V6-standard` tag.

BillWagner: I would change this to the C# 7 standard. (It's been published, and the branch name has changed:

```suggestion
### C# 7 standard
```

README.md
BillWagner: right:

```suggestion
A home for administrative files.
```

BillWagner (approved):
I made a few suggestions @RexJaeschke 

And then, this is ready to merge!

RexJaeschke:
@KalleOlaviNiemitalo, re the tables, they are maintained by the Ecma office, and I'll ask them to make the following changes:

1. Remove the .x version numbers from the left-most column of both tables.
2. In the TG2 table, change the "TBD" for V7 to "ECMA-334:2023".
3. In both tables, change the right-most column for V7 to "ISO/IEC 20619:2023 (which includes by reference the latest edition of ECMA-334).

Re ISO/IEC 20619, yes that does mirror ECMA-422, but 422 was created just to get the ISO version thereof, so I won't mention that Ecma spec here, as that adds no value and might lead to confusion.

KalleOlaviNiemitalo:
My concern is that it will be more confusing when C# 8 is published and there is no new version of ISO/IEC 20619 because the set of documents did not change; then the table will have the same version of ISO/IEC 20619 for both C# 7 and C# 8.

RexJaeschke:
@KalleOlaviNiemitalo You are correct; going forward, we do not expect 20619 to be updated, which is why I added the text "which includes by reference the **latest edition** of ECMA-334". This is achieved by using an undated reference in 20619, as shown below:

> 2 Normative references
>
> The following documents are referred to in the text in such a way that some or all of their content constitutes requirements of this document. For dated references, only the edition cited applies. For undated references, the latest edition of the referenced document (including any amendments) applies.
>
> ECMA-334, C# language specification, https://www.ecma-international.org/publications-and-standards/standards/ecma-334/
> ...

where the link to 334 leads to the latest published version.

PR #1084, opened 2024-04-20
[v8] Annex B: add 'mechanism by which linkage to an external method is achieved' to the list of implementation defined behaviors

jkrejcha:
While it's currently in the standard (§15.6.8), it's not currently enumerated on the list of implementation-defined behaviors


RexJaeschke:
@jkrejcha Good catch. However, the entries in this annex's sections are supposed to be in ascending section-number order, so please move your new line up earlier in the list. And at the same time, please move the existing final entry about stack allocation of size zero as well, as it is is also out-of-order.

jkrejcha:
@RexJaeschke Done!

RexJaeschke (approved):

PR #1100, opened 2024-05-01
Consolidate namespace System member sets

RexJaeschke:
In V7, we added a second block of members to namespace `System` **at the end of C.3 when there was already a first block at the start, which we should have added to instead**. This PR simply consolidates those two blocks (in alpha-member order).

It also removes some blank lines for consistency in this section.

(I found this while looking at adding new members to this namespace for a V11 feature.)


BillWagner (approved):
This LGTM. I'll :shipit:

I'll figure out why markdown lint is failing now later today.

PR #1104, opened 2024-05-08
Reserve operator method names

RexJaeschke:
While the mapping of C# operators to compiler-generated method names is *not* stated in the C# spec, it *is* in the CLI spec. (Partition I, Table I.4 [I.10.3.1 Unary operators] and Table I.5 [I.10.3.2 Binary operators].) Unfortunately, this is insufficient, for the following reasons:

1. The C# spec normatively references Partition IV only: ISO/IEC 23271:2012, Common Language Infrastructure (CLI), Partition IV: Base Class Library (BCL), Extended Numerics Library, and Extended Array Library. In the C# spec Introduction we say, “Although Microsoft’s implementation of C# relies on CLI for library and run-time support, other implementations of C# need not, provided they support an alternate way of getting at the minimum CLI features required by this C# standard (see Annex C).” And Annex C is all about the library in Partition IV. So, we can’t simply defer to Partition I without making it a Normative Reference (which I’m sure we don’t want to do).
1. In any event, those tables do *not* actually show the C# mapping! Instead, their right-most columns imply a mapping to C++ and are marked as *informative*.
1. Even if we could defer to the CLI for this information, we’d still have to augment it for the new operator function names resulting from V11’s “Checked user-defined operators” as well as `>>>`.

This PR replicates the tables from the CLI spec, replacing C++ with C# in the (now normative) right column, and puts them in a new section, 15.3.10.6, “Member names reserved for operators.” And as we already know the extra names that V11 uses, I’ve put those in the current spec, but tagged with “reserved for future use” to warn people to stay away from defining methods with those names.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -1390,7 +1390,7 @@ The following method names are reserved. While many have corresponding operators
 
 | **Method Name** | **C# Operator** |
 |--------------|----------------------|
-| `op_Addition` | `+` (binary) | 
+| `op_Addition` | `+` (binary) |
 | `op_AdditionAssignment` | (reserved) |
 | `op_AddressOf` | (reserved) |
 | `op_Assign` | (reserved) |
@@ -1433,7 +1433,7 @@ The following method names are reserved. While many have corresponding operators
 | `op_MultiplicationAssignment` | (reserved) |
 | `op_Multiply` | `*` (binary) |
 | `op_OnesComplement` | `~` |
-| `op_PointerDereference` | (reserved)
+| `op_PointerDereference` | (reserved) |
 | `op_PointerToMemberSelection` | (reserved) |
 | `op_RightShift` | `>>` |
 | `op_RightShiftAssignment` | (reserved) |
@@ -1443,7 +1443,7 @@ The following method names are reserved. While many have corresponding operators
 | `op_True` | `true` |
 | `op_UnaryNegation` | `-` (unary) |
 | `op_UnaryPlus` | `+` (unary) |
-| `op_UnsignedRightShift ` | (reserved for future use) |
+| `op_UnsignedRightShift` | (reserved for future use) |
 | `op_UnsignedRightShiftAssignment` | (reserved) |
 
 ## 15.4 Constants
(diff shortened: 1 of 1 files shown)


standard/classes.md
```diff
@@ -1382,6 +1384,68 @@ For a class containing a finalizer ([§15.13](classes.md#1513-finalizers)), the
 void Finalize();
 ```
 
+#### §Names-Reserved-For-Ops Member names reserved for operators
+
+The following method names are reserved. While many have corresponding operators in this specification, some are reserved for use by future versions, while some are reserved for interop with other languages.
```
KalleOlaviNiemitalo: Please clarify whether these names are reserved for methods only, or also for other kinds of members (fields, properties, events, nested types). "Member names" in the heading suggests they are reserved for those too, but "method names" in the paragraph says otherwise.

RexJaeschke: @KalleOlaviNiemitalo I'm not seeing a problem. I chose the new section heading based on the pattern of the previous 3 sections, "Member names reserved for ...," which contain only method signatures. Now as the set of operator-related method signatures is open-ended, I can't enumerate all those signatures, which is why I avoided using the term *signature* in the new section. I do say explicitly that these are method names, which of course are member names, so there is no conflict with the heading. Would you prefer the new section heading to say, "Method names reserved for operators"? If so, does that suggest the previous 3 headings also be changed to say "signatures" instead of "members"?

BillWagner: I think the best resolution is to change the header from "member" to "method". I agree that enumerating all the signatures might be problematic.

BillWagner (approved):
This LGTM @RexJaeschke 

I'd like Jon or Nigel to weigh in, and then let's ship it.

standard/classes.md
jskeet: Yes, I think method name would be fine here.

jskeet (approved):
Approved with the change to "Method names"

Nigel-Ecma (approved):

standard/classes.md
RexJaeschke: Change the header from "Member" to "Method".

PR #1105, opened 2024-05-09
C# 8 Nullable reference types: Define different reference types:  oblivious, nullable, and non-nullable

### LINKED ISSUE #1089: [Nullable Reference Types] Nullable, Non-nullable, and "Nullable oblivous" reference types
Update the clause on types to describe the reference types for "nullable reference types", "non-nullable reference types", and reference types in a "nullable oblivous context".

This should include that there is an identity conversion among the nullable reference type, its non-nullable type and the corresponding "nullable oblivious" type. (For example between `string` and `string?`, note that `string` could mean a non-nullable reference type, or a nullable oblivious type). 

It should also add a note (or possibly a normative statement) that the nullability of a reference type argument can't participate in overload resolution. Members can't be overloaded based on the nullability of a reference parameter.

BillWagner:
Related to #700 

Fixes #1089

Add text to define the different "flavors" of reference types:  "Nullable oblivious" (class reference types), "nullable reference types", and "non-nullable reference types".  

Commit history is less clean than I'd like. I'd start thinking that one PR for the contexts `#nullable` changes and the types would be fine. That was getting larger. So I split the additional changes for `#nullable` into #1108

### DIFF
--- a/standard/types.md
+++ b/standard/types.md
@@ -27,7 +27,7 @@ For convenience, throughout this specification, some library type names are writ
 
 ### 8.2.1 General
 
-A reference type is a class type, an interface type, an array type, a delegate type, or the `dynamic` type.
+A reference type is a class type, an interface type, an array type, a delegate type, a nullable reference type, or the `dynamic` type.
 
 ```ANTLR
 reference_type
@@ -35,6 +35,7 @@ reference_type
     | interface_type
     | array_type
     | delegate_type
+    | nullable_reference_type
     | 'dynamic'
     ;
 
@@ -69,9 +70,19 @@ rank_specifier
 delegate_type
     : type_name
     ;
+
+nullable_reference_type
+    : non_nullable_reference_type '?'
+    ;
+
+non_nullable_reference_type
+    : reference_type
+    ;
 ```
 
 *pointer_type* is available only in unsafe code ([§23.3](unsafe-code.md#233-pointer-types)).
+
+The *non_nullable_reference_type* in *nullable_reference_type* shall be a non-nullable class, a non-nullable interface, a non-nullable delegate, a non-nullable array type, or `dynamic`. *nullable_reference_type* is discussed further in §Types-And-Nullability.
 
 A reference type value is a reference to an ***instance*** of the type, the latter known as an object. The special value `null` is compatible with all reference types and indicates the absence of an instance.
 
@@ -431,7 +442,7 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can 
 
 ### 8.3.12 Nullable value types
 
-A nullable value type can represent all values of its underlying type plus an additional null value. A nullable value type is written `T?`, where `T` is the underlying type. This syntax is shorthand for `System.Nullable<T>`, and the two forms can be used interchangeably.
+A ***nullable value type*** can represent all values of its underlying type plus an additional null value. A nullable value type is written `T?`, where `T` is the underlying type. This syntax is shorthand for `System.Nullable<T>`, and the two forms can be used interchangeably.
 
 Conversely, a ***non-nullable value type*** is any value type other than `System.Nullable<T>` and its shorthand `T?` (for any `T`), plus any type parameter that is constrained to be a non-nullable value type (that is, any type parameter with a value type constraint ([§15.2.5](classes.md#1525-type-parameter-constraints))). The `System.Nullable<T>` type specifies the value type constraint for `T`, which means that the underlying type of a nullable value type can be any non-nullable value type. The underlying type of a nullable value type cannot be a nullable value type or a reference type. For example, `int??` and `string?` are invalid types.
 
@@ -454,7 +465,7 @@ creates a non-null instance of `T?` for which the `Value` property is `x`. The p
 
 Implicit conversions are available from the `null` literal to `T?` ([§10.2.7](conversions.md#1027-null-literal-conversions)) and from `T` to `T?` ([§10.2.6](conversions.md#1026-implicit-nullable-conversions)).
 
-The nullable type `T?` implements no interfaces ([§18](interfaces.md#18-interfaces)). In particular, this means it does not implement any interface that the underlying type `T` does.
+The nullable value type `T?` implements no interfaces ([§18](interfaces.md#18-interfaces)). In particular, this means it does not implement any interface that the underlying type `T` does.
 
 ### 8.3.13 Boxing and unboxing
 
@@ -705,3 +716,122 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
 - `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`.
 - Any *enum_type*.
 - Any user-defined *struct_type* that is not a constructed type and contains instance fields of *unmanaged_type*s only.
+- In unsafe code ([§23.2](unsafe-code.md#232-unsafe-contexts)), any *pointer_type* ([§23.3](unsafe-code.md#233-pointer-types)).
+
+## §Types-And-Nullability Types and nullability
+
+### §Nullable-Types-General General
+
+C# has long supported the notion of a nullable value type; for example, the type `int?` is the nullable version of the value type `int`. (Value types and their nullable variants are described in [§8.3](types.md#83-value-types) and [§8.3.12](types.md#8312-nullable-value-types).) Almost all of the remaining text in this subclause pertains to the addition of support for nullable reference types, with only occasional mention of nullable value types.
+
+> *Note*: By definition, a reference type is nullable; that is, a variable of that type can either contain a reference to an object or be the value `null`, which indicates “no reference.” However, in many applications, the value of a reference should never be `null`, yet for the first 20-odd years of C#’s existence, there was no way to indicate this, or to have attempts to use such a reference in a `nonnull`-related context be diagnosed. Now that support for nullable reference types exists, such indications and diagnosis are possible by setting the nullable contexts (§Nullable-Contexts). *end note*
+
+A reference type of the form `T?` (such as `string?`) is an ***annotated reference type***. The annotation `?` indicates the intent that variables of this type are nullable. A reference type of the form `T` (such as `string`) is an ***unannotated reference type***. The absence of the annotation `?` indicates the intent that variables of this type are non-nullable. When the nullable annotation context (§Nullable-Annotation-Context) is enabled, the compiler shall recognize these intents. When the nullable annotation context is disabled, the compiler shall ignore these intents, thereby treating `T` and `T?` in the same (nullable) manner, and generate a warning.
+
+A ***non-nullable reference type*** is a reference type whose variables should not contain the value `null`, and it is an unannotated reference type whose non-null value intent is recognized. A ***nullable reference type*** is any reference type that is not a non-nullable reference type.
+
+> *Note*: Here, “should” is used rather than “shall,” as this a declaration of intent. The presence of a `?` suffix is not enough to make it a non-nullable type; for that, the annotation context has to be enabled. If it is not, the `?` is ignored, and the type is nullable. *end note*
+
+Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the exact same type; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. Unlike a nullable value type, a nullable reference type has no relationship to the type `System.Nullable<T>`.
+
+### §Nullable-Contexts Nullable contexts
+
+#### §Nullable-Contexts-General General
+
+Every line of source code has a ***nullable annotation context*** and a ***nullable warning context***. The former controls whether nullable annotations (§Nullable-Annotation-Context) have effect; the latter controls whether nullable warnings (§Nullable-Warning-Context) are issued by the compiler. The state of each of the annotation and warning contexts of a given line is disabled or enabled.
+
+Both nullable contexts may be specified within source code via nullable directives (§Nullable-Directives) and/or via some implementation-specific mechanism external to the source code. If both approaches are used, nullable directives supersede the settings made via an external mechanism.
+
+If a nullable context is not specified, its default state is disabled.
+
+Throughout this specification, all C# code that does not contain nullable directives, or about which no statement is made regarding the current nullable context state, shall be assumed as having been compiled with nullable annotation and nullable warning contexts enabled.
+
+#### §Nullable-Annotation-Context Nullable annotation context
+
+This context controls whether nullable annotations have any effect.
+
+When the nullable annotation context is disabled
+
+- A variable of any reference type is nullable, so that variable may be initialized with, or assigned a value of, `null`.
+- No warning shall be generated when a variable of a reference type that possibly has the null value, is dereferenced.
+- For any reference type `T`, the annotation `?` in `T?` is ignored, as `T` is already a nullable type. An informational message shall be generated to that effect.
+  > *Note*: This message is characterized as “informational” rather than “warning,” so as not to confuse it with the state of the nullable warning context, which is unrelated.  *end note*
+- The null-forgiving operator `!` (§Null-Forgiving-Expressions) is ignored.
+
+> *Example*:
+>
+> <!-- Example: {template:"code-in-main-without-using", name:"NullableAnnotationContext1", ignoredWarnings:["CS0219","CS8632"], expectedException:"NullReferenceException"} -->
+> ```csharp
+> #nullable disable
+> string? s1 = null;    // Informational message; ? is ignored; s2 already nullable
+> string s2 = null;     // OK; null initialization of a nullable reference
+> s2 = null;            // OK; null assignment to a nullable reference
+> char c1 = s2[1];      // OK; no warning on dereference of a possible null; throws NullReferenceException
+> c1 = s2![1];          // OK; ! is ignored
+> ```
+>
+> *end example*
+
+When the nullable annotation context is enabled
+
+- For any reference type `T`, the annotation `?` in `T?` makes `T?` a nullable type, whereas the unannotated `T` is non-nullable.
+- A warning shall be generated when a variable of an unannotated reference type is initialized with, or assigned a value of, `null`.
+- A warning shall be generated when a variable of a nullable or non-nullable reference type, whose value might be `null`, is dereferenced; otherwise, that variable may be dereferenced safely.
+- The null-forgiving operator `!` (§Null-Forgiving-Expressions) suppresses warnings about dereferencing a possible null reference.
+
+> *Example*:
+>
+> <!-- Example: {template:"code-in-main-without-using", name:"NullableAnnotationContext2", ignoredWarnings:["CS0219"], expectedWarnings:["CS8600","CS8600","CS8602"], expectedException:"NullReferenceException"} -->
+> ```csharp
+> #nullable enable
+> string? s1 = null;    // OK; ? makes s2 nullable
+> string s2 = null;     // Warning; s1 is non-nullable
+> s2 = null;            // Warning; s1 is non-nullable
+> char c1 = s2[1];      // Warning; dereference of a possible null; throws NullReferenceException
+> c1 = s2![1];          // The warning is suppressed
+> ```
+>
+> *end example*
+
+#### §Nullable-Warning-Context Nullable warning context
+
+This context controls whether nullable warnings are issued by the compiler.
+
+The compiler uses static flow analysis to determine the null state of any reference variable. When the nullable warning context is enabled, a reference variable’s null state is either *not null* or *maybe null*, and
+
+- A warning shall be generated if a reference variable that has been determined to be *maybe null*, is dereferenced.
+- The state of a reference variable is *maybe null* unless the compiler can determine one of two conditions:
+  1. The variable has been definitely assigned a non-`null` value.
+  1. The variable or expression has been checked against `null` before dereferencing it.
+
+Some warnings are generated regardless of the state of the nullable annotation context. (Examples include throwing a value that maybe `null`, and dereferencing a possibly null reference.) Other warnings are generated only if both the nullable warning context and the nullable annotation context are enabled. (Examples include converting a null literal or possible null value to a non-nullable type, possible null reference assignment, and possible null reference return.)
+
+### §Nullabilities-And-Null-States Nullabilities and null states
+
+A type can have one of three kinds of ***nullability***: “non-nullable”, “nullable,” or “oblivious,” and this is determined by the compiler using static flow analysis. For a reference type `T`, if the nullable annotation context (§Nullable-Annotation-Context) is enabled, `T` is non-nullable and `T?` is nullable. If the nullable annotation context is disabled, `T` is nullable and the `?` in `T?` is ignored. If analysis cannot determine the nullability, the nullability is oblivious.
+
+A warning might result if a variable of a non-nullable type has a potential `null` value assigned to it. However, oblivious and nullable variables can have `null` values freely assigned to them.
+
+Values of oblivious and non-nullable types can be dereferenced or assigned without generating a warning. Values of nullable types, however, are “null-yielding” and can cause warnings when dereferenced or assigned without proper null checking.
+
+Every expression has a ***null state*** that indicates whether that expression might potentially evaluate to `null`. The null state is either “not null,” “maybe null,” or “maybe default.” The null state is used to determine whether a warning should be given about null-unsafe conversions and dereferences. The ***default null state*** of a null-yielding variable is “maybe null” or “maybe default.” The default null state of a non-null-yielding variable is “not null.”
+
+Nullability is determined, as follows:
+
+- A non-nullable value type `S` is always non-nullable
+- A nullable value type `S?` is always nullable
+- An unannotated reference type `C` in a disabled nullable annotation context is oblivious
+- An unannotated reference type `C` in an enabled nullable annotation context is non-nullable
+- A nullable reference type `C?` is nullable; however, a warning shall be generated in a disabled nullable annotation context
+
+With a type parameter, any constraints are taken into account:
+
+- A type parameter `T` where all constraints (if any) are either nullable types or the `class?` constraint, is `nullable`
+- A type parameter `T` where at least one constraint is either oblivious or non-nullable, or one of the constraints `struct`,  `class`, or `notnull`, is
+  - oblivious in a disabled nullable annotation context
+  - non-nullable in an enabled nullable annotation context
+- A nullable type parameter `T?` is nullable, but a warning shall be generated in a disabled nullable annotation context if `T` isn't a value type
+
+A type is deemed to occur in a given annotation context when the final token of that type is within that context.
+
+Whether a given reference type `C` is interpreted as oblivious or non-nullable depends on the nullable annotation context. However, once the oblivious or non-nullable nature has been established, it is considered part of that type, and “travels with it” (during substitution of generic type arguments, for example). It is as if there is an (invisible) annotation like `?` on the type.
--- a/standard/types.md
+++ b/standard/types.md
@@ -756,7 +756,7 @@ When the nullable annotation context is disabled
 - No warning shall be generated when a variable of a reference type that possibly has the null value, is dereferenced.
 - For any reference type `T`, the annotation `?` in `T?` is ignored, as `T` is already a nullable type. An informational message shall be generated to that effect.
   > *Note*: This message is characterized as “informational” rather than “warning,” so as not to confuse it with the state of the nullable warning context, which is unrelated.  *end note*
-- The null-forgiving operator `!` (§Null-Forgiving-Expressions) is ignored.
+- The null-forgiving operator `!` is ignored.
 
 > *Example*:
 >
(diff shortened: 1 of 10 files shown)


jskeet:
Worth labeling for discussion next week, or is it too early for that?

BillWagner:
> Worth labeling for discussion next week, or is it too early for that?

I'll add the label after making edits. I *hope* we can discuss it next week.

BillWagner:
@jskeet @RexJaeschke 

As I start on this, I wonder if a separate PR that updates all test examples so they generate the existing output in a nullable context is a good idea. It would separate out that work into one PR to review. 

If you both agree, I'll add that PR for Wednesday as well.

jskeet:
I'm not entirely sure whether that's a good idea. It might end up making a lot of examples that have nothing to do with nullability very null-syntax heavy. But I could be wrong.

Maybe make a start but see how you feel without doing it for *all* examples?

standard/types.md
```diff
@@ -705,3 +705,121 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+A ***non-nullable reference type*** is a reference type whose variables should not contain the value `null`, and it is an unannotated reference type whose non-null value intent is recognized. A ***nullable reference type*** is any reference type that is not a non-nullable reference type.
+
+> *Note*: Here, “should” is used rather than “shall,” as this a declaration of intent. The presence of a `?` suffix is not enough to make it a non-nullable type; for that, the annotation context has to be enabled. If it is not, the `?` is ignored, and the type is nullable. *end note*
```
KalleOlaviNiemitalo: Surely the **presence** of a `?` suffix won't make it a <strong>non-</strong>nullable type, even if the annotation context is enabled.

standard/types.md
```diff
@@ -705,3 +705,121 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the exact same type; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. Unlike a nullable value type, a nullable reference type has no relationship to the type `System.Nullable<T>`.
```
KalleOlaviNiemitalo: The "exact same type" wording feels troublesome.  If the `?` nullable-reference-type annotation is used in an alias 

```csharp
#nullable enable

using ListOfNullableString = System.Collections.Generic.List<string?>;

class C
{
    // warning CS8619
    System.Collections.Generic.List<string> sl = new ListOfNullableString();
}
```

then the standard needs some wording to specify that the initializer of the field `sl` gets both its _type_ and its _nullability annotations_  from the alias `ListOfNullableString`.  Currently, §14.5.2 (Using alias directives) just says "the identifier introduced by the *using_alias_directive* can be used to reference the given namespace or type" and does not mention nullability.

The alternative would be to specify `R` and `R?` as different types that have two-way identity conversions between them, like there are between `object` and `dynamic`, or between similar tuple types.  But perhaps there's something about flow analysis that makes this not possible.

```csharp
#nullable enable

class TypeArgumentDeductionWithIdentityConversion
{
    static void OneNrt(string? a, string b)
    {
        N(ref a, ref b); // warning CS8601
    }
    
    static void TwoNrt(string? a, string? b)
    {
        N(ref a, ref b);
    }
    
    static void ObjectVsDynamic(object[] a, dynamic[] b)
    {
        N(ref a, ref b);
    }
    
    static void TupleElementNames((int X, int Y) a, (int Z, int W) b)
    {
        N(ref a, ref b);
    }
    
    static void N<T>(ref T a, ref T b)
    {
    }
}
```

RexJaeschke:
@BillWagner As we discussed privately recently, I'm OK with spinning off the examples into a separate PR. I recall there were many, and I applied them in several large commits.

BillWagner:
> As we discussed privately recently, I'm OK with spinning off the examples into a separate PR. I recall there were many, and I applied them in several large commits.

@RexJaeschke @jskeet 

I'm leaning that way now too! I updated one sample in this PR because it made sense to show the nullable annotations as part of this PR. That didn't compile in the sample extractor, because nullable was off. Then, when I turned it on, several things failed.

The best way to handle this may be starting with #1106.  In that PR, I set the `Nullable` flag to `annotations`. Then, as we add other PRs for the nullable reference type feature, we can show it, and it will compile in our test harness. That scopes the changes so that PRs are easier to review.

Then, later, we can change the setting to `enable` and address any warnings as we see fit.

BillWagner:
Note: the experiment I tried in this PR and #1106 worked. In #1106, all existing samples compile with the same result as before. And in this PR, I was able to add a nullable annotation where needed. Let's make that the plan of record and move forward.

standard/types.md
BillWagner: Fixed.

standard/types.md
BillWagner: We're thinking that the identity conversion path is the correct one. (We haven't speced out the warning section yet, but there are other identity conversions that can produce warnings).

standard/types.md
```diff
@@ -431,7 +440,7 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+A ***nullable value type*** can represent all values of its underlying type plus an additional null value. A nullable value type is written `T?`, where `T` is the underlying type. This syntax is shorthand for `System.Nullable<T>`, and the two forms can be used interchangeably.
 
 Conversely, a ***non-nullable value type*** is any value type other than `System.Nullable<T>` and its shorthand `T?` (for any `T`), plus any type parameter that is constrained to be a non-nullable value type (that is, any type parameter with a value type constraint ([§15.2.5](classes.md#1525-type-parameter-constraints))). The `System.Nullable<T>` type specifies the value type constraint for `T`, which means that the underlying type of a nullable value type can be any non-nullable value type. The underlying type of a nullable value type cannot be a nullable value type or a reference type. For example, `int??` and `string?` are invalid types.
```
KalleOlaviNiemitalo: `string?` is no longer an invalid type, although `Nullable<string>` still is.

standard/types.md
```diff
@@ -705,3 +714,138 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+### §Nullable-Types-General General
+
+C# has long supported the notion of a nullable value type; for example, the type `int?` is the nullable version of the value type `int`. (Value types and their nullable variants are described in [§8.3](types.md#83-value-types) and [§8.3.12](types.md#8312-nullable-value-types).) Almost all of the remaining text in this subclause pertains to the addition of support for nullable reference types, with only occasional mention of nullable value types.
+
+There are three forms of nullability for reference types:
+
+- *null-oblivious*: A *null-oblivious* type can be assigned `null`. It's default null state is *not-null*.
+- *nullable*: A *nullable-reference-type* can be assigned `null`. It's default null state is *maybe-null*.
+- *non-nullable*" A *non-nullable reference* should not be assigned a `null` value. It's default null state is *not-null*.
```
KalleOlaviNiemitalo: "Its", not "It's".

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+> *Note:* When the nullable annotation context is disabled, `T` and `T?` both declare null-oblivious reference types. *end note*
+
+### §Nullable-Contexts Nullable contexts
+
+#### §Nullable-Contexts-General General
+
+Every line of source code has a ***nullable annotation context*** and a ***nullable warning context***. The former controls whether nullable annotations (§Nullable-Annotation-Context) have effect; the latter controls whether nullable warnings (§Nullable-Warning-Context) are issued by the compiler. The state of each of the annotation and warning contexts of a given line is disabled or enabled.
+
+Both nullable contexts may be specified within source code via nullable directives (§Nullable-Directives) and/or via some implementation-specific mechanism external to the source code. If both approaches are used, nullable directives supersede the settings made via an external mechanism.
```
BillWagner, line 755: This is causing the section renumber failure. That PR should merge first, then this can be updated as needed.

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+The default state for the nullable annotation context and the nullable warning context is implementation defined.
```
BillWagner, line 757: Rex had originally written this to say the default state is "disabled". I prefer implementation defined, so in the future, if the implementation changes, it won't require changing the standard. (Yes, it's already documented for roslyn).

BillWagner:
@jskeet 

Adding the meeting discuss label. I don't think this is ready to merge. But, we should discuss terms, definitions, and concepts so that the edits can be refined after this next meeting.

standard/types.md
```diff
@@ -431,9 +440,9 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+Conversely, a ***non-nullable value type*** is any value type other than `System.Nullable<T>` and its shorthand `T?` (for any `T`), plus any type parameter that is constrained to be a non-nullable value type (that is, any type parameter with a value type constraint ([§15.2.5](classes.md#1525-type-parameter-constraints))). The `System.Nullable<T>` type specifies the value type constraint for `T`, which means that the underlying type of a nullable value type can be any non-nullable value type. The underlying type of a nullable value type cannot be a nullable value type or a reference type. For example, `int??` is an invalid types. Nullable reference types are covered in §Types-And-Nullability.
```
```suggestion
Conversely, a ***non-nullable value type*** is any value type other than `System.Nullable<T>` and its shorthand `T?` (for any `T`), plus any type parameter that is constrained to be a non-nullable value type (that is, any type parameter with a value type constraint ([§15.2.5](classes.md#1525-type-parameter-constraints))). The `System.Nullable<T>` type specifies the value type constraint for `T`, which means that the underlying type of a nullable value type can be any non-nullable value type. The underlying type of a nullable value type cannot be a nullable value type or a reference type. For example, `int??` is an invalid type. Nullable reference types are covered in §Types-And-Nullability.
```

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
 - `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, or `bool`.
 - Any *enum_type*.
 - Any user-defined *struct_type* that is not a constructed type and contains instance fields of *unmanaged_type*s only.
+- In unsafe code ([§23.2](unsafe-code.md#232-unsafe-contexts)), any *pointer_type* ([§23.3](unsafe-code.md#233-pointer-types)).
+
+## §Types-And-Nullability Reference Types and nullability
+
+### §Nullable-Types-General General
+
+Any expression of a reference type can either refer to an object, or be the value `null`, which indicates "no reference". The terms "nullable reference type" and "non-nullable reference type" refer to annotations that declare whether an expression is intended to be null or not. A compiler can provide diagnostics when a expression is not used according to that intent.
```
```suggestion
Any expression of a reference type can either refer to an object, or be the value `null`, which indicates "no reference". The terms "nullable reference type" and "non-nullable reference type" refer to annotations that declare whether an expression is intended to permit null values or not. A compiler can provide diagnostics when a expression is not used according to that intent.
```

(Otherwise it sounds like any expression of type `string?` is expected to be null, rather than just null*able*...)

jskeet: (I'm not overly fond of "permit null values" either - happy to bikeshed on this.)

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+There are three forms of nullability for reference types:
+
+- *null-oblivious*: A *null-oblivious* type can be assigned `null`. It's default null state is *not-null*.
+- *nullable*: A *nullable-reference-type* can be assigned `null`. It's default null state is *maybe-null*.
```
```suggestion
- *null-oblivious*: A *null-oblivious* type can be assigned `null`. Its default null state is *not-null*.
- *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
```

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+- *non-nullable*" A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
+
+Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same underlying type; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type, its corresponding non-nullable reference type, and its corresponding null-oblivious reference type (§10.2.2).
```
jskeet, line 731: We've said they're different types - I think we should reword "same underlying type", but it will require careful wordsmithing. We should see what we say about `dynamic`.

standard/types.md
```diff
@@ -27,10 +27,15 @@ For convenience, throughout this specification, some library type names are writ
 
 ### 8.2.1 General
 
-A reference type is a class type, an interface type, an array type, a delegate type, or the `dynamic` type.
+A reference type is a class type, an interface type, an array type, a delegate type, or the `dynamic` type. For each non-nullable reference type, there is a corresponding nullable reference type noted by appending the `?` to the type name.
```
jskeet, line 30: And a corresponding null-oblivious type? Later on we seem to be treating them as three different types. Let's discuss how we want to describe this, and then try really hard to be consistent.

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same underlying type; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type, its corresponding non-nullable reference type, and its corresponding null-oblivious reference type (§10.2.2).
+
+> *Note:* By definition, a reference type is nullable; that is, a variable of that type can either contain a reference to an object or be the value `null`, which indicates “no reference.” The disabled nullable context matches the previous standard behavior for reference types. All reference types in these programs are *null-oblivious* reference types. *end note*
+
+### §Null-oblivious-reference-types Null oblivious reference types
+
+A *null-oblivious reference type*** is any reference type declared in a disabled nullable annotation context (§Nullable-Annotation-Context). Variables of the type `T` are allowed to be assigned to `null`. The default null-state of a null-oblivious variable is *not-null*.
```
jskeet, line 737: Is the important part where the type itself is declared, or where some usage is declared? (For example, the nullability of the return type of `public Foo M()` depends on the nullable annotation context of where `M` is declared, not where `Foo` is declared, right?

jskeet: (Likewise for the sections below.)

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+#### §Nullable-Contexts-General General
+
+Every line of source code has a ***nullable annotation context*** and a ***nullable warning context***. The former controls whether nullable annotations (§Nullable-Annotation-Context) have effect; the latter controls whether nullable warnings (§Nullable-Warning-Context) are issued by the compiler. The state of each of the annotation and warning contexts of a given line is disabled or enabled.
+
+Both nullable contexts may be specified within source code via nullable directives (§Nullable-Directives) and/or via some implementation-specific mechanism external to the source code. If both approaches are used, nullable directives supersede the settings made via an external mechanism.
+
+The default state for the nullable annotation context and the nullable warning context is implementation defined.
+
+Throughout this specification, all C# code that does not contain nullable directives, or about which no statement is made regarding the current nullable context state, shall be assumed as having been compiled with nullable annotation and nullable warning contexts enabled.
```
```suggestion
Throughout this specification, all C# code that does not contain nullable directives, or about which no statement is made regarding the current nullable context state, shall be assumed to have been compiled with nullable annotation and nullable warning contexts enabled.
```

(I think?)

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+#### §Nullable-Annotation-Context Nullable annotation context
+
+This context controls whether nullable annotations have any effect.
+
+When the nullable annotation context is disabled
+
+- A variable of any reference type is nullable, so that variable may be initialized with, or assigned a value of, `null`.
```
jskeet, line 767: All variables, or those declared within the context?

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+- No warning shall be generated when a variable of a reference type that possibly has the null value, is dereferenced.
```
jskeet, line 768: This is weird, given that we've said that the nullable warning context controls whether nullable warnings are issued.

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+- For any reference type `T`, the annotation `?` in `T?` is ignored, as `T` is already a nullable type. An informational message should be generated to that effect.
+  > *Note*: This message is characterized as “informational” rather than “warning,” so as not to confuse it with the state of the nullable warning context, which is unrelated.  *end note*
```
jskeet: Is the double space before "end note" deliberate?

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+- The null-forgiving operator `!` is ignored.
+
+> *Example*:
+>
+> <!-- Example: {template:"code-in-main-without-using", name:"NullableAnnotationContext1", ignoredWarnings:["CS0219","CS8632"], expectedException:"NullReferenceException"} -->
+> ```csharp
+> #nullable disable
```
jskeet, line 777: I think this example would be better if it *just* disabled the nullable annotation context.

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+> *end example*
+
+When the nullable annotation context is enabled
+
+- For any reference type `T`, the annotation `?` in `T?` makes `T?` a nullable type, whereas the unannotated `T` is non-nullable.
+- A warning shall be generated when a variable of an unannotated reference type is initialized with, or assigned a value of, `null`.
+- A warning shall be generated when a variable of a nullable or non-nullable reference type, whose value might be `null`, is dereferenced; otherwise, that variable may be dereferenced safely.
+- The null-forgiving operator `!` suppresses warnings about dereferencing a possible null reference.
+
+> *Note*: A Nullable context (§Nullable-Contexts) determines if reference types are considered nullable and non-nullable, or null-oblivious. Switching from a disabled nullable context to an enabled nullable context has the effect of changing the type of reference variables from *null-oblivious* to *non-nullable*. *end note*
```
jskeet, line 794: Nullable annotation context? (And is the capitalized "N" deliberate?)

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+> s2 = null;            // Warning; s1 is non-nullable
+> char c1 = s2[1];      // Warning; dereference of a possible null; throws NullReferenceException
+> c1 = s2![1];          // The warning is suppressed
+> ```
+>
+> *end example*
+
+#### §Nullable-Warning-Context Nullable warning context
+
+This context controls whether nullable warnings are issued by the compiler.
```
jskeet, line 814: Maybe "The nullable warning context controls [...]"? (Same for the annotation context intro.)

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+The compiler uses static flow analysis to determine the null state of any reference variable. When the nullable warning context is enabled, a reference variable’s null state (§Nullabilities-And-Null-States) is either *not null* or *maybe null*, and
```
jskeet, line 816: Variable, or expression? (I'm thinking about properties here.) I'll need to look at the notes from before to see what we said about this. Maybe variable is correct, and we'll just add something saying that an implementation can be smarter.

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+- A warning shall be generated if a reference variable that has been determined to be *maybe null*, is dereferenced.
```
jskeet, line 818: I think we probably need something about `x.Property.M()` regardless of my comment above - if `Property` is declared to be nullable, there should be *something* about the compiler being able to warn, right?

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+- The state of a reference variable is *maybe null* unless the compiler can determine one of two conditions:
+  1. The variable has been definitely assigned a non-`null` value.
+  1. The variable or expression has been checked against `null` before dereferencing it.
+
+Some warnings are generated regardless of the state of the nullable annotation context. (Examples include throwing a value that maybe `null`, and dereferencing a possibly null reference.) Other warnings are generated only if both the nullable warning context and the nullable annotation context are enabled. (Examples include converting a null literal or possible null value to a non-nullable type, possible null reference assignment, and possible null reference return.)
```
```suggestion
Some warnings are generated regardless of the state of the nullable annotation context. (Examples include throwing a value that may be `null`, and dereferencing a possibly null reference.) Other warnings are generated only if both the nullable warning context and the nullable annotation context are enabled. (Examples include converting a null literal or possible null value to a non-nullable type, possible null reference assignment, and possible null reference return.)
```

standard/types.md
jskeet: (There's something later that somewhat explains this, but it's quite a bit later.)

standard/types.md
jskeet: (I see now that we talk about the null state of *expressions* later...)

standard/types.md
BillWagner: I accepted the text, and included it in the related PR.

standard/types.md
BillWagner: nope:

```suggestion
  > *Note*: This message is characterized as “informational” rather than “warning,” so as not to confuse it with the state of the nullable warning context, which is unrelated. *end note*
```

BillWagner:
PR #1124 is an alternative that removes the type "null oblivious".
Either this PR or #1124 should be merged.

standard/types.md
Nigel-Ecma: @jskeet – Does the above say they are different types? We have:

> The terms "nullable reference type" and "non-nullable reference type" refer to annotations that declare whether an expression is intended to permit null values or not.

This language suggests there is one type which can have different annotations, not that there are multiple distinct types.

Maybe the term “underlying type” should not be overloaded to deal with (non-)nullable reference types? Given the text is leaning on annotations maybe a better term to use here is “non annotated type” (is it *non annotated*, *nonannotated* or *unannotated*?). This would give:

> […] given a reference type `R`, the notations `R` and `R?` denote the same non annotated type […]

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+There are three forms of nullability for reference types:
+
+- *null-oblivious*: A *null-oblivious* type can be assigned `null`. Its default null state is *not-null*.
+- *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
+- *non-nullable*" A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
+
+Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same underlying type; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type, its corresponding non-nullable reference type, and its corresponding null-oblivious reference type (§10.2.2).
```
Nigel-Ecma, line 731: > […] and allows for static flow analysis.

Reads wrong, I don’t think the annotations allow static flow analysis *per se*; they may assist/make possible a *particular* static flow analysis.

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+### §Nullable-Types-General General
+
+Any expression of a reference type can either refer to an object, or be the value `null`, which indicates "no reference". The terms "nullable reference type" and "non-nullable reference type" refer to annotations that declare whether an expression is intended to permit null values or not. A compiler can provide diagnostics when a expression is not used according to that intent.
+
+There are three forms of nullability for reference types:
+
+- *null-oblivious*: A *null-oblivious* type can be assigned `null`. Its default null state is *not-null*.
+- *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
+- *non-nullable*" A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
```
Nigel-Ecma, line 729: This is referring to “null state” which is not yer defined (it’s ~100 lines further down) – a cross-reference at least should be added.

I also think the default null state difference between *null-oblivious*, which is *not-null*, and *nullable*, which is *maybe-null*, is jarring enough to require an explanation, even if that is just a forward reference.

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same underlying type; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type, its corresponding non-nullable reference type, and its corresponding null-oblivious reference type (§10.2.2).
+
+> *Note:* By definition, a reference type is nullable; that is, a variable of that type can either contain a reference to an object or be the value `null`, which indicates “no reference.” The disabled nullable context matches the previous standard behavior for reference types. All reference types in these programs are *null-oblivious* reference types. *end note*
```
Nigel-Ecma, line 733: Forward reference to *nullable context* should be added.

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+A reference type of the form `T?` (such as `string?`) is a ***nullable reference type***. The annotation `?` indicates the intent that variables of this type are nullable. When the nullable annotation context (§Nullable-Annotation-Context) is enabled, the compiler shall recognize these intents. When the nullable annotation context is disabled, the compiler shall ignore these intents. When the nullable annotation context is disabled, a reference type of the form `T?` shall declare a null-oblivious reference type.
+
+> *Note:* When the nullable annotation context is disabled, `T` and `T?` both declare null-oblivious reference types. *end note*
+
+### §Nullable-Contexts Nullable contexts
+
+#### §Nullable-Contexts-General General
+
+Every line of source code has a ***nullable annotation context*** and a ***nullable warning context***. The former controls whether nullable annotations (§Nullable-Annotation-Context) have effect; the latter controls whether nullable warnings (§Nullable-Warning-Context) are issued by the compiler. The state of each of the annotation and warning contexts of a given line is disabled or enabled.
```
Nigel-Ecma, line 753: I’m not sure about the wording of (*emphasis* added):

> The former controls whether nullable annotations (§Nullable-Annotation-Context) have *effect*; the latter controls whether nullable warnings (§Nullable-Warning-Context) *are issued* by the compiler.

I would think the word “effect” in the context of a language specification would by default be taken to imply some change to the interpretation of the syntax or semantics. NRB is, I understand, being presented as an *optional* analysis which may produce useful comments (aka warnings) about (a fragment of) the language, but *not* have any impact at all on the *meaning* of that (fragment of) the language. In this sense the annotation and context have *no* direct effect at all – they may have an indirect effect in inducing the human/AI who wrote to the code to revise it.

standard/types.md
```diff
@@ -705,3 +714,150 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+> c1 = s2![1];          // The warning is suppressed
+> ```
+>
+> *end example*
+
+#### §Nullable-Warning-Context Nullable warning context
+
+This context controls whether nullable warnings are issued by the compiler.
+
+The compiler uses static flow analysis to determine the null state of any reference variable. When the nullable warning context is enabled, a reference variable’s null state (§Nullabilities-And-Null-States) is either *not null* or *maybe null*, and
```
Nigel-Ecma, line 816: > The compiler uses static flow analysis […]

Is that a requirement of the Standard? I’m a bit fuzzy on where the (fuzzy) line is here.

As I currently understand an implementation must accept the annotations and pragmas, but may produce any warnings based on those that it likes – specific warnings are not being required.

The hope/allowance is that over time implementations will produce better warnings (and no “hallucinations” to use the in-phrase ;-)) from improving their analyses. But this also implies an implementation could choose to produce no warnings and still be compliant, and if it chooses to do so no analyses would be required…

That the analysis must be *static* could be questioned (consider typing where some limited dynamic typing is supported) but I expect it's unlikely to be an issue for the foreseeable future. But what about it being a *flow* analysis – could a new algorithm be developed that doesn’t fit that category?

TL;DR: I wondering whether, given the lack of requirements to meet conformance for this feature, why the language is not about “an implementation *may*” do such and such, like use a static flow analysis… produce a warning… or even determine that something definitely *is-null* and not just *maybe-null*

Nigel-Ecma (commented):
Overall, as my comments undoubtedly show, I’m not sure it is yet clear what the boundary is between what an implementation (and we probably shouldn't say “compiler” here) must do and may do to be conformant.

BillWagner:
Closing in favor of #1124

PR #1110, opened 2024-05-15
Fix #1109

### LINKED ISSUE #1109: Base access in expression body
**Describe the bug**

In the C# 7 standard, §12.8.14 (Base access) says:

> A *base_access* is permitted only in the *block* of an instance constructor, an instance method, an instance accessor (§12.2.1), or a finalizer.

This should allow *base_access* also in a *method_body*, *ref_method_body*, *property_body*, *ref_property_body*, *indexer_body*, *ref_indexer_body*, *constructor_body*, or *finalizer_body* that is not a *block*.

**Example**

This should be allowed:

```csharp
class A {
    private int i;

    public A() {}
    
    public int M() => 0;
    public ref int RM() => ref this.i;

    public int P => 0;
    public ref int RP => ref this.i;
    public int PA {
        get => 0;
        set {}
    }
    public ref int RPA {
        get => ref this.i;
    }

    public int this[short index] => 0;
    public ref int this[char index] => ref this.i;

    public event System.Action EA {
        add {}
        remove {}
    }
}

class B : A {
    public B() => base.M();

    public new int M() => base.M();
    public new ref int RM() => ref base.RM();

    public new int P => base.P;
    public ref int RP => ref base.RP;
    public new int PA {
        get => base.PA;
        set => base.PA = value;
    }
    public new ref int RPA {
        get => ref base.RPA;
    }

    public new int this[short index] => base[index];
    public new ref int this[char index] => ref base[index];
    
    public new event System.Action EA {
        add => base.EA += value;
        remove => base.EA -= value;
    }
    
    ~B() => base.M();
}
```

**Expected behavior**

Specify that the preceding is allowed.

**Additional context**

Exclude *anonymous_function_body*, *local_function_body*, and *ref_local_function_body*.  They can contain a *base_access* only when they occur within some other body.

Exclude *static_constructor_body* and *operator_body*.  They are not used in instance members.

Nigel-Ecma:
Fixes #1109
§12.8.14 Base access was not updated when expression bodies were introduced and still referred to *block* rather than body (which is now a block or expression).

Reported by @KalleOlaviNiemitalo

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -2218,7 +2218,7 @@ base_access
     ;
 ```
 
-A *base_access* is used to access base class members that are hidden by similarly named members in the current class or struct. A *base_access* is permitted only in the *block* of an instance constructor, an instance method, an instance accessor ([§12.2.1](expressions.md#1221-general)), or a finalizer. When `base.I` occurs in a class or struct, I shall denote a member of the base class of that class or struct. Likewise, when `base[E]` occurs in a class, an applicable indexer shall exist in the base class.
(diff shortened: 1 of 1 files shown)


jskeet (approved):

BillWagner (approved):
This LGTM.

PR #1108, opened 2024-05-10
C# 8 nullable reference types - Preprocessor symbols for nullable reference types

BillWagner:
Split from #1105

Fixes #1088 

Separate the portion of 1105 that defines the nullable context. This includes only the preprocessor symbols, update to one introductory sample, and a related fix that updates one sample test.

PR #1112, opened 2024-05-16
Fix explicit interface example

### LINKED ISSUE #1087: Example for explicit interface member implemenation
### Type of issue

Other (describe below)

### Description

From: [Explicit interface member implementations](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/interfaces#1862-explicit-interface-member-implementations)

``` C#
interface IList<T>
{
    T[] GetElements();
}

interface IDictionary<K, V>
{
    V this[K key] { get; }
    void Add(K key, V value);
}

class List<T> : IList<T>, IDictionary<int, T>
{
    T[] IList<T>. GetElements() {...}
    T IDictionary<int, T>.this[int index] {...}
    void IDictionary<int, T>.Add(int index, T value) {...}
}
```
Here IDictionary<int,T>.this and IDictionary<int,T>.Add are explicit interface member implementations.

---

Question: Is `IList<T>. GetElements()` in `T[] IList<T>. GetElements() {...}` not an explicit interface member implementation?

### Page URL

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/interfaces

### Content source URL

https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/interfaces.md

jskeet:
Fixes #1087.

### DIFF
--- a/standard/interfaces.md
+++ b/standard/interfaces.md
@@ -627,7 +627,7 @@ For purposes of implementing interfaces, a class or struct may declare ***explic
 >
 > class List<T> : IList<T>, IDictionary<int, T>
 > {
->     T[] IList<T>. GetElements() {...}
+>     public T[] GetElements() {...}
 >     T IDictionary<int, T>.this[int index] {...}
 >     void IDictionary<int, T>.Add(int index, T value) {...}
 > }


jskeet:
I think the test failure is unrelated to this change. @BillWagner are you happy to address that? Let me know if you'd like me to do so.

BillWagner:
> I think the test failure is unrelated to this change. @BillWagner are you happy to address that? Let me know if you'd like me to do so.

That's already addressed in #1108 

We should be good once we get the remaining question there resolved.

BillWagner (approved):
This LGTM. I'll :shipit:

PR #1119, opened 2024-05-25
Minor editorial tweaks to the Attributes chapter

RexJaeschke:
Bill, this is the extended-attribute example we spoke of yesterday. Plus, I noted a need for clarity in a neighboring section.


BillWagner (approved):
Thanks @RexJaeschke 

This LGTM. I'll :shipit: now, so our builds are fixed.

Issue #1085, opened 2024-04-24
[Epic] Specify nullable reference types in C# 8


jskeet:
#700 is a draft which will be useful, but we're going to try to do this in multiple small PRs.

The current planned task list is:

- [x] #1088
- [x] #1089 
- [x] #1090 
- [x] #1091
- [x] #1092
- [x] #1093
- [x] #1094

KalleOlaviNiemitalo:
Are [MaybeNullWhenAttribute](https://learn.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.maybenullwhenattribute?view=netstandard-2.1) and others going to be standardised?

BillWagner:
> Are [MaybeNullWhenAttribute](https://learn.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.maybenullwhenattribute?view=netstandard-2.1) and others going to be standardised?

Yes, see #1092

KalleOlaviNiemitalo:
Does any of those issues cover how variance of generic parameters should affect warnings about identity conversions:

```csharp
using System;
public class C {
    // f1 is a delegate that takes string? and returns string.
    public void M(Func<string?, string> f1) {
        // Nullability of type arguments does not match,
        // but there is no warning, because of variance.
        Func<string, string?> f2 = f1;

        // Convert back to the original type.
        // This causes a warning.
        Func<string?, string> f3 = f2;
    }
}
```


<https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/nullable-reference-types-specification.md> doesn't seem to mention this.

I'm not sure whether this behaviour is C# 8 or 9, but I guess 8.

BillWagner:
Terminology question: Do we want to use "nullable context" or "null analysis context"?  Alternative is "set the flags in the nullable context".

See https://github.com/dotnet/csharpstandard/pull/1123/files#r1672741166

However, note that the null state analysis is *always* being done. The comment states that the context may impact that.

PR #1123, opened 2024-05-29
Preprocessor symbols for nullable reference types: Define a single nullable context.

### LINKED ISSUE #1088: [Nullable Reference Types] Compiler directives for nullable reference types
Specify the compiler directive `#nullable` that controls the interpretation of reference types and nullability. The effect of the directives is as follows:

- `#nullable disable`: Sets the nullable annotation and warning contexts to *disabled*.
- `#nullable enable`: Sets the nullable annotation and warning contexts to *enabled*.
- `#nullable restore`: Restores the nullable annotation and warning contexts to project settings.
- `#nullable disable annotations`: Sets the nullable annotation context to *disabled*.
- `#nullable enable annotations`: Sets the nullable annotation context to *enabled*.
- `#nullable restore annotations`: Restores the nullable annotation context to project settings.
- `#nullable disable warnings`: Sets the nullable warning context to *disabled*.
- `#nullable enable warnings`: Sets the nullable warning context to *enabled*.
- `#nullable restore warnings`: Restores the nullable warning context to project settings.

In addition, the spec should define the behavior of reference types in those contexts.

---
[Associated WorkItem - 248771](https://dev.azure.com/msft-skilling/Content/_workitems/edit/248771)

BillWagner:
Fixes #1088 

Alternative PR to #1108

This branch builds on the work in #1108. This is an alternative approach that defines a single `nullable` context. The nullable context has two properties: `annotations` and `warnings` that can be enabled or disabled separately.

### DIFF
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -1529,19 +1529,17 @@ fragment PP_Nullable_Target
     ;
 ```
 
-A nullable directive sets the denoted nullable context for subsequent lines of code, until another nullable directive overrides it, or until the end of the source code is reached. The effect of each form of nullable directive is, as follows:
+A nullable directive sets the denoted nullable context for subsequent lines of code, until another nullable directive overrides it, or until the end of the source code is reached. The nullable context contains two flags: *annotations* and *warnings*. The effect of each form of nullable directive is, as follows:
 
-- `#nullable disable`: Sets the nullable context to “disabled”.
-- `#nullable enable`: Sets the nullable context to “enabled”.
-- `#nullable restore`: Restores the nullable context to the state specified by the external mechanism, if any.
-- `#nullable disable annotations`: Disables nullable annotations. Nullable warnings are unaffected.
-- `#nullable enable annotations`: Enables nullable annotations. Nullable warnings are unaffected.
-- `#nullable restore annotations`: Restores nullable annotations to the state specified by the external mechanism, if any. Warnings are unaffected.
-- `#nullable disable warnings`: Disables nullable warnings. Nullable annotations are unaffected.
-- `#nullable enable warnings`: Enables nullable warnings. Nullable annotations are unaffected.
-- `#nullable restore warnings`: Restores nullable warnings to the state specified by the external mechanism, if any. Nullable annotations are unaffected.
-
-Disabling the nullable context when it is already disabled has no effect. Likewise, enabling the nullable context when it is already enabled has no effect.
+- `#nullable disable`: Disables both nullable annotations and nullable warnings flags.
+- `#nullable enable`: Enables both nullable annotations and nullable warnings flags.
+- `#nullable restore`: Restores both the annotations and warnings flags to the state specified by the external mechanism, if any.
+- `#nullable disable annotations`: Disables the nullable annotations flag. The nullable warnings flag is unaffected.
+- `#nullable enable annotations`: Enables the nullable annotations flag. The nullable warnings flag is unaffected.
+- `#nullable restore annotations`: Restores the nullable annotations flag to the state specified by the external mechanism, if any. The nullable warnings flag is unaffected.
+- `#nullable disable warnings`: Disables the nullable warnings flag. The nullable annotations flag is unaffected.
+- `#nullable enable warnings`: Enables the nullable warnings flag. The nullable annotations flag is unaffected.
+- `#nullable restore warnings`: Restores the nullable warnings flag to the state specified by the external mechanism, if any. The nullable annotations flag is unaffected.
 
 ### 6.5.9 Pragma directives
 
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -1529,7 +1529,7 @@ fragment PP_Nullable_Target
     ;
 ```
 
-A nullable directive sets the denoted nullable context for subsequent lines of code, until another nullable directive overrides it, or until the end of the source code is reached. The nullable context contains two flags: *annotations* and *warnings*. The effect of each form of nullable directive is, as follows:
+A nullable directive sets the available flags for subsequent lines of code, until another nullable directive overrides it, or until the end of the source code is reached. The nullable context contains two flags: *annotations* and *warnings*. The effect of each form of nullable directive is, as follows:
 
 - `#nullable disable`: Disables both nullable annotations and nullable warnings flags.
 - `#nullable enable`: Enables both nullable annotations and nullable warnings flags.
@@ -1540,6 +1540,24 @@ A nullable directive sets the denoted nullable context for subsequent lines of c
 - `#nullable disable warnings`: Disables the nullable warnings flag. The nullable annotations flag is unaffected.
 - `#nullable enable warnings`: Enables the nullable warnings flag. The nullable annotations flag is unaffected.
 - `#nullable restore warnings`: Restores the nullable warnings flag to the state specified by the external mechanism, if any. The nullable annotations flag is unaffected.
+
+The nullable state of expressions is tracked at all times. The state of the annotation flag determines the initial null state of a variable declaration. Warnings are only issued when the warnings flag is enabled.
+
+> *Example*: The example
+>
+> <!-- Example: {template:"standalone-console", name:"InitialWarning", ignoredWarnings:["CS8602"], expectedException:"NullReferenceException"}} -->
+> ```csharp
+> #nullable disable
+> string x = null;
+> string y = "";
+> #nullable enable
+> Console.WriteLine(x.Length); // Warning
+> Console.WriteLine(y.Length);
+> ```
+>
+> produces a compile-time warning (“dereference of a possible null reference”). The nullable state of `x` is tracked everywhere. A warning is issued when the warnings flag is enabled.
+>
+> *end example*
 
 ### 6.5.9 Pragma directives
 
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -1026,7 +1026,7 @@ right_shift_assignment
 
 ### 6.5.1 General
 
-The pre-processing directives provide the ability to conditionally skip sections of compilation units, to report error and warning conditions, and to delineate distinct regions of source code.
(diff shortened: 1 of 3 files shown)


jskeet (approved):
Yup, personally I like this better than having multiple separate notional contexts.

standard/lexical-structure.md
```diff
@@ -1507,6 +1509,40 @@ A `#line hidden` directive has no effect on the compilation unit and line number
...
+fragment PP_Nullable_Target
+    : 'warnings'
+    | 'annotations'
+    ;
+```
+
+A nullable directive sets the denoted nullable context for subsequent lines of code, until another nullable directive overrides it, or until the end of the source code is reached. The effect of each form of nullable directive is, as follows:
+
+- `#nullable disable`: Sets the nullable context to “disabled”.
+- `#nullable enable`: Sets the nullable context to “enabled”.
```
Nigel-Ecma: Needs to say both annotations and warnings are enabled.

The previous bullet could also say it disables both, though that is not so important if the  options are:

- null context disabled, or
- null context enabled with annotations and/or warnings

standard/lexical-structure.md
```diff
@@ -1507,6 +1509,40 @@ A `#line hidden` directive has no effect on the compilation unit and line number
...
+- `#nullable restore`: Restores the nullable context to the state specified by the external mechanism, if any.
+- `#nullable disable annotations`: Disables nullable annotations. Nullable warnings are unaffected.
+- `#nullable enable annotations`: Enables nullable annotations. Nullable warnings are unaffected.
+- `#nullable restore annotations`: Restores nullable annotations to the state specified by the external mechanism, if any. Warnings are unaffected.
+- `#nullable disable warnings`: Disables nullable warnings. Nullable annotations are unaffected.
+- `#nullable enable warnings`: Enables nullable warnings. Nullable annotations are unaffected.
+- `#nullable restore warnings`: Restores nullable warnings to the state specified by the external mechanism, if any. Nullable annotations are unaffected.
+
+Disabling the nullable context when it is already disabled has no effect. Likewise, enabling the nullable context when it is already enabled has no effect.
```
Nigel-Ecma: `#nullable enable` has an effect if only one of annotations and warnings are currently enabled as it enables the other.

Nigel-Ecma (changes requested):
A couple of language fixes/clarifications required so it describes the same overall semantics as #1108.

Nigel-Ecma:
> Yup, personally I like this better than having multiple separate notional contexts.

I probably agree (yes I'm hedging ;-)) – If a file starts with an enable/disable and never changes it makes little difference. However I can see that when nullable state is enabled & disabled at different points in a file having one context might make that semantic description simpler.

standard/lexical-structure.md
jskeet: Personally I'd prefer the approach of saying "enables both annotations and warnings".

jskeet:
Example of something that might be hard to understand:

```csharp
#nullable disable
string x = null;
string y = "";
#nullable enable
Console.WriteLine(x.ToString());
Console.WriteLine(y.ToString());
```

Should that give a warning for `x.ToString()`?

#1124 states: "However, the disabled context means the compiler does not track null state and does not issue nullable related warnings" - so that would suggest no warning.

However, (with .NET 8 SDK) it issues a warning:

warning CS8602: Dereference of a possibly null reference.

standard/lexical-structure.md
jskeet: In meeting: no need for this.

standard/lexical-structure.md
jskeet: In meeting: yes, make each of these specify the effect on both flags.

jskeet (approved):
Approved but with changes as described.

BillWagner:
I updated everything from the last meeting (both comments, and from the in-meeting discussion). This should be ready to merge @jskeet 

We can discuss at the meeting, or merge before.

jskeet:
> We can discuss at the meeting, or merge before.

Let's get a final review at the meeting just in case. It would be handy if everyone could review beforehand though.

standard/lexical-structure.md
```diff
@@ -1026,7 +1026,7 @@ right_shift_assignment
 
 ### 6.5.1 General
 
-The pre-processing directives provide the ability to conditionally skip sections of compilation units, to report error and warning conditions, and to delineate distinct regions of source code.
+The pre-processing directives provide the ability to conditionally skip sections of compilation units, to report error and warning conditions, to delineate distinct regions of source code, and to set the nullable context.
```
Nigel-Ecma, line 1029: Reading this now “nullable context” caused a pause. We have a null analysis, which may be enabled or disabled; and certain annotations (like adding a `?` to a type) etc. which both provide input to that analysis and which are treated differently depending on whether the analysis is being done. Is this a “nullable context”, a ”null analysis context”, something else?

standard/lexical-structure.md
```diff
@@ -1507,6 +1509,56 @@ A `#line hidden` directive has no effect on the compilation unit and line number
...
+    | 'enable'
+    | 'restore'
+    ;
+fragment PP_Nullable_Target
+    : 'warnings'
+    | 'annotations'
+    ;
+```
+
+A nullable directive sets the available flags for subsequent lines of code, until another nullable directive overrides it, or until the end of the source code is reached. The nullable context contains two flags: *annotations* and *warnings*. The effect of each form of nullable directive is, as follows:
```
Nigel-Ecma: “until the end of the compilation unit is reached”? (or *compilation_unit*)

standard/lexical-structure.md
```diff
@@ -1507,6 +1509,56 @@ A `#line hidden` directive has no effect on the compilation unit and line number
...
+- `#nullable restore`: Restores both the annotations and warnings flags to the state specified by the external mechanism, if any.
+- `#nullable disable annotations`: Disables the nullable annotations flag. The nullable warnings flag is unaffected.
+- `#nullable enable annotations`: Enables the nullable annotations flag. The nullable warnings flag is unaffected.
+- `#nullable restore annotations`: Restores the nullable annotations flag to the state specified by the external mechanism, if any. The nullable warnings flag is unaffected.
+- `#nullable disable warnings`: Disables the nullable warnings flag. The nullable annotations flag is unaffected.
+- `#nullable enable warnings`: Enables the nullable warnings flag. The nullable annotations flag is unaffected.
+- `#nullable restore warnings`: Restores the nullable warnings flag to the state specified by the external mechanism, if any. The nullable annotations flag is unaffected.
+
+The nullable state of expressions is tracked at all times. The state of the annotation flag determines the initial null state of a variable declaration. Warnings are only issued when the warnings flag is enabled.
```
Nigel-Ecma: “The nullable state of expressions…” vs. “…initial null state of a variable…”
Change the former to “null”?
“is tracked” -> ”shall be tracked” – that’s one for @RexJaeschke to determine, but the tense seems wrong to me and maybe the Standardese.

standard/lexical-structure.md
```diff
@@ -1507,6 +1509,56 @@ A `#line hidden` directive has no effect on the compilation unit and line number
...
+> ```csharp
+> #nullable disable
+> string x = null;
+> string y = "";
+> #nullable enable
+> Console.WriteLine(x.Length); // Warning
+> Console.WriteLine(y.Length);
+> ```
+>
+> produces a compile-time warning (“dereference of a possible null reference”). The nullable state of `x` is tracked everywhere. A warning is issued when the warnings flag is enabled.
```
Nigel-Ecma: I would replace:

> (“dereference of a possible null reference”)

with

> (as `x` is `null`)

as the former reads like it is specifying the error text, including “possible”, even though we know its non-normative.

Of course this also states *non normatively* that there will be a warning, we is also not currently required…

standard/lexical-structure.md
gafter: The state of the annotation flag *and* the presence or absence of a nullable annotation `?`.

Nigel-Ecma (dismissed):
A few language issues.

I also do not think we should approve/merge this, it is one of a set and as the others are developed we might need to circle around to this or any of them. So we should leave it as "done for now", a well known Github state ;-)

standard/lexical-structure.md
BillWagner:
```suggestion
The nullable state of expressions is tracked at all times. The state of the annotation flag and the presence or absence of a nullable annotation, `?`, determines the initial null state of a variable declaration. Warnings are only issued when the warnings flag is enabled.
```

standard/lexical-structure.md
BillWagner:
```suggestion
> produces a compile-time warning (“as `x` is `null`”). The nullable state of `x` is tracked everywhere. A warning is issued when the warnings flag is enabled.
```

standard/lexical-structure.md
BillWagner: Tracked in the epic.

standard/lexical-structure.md
BillWagner:
```suggestion
A nullable directive sets the available flags for subsequent lines of code, until another nullable directive overrides it, or until the end of the *compilation _unit* is reached. The nullable context contains two flags: *annotations* and *warnings*. The effect of each form of nullable directive is, as follows:
```

jskeet (approved):

jskeet:
> I also do not think we should approve/merge this, it is one of a set and as the others are developed we might need to circle around to this or any of them. So we should leave it as "done for now", a well known Github state ;-)

This was discussed unfortunately in Nigel's absence due to connectivity issues. We agreed that it would be better to avoid having multiple PRs hanging around and getting stale - the spec is "always inconsistent" in a draft form, as it doesn't accurately describe any specific version of C#.

If Nigel wishes to convince us otherwise, we could always revert the commit and open another PR :)

We will dismiss Nigel's review in order to merge now.

standard/lexical-structure.md
Nigel-Ecma: Re: [epic comment](https://github.com/dotnet/csharpstandard/issues/1085#issuecomment-2221397275) – yes “is being done” should be worded ”is enabled/disabled”. Which unfortunately highlights something we are stuck with – the analysis is always being done but the pragmas enable/disable it… C’est la vie

Issue #1093, opened 2024-04-26
[Nullable Reference Types] Specify nullable flow analysis lattice


BillWagner:
Specify the general rules for static flow analysis of nullable states. Namely, the standard should specify:

- The possible null-state of a variable or expression (*not-null* or *maybe-null*)
- The initial nullable state of a variable, based on the nullable context and its declared type
- Some general rules about flow analysis and diagnostics.

That final bullet is fraught with risk. We don't want to constrain any future compiler innovations that improve static analysis. However, we need to paint some general view of the diagnostics to provide useful information both for compiler writers and C# programmers.

BillWagner:
See https://github.com/dotnet/csharpstandard/pull/1124/files#r1632781297 for more context here.

Depending on the resolution of that conversation, this PR should clarify how property accessors are handled. The result of invoking the `set` accessor, or the null-state of the value returned by a `get` accessor impacts the null-state of subsequent `get` accessors.

Stated non-normatively, the null state of a property is treated as though it is a variable.

That is not true for method invocations: Their result is not considered consistent between invocations.

BillWagner:
From discussion during the 7/10 meeting:

We may to have three state:

- *not-null*
- *maybe-null*
- *oblivious*

The existence of the *oblivious* state comes from any declaration when the annotation flag is *disabled*

BillWagner:
We should discuss this issue and #1092 in our next meeting.

Does any of the information on the static analysis belong in the standard?

If so, what?

BillWagner:
While working on the PR for generic type constraints, I looked again at #700. 

That contains normative language on requirements to initialize static and instance fields either in initializers or all constructors. Should we add either conditionally normative or informative language regarding those rules?

jskeet:
Decision 2024-10-02: closing as not needed

Issue #1091, opened 2024-04-26
[Nullable Reference Types] Specify new generic constraints necessary for nullable reference types


BillWagner:
New constraints have been defined to support nullable reference types. These include `notnull`, `class?`, `default`

BillWagner:
This has been completed

PR #1124, opened 2024-05-29
Different reference types: nullable and non-nullable

### LINKED ISSUE #1089: [Nullable Reference Types] Nullable, Non-nullable, and "Nullable oblivous" reference types

### LINKED ISSUE #1090: [Nullable Reference Types] Specify syntax for nullable reference types
This would specify any additional syntax and grammar needed for the new tokens for nullable reference types. This includes the `?` and `!` tokens, and any restrictions on where nullable reference types can be used in a valid C# program.

Notes that the `?` may be mostly covered by #1089

BillWagner:
Fixes #1089
Fixes #1090 

This is an alternative to #1105 

The changes were done on top of the branch for #1105. The final commit has the updates to remove null oblivious types.

It removes the text for a "null-oblivious" type. Instead, it relies on the nullable context for how to interpret the nullability of a reference type `T`.


standard/types.md
```diff
@@ -722,19 +722,14 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
 
-> *Note:* By definition, a reference type is nullable; that is, a variable of that type can either contain a reference to an object or be the value `null`, which indicates “no reference.” The disabled nullable context matches the previous standard behavior for reference types. All reference types in these programs are *null-oblivious* reference types. *end note*
-
-### §Null-oblivious-reference-types Null oblivious reference types
-
-A *null-oblivious reference type*** is any reference type declared in a disabled nullable annotation context (§Nullable-Annotation-Context). Variables of the type `T` are allowed to be assigned to `null`. The default null-state of a null-oblivious variable is *not-null*.
+> *Note:* By definition, a reference type is nullable; that is, a variable of that type can either contain a reference to an object or be the value `null`, which indicates “no reference.” The disabled nullable context matches the previous standard behavior for reference types. All reference types in these programs can be considered *null-oblivious*. Their default null state is *not-null*, and they can be assigned a `null` value. *end note*
```
jskeet: This still talks about a *type* as being null-oblivious.

standard/types.md
```diff
@@ -722,19 +722,14 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
 
-There are three forms of nullability for reference types:
+There are two forms of nullability for reference types:
 
-- *null-oblivious*: A *null-oblivious* type can be assigned `null`. Its default null state is *not-null*.
 - *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
 - *non-nullable*" A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
 
 Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same underlying type; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type, its corresponding non-nullable reference type, and its corresponding null-oblivious reference type (§10.2.2).
```
jskeet: We need to remove the end of this sentence.

standard/types.md
```diff
@@ -809,18 +804,18 @@ When the nullable annotation context is enabled
...
 > *end example*
 
-#### §Nullable-Warning-Context Nullable warning context
+#### §Nullable-Warning-Context Nullable warning setting
 
-This context controls whether nullable warnings are issued by the compiler.
+This settings controls whether nullable warnings are issued by the compiler.
 
-The compiler uses static flow analysis to determine the null state of any reference variable. When the nullable warning context is enabled, a reference variable’s null state (§Nullabilities-And-Null-States) is either *not null* or *maybe null*, and
+The compiler uses static flow analysis to determine the null state of any reference variable. When nullable warnings are enabled, a reference variable’s null state (§Nullabilities-And-Null-States) is either *not null*, *maybe null*, or *maybe default* and
```
jskeet: We don't appear to reference *maybe default* anywhere else. Is that intended?

standard/types.md
```diff
@@ -722,19 +722,14 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
 - *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
 - *non-nullable*" A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
 
 Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same underlying type; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type, its corresponding non-nullable reference type, and its corresponding null-oblivious reference type (§10.2.2).
 
-> *Note:* By definition, a reference type is nullable; that is, a variable of that type can either contain a reference to an object or be the value `null`, which indicates “no reference.” The disabled nullable context matches the previous standard behavior for reference types. All reference types in these programs are *null-oblivious* reference types. *end note*
-
-### §Null-oblivious-reference-types Null oblivious reference types
```
jskeet: I think we will still need some kind of section defining what null-oblivious means - it's just that instead of referring to a type, it'll be to the state of an expression. (I *think*, anyway...)

standard/types.md
```diff
@@ -836,28 +831,27 @@ The ***default null state*** of an expression is determined by its type:
...
-- A reference type `C` in a disabled nullable annotation context is oblivious
+- A reference type `C` in a disabled nullable annotation context is nullable, however its default null state is *not-null*.
 - A reference type `C` in an enabled nullable annotation context is non-nullable
 - A nullable reference type `C?` is nullable; however, a warning shall be generated in a disabled nullable annotation context
 
 With a type parameter, any constraints are taken into account:
 
 - A type parameter `T` where all constraints (if any) are either nullable types or the `class?` constraint, is `nullable`
-- A type parameter `T` where at least one constraint is either oblivious or non-nullable, or one of the constraints `struct`,  `class`, or `notnull`, is
+- A type parameter `T` where at least one constraint is non-nullable, or one of the constraints `struct`,  `class`, or `notnull`, is
```
jskeet: Here's another example of where we refer to "oblivious" without really defining it.

standard/types.md
BillWagner: Yes, for now. It will be expanded when we add the PR for generics and nullable reference types. That's when *maybe default* impacts null state analysis.

standard/types.md
BillWagner: I think I finessed this is in the latest commit. Let me know what you think.

standard/types.md
BillWagner: Should be fixed.

standard/types.md
BillWagner: Updated in the latest commit.

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+### §Nullable-Types-General General
+
+Any expression of a reference type can either refer to an object, or be the value `null`, which indicates "no reference". The terms "nullable reference type" and "non-nullable reference type" refer to annotations that declare whether an expression is intended to permit null values or not. A compiler can provide diagnostics when a expression is not used according to that intent.
+
+There are two forms of nullability for reference types:
+
+- *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
+- *non-nullable*" A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
+
+Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same underlying type; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type and its corresponding non-nullable reference type (§10.2.2).
```
jskeet: Do we need to define what "underlying type" means in this context?

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+> *Note:* By definition, a reference type is nullable; that is, a variable of that type can either contain a reference to an object or be the value `null`, which indicates “no reference.” In a disabled nullable context, reference types aren't annotated. Therefore, they are non-nullable reference types. However, the disabled context means the compiler does not track null state and does not issue nullable related warnings. The disabled nullable context matches the previous standard behavior for reference types. *end note*
+
+### §Non-nullable-reference-types Non-nullable reference types
+
+A ***non-nullable reference type*** is a reference type declared in an enabled nullable annotation context (§Nullable-Annotation-Context). A variable of a non-nullable reference type of the form `T` should not contain the value `null`. Warnings are generated when a variables of the type `T` is assigned to `null`. The default null-state of a non-nullable variable is *not-null*.
```
jskeet: This seems to go against the description above:

> In a disabled nullable context, reference types aren't annotated. Therefore, they are non-nullable reference types.

That's not "a reference type declared in an enabled nullable annotation context".

No proposal for a change yet, but if we agree it's an issue I can put my thinking cap on.

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+### §Nullable-reference-types Nullable reference types
+
+A reference type of the form `T?` (such as `string?`) is a ***nullable reference type***. The annotation `?` indicates the intent that variables of this type are nullable. When the nullable annotation context (§Nullable-Annotation-Context) is enabled, the compiler shall recognize these intents. When the nullable annotation context is disabled, the compiler shall ignore these intents. When the nullable annotation context is disabled, a reference type of the form `T?` shall declare a null-oblivious reference type.
```
jskeet: This still refers to null-oblivious, as does the note.

standard/types.md
jskeet: SGTM.

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+> s2 = null;            // OK; null assignment to a reference
+> char c1 = s2[1];      // OK; no warning on dereference of a possible null; throws NullReferenceException
+> c1 = s2![1];          // OK; ! is ignored
+> ```
+>
+> *end example*
+
+#### §Nullable-Annotation-Context Nullable annotations
+
+When the nullable context is annotations:
```
jskeet: I wonder whether "disabled", "annotations", "warnings" etc should be in code font, or quotes, or *something* like that. It looks odd at the moment.

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+> *end example*
+
+#### §Nullable-Warning-Context Nullable warnings
+
+When the nullable context is warnings:
+
+The compiler uses static flow analysis to determine the null state of any reference variable. When nullable warnings are enabled, a reference variable’s null state (§Nullabilities-And-Null-States) is either *not null*, *maybe null*, or *maybe default* and
+
+- A warning shall be generated if a reference variable that has been determined to be *maybe null*, is dereferenced.
+- The state of a reference variable is *maybe null* unless the compiler can determine one of two conditions:
```
jskeet: Can we use "expression" instead of "variable" uniformly here?

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+> s2 = null;            // OK; null-state of s2 is "maybe null"
+> char c1 = s2[1];      // Warning; dereference of a possible null; throws NullReferenceException
+> c1 = s2![1];          // The warning is suppressed
+> ```
+>
+> *end example*
+
+#### §Nullable-enable-Context Nullable enable
+
+When the nullable context is enable:
```
jskeet: Should this be "enabled"? (It seems odd to have "disabled" but "enable".)

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+- For any reference type `T`, the annotation `?` in `T?` makes `T?` a nullable type, whereas the unannotated `T` is non-nullable.
+- The compiler uses static flow analysis to determine the null state of any reference variable. When nullable warnings are enabled, a reference variable’s null state (§Nullabilities-And-Null-States) is either *not null*, *maybe null*, or *maybe default* and
```
jskeet: Again, let's see if we can just use expression here.

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+  - The state of a non-nullable reference variable is *not null* unless the compiler can determine that the variable was assigned to an expression that is *maybe null*.
+  - The state of a nullable reference variable is *maybe null* unless the compiler can determine one of two conditions:
+    1. The variable has been definitely assigned a non-`null` value.
+    1. The variable or expression has been checked against `null` before dereferencing it.
+
+> *Example*:
+>
+> <!-- Example: {template:"code-in-main-without-using", name:"NullableAnnotationContext4", ignoredWarnings:["CS0219"], expectedWarnings:["CS8600","CS8600","CS8602"], expectedException:"NullReferenceException"} -->
+> ```csharp
+> #nullable warnings
```
jskeet: Should this be "enabled"?

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+>
+> *end example*
+
+### §Nullabilities-And-Null-States Nullabilities and null states
+
+Every expression has one of three ***null state***s:
+
+- *maybe null*: The value of the expression may evaluate to null.
+- *maybe default*: The value of the expression may evaluate to the default value for that type.
+- *not null*: The value of the expression definitely won't evaluate to null.
```
jskeet: I think we should be more circumspect than "definitely" here, given that there are ways in which this could be incorrect.

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+A warning might result if a variable of a non-nullable type has a potential `null` value assigned to it. Nullable variables can have `null` values freely assigned to them.
+
+> *Note*: Assigning a maybe null expression to a variable `v` of a non-nullable reference type sets the null state of `v` to maybe null. *end note*
+
+Nullability is determined, as follows:
+
+- A non-nullable value type `S` is always non-nullable
+- A nullable value type `S?` is always nullable
+- A reference type `C` in a disabled nullable annotation context is nullable, however its default null state is *not-null*.
```
jskeet: This is the change for null-oblivious, right?

standard/types.md
jskeet: Still says oblivious in the line below...

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+- A reference type `C` in an enabled nullable annotation context is non-nullable
+- A nullable reference type `C?` is nullable; however, a warning shall be generated in a disabled nullable annotation context
+
+With a type parameter, any constraints are taken into account:
+
+- A type parameter `T` where all constraints (if any) are either nullable types or the `class?` constraint, is `nullable`
+- A type parameter `T` where at least one constraint is non-nullable, or one of the constraints `struct`,  `class`, or `notnull`, is
+  - oblivious in a disabled nullable annotation context
+  - non-nullable in an enabled nullable annotation context
+- A nullable type parameter `T?` is nullable, but a warning shall be generated if nullable annotations are disabled and `T` isn't a value type
```
jskeet: I suspect we'll need more details on what "T isn't a value type" means.

standard/types.md
```diff
@@ -705,3 +714,184 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+A type is deemed to occur in a given annotation context when the final token of that type is within that context.
```
jskeet: Should this be something like a "type reference"? We need some way of distinguishing between "where the name of the type appears" and "where the type is declared" I suspect.

jskeet (commented):
More comments, but I think we're making progress :)

standard/types.md
BillWagner: Yes.

standard/types.md
BillWagner: Added that the "underlying type" is `R`

standard/types.md
BillWagner: I put them in bold italics in the section. The sections serve as the definition of those contexts.

standard/types.md
BillWagner: Mostly true. Some uses are specific to "variables" (e.g. tracking the state from line to line)

standard/types.md
BillWagner: There are places where "variable" is needed.  Consider:

```csharp
if (M() is not null)
    Console.WriteLine(M().Length); // Warning! Dereference a possible null value.

var s = M();
if (s is not null)
    Console.WriteLine(s.Length); // No warning.
string? M() => someVar;
```

See [SharpLab](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK4wD4AEBMAGAWAChcBGY4rUgTgAoAiQHg3AQffoEoBuCoqAMwAJaAWVpt+UAM78YAewTTUAG0Vti/dfyp0RbAHQAZAKYwA5ggAWnfgHpr/AOoRkMWCYCE/ACKHkh3j+MAY0N+CH4ABxkJCShgRRCYJUV+ADcIRVRDXW405H4pAF5+HS4eAVopSWk5BWVVIg1NGgqDYzNLDhs7ADkZfgB3JxdTbJJSPAB+YtF+AoA+fgATPwglBA4gA===)

standard/types.md
BillWagner: Yes. I'd like to defer this to #1091

standard/types.md
BillWagner: I'm deleting this and the following paragraph from this PR. They are related to type parameters, and will be added in the PR that addresses #1091

standard/types.md
BillWagner: I tried a slightly different tack in the latest commit. Not sure I like it better.

BillWagner (commented):
Comments on updates per feedback.

standard/types.md
```diff
@@ -727,35 +727,37 @@ There are two forms of nullability for reference types:
...
 
 ### §Non-nullable-reference-types Non-nullable reference types
 
-A ***non-nullable reference type*** is a reference type declared in an enabled nullable annotation context (§Nullable-Annotation-Context). A variable of a non-nullable reference type of the form `T` should not contain the value `null`. Warnings are generated when a variables of the type `T` is assigned to `null`. The default null-state of a non-nullable variable is *not-null*.
+A ***non-nullable reference type*** is a reference type of the form `T`, where `T` is the name of the type. A variable of a non-nullable reference type should not contain the value `null`. Warnings are generated when a variables of the type `T` is assigned to `null`. The default null-state of a non-nullable variable is *not-null*.
+
+> *Note:* A non-nullable reference type is defined by the absence of the `?` annotation on the type declaration. When the nullable annotation context is disabled, `T` and `T?` both declare the same types. *end note*
```
jskeet: "Refer to" rather than "declare" perhaps?

standard/types.md
```diff
@@ -727,35 +727,37 @@ There are two forms of nullability for reference types:
...
+A reference type of the form `T?` (such as `string?`) is a ***nullable reference type***. The annotation `?` indicates the intent that variables of this type are nullable. When the nullable annotation context (§Nullable-Annotation-Context) is enabled, the compiler shall recognize these intents. When the nullable annotation context is disabled, the compiler shall ignore these intents. When the nullable annotation context is disabled, a reference type of the form `T?` is the same as a reference of the form `T`.
 
-> *Note:* When the nullable annotation context is disabled, `T` and `T?` both declare null-oblivious reference types. *end note*
+> *Note:* When the nullable annotation context is disabled, `T` and `T?` both declare the same types. *end note*
```
jskeet: Again, refer to - sorry for missing this before.

standard/types.md
```diff
@@ -804,14 +806,14 @@ When the nullable context is annotations:
...
+When the nullable context is ***warnings***:
 
 The compiler uses static flow analysis to determine the null state of any reference variable. When nullable warnings are enabled, a reference variable’s null state (§Nullabilities-And-Null-States) is either *not null*, *maybe null*, or *maybe default* and
 
 - A warning shall be generated if a reference variable that has been determined to be *maybe null*, is dereferenced.
-- The state of a reference variable is *maybe null* unless the compiler can determine one of two conditions:
+- The state of a reference expression is *maybe null* unless the compiler can determine one of two conditions:
   1. The variable has been definitely assigned a non-`null` value.
-  1. The variable or expression has been checked against `null` before dereferencing it.
+  1. The variable has been checked against `null` before dereferencing it.
```
jskeet: Expression for this and the line above?

jskeet (commented):
Getting there :) Sorry to keep picking nits, but I'm definitely keen on getting this one over the line rather than the alternative that has null-oblivious.

standard/types.md
BillWagner: I think this has to be "variable", not expression. It's for the same reason as https://github.com/dotnet/csharpstandard/pull/1124/files#r1626266110

However, changing to a property removes the warnings:

See [Sharplab](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK4wD4AEBMAGAWAChcBGY4gNwmQAIBjWgXlpgFMB3WgYQAoBKANzEoAM1q96AOgCytKAGdWAewStUAGw39itPbSykAnJNlSAMmxgBzBAAshtAPRPaAOWW0ONGLGtSKImo6JRZpGWEiMQklRRU1GE1tXX1DEwULK1sHQWdXDy8fPwCiYiwAZgMcHmIAbxS9CoNSPAB+Wjla2ms2BFyFXtyAX2IhoA==)

jskeet: But a property isn't a variable. Let's talk about this more - my guess is that we're in agreement with what actually happens, but need to figure out what it should look like in the specs.

BillWagner: Right. I added a note here: https://github.com/dotnet/csharpstandard/issues/1093#issuecomment-2158506889 because we may want to handle this in one of the two later PRs on the null-state analysis.

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+## §Types-And-Nullability Reference Types and nullability
+
+### §Nullable-Types-General General
+
+Any expression of a reference type can either refer to an object, or be the value `null`, which indicates "no reference". The terms "nullable reference type" and "non-nullable reference type" refer to annotations that declare whether an expression is intended to permit null values or not. A compiler can provide diagnostics when a expression is not used according to that intent.
+
+There are two forms of nullability for reference types:
+
+- *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
+- *non-nullable*" A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
```
Nigel-Ecma, line 728: Needs a forward reference to *null state*.

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same underlying type, `R`; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type and its corresponding non-nullable reference type (§10.2.2).
```
Nigel-Ecma: See comments (@jskeet, @Nigel-Ecma x 2) for this para in #1105, they apply to this version as well.

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+> *Note:* By definition, a reference type is nullable; that is, a variable of that type can either contain a reference to an object or be the value `null`, which indicates “no reference.” In a disabled nullable context, reference types aren't annotated. Therefore, they are non-nullable reference types. However, the disabled context means the compiler does not track null state and does not issue nullable related warnings. The disabled nullable context matches the previous standard behavior for reference types. *end note*
```
Nigel-Ecma: a) Forward reference to *nullable context* should be added

b) This reads wrong to me. The previous standard behaviour was that all reference types were *nullable*, i.e. `T` was nullable and there was no `T?`. In the new system `T` becomes non-nullable and `T?` is nullable.

Let’s try some wordsmithing – as a *starting point only*:

> *Note:* By definition, a reference type is nullable; that is, a variable of that type can either contain a reference to an object or be the value `null`, which indicates “no reference.” In a disabled nullable context, no distinction is made between `T` and `T?`, variables declared either way are treated as nullable, no null state tracking is required, and therefore no consequential warnings need be issued. The disabled nullable context matches the previous standard behavior for reference types. *end note*

This uses “is required” and “need” but it’s not clear to me they are correct. Under the current system I do not believe that an implementation which chooses to track the possible null value of reference variables, and issue warnings that a value might be/is null, would be non-compliant.

The new system assists, but does not require, an implementation to track null state and issue implementation-specific warnings – so what does it mean to “disable” this non-compulsory system? Does an implementation which already did its own tracking and warning, or has developed better tracking (say by tracking known null values) now have to disable it if the new optional, but specified, null state is disabled?

So I went with ”no [...] is required” for the tracking, and “no [...] warnings need be issued which does allow tracking and warnings, but would that be non-compliant?

> *“I am a compliant implementation and you have disabled nullability, so with apologies I cannot tell you why this bit of code will fail. If you'd like to know enable nullability checking and then I can tell you.”* some future anthropomorphised implementation ;-) [yes, it's been a long day!]

Maybe “disabled” in this context should mean ”don’t flood me with inaccurate warnings but if you are *certain* you may tell me”? But I haven't spent any time on how we might specify that!

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+### §Non-nullable-reference-types Non-nullable reference types
+
+A ***non-nullable reference type*** is a reference type of the form `T`, where `T` is the name of the type. A variable of a non-nullable reference type should not contain the value `null`. Warnings are generated when a variables of the type `T` is assigned to `null`. The default null-state of a non-nullable variable is *not-null*.
+
+> *Note:* A non-nullable reference type is defined by the absence of the `?` annotation on the type declaration. When the nullable annotation context is disabled, `T` and `T?` both refer to the same types. *end note*
```
Nigel-Ecma: Above “annotation” was used, here we have “both refer to the same types” *if* the context is disabled but under the “annotation” model both always refer to the same type, just to instances with different annotations.

Nigel-Ecma (commented):
I think I like this approach better than #1105, but do see some issues.

standard/types.md
jskeet: My comment at the (current) end of #1123 seems to contradict the "does not track null state" part.

standard/types.md
jskeet: Yup, this it the only time we talk about an "underlying" type for reference types. Will add a suggestion.

```suggestion
Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same unannotated type, `R`; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type and its corresponding non-nullable reference type (§10.2.2).
```

standard/types.md
jskeet: Things we're concerned about in 732:

- Null state tracking is always enabled, regardless of context
- Now we have two flags in one context, we shouldn't just talk about the context being disabled
- Nigel's point around "everything was previously nullable" is a natural reaction, and we need to work hard at figuring out how to square that circle.

Might be worth pairing on this.

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+### §Nullable-reference-types Nullable reference types
+
+A reference type of the form `T?` (such as `string?`) is a ***nullable reference type***. The annotation `?` indicates the intent that variables of this type are nullable. When the nullable annotation context (§Nullable-Annotation-Context) is enabled, the compiler shall recognize these intents. When the nullable annotation context is disabled, the compiler shall ignore these intents. When the nullable annotation context is disabled, a reference type of the form `T?` is the same as a reference of the form `T`.
```
jskeet: I'm not sure what "shall ignore these intents" means. I suspect I'll become more aware over time, but maybe an example would help.

jskeet: (In fact, using `string?` in a warnings-only or fully-disabled generates a warning - CS8632 - which isn't really the same as *ignoring* the intent.)

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+> *Note:* When the nullable annotation context is disabled, `T` and `T?` both refer to the same types. *end note*
```
jskeet: Let's crisp up what we mean here, given the two flags we now have.

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+### §Nullable-Contexts Nullable context
+
+#### §Nullable-Contexts-General General
+
+Every line of source code has a ***nullable context***. The nullable context controls whether nullable annotations (§Nullable-Annotation-Context) have effect and whether nullable warnings (§Nullable-Warning-Context) are issued by the compiler. The nullable context can be one of *disabled*, *enabled*, *annotations*, or *warnings*. The *enabled* setting combines both *annotations* and *warnings*.
```
jskeet: I think we should not refer to the context as being "enabled" or "disabled", but always and only the specific flags.

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+The nullable context may be specified within source code via nullable directives (§Nullable-Directives) and/or via some implementation-specific mechanism external to the source code. If both approaches are used, nullable directives supersede the settings made via an external mechanism.
+
+The default state of the nullable context is implementation defined.
+
+Throughout this specification, all C# code that does not contain nullable directives, or about which no statement is made regarding the current nullable context state, shall be assumed to have been compiled with nullable context enabled.
+
+#### §Nullable-Disable-Context Nullable disable
+
+When the nullable context is ***disabled***:
```
jskeet, line 760: See above - I think this should be "when both annotations and warnings are disabled the nullable context" or similar.

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+> string? s1 = null;    // Informational message; ? is ignored
+> string s2 = null;     // OK; null initialization of a reference
+> s2 = null;            // OK; null assignment to a reference
+> char c1 = s2[1];      // OK; no warning on dereference of a possible null; throws NullReferenceException
+> c1 = s2![1];          // OK; ! is ignored
+> ```
+>
+> *end example*
+
+#### §Nullable-Annotation-Context Nullable annotations
```
jskeet, line 782: Anchoring this here somewhat randomly... the annotation aspect affects "question marks in this bit of code" not "whether or not the compiler understands that a type may be nullable". Example:

```csharp
#nullable enable
string? M2() => null;

#nullable disable annotations
// Warning, because this code still "knows" that M2 returns string?, even if we couldn't declare a variable that way explicitly.
Console.WriteLine(M2().ToString());
```

standard/types.md
jskeet: In-meeting discussion: we can probably use variables for now, but revisit later. Some options:

- Define what a "dotted sequence of variables then field/property names" is and use that
- Talk about "expressions in general" but without expecting compilers to track *all* expressions (but go into no further detail)
- Talk about variables specifically, but give compilers enough options to add extra diagnostics (without being specific about it being related to null state tracking)

Happy to talk about that later when we know more about the shape of things.

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+#### §Nullable-enable-Context Nullable enable
+
+When the nullable context is ***enabled***:
+
+- For any reference type `T`, the annotation `?` in `T?` makes `T?` a nullable type, whereas the unannotated `T` is non-nullable.
+- The compiler uses static flow analysis to determine the null state of any reference variable. When nullable warnings are enabled, a reference variable’s null state (§Nullabilities-And-Null-States) is either *not null*, *maybe null*, or *maybe default* and
+  - A warning shall be generated when a possible null value is assigned to a variable of a non-nullable type.
+  - A warning shall be generated if a reference expression that has been determined to be *maybe null*, is dereferenced.
+  - The state of a non-nullable reference variable is *not null* unless the compiler can determine that the variable was assigned to an expression that is *maybe null*.
```
```suggestion
  - The state of a non-nullable reference variable is *not null* unless the compiler can determine that the variable was assigned an expression that is *maybe null*.
```

jskeet: (Or "assigned with the result of an expression" or even "assigned with the result of evaluating an expression" - we should look at what terminology we use elsewhere for assignments.)

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+  - The state of a nullable reference expression is *maybe null* unless the compiler can determine one of two conditions:
+    1. The variable has been definitely assigned a non-`null` value.
+    1. The variable has been checked against `null` before dereferencing it.
```
```suggestion
    1. The variable has been checked against `null`, found not to be null, and will not have be assigned a new value since that check.
```

jskeet: (We don't need to care about dereferencing, but we do need to care about reassignment and the result of the check.)

standard/types.md
```diff
@@ -705,3 +714,182 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
```
jskeet: Extra note from Mads: checking against null affects things even outside the branch:

```#nullable enable
string x = "";
// With this line, there's a warning on x.Length.
// Without this line, there's no warning
if (x is null) { } // Deliberately empty block

Console.WriteLine(x.Length);
```

jskeet: This *could* be something that we don't make normative - but might be a helpful example of "additional implementation-defined null state tracking".

standard/types.md
BillWagner: Listening to the meeting, I am coming to this conclusion for expressions vs. variables:

- The result of evaluating an expression has a *null state*. In other words, there's a known null state for any expression and known operands.
- Variables have a *null state* associated with them on any line of source code. 

(Need to wordsmith, but the idea is that a variable has a null state, and any number of operations like assignment, or performing a null check impacts the null state of the variable. That same concept won't apply to an expression. For example:

```csharp
// declaration:

object? M() => null;

var n = M();
if (n is not null)
{
    // null state of n is not null
} 
if (M()) is not null)
{
    var v = M();
    // null state of v is Maybe null.
}
```

}

Nigel-Ecma: Extending the last example:

```csharp
if (n is null)
{
   // null state of n is Maybe null
}
```

The “maybe null” is at least curious… the issue is more than the null state of variables vs. expressions; it is the available null states themselves, or at least how they are described/explained.

We’re developing this collection of PRs to answer questions as well as coming up with possible wording (and consequentially we shouldn’t merge any of them until we’re happy to merge all of them).

A fundamental question which I don’t think we know the answer to yet is whether the approach of not specifying the algorithm that determines null states, while limiting it to determine only 2 out of the possible 3 states (is null, is not null, unknown), is that the right balance between specifying too little and too much?

Unless we do have our answer to that question?

standard/types.md
```diff
@@ -727,7 +727,7 @@ There are two forms of nullability for reference types:
 - *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
 - *non-nullable*" A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
 
-Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same underlying type, `R`; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type and its corresponding non-nullable reference type (§10.2.2).
+Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same unannotated type, `R`; the difference in their notations indicates, at compile time, only the intent of their usage, and allows for static flow analysis. An identity conversion exists among a nullable reference type and its corresponding non-nullable reference type (§10.2.2).
```
Nigel-Ecma: The annotation doesn’t *allow* analysis, it is possible without it. Maybe:

```suggestion
Unlike with nullable value types, were value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same unannotated type, `R`; the difference in their notations indicates, at compile time, only the intent of their usage to inform the null state analysis. An identity conversion exists among a nullable reference type and its corresponding non-nullable reference type (§10.2.2).
```

Nigel-Ecma: (Different issue on the same line)

In the annotations model are there any identity conversions? An identity conversion exists from `T` to `T` already, why would annotating the type change that? The final sentence also contradicts the starting “Unlike”.

So a second attempt at wordsmithing:

```suggestion
Unlike with nullable value types, where value types `V` and `V?` denote different types, given a reference type `R`, the notations `R` and `R?` denote the same type, `R`; the difference in their notations indicates, at compile time, only the intent of their usage to inform the null state analysis. As the notations denote the same reference type there is an identity conversion between them (§10.2.2).
```

Not that happy with the last sentence but it does avoid referring to them as distinct types.

standard/types.md
Nigel-Ecma: I think it *is* something we don’t make normative.

As this collection of PRs develops we need to consider how many requirements/constraints are being place on a null analysis that itself is not being specified. My current feeling is we're doing erring on the side of specifying too much, but things may change…

Nigel-Ecma (dismissed):
A few language changes, some comments to consider.

As with #1123 I am recommending we don’t merge this, we get it to a state we’re currently happy with, move on to others in the group, circle back around as needed, and then merge/reject the group as a whole.

gafter:
We may need three types, not two.  For example:

```csharp
void f(
#nullable disable
string x, // could x (oblivious) be seen as nullable type but non-nullable initial state?
#nullable enable
string y,
string? z)
{
 if (...) {
   // x state is not null? or oblivious?
   g1(ref x); // okay
   g1(ref y); // not okay
   g1(ref z); // okay
   // x state is possibly null
 } else {
   // x state is not null? or oblivious?
   g2(ref x); // okay
   g2(ref y); // okay
   g2(ref z); // not okay
   // x state is possibly null
 }
}
void g1(ref string? x) {}
void g2(ref string x) {}
```

BillWagner:
Very short version:

```csharp
f("x", "y", "z");
void f(
#nullable disable
    string x,
#nullable enable
    string y,
    string? z)
{
    y = x; // no warning
    y = z; // warning. Assignment of null to non-nullable variable.
}
```

jskeet:
Example of why I believe it makes sense to view a variable of type T declared in a nullable-disabled context as a nullable type:

```csharp
# nullable disable
string x;
# nullable enable
x = "some value";
x = null;
```

This does not issue a warning, even for the last line. The state of `x` just before the last line is surely "definitely not null" as we've assigned a string literal to it... so to be able to assign `null` to it means its type can't be a not-null type.

BillWagner:
```csharp
#nullable enable
string s = M();

#nullable disable
// What is the type of the return value?
string M() => null;
```

RexJaeschke:
Note that from 2024-8-20 through 2024-09-03, there was a lengthy thread on email re this topic, under the subject "ECMA TC49-TG2 agenda for September 4th."

standard/types.md
```diff
@@ -27,10 +27,15 @@ For convenience, throughout this specification, some library type names are writ
...
-A reference type is a class type, an interface type, an array type, a delegate type, or the `dynamic` type.
+A reference type is a class type, an interface type, an array type, a delegate type, or the `dynamic` type. For each non-nullable reference type, there is a corresponding nullable reference type noted by appending the `?` to the type name.
 
 ```ANTLR
 reference_type
+    : non_nullable_reference_type
+    | nullable_reference_type
+    ;
+
+non_nullable_reference_type
```
Nigel-Ecma, line 38: We’re stretching for rules names, and I’m not sure of a good solution. This one isn't tortuous in the way *non_nullable_non_value_type_parameter* (#1178), rather it is somewhat misleading.

If NRT checking is disabled all these types are nullable. If it is enabled then these types are still nullable but may raise warnings if an implementation determines they are assigned null.

But what alternatives can we come up with?

**Just musing**: *basic_reference_type*? Maybe *reference_type* ⇒ *general_reference_type* allowing *non_nullable_reference_type* ⇒ *reference_type*? The latter would allow text to state that something of *reference_type* may be assigned null, which in turn may elicit a warning if NRT is enabled. That has to be better than saying something of *non_nullable_reference_type* may be assigned null surely?

Of course in some future version of the Standard we may well be describing the disabled NRT state as a little used feature for backward compatibility and then what is being written now won’t be so jarring…

standard/expressions.md
```diff
@@ -1821,6 +1822,49 @@ A  *null_conditional_member_access* expression `E` is of the form `P?.A`. The me
...
+null_forgiving_expression
+    : primary_no_array_creation_expression suppression
+    ;
+
+suppression
+    : '!'
+    ;
+```
+
+The *primary_expression* in *null_forgiving_expression* shall not be a *null_forgiving_expression*.
```
```suggestion
The *primary_no_array_creation_expression* in *null_forgiving_expression* shall not be a *null_forgiving_expression*.
```

jskeet: Or perhaps remove this sentence entirely? Basically the grammar already handles this. It does mean that `primary_no_array_creation_expression` is now really `primary_neither_array_creation_nor_null_forgiving_expression`...

standard/expressions.md
```diff
@@ -1821,6 +1822,49 @@ A  *null_conditional_member_access* expression `E` is of the form `P?.A`. The me
...
+
+This operator has no runtime effect; it evaluates to the result of its operand, and that result retains that operand’s classification.
+
+The null-forgiving operator is used to declare that an expression having a reference type isn't null.
```
jskeet: Can it be used on expressions whose type is a generic type parameter not known to be a reference type? I think so, and that would mean that the "having a reference type" isn't always true.

standard/statements.md
```diff
@@ -366,7 +366,7 @@ implicitly_typed_local_variable_declarator
     ;
 ```
 
-An *implicity_typed_local_variable_declaration* introduces a single local variable, *identifier*. The *expression* or *variable_reference* shall have a compile-time type, `T`. The first alternative declares a variable with type `T` and an initial value of *expression*. The second alternative declares a ref variable with type `ref T` and an initial value of `ref` *variable_reference*.
+An *implicity_typed_local_variable_declaration* introduces a single local variable, *identifier*. The *expression* or *variable_reference* shall have a compile-time type, `T`. The first alternative declares a variable with an initial value of *expression*; its type `T?` when `T` is a reference type and `T` when `T` is a value type. The second alternative declares a ref variable with an initial value of `ref` *variable_reference*; its type is `ref T?` when `T` is a reference type and `ref T` when `T` is a value type.
```
jskeet: What about when the expression already has a type which is a nullable reference type? For example:

```csharp
string? x = null;
var y = x;
```

Doesn't this paragraph suggest that `y` has a type of `string??` ?

The `ref` version of this feels dangerous to me - but seems to be "correct". This code compiles with no warnings, and throws at runtime:

```csharpstring x = "";
ref var y = ref x;
y = null;

Console.WriteLine(x.Length);
```

standard/types.md
jskeet: > If NRT checking is disabled all these types are nullable.

That's one approach - another approach would be to say that they're all non-nullable, but nothing checks them. That might be a simpler approach in terms of the spec. I think it's worth discussing this.

standard/types.md
```diff
@@ -705,3 +714,155 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+There are two forms of nullability for reference types:
+
+- *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
+- *non-nullable*" A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
+
+> *Note:* The types `R` and `R?` are represented by the same underlying type, `R`. A variable of that underlying type can either contain a reference to an object or be the value `null`, which indicates “no reference.” *end note*
+
+The syntactic distinction between a *nullable reference type* and its corresponding *non-nullable reference type* enables a compiler to generate diagnostics. A compiler must allow the `?` annotation as defined in §8.2.1. The diagnostics must be limited to warnings. Neither the presence or absence of nullable annotations, nor the state of the nullable context can change the compile time or runtime behavior of a program except for changes in any diagnostic messages generated at compile time.
```
jskeet, line 732: Does "warnings as errors" affect this guidance? I suspect not - just think it's worth discussing.

standard/types.md
```diff
@@ -705,3 +714,155 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+### §Non-nullable-reference-types Non-nullable reference types
+
+A ***non-nullable reference type*** is a reference type of the form `T`, where `T` is the name of the type. The default null-state of a non-nullable variable is *not-null*. Warnings can be generated when a variables of the type `T` is assigned to an expression that may be `null`.
```
```suggestion
A ***non-nullable reference type*** is a reference type of the form `T`, where `T` is the name of the type. The default null-state of a non-nullable variable is *not-null*. Warnings can be generated when an expression that may be `null` is assigned to a variable of the type `T`.
```

"X assigned to Y" sounds like Y is the *target* of the assignment, i.e. the variable.

standard/types.md
```diff
@@ -705,3 +714,155 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+### §Nullable-reference-types Nullable reference types
+
+A reference type of the form `T?` (such as `string?`) is a ***nullable reference type***. The default null-state of a nullable variable is *maybe null*. The annotation `?` indicates the intent that variables of this type are nullable. The compiler can recognize these intents to issue warnings. When the nullable annotation context is disabled, using this annotation can generate a warning.
+
+### §Nullable-Contexts Nullable context
+
+#### §Nullable-Contexts-General General
+
+Every line of source code has a ***nullable context***. The annotations and warnings flags for the nullable context control whether nullable annotations (§Nullable-Annotation-Context) are allowed and whether nullable warnings (§Nullable-Warning-Context) are issued by the compiler, respectively. Each flag can be *enabled* or *disabled*. The compiler can use static flow analysis to determine the null state of any reference variable. A reference variable’s null state (§Nullabilities-And-Null-States) is either *not null*, *maybe null*, or *maybe default*.
```
jskeet: I think "allowed" is misleading here - it sounds like `string? x = null;` should fail in a nullable-disabled context, and we've discussed before that that's not the case. Let's discuss a better word in the meeting.

standard/types.md
```diff
@@ -705,3 +714,155 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+When both the warning and annotations flags are disabled, the nullable context is *disabled*.
+
+When the nullable context is ***disabled***:
+
+- No warning shall be generated when a variable of an unannotated reference type is initialized with, or assigned a value of, `null`.
+- No warning shall be generated when a variable of a reference type that possibly has the null value.
+- For any reference type `T`, the annotation `?` in `T?` generates a message and the type `T?` is the same as `T`.
+  > *Note*: This message is characterized as “informational” rather than “warning,” so as not to confuse it with the state of the nullable warning setting, which is unrelated. *end note*
+- The null-forgiving operator `!` (§Null-Forgiving-Expressions) generates a message, and has no effect.
```
jskeet: It doesn't look like it *does* generate a message...

standard/types.md
```diff
@@ -705,3 +714,155 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+
+- For any reference type `T`, the annotation `?` in `T?` makes `T?` a nullable type, whereas the unannotated `T` is non-nullable.
+- The compiler can use static flow analysis to determine the null state of any reference variable. When nullable warnings are enabled, a reference variable’s null state (§Nullabilities-And-Null-States) is either *not null*, *maybe null*, or *maybe default* and
+- The null-forgiving operator `!` (§Null-Forgiving-Expressions) sets the null state of its operand to *not null*.
+
+### §Nullabilities-And-Null-States Nullabilities and null states
+
+A compiler is not required to perform any static analysis nor is it required to generate any diagnostic messages related to nullability.
+
+**The remainder of this subclause is conditionally normative.**
```
jskeet, line 847: Should we say what it's conditional on?

standard/types.md
```diff
@@ -705,3 +714,155 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+The ***default null state*** of an expression is determined by its type, and the state of the annotations flag when it is declared:
+
+- The default null state of a nullable reference type is:
+  - Maybe null when its declaration is in text where the annotations flag is enabled.
+  - Not null when its declaration is in text where the annotations flag is disabled.
+- The default null state of a non-nullable reference type is not null.
+
+A diagnostic can be produced when a variable (§9.2.1) of a non nullable reference type is initialized or assigned to an expression that is maybe null when that variable is declared in text where the annotation flag is enabled.
+
+The compiler can update the null state of a variable when the variable is read or written to.
```
jskeet: Only then? How about this:

```csharp
string? x = null;
ref string? y = ref x;
x = "hello";
Console.WriteLine(y.Length);
```

Can the compiler (theoretically) update the null state of `y` on the assignment to `x`, if it's smart enough to work out that they refer to the same storage location?

standard/types.md
jskeet: We discussed this; Nigel will suggest a new name if he thinks of a better one.

standard/expressions.md
BillWagner, line 1839: This can be removed with the latest grammar change:

```suggestion
```

BillWagner: This sentence is unnecessary:

```suggestion
```

standard/expressions.md
BillWagner: Suggestion:

```suggestion

The null-forgiving operator is used to declare that an expression not known to be a value type is not null.
```

standard/types.md
BillWagner: This is below the bar.

jskeet: Discussed - warnings as errors is out of scope.

standard/types.md
BillWagner: Alternate text:

```suggestion
A ***non-nullable reference type*** is a reference type of the form `T`, where `T` is the name of the type. The default null-state of a non-nullable variable is *not-null*. Warnings can be generated when an expression that is *maybe-null* is used where a *not-null* value is required.
```

standard/types.md
```diff
@@ -705,3 +714,155 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+- *not null*: The value of the expression isn't null.
+
+The ***default null state*** of an expression is determined by its type, and the state of the annotations flag when it is declared:
+
+- The default null state of a nullable reference type is:
+  - Maybe null when its declaration is in text where the annotations flag is enabled.
+  - Not null when its declaration is in text where the annotations flag is disabled.
+- The default null state of a non-nullable reference type is not null.
+
+A diagnostic can be produced when a variable (§9.2.1) of a non nullable reference type is initialized or assigned to an expression that is maybe null when that variable is declared in text where the annotation flag is enabled.
```
```suggestion
A diagnostic can be produced when a variable (§9.2.1) of a non-nullable reference type is initialized or assigned to an expression that is maybe null when that variable is declared in text where the annotation flag is enabled.
```

standard/types.md
```suggestion
- The null-forgiving operator `!` (§Null-Forgiving-Expressions) has no effect.
```

standard/types.md
BillWagner: This is below the bar for this PR, but should be unresolved if we go back to it.

standard/types.md
```suggestion
The compiler can update the null state of a variable as part of its analysis.
```

BillWagner (commented):
comments during our meeting

jskeet (approved):
Thanks! Big step forward :)

standard/types.md
BillWagner: I think we can resolve this. I made an edit that simplifies the description here. The normative behavior is described in the referenced sections.

BillWagner:
I've addressed all open comments, and resolved all conversations. This is ready for a final (asynchronous) pass and approval.

@jskeet @Nigel-Ecma @MadsTorgersen @ericlippert @RexJaeschke @gafter

standard/statements.md
```diff
@@ -366,7 +366,7 @@ implicitly_typed_local_variable_declarator
     ;
 ```
 
-An *implicity_typed_local_variable_declaration* introduces a single local variable, *identifier*. The *expression* or *variable_reference* shall have a compile-time type, `T`. The first alternative declares a variable with an initial value of *expression*; its type `T?` when `T` is a reference type and `T` when `T` is a value type. The second alternative declares a ref variable with an initial value of `ref` *variable_reference*; its type is `ref T?` when `T` is a reference type and `ref T` when `T` is a value type.
+An *implicity_typed_local_variable_declaration* introduces a single local variable, *identifier*. The *expression* or *variable_reference* shall have a compile-time type, `T`. The first alternative declares a variable with an initial value of *expression*; its type `T?` when `T` is a non-nullable reference type, otherwise its type is `T`. The second alternative declares a ref variable with an initial value of `ref` *variable_reference*; its type is `ref T?` when `T` is a non-nullable reference type, otherwise its type is `ref T`.
```
Nigel-Ecma: Missing word:

```suggestion
An *implicity_typed_local_variable_declaration* introduces a single local variable, *identifier*. The *expression* or *variable_reference* shall have a compile-time type, `T`. The first alternative declares a variable with an initial value of *expression*; its type is `T?` when `T` is a non-nullable reference type, otherwise its type is `T`. The second alternative declares a ref variable with an initial value of `ref` *variable_reference*; its type is `ref T?` when `T` is a non-nullable reference type, otherwise its type is `ref T`.
```

standard/types.md
```diff
@@ -733,7 +733,7 @@ The syntactic distinction between a *nullable reference type* and its correspond
 
 ### §Non-nullable-reference-types Non-nullable reference types
 
-A ***non-nullable reference type*** is a reference type of the form `T`, where `T` is the name of the type. The default null-state of a non-nullable variable is *not-null*. Warnings can be generated when a variables of the type `T` is assigned to an expression that may be `null`.
+A ***non-nullable reference type*** is a reference type of the form `T`, where `T` is the name of the type. The default null-state of a non-nullable variable is *not-null*. Warnings can be generated when an expression that is *maybe-null* is used where a *not-null* value is required.
```
Nigel-Ecma: “Warnings may be generated…”? What do you say @RexJaeschke?

Nigel-Ecma (commented):
Just a missing word and a one word change suggestion.

standard/types.md
jskeet: May sounds better to me, yes.

RexJaeschke: Agreed; "may" is formal permission, and is conformance-related.

BillWagner:
```suggestion
A ***non-nullable reference type*** is a reference type of the form `T`, where `T` is the name of the type. The default null-state of a non-nullable variable is *not-null*. Warnings may be generated when an expression that is *maybe-null* is used where a *not-null* value is required.
```

RexJaeschke (approved):

PR #1126, opened 2024-06-05
Use correct grammar rule for refs w.r.t variant type parameter lists in interfaces

RexJaeschke:
During the creation of V5, we added variance to interface types, which resulted in the creation of a new variety of *type_parameter_list* called *variant_type_parameter_list*. However, the prose in some interface-clause sections still uses the pre-variant grammar rule. This PR replaces those with the variant one.

**PLEASE CHECK:** There is one change I did *not* make, and that is in the grammar in 18.4.2 Interface methods:

```ANTLR
interface_method_header
    : identifier '(' formal_parameter_list? ')' ';'
    | identifier type_parameter_list
```

I left the *type_parameter_list* as is, *guessing* it should not have a "variant_* prefix.


KalleOlaviNiemitalo:
The guess is correct; this is not valid:

```csharp
interface I
{
    T M<out T>();
}
```

*variant_type_parameter_list*  is not appropriate here because M is a generic method rather than a generic type.

BillWagner (approved):
This LGTM @RexJaeschke 

Tagging @Nigel-Ecma for review as well.

Nigel-Ecma (approved):
LGTM

PR #1128, opened 2024-06-06
Try creating the checks from running the anchor tags workflow

BillWagner:
Creating a draft PR to test the workflow.


BillWagner:
Closing in favor of #1129

Because this PR adds "write" permissions on "checks", GitHub only honors that change on the main repo, not when the PR is in a fork.

Issue #1125, opened 2024-06-03
Our tools should format messages to match GitHub's UI format


BillWagner:
We should format the error output for the TOC / anchor generator, Word generator, and the test extractor so that they participate in the GitHub PR Review process as markdownlint lint does.

@jskeet I'll take this on in a combination of 3 PRs if you want.

jskeet:
Sounds good to me, thanks :)

BillWagner:
The only tool remaining to hook into this mechanism is the grammar validator. That might be able to do it using curl, but let's wait a bit until the next updates are in place.

PR #1129, opened 2024-06-06
Use Check Runs API to report on status checks (success and failures)

BillWagner:
Replaces #1128 (because `checks: write` permissions is honored when the permission is added on a fork - details below)

This updates the TOC / Anchor update tool to create a Status Check payload and upload that as part of the workflow. A new library classes captures all warnings, errors, and information messages and adds them to an "annotations" collection. When the workflow completes, the annotations and status are sent to GitHub as a status check. That enables the errors and warnings from this tool to be displayed inline in the files tab on the PR.

My tests using #1128 failed because this PR adds the `checks:write` permission to this workflow. As a security precaution, when a PR from a fork adds write permissions to any workflow, those permissions are disabled in the PR. That prevents a malicious user from elevating their permissions.


BillWagner:
Woo hoo:

image

tools/StandardAnchorTags/DiagnosticIDs.cs
```diff
@@ -0,0 +1,6 @@
+﻿namespace StandardAnchorTags;
+internal class DiagnosticIDs
+{
+    public const string TOC001 = nameof(TOC001);
+    public const string TOC002 = nameof(TOC002);
+}
```
BillWagner, line 6: I'm not sure how useful this is, but it seems reasonable to sort errors and warnings over time.

tools/StandardAnchorTags/Program.cs
```diff
@@ -11,10 +14,20 @@ public class Program
...
+
+    /// <summary>
+    /// Update section numbers, anchor tags, and the TOC
+    /// </summary>
+    /// <param name="owner">The GitHub owner org (for example, "dotnet")</param>
+    /// <param name="repo">The GitHub repo name (for example, "csharpstandard")</param>
+    /// <param name="headSha">The commit sha, when run as a GitHub action</param>
+    /// <param name="dryrun">True for a dry run, false to update the text in all files</param>
+    /// <returns>0 on success, non-zero on failure</returns>
+    static async Task<int> Main(string owner, string repo, string? headSha = null, bool dryrun =false)
```
BillWagner, line 27: There's a lot less actual changes than appears in this file. The `logger` is new. And, it's passed to the objects that do the work. Finally, it logs the errors and uploads the collection of annotations.

The extra diffs are because the `Main` method was getting long, so I broke it into 3 distinct functions.

tools/Utilities/StatusCheckLogger.cs
```diff
@@ -0,0 +1,146 @@
+﻿using Octokit;
```
BillWagner, line 1: This file is the primary new functionality. It builds that annotation collection and provides consistent output for error, warning, and informational messages. It also uploads the annotations to GitHub when running in an actions context.

tools/run-section-renumber.sh
```diff
@@ -4,12 +4,7 @@ set -e
 
 declare -r PROJECT=StandardAnchorTags
 
-if [ "$1" == "--dryrun" ]
-then
-    echo "Performing a dry run"
-fi
-
-dotnet run --project $PROJECT -- $1
+dotnet run --project $PROJECT -- --owner dotnet --repo csharpstandard --dryrun true --head-sha $1
```
BillWagner, line 7: Yeah, I hardcoded a few options. And added the head sha for the PR.

BillWagner (commented):
Added notes for reviewers.

BillWagner:
ping @jskeet for review.

Once this one is merged, I'll make similar changes to the other tools.

jskeet:
Will have a look on Monday - busy as a dad taxi today I'm afraid.

tools/Utilities/StatusCheckLogger.cs
```diff
@@ -0,0 +1,146 @@
...
+    /// </remarks>
+    public void LogWarning(Diagnostic d)
+    {
+        WriteMessageToConsole("⚠️", d);
+        annotations.Add(
+            new(FormatPath(d.file),
+            d.StartLine, d.EndLine,
+            CheckAnnotationLevel.Warning, $"{d.Id}::{d.Message}")
+        );
+        Success = false;
```
jskeet: This looks like it's treating warnings as errors... I don't know whether we want that.

jskeet (approved):
One thing to check on whether warnings should count as errors, but other than that looks okay. I can't say I've reviewed *hugely* thoroughly.

BillWagner:
I did update to succeed with warnings.

We'll see if we want to replace that.

BillWagner:
Contributes to #1125

PR #1134, opened 2024-06-24
Update dependencies, plus merge fix

BillWagner:
Two distinct changes here.

First, update all dependencies. Dependabot should have updated these, but that PR is failing the build. It appears that some updates weren't correctly applied.

Second, a quick work-around for the "update on merge" action so that it runs correctly. It's block, so I just copied the shell script and made mods. I'll take a bit longer to create a single script that reports errors and warnings correctly when the actions run not as part of a PR validation. I need to do some research there, and this is blocking us.

### DIFF
--- a/.github/workflows/update-on-merge.yaml
+++ b/.github/workflows/update-on-merge.yaml
@@ -50,7 +50,7 @@ jobs:
       id: renumber-sections
       run: |
         cd tools
-        ./run-section-renumber.sh
+        ./run-section-renumber-update.sh
       shell: bash
 
     - name: Create pull request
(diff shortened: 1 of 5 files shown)


.github/workflows/update-on-merge.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
       id: renumber-sections
       run: |
         cd tools
-        ./run-section-renumber.sh
+        ./run-section-renumber-update.sh
```
BillWagner, line 53: Yes, this is a hack workaround. I'll fix when I update the reporting for the word converter and the test runner.

jskeet (approved):
Took me a little while to work out what the differences were in the script, but I've now figured it out :)

PR #1136, opened 2024-06-24
Add status checks to the word converter

BillWagner:
Add status checks to the Word converter.

Since this also drives some changes on the shell script, so fix the hack in #1134 with more reasonable shell programming.

Contributes to #1125

The bulk of the changes are hooking up the reporter type to the new status check library. The reporter already does most of the work to report errors and warnings.

The final set of changes is to replace the hardcoded strings with constants for diagnostic ids.


jskeet (commented):
This PR currently seems to *only* be the shell/yaml changes. Was there meant to be something in the Word converter itself?

BillWagner:
> This PR currently seems to _only_ be the shell/yaml changes. Was there meant to be something in the Word converter itself?

Not yet. I wanted to check the shell script updates first.

.github/workflows/renumber-sections.yaml
```diff
@@ -19,6 +19,7 @@ jobs:
     env:
       DOTNET_NOLOGO: true
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      HEAD_SHA: ${{ github.event.pull_request.head.sha }}
```
BillWagner, line 22: Here, I moved the SHA from a command line parameter to an environment variable. It will get added to the environment on the container, and then be destroyed when the container is.

It's simpler. Also, while the SHA isn't a secret, it's consistent with a safer way to pass secrets to the action.

tools/MarkdownConverter/Converter/DiagnosticIDs.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace MarkdownConverter.Converter;
+internal class DiagnosticIDs
+{
+    public const string MDC002 = nameof(MDC002);
```
BillWagner, line 5: I replaced all the hard coded string IDs with constants. Also, I changed the prefix from "MD" to "MDC" for "markdown converter". (Markdown lint is already using "MD").

tools/MarkdownConverter/Spec/Reporter.cs
```diff
@@ -10,10 +11,7 @@ namespace MarkdownConverter.Spec;
 /// </summary>
 public class Reporter
 {
-    /// <summary>
-    /// The text writer to write Messages to.
-    /// </summary>
-    private readonly TextWriter writer;
+    private readonly StatusCheckLogger githubLogger;
```
BillWagner, line 14: Replace the log to console with the status check update

tools/MarkdownConverter/Converter/MarkdownSpecConverter.cs
```diff
@@ -32,7 +32,7 @@ public static void ConvertToWord(MarkdownSpec spec, string templateFile, string
                     wordDocument: resultDoc,
                     spec: spec,
                     filename: fileName,
-                    reporter.WithFileName(fileName));
+                    reporter.WithFileName(src.Item1));
```
BillWagner, line 35: The updated report needs the relative path, not just the file name. That's how GitHub knows where to draw the annotations in the PR review window.

tools/MarkdownConverter/Spec/SourceLocation.cs
```diff
@@ -26,53 +27,92 @@ public string Description
     {
         get
         {
-            if (_loc != null)
-            {
-                return _loc;
-            }
+            ComputeLazyFields();
```
BillWagner, line 30: The diff here is much less useful than I'd hope. Conceptually, here are the changes:

1. Add two new properties for the start and end line of the span so that errors are highlighted correctly.
2. Factor out a helper method to initialize the three private backing fields. It's effectively the same logic in all three cases.

BillWagner (commented):
Adding roadmap comments for reviewers.

BillWagner:
@jskeet 

This is now ready.

jskeet:
Will look tomorrow morning UK time.

jskeet (approved):
Yup, looks good :)

PR #1141, opened 2024-07-09
Update quest-config.json

adegeo:
- Add the content curation tag for quest items.
- Set the default parent node to C# work item

The actual github label used to trigger the quest tag is something you can change. I used `🏷️ content-curation` but feel free to edit this PR to change it to what you want.

You'll need to add this label to this repo and then use it for curation items.


BillWagner (approved):
This LGTM.

I'll :shipit:

PR #1142, opened 2024-07-09
Update quest-config.json

adegeo:
After reviewing the purpose of the curation tag/parent, it doesn't really apply in this repo. The default parent still applies.


BillWagner (approved):
This LGTM.

Issue #1146, opened 2024-07-10
Annex C.2 isn't necessarily correct (in terms of only being things in ISO)


jskeet:
ValueTuple is in Annex C.2 but it's not defined in ISO.

There may be others that should be in C.3.

RexJaeschke:
@jskeet I just compared the current C.2 with the CLI library type list, and found the following incorrectly located types, which I'll move to C.3:

```csharp
namespace System.Collections.Generic
{
    public interface IReadOnlyCollection<out T> : IEnumerable<T>
    {
        int Count { get; }
    }

    public interface IReadOnlyList<out T> : IReadOnlyCollection<T>
    {
        T this [int index] { get; }
    }
}
```

PR #1027, opened 2023-12-07
Address async issues

BillWagner:
Each fix is in a single commit. In order:

Fixes #857: produces a result vs. returns a value.

> For an async method, say "produces a result value" vs. "return a value" when describing the rules for the returned "task type".

Fixes #878

> Add a note that the builder pattern supports identity convertible types. I made this a note, rather than normative language. I read the existing text to allow the example in the issue (as does the roslyn implementation).

Fixes #775

> Add a bullet to the list specifying that an `OperationCanceledException` puts the returned task in the cancelled state.
>
> This meant adding `OperationCanceledException` to the types in Annex C.
> 
> I chose this fix because all known task builders conform to this behavior. It seems reasonable to standardize it.

Fixes #879 

> Clarify the required accessibility of any task builder type and its required members. The members of the task builder type must have declared public accessibility. The accessibility of the task builder type must match the accessibility of its corresponding task type.
> 
> Note that we will need to update this text once dotnet/roslyn#71283.

The fifth commit adds a note clarifying the meaning of the accessibility requirements for the task builder type and its corresponding task type.

PR #1150, opened 2024-07-15
Move types from C.2 to C.3

RexJaeschke:
This addresses Issue #1146, and one other like scenario.

PR #664 incorrectly added type `ValueTuple` to C.2. However, that type is not defined in the CLI spec.

PR #1027  incorrectly added type `OperationCanceledException` to C.2. However, that type is not defined in the CLI spec.

Jon, I looked at the "blame" for each change to Annex C during the past year, and found only one other type, `OperationCanceledException`, in the wrong section.

### DIFF
--- a/standard/standard-library.md
+++ b/standard/standard-library.md
@@ -168,13 +168,6 @@ namespace System
             Exception innerException);    
     }
 
-    public class OperationCanceledException : Exception
-    {
-        public OperationCanceledException();
-        public OperationCanceledException(string message);
-        public OperationCanceledException(string message, Exception innerException);
-    }
-
     public struct Nullable<T>
     {
         public bool HasValue { get; }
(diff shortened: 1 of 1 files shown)


standard/standard-library.md
```diff
@@ -464,6 +385,13 @@ A conforming implementation may provide `Task.GetAwaiter()` and `Task<TResult>.G
 ```csharp
 namespace System
 {
+    public class OperationCanceledException : Exception
```
jskeet, line 390: Can we keep these declarations in alphabetical order? I think we just need this after FormattableString.

jskeet (approved):
One nit about order, but I'm otherwise happy.

I do wonder whether we should have more blank lines for the sake of readability, as we do in C.2. I'll file a separate issue for that and send a PR when this is fixed and merged.

standard/standard-library.md
RexJaeschke: @jskeet I moved that type per your suggestion, and I added numerous blank lines throughout C.3 to match the C.2 approach.

RexJaeschke:
Jon's comment above (which was also raised by PR #1151) re making sure there is a blank line between types was handled by the final commit above.

PR #1153, opened 2024-07-30
Removed *end note*

sguitardude:
Not sure if *end note* is intentional or an artifact from a past system.  It seems like we could remove it.


jskeet:
No, it's very intentional, to show where each note ends. That ensures we know what text is normative and what text is informative. I'll close this PR now, as we definitely don't want to remove "end note" or "end example".

PR #1121, opened 2024-05-27
Remove '>' from sequence of tokens disambiguating type_argument_list

logeshkumars0604:
Fixes #1115

PR #1164, opened 2024-08-23
Remove '>' from sequence of tokens disambiguating type_argument_list

Nigel-Ecma:
Replacement for #1121 which I suspect was mistakenly closed. The original was approved back in May but merging seems to have slipped through the cracks – stuff happens.


BillWagner (approved):
Thanks again @Nigel-Ecma 

I'll merge this one now.

Issue #1159, opened 2024-08-15
Parameters #2: Alternate Terms for a Parameter Kind


RexJaeschke:
Recently, I had cause to review the spec in terms of function parameters, and argument passing. [This was regarding adding future support for `ref readonly` parameters.] I found a lot of places where the text could be considerably simplified, made consistent, or have duplication of normative text removed. This issue deals with one aspect, and is *not* intended to result in any semantic changes. Rather, it is intended to create a better base on which to add future feature specs. Hopefully, we can decide offline on each proposal without taking up TG2 meeting time.

## Issue

Variables.md, [9.2.8 Input parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#928-input-parameters), states

>A parameter declared with an `in` modifier is an ***input parameter***.

So, the term “input parameter” is defined (and matches the parent section name). However, while 61 occurrences of text about input parameters uses that term, 33 occurrences use “`in` parameter” instead, for no obvious reason and with no value added.

The same situation exists for “output/`out` parameter” ([9.2.7 Output parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#927-output-parameters)) and “reference/`ref` parameter” ([9.2.6 Reference parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#926-reference-parameters)).

[There is no such duplication for “value parameter.” [9.2.5 Value parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#925-value-parameters) states: “A parameter declared without an `in`, `out`, or `ref` modifier is a ***value parameter***, and “value parameter” is used throughout.]

## Proposal

- Change all “`in` parameter” to “input parameter.”
- Change all “`out` parameter” to “output parameter.” 
- Change all “`ref` parameter” to “reference parameter.”
- Also, change such text not having the backticks, as in “out parameter.”
- Change the term definition “***by-reference*** parameter” to “***by-reference parameter***.”
- Change all enumerated lists “input, output, and/or reference parameter” to the corresponding group term “by-reference parameter.”

## Discussion

The term “by-reference parameters” is defined in [15.6.2.3.1 General](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#156231-general) and only referenced in Clause 15. It’s simply used to organize related subclauses as a group, so let’s use it instead of an enumerated list of its members.

[...]

PR #1172, opened 2024-09-11
Use consistent terminology for parameter kinds

RexJaeschke:
This PR address Issue #1159.

This makes wording re parameters tighter, but because we don't define corresponding terms such as "input argument," "output argument," or "reference argument," we still have many mentions of `in`, `out`, and `ref`.


standard/classes.md
```diff
@@ -2208,19 +2208,19 @@ A method is permitted to assign new values to a value parameter. Such assignment
 
 ##### 15.6.2.3.1 General
 
-A parameter declared with one of the `in`, `ref` or `out` modifiers is a ***by-reference*** parameter. A by-reference parameter is a local reference variable ([§9.7](variables.md#97-reference-variables-and-returns)); the initial referent is obtained from the corresponding argument supplied in the method invocation.
+Any input parameter, output parameter, or reference parameter is also a ***by-reference parameter***. A by-reference parameter is a local reference variable ([§9.7](variables.md#97-reference-variables-and-returns)); the initial referent is obtained from the corresponding argument supplied in the method invocation.
```
Nigel-Ecma: The addition of “also” doesn’t sound correct to me in the same way “cod and haddock are also fish” doesn’t – they *are*, not *also*, fish. But how to reword? I’ll offer:

```suggestion
Input, output, and reference parameters are ***by-reference parameter***s. A by-reference parameter is a local reference variable ([§9.7](variables.md#97-reference-variables-and-returns)); the initial referent is obtained from the corresponding argument supplied in the method invocation.
```

Feel free to wordsmith!

standard/classes.md
```diff
@@ -3050,8 +3050,8 @@ class Customer
 
 When the first parameter of a method includes the `this` modifier, that method is said to be an ***extension method***. Extension methods shall only be declared in non-generic, non-nested static classes. The first parameter of an extension method is restricted, as follows:
 
-- It may have the parameter modifier `in` only if the parameter has a value type
-- It may have the parameter modifier `ref` only if the parameter has a value type or is a generic type constrained to struct
+- It may be an input parameter if the parameter has a value type
```
Nigel-Ecma: The rewording has lost “only” which to me links to the lead-in’s “restriction”. I'll offer:

```suggestion
- It may only be an input parameter if it has a value type
```

As usual feel free to wordsmith.

standard/classes.md
```diff
@@ -3050,8 +3050,8 @@ class Customer
...
+- It may be be a reference parameter if the parameter has a value type or is a generic type constrained to struct
```
Nigel-Ecma: As previous but also remove the “be be” typo:

```suggestion
- It may only be a reference parameter if it has a value type or has a generic type constrained to struct
```

standard/variables.md
```diff
@@ -165,7 +165,7 @@ A ***discard*** is a local variable that has no name. A discard is introduced by
   
 Because a discard has no name, the only reference to the variable it represents is the expression that introduces it.
 
-> *Note*: A discard can however be passed as an out argument, allowing the out parameter to denote its associated storage location. *end note*
+> *Note*: A discard can however be passed as an `out` argument, allowing the corresponding output parameter to denote its associated storage location. *end note*
```
Nigel-Ecma: I’d suggest replacing the keyword:

```suggestion
> *Note*: A discard can however be passed as an output argument, allowing the corresponding output parameter to denote its associated storage location. *end note*
```
As it reads more consistently, and while we may not define *output argument* I think the meaning is obvious from the context!

Nigel-Ecma (changes requested):
Overall this is an improvement in readability.

I’ve made a few minor wording suggestions, the original edits are not egregious so once those suggestions are accepted/rejected/wordsmithed this review is an Approve.

Nigel-Ecma (changes requested):
_Deleted_

Apologies, I broke GitHub, again…

Nigel-Ecma (approved):
This is a good improvement in readability 👍

Issue #1160, opened 2024-08-15
Parameters #3: Duplicate normative parameter-kind definitions 


RexJaeschke:
Recently, I had cause to review the spec in terms of function parameters, and argument passing. [This was regarding adding future support for `ref readonly` parameters.] I found a lot of places where the text could be considerably simplified, made consistent, or have duplication of normative text removed. This issue deals with one aspect, and is *not* intended to result in any semantic changes. Rather, it is intended to create a better base on which to add future feature specs. Hopefully, we can decide offline on each proposal without taking up TG2 meeting time.

## Issue

The variables.md subclause [9.2.5 Value parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#925-value-parameters) and the classes.md subclause [15.6.2.2 Value parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#15622-value-parameters) contain duplicate text. Likewise for [9.2.6 Reference parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#926-reference-parameters) and [15.6.2.3.3 Reference parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#156233-reference-parameters),  [9.2.8 Input parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#928-input-parameters) and [15.6.2.3.2 Input parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#156232-input-parameters), and [9.2.7 Output parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#927-output-parameters) and [15.6.2.3.4 Output parameters](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#156234-output-parameters).

## Proposal

- From 9.2.x remove all text except that related to definite-assignment and variables, in general, and merge the removed text with that in 15.6.2.x. 
- Add to 9.2.x a forward pointer to 15.6.2.x for details about that kind of parameter.

## Discussion

As definite-assignment appears to be covered in 9.2.* for other kinds of variables, it makes sense to leave this text in those subclauses. However, all semantics specific to arguments/parameters should go in 15.6.2.*.

RexJaeschke:
[...]

PR #1174, opened 2024-09-13
Consolidate parameter-kinds text w.r.t Variables and Classes clauses

RexJaeschke:
Instead of starting with only one of the 4 parameter kinds, the solution was found to be much simpler than first thought, so I addressed all 4.

This PR address all the concerns of Issue #1160.

1. Variables subclauses 9.x cover lifetime, (sometimes) initialization, and definite assignment. All other text was moved to the Classes 15.x counterpart places with forward pointers to them, including the definitions of the terms value/input/output/reference parameters.
2.  Classes subclauses 15.x now point back to their 9.x counterpart places w.r.t definite assignment.
3. Various links were updated due to the 9.x-->15.x move.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -2192,7 +2192,7 @@ The following kinds of parameters exist:
 - Input parameters ([§15.6.2.3.2](classes.md#156232-input-parameters)).
 - Output parameters ([§15.6.2.3.4](classes.md#156234-output-parameters)).
 - Reference parameters ([§15.6.2.3.3](classes.md#156233-reference-parameters)).
-- Parameter arrays ([§15.6.2.](classes.md#15624-parameter-arrays)).
+- Parameter arrays ([§15.6.2.4](classes.md#15624-parameter-arrays)).
 
 > *Note*: As described in [§7.6](basic-concepts.md#76-signatures-and-overloading), the `in`, `out`, and `ref` modifiers are part of a method’s signature, but the `params` modifier is not. *end note*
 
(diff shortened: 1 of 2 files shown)


jskeet (approved):

standard/classes.md
```diff
@@ -2220,15 +2222,15 @@ In a method that takes multiple by-reference parameters, it is possible for mult
 
 ##### 15.6.2.3.2 Input parameters
 
-A parameter declared with an `in` modifier is a input parameter. The argument corresponding to an `in` parameter is either a variable existing at the point of the method invocation, or one created by the implementation ([§12.6.2.3](expressions.md#12623-run-time-evaluation-of-argument-lists)) in the method invocation. A variable shall be definitely assigned before it can be passed as an argument for an input parameter. Within a method, an input parameter is always considered definitely assigned.
+A parameter declared with an `in` modifier is an ***input parameter***. The argument corresponding to an `in` parameter is either a variable existing at the point of the method invocation, or one created by the implementation ([§12.6.2.3](expressions.md#12623-run-time-evaluation-of-argument-lists)) in the method invocation. For definite-assignment rules, see [§9.2.8](variables.md#928-input-parameters).
```
jskeet: > The argument corresponding to an `in` parameter

Should this be "input parameter" as well, or was that deliberately using the modifier?

RexJaeschke: I'll change that too; thanks.

BillWagner (approved):
This LGTM as well. Let's :shipit:

PR #1175, opened 2024-09-16
enhance trace-back message when exception thrown by StandardAnchorTags tool

RexJaeschke:
Now displays the source line and the text that could not be found there.


BillWagner (approved):
Thanks @RexJaeschke 

This LGTM. I'll :shipit: now.

Issue #79, opened 2014-08-12
9.2.8 Overflow exceptions when converting *from* decimal?


jskeet:
10.2.8 includes:

> and conversions from decimal to the floating-point types might cause loss of precision or overflow exceptions.

Given that the range of each of float and double is greater than the range of decimal, I can't see how a conversion from decimal to float or double could cause an overflow.

I propose we remove "or overflow exceptions".

PR #1178, opened 2024-09-18
Specify generic constraints added to support nullable reference types in C# 8

BillWagner:
Add the `class?` and `notnull` constraint.

The `default` constraint is added in C# 9, so is not included.


BillWagner:
@jskeet I added the meeting discuss label on this. I doubt that it's final, but it's worth a first look.

standard/classes.md
```diff
@@ -412,17 +412,18 @@ type_parameter_constraints
...
+    | 'class' '?'?
     | 'struct'
+    | 'notnull'
     | 'unmanaged'
     ;
 
 secondary_constraints
-    : interface_type
+    : interface_type '?'?
     | type_parameter
```
KalleOlaviNiemitalo: Shouldn't *secondary_constraints* allow *type_parameter* `?` as the first secondary constraint?  Seems strange to only allow the `?` in subsequent secondary constraints.  E.g.

```csharp
public class C<T, U, V>
    where T : U?, V
{
}
```

vs.

```csharp
public class C<T, U, V>
    where T : U, V?
{
}
```

standard/classes.md
```diff
@@ -434,14 +435,18 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
 
 The value type constraint specifies that a type argument used for the type parameter shall be a non-nullable value type. All non-nullable struct types, enum types, and type parameters having the value type constraint satisfy this constraint. Note that although classified as a value type, a nullable value type ([§8.3.12](types.md#8312-nullable-value-types)) does not satisfy the value type constraint. A type parameter having the value type constraint shall not also have the *constructor_constraint*, although it may be used as a type argument for another type parameter with a *constructor_constraint*.
 
+The ***not null*** constraint specifies that a type argument used for the type parameter shall be a non-nullable value type or a non-nullable reference type. All non-nullable class types, interface types, delegate types, array types, struct types, enum types, and type parameters known to be a non-nullable value type or non-nullable reference type satisfy this constraint.
+
 > *Note*: The `System.Nullable<T>` type specifies the non-nullable value type constraint for `T`. Thus, recursively constructed types of the forms `T??` and `Nullable<Nullable<T>>` are prohibited. *end note*
```
KalleOlaviNiemitalo, line 498: This note would be easier to understand if it were above the `not null` constraint description.

standard/types.md
```diff
@@ -532,7 +532,7 @@ type_argument
     ;
 ```
 
-Each type argument shall satisfy any constraints on the corresponding type parameter ([§15.2.5](classes.md#1525-type-parameter-constraints)).
+Each type argument shall satisfy any constraints on the corresponding type parameter ([§15.2.5](classes.md#1525-type-parameter-constraints)). A type argument whose nullability doesn't match the nullability of the type parameter satisfies the constraint with the exception that a nullable value type does not satisfy the struct constraint. A warning may be issued when the nullability of a type argument does not satisfy the nullability requirements of the constraint.
```
KalleOlaviNiemitalo: Does not satisfy the **value type constraint**.  `struct` is the keyword in the constraint syntax but it's not the name of the constraint.

standard/types.md
```diff
@@ -601,8 +601,20 @@ A type parameter is an identifier designating a value type or reference type tha
...
+nullable_type_parameter
+    : non_nullable_non_value_type_parameter '?'
+    ;
+
+non_nullable_non_value_type_parameter
+    : type_parameter
+    ;
 ```
 
+The *non_nullable_non_value_type_parameter* in *nullable_type_parameter* shall be a type parameter that isn’t constrained to be a value type.
```
KalleOlaviNiemitalo: "shall" can give the wrong idea that the compiler would report an error otherwise.  Could clarify with a note saying that the same syntax can be used with a type parameter that is constrained to be a value type, but it means Nullable\<T\> in that case.

standard/types.md
```diff
@@ -601,8 +601,20 @@ A type parameter is an identifier designating a value type or reference type tha
...
+
+In *nullable_type_parameter*, the annotation `?` indicates the intent that type arguments of this type are nullable. The absence of the annotation `?` indicates the intent that type arguments of this type are non-nullable.
```
KalleOlaviNiemitalo: I'm not sure that the intention is for the **type argument** to be nullable.

```csharp

class C {
    void M<T>()
        where T : class
    {
        T?[] a = {};
    }
    
    void N()
    {
        M<object>();
    }
}
```

In the declaration `T?[] a = {};`, `T?` is a *nullable_type_parameter*.  The corresponding type argument is `object` in `M<object>();`, but that one isn't nullable.

jskeet:
(I'll wait for responses to @KalleOlaviNiemitalo's questions before reviewing, if that's okay.)

standard/types.md
KalleOlaviNiemitalo: On second thought, I'm not sure `T?` is a *nullable_type_parameter* after all.  It matches the *nullable_type_parameter* grammar but *nullable_type_parameter* doesn't seem to be referenced by any other grammar rule, so it is not applicable in this context (nor in any other).

BillWagner: I'm waiting for others to weigh in here as well. I see your point, but I don't have an obvious suggestion yet.

standard/types.md
BillWagner: Also waiting for more eyes here. I see your point, but I don't have the right suggestion yet.

standard/classes.md
```diff
@@ -412,17 +412,18 @@ type_parameter_constraints
...
 
 secondary_constraints
-    : interface_type
-    | type_parameter
-    | secondary_constraints ',' interface_type
-    | secondary_constraints ',' type_parameter
+    : interface_type '?'?
+    | type_parameter '?'?
+    | secondary_constraints ',' interface_type '?'?
+    | secondary_constraints ',' type_parameter '?'?
```
Nigel-Ecma: I think we should review the grammar here. I can't add this as a suggested changed as it covers non-altered lines, so just code. The rules here pre-date the use of ANTLR, if we ANTLRize them and remove unneeded left-recursion (as we do elsewhere) we get:

```ANTLR
type_parameter_constraints
    : primary_constraint (',' secondary_constraints)? (',' constructor_constraint)?
    | secondary_constraints (',' constructor_constraint)?
    | constructor_constraint
    ;

primary_constraint
    : class_type '?'?
    | 'class' '?'?
    | 'struct'
    | 'notnull'
    | 'unmanaged'
    ;

secondary_constraint
    : interface_type '?'?
    | type_parameter '?'?
    ;

secondary_constraints
    : secondary_constraint (',' secondary_constraint)*
    ;
```

Optionally to improve readability we could replace the `'?'?` by adding a rule:

```ANTLR
nullable_type_attribute
    : '?'
    ;

type_parameter_constraints
    : primary_constraint (',' secondary_constraints)? (',' constructor_constraint)?
    | secondary_constraints (',' constructor_constraint)?
    | constructor_constraint
    ;

primary_constraint
    : class_type nullable_type_attribute?
    | 'class' nullable_type_attribute?
    | 'struct'
    | 'notnull'
    | 'unmanaged'
    ;

secondary_constraint
    : interface_type nullable_type_attribute?
    | type_parameter nullable_type_attribute?
    ;

secondary_constraints
    : secondary_constraint (',' secondary_constraint)*
    ;
```

The introduced *nullable_type_attribute* would then also need to be used in other grammar rules where the literal `'?'` is used as a type attribute and not in places where it is not, e.g. for the null conditional operations.

standard/classes.md
```diff
@@ -434,16 +435,20 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
 
 The list of constraints given in a `where` clause can include any of the following components, in this order: a single primary constraint, one or more secondary constraints, and the constructor constraint, `new()`.
 
-A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, or the ***unmanaged type constraint*** `unmanaged`.
+A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***nullable reference type constraint*** `class?`, the ***not null*** constraint `notnull`, the ***value type constraint*** `struct` or the ***unmanaged type constraint*** `unmanaged`.
 
-A secondary constraint can be a *type_parameter* or *interface_type*.
+A secondary constraint can be a *type_parameter* or *interface_type*, either optionally followed by `?`.
```
Nigel-Ecma: This doesn’t read well to me in this context, given the comment on 426 above I’ll offer two alternatives:

```suggestion
A secondary constraint can be a *type_parameter* or *interface_type*, either optionally nullable.
```

Or if the line 426 added rule is adopted:

```suggestion
A secondary constraint can be a *type_parameter* or *interface_type*, optionally followed by a *nullable_type_attribute*.
```

standard/classes.md
```diff
@@ -604,7 +609,9 @@ The ***effective interface set*** of a type parameter `T` is defined as follows
 - If `T` has no *interface_type* constraints but has *type_parameter* constraints, its effective interface set is the union of the effective interface sets of its *type_parameter* constraints.
 - If `T` has both *interface_type* constraints and *type_parameter* constraints, its effective interface set is the union of the set of dynamic erasures of its *interface_type* constraints and the effective interface sets of its *type_parameter* constraints.
 
-A type parameter is *known to be a reference type* if it has the reference type constraint or its effective base class is not `object` or `System.ValueType`.
+A type parameter is *known to be a non-nullable reference type* if it has the non-nullable reference type constraint or its effective base class is not `object` or `System.ValueType`.
+
+A type parameter is *known to be a reference type* if it has the non-nullable reference type constraint, reference type constraint or its effective base class is not `object` or `System.ValueType`.
```
Nigel-Ecma: Could this be simplified to:

```suggestion
A type parameter is *known to be a non-nullable reference type* if it has the non-nullable reference type constraint or its effective base class is not `object` or `System.ValueType`.

A type parameter is *known to be a reference type* if it has the reference type constraint or it is known to be a non-nullable reference type.
```

standard/classes.md
```diff
@@ -878,7 +885,7 @@ All members of a generic class can use type parameters from any enclosing class,
 > class C<V>
 > {
 >     public V f1;
->     public C<V> f2 = null;
+>     public C<V> f2 = null!;
```
Nigel-Ecma: It is unclear to me why in the original `f2` was initialised, and `f1` not, when both are set in the constructor. Now a *suppression* has been added…

I suggest a comment explaining the *suppression*, or the use of `default` (which leaves the explanation for the section on `default` to handle), or just drop the initialisation.

standard/types.md
```diff
@@ -532,7 +532,7 @@ type_argument
     ;
 ```
 
-Each type argument shall satisfy any constraints on the corresponding type parameter ([§15.2.5](classes.md#1525-type-parameter-constraints)).
+Each type argument shall satisfy any constraints on the corresponding type parameter ([§15.2.5](classes.md#1525-type-parameter-constraints)). A type argument whose nullability doesn't match the nullability of the type parameter satisfies the constraint with the exception that a nullable value type does not satisfy the value type constraint. A warning may be issued when the nullability of a type argument does not satisfy the nullability requirements of the constraint.
```
Nigel-Ecma: I don’t think this needs to be phrased in terms of value type exceptions, the point being that reference type nullability is a “soft” constraint. Maybe:

```suggestion
Each type argument shall satisfy any constraints on the corresponding type parameter ([§15.2.5](classes.md#1525-type-parameter-constraints)). A reference type argument whose nullability doesn’t match the nullability of the type parameter satisfies the constraint; however a warning may be issued.
```

I see that @KalleOlaviNiemitalo has commented on the use of “shall” here, I'll defer that to @RexJaeschke.

standard/types.md
```diff
@@ -601,8 +601,20 @@ A type parameter is an identifier designating a value type or reference type tha
...
     : identifier
     ;
+
+nullable_type_parameter
+    : non_nullable_non_value_type_parameter '?'
+    ;
+
+non_nullable_non_value_type_parameter
+    : type_parameter
+    ;
```
Nigel-Ecma: This grammar is wrong/pointless, as @KalleOlaviNiemitalo has commented the added rules are not used anywhere.

Unfortunately I’m not sure what the intended semantics are here.

Guessing: is the addition of the nullable type attribute meant to require the type argument to be a nullable reference and/or value type, i.e. it’s an additional kind of *primary_constraint* or *secondary_constraint*?

Nigel-Ecma (changes requested):
Some comments plus ANTLRizing some grammar rules.

I haven’t considered *per se* what should (not) be in the Standard in regards to the specification of NRT.

standard/classes.md
BillWagner: I like the second one. When I was working through the grammar, I found `'?'?` hard to mentally parse and read through the grammar.

jskeet: +1 for a rule for nullable_type_attribute.

standard/classes.md
jskeet: `its` => `it's` in the suggestion (line 614) - but then I like it.

standard/classes.md
jskeet: I think it would be better to drop the initialization.

jskeet (commented):
I like Nigel's suggestions in general. My guess is that the currently-unused rules are intended to be used in the future? Would be good to know.

standard/classes.md
BillWagner: I think we're not to use contractions, so "it is".

BillWagner:
@jskeet 

> My guess is that the currently-unused rules are intended to be used in the future? Would be good to know.

Which rules do you think are unused? I didn't see any.

jskeet:
> @jskeet
> 
> > My guess is that the currently-unused rules are intended to be used in the future? Would be good to know.
> 
> Which rules do you think are unused? I didn't see any.

`non_nullable_non_value_type_parameter` and `nullable_type_parameter` - see Nigel's comments.

standard/types.md
BillWagner: That was my mistake. I added the additional rules in the latest commit.

Note that the grammar additions for `non_nullable_reference_type` and `nullable_reference_type` are in #1089

BillWagner:
> > @jskeet
> > > My guess is that the currently-unused rules are intended to be used in the future? Would be good to know.
> > 
> > 
> > Which rules do you think are unused? I didn't see any.
> 
> `non_nullable_non_value_type_parameter` and `nullable_type_parameter` - see Nigel's comments.

Got it. I'd missed some edits. See latest commit.

standard/types.md
```diff
@@ -597,12 +598,25 @@ Since type parameters are not inherited, constraints are never inherited either.
 
 A type parameter is an identifier designating a value type or reference type that the parameter is bound to at run-time.
 
+type_parameter, nullable_type_parameter, non_nullable_non_value_type_parameter
+
```
Nigel-Ecma: Is this a writer's note left in by accident?

```suggestion
```

BillWagner: Yes.

standard/types.md
```diff
@@ -6,10 +6,11 @@ The types of the C# language are divided into two main categories: ***reference
...
 ```ANTLR
 type
-    : reference_type
-    | value_type
+    : value_type
+    | reference_type
+    | nullable_type_parameter
     | type_parameter
-    | pointer_type     // unsafe code support
+    | type_unsafe
```
jskeet: Was this change from `pointer_type` to `type_unsafe` deliberate? Is there another change coming to clause 23?

standard/types.md
jskeet: Calling @Nigel-Ecma as this is on the blurry line between prose and grammar. I'm only 80% convinced I understand what's going on here in the first place, so definitely don't have a suggestion to improve it.

Nigel-Ecma: @jskeet – Hmm, I added further comments days ago, but after my review, and they seem to have gone AWOL. Off to make coffee and then I’ll take another shot, no promises that they’ll help here of course…

standard/classes.md
Nigel-Ecma: Edited above suggestion as per preceeding two comments from @jskeet & @BillWagner.

standard/types.md
```diff
@@ -600,9 +601,20 @@ A type parameter is an identifier designating a value type or reference type tha
 ```ANTLR
 type_parameter
     : identifier
+
+nullable_type_parameter
+    : non_nullable_non_value_type_parameter '?'
+    ;
+
+non_nullable_non_value_type_parameter
```
Nigel-Ecma: Our rule names are getting rather tortuous, see comments on §8.2.1 in #1124

standard/types.md
Nigel-Ecma: OK, I suggested that this was written WRT to the wong rule, offering instead:

```suggestion
The *type_parameter* in *non_nullable_non_value_type_parameter* shall be a type parameter that isn’t constrained to be a value type.
```
However that doesn’t address, as @KalleOlaviNiemitalo points out, that a nullable type parameter can be a nullable reference or nullable value type.

Should the rules not be:

```ANTLR
type_parameter
    : identifier
    ;

nullable_type_parameter
    : type_parameter nullable_type_attribute
    ;
```

and the para then:

```suggestion
The *type_parameter* in *nullable_type_parameter* shall be a type parameter that is constrained (§15.2.5) to be a non-nullable reference or non-nullable value type.
```

Leaving the explanation of what constraints (`struct`, `class` or `notnull`) are valid to §15.2.5.

standard/types.md
```diff
@@ -600,9 +601,20 @@ A type parameter is an identifier designating a value type or reference type tha
...
+    : type_parameter
     ;
 ```
 
+The *non_nullable_non_value_type_parameter* in *nullable_type_parameter* shall be a type parameter that isn’t constrained to be a value type.
+
+In *nullable_type_parameter*, the annotation `?` indicates the intent that nullable type corresponding to the type arguments of this type are nullable. The absence of the annotation `?` indicates the intent that type arguments of this type are non-nullable.
```
Nigel-Ecma: This seems to imply that the *type_parameter_list* of a generic declaration can contain *nullable_type_parameter* but that is not (currently?) the case. *nullable_type_parameter* is a *type* and used in such things as the declaration of variables.

Is not the intent described here conveyed by the constraints §15.2.5?

Suggestion: delete this

```suggestion
```

standard/classes.md
```diff
@@ -412,17 +412,18 @@ type_parameter_constraints
...
     | 'unmanaged'
     ;
 
 secondary_constraints
-    : interface_type
-    | type_parameter
-    | secondary_constraints ',' interface_type
-    | secondary_constraints ',' type_parameter
+    : interface_type '?'?
+    | type_parameter '?'?
```
Nigel-Ecma: What is the intended meaning of *type_parameter* *nullable_type_attribute* as a constraint? The following text does not explain and while the current version of Roslyn accepts:

```
class Curious<S, T> where S : T?
```

It only accepts reference types for `S` & `T`.

Maybe some form of the (suggested removed) text on line 616 of types.md was meant to be here?

BillWagner:
Example discussed in the meeting:

```
M("Hello", null);
M(7, 3);
M<int?>(8, null);    // Warning

void M<T>(T t, T? n) where T : notnull 
{
    T t1 = t;
    T t2 = n;        // Warning
    T t3 = default;  // Warning
    T? n1 = t;
    T? n2 = n;
    T? n3 = default;
    T? n4 = null;    // Error

}
```

standard/classes.md
BillWagner: I added the second set of rules.

standard/types.md
BillWagner: I think that's a merge mistake. Reverted.

jskeet:
@BillWagner Is this ready for another round of review?

BillWagner:
> Is this ready for another round of review?

@jskeet ~~Not yet. But on my list for this week.~~

Edits made and pushed this morning (my time). It's ready for review.

standard/types.md
BillWagner: done (in next commit)

standard/types.md
```diff
@@ -537,11 +537,15 @@ type_arguments
     ;   
 
 type_argument
-    : type
+    : type nullable_argument_attribute?
+    ;
+
+nullable_argument_attribute
```
BillWagner: Note for Nigel: This enables nullable annotations on a type argument. It follows the structure you suggested in types for nullable annotation on type parameters. 

It removes the `??` from the grammar, but also does repeat the same syntax as different productions. Suggestions welcome.

Nigel-Ecma: @BillWagner – The attribute applies to types and we don't need two rules. Use *nullable_type_attribute* here. At the first use in the Standard (probably here?) define it and just use it elsewhere (so probably remove the definition in classes.md).

standard/types.md
```diff
@@ -6,8 +6,8 @@ The types of the C# language are divided into two main categories: ***reference
 
 ```ANTLR
 type
-    : reference_type
-    | value_type
+    : value_type
```
jskeet: Is this ordering change meaningful? (I'm guessing at one point it was more than this, and then we removed bits? Or is it genuinely meaningful due to precedence of some kind?)

jskeet (commented):
Just one comment for now. I need to have another look when more awake...

standard/classes.md
```diff
@@ -434,12 +433,16 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
 
 The list of constraints given in a `where` clause can include any of the following components, in this order: a single primary constraint, one or more secondary constraints, and the constructor constraint, `new()`.
 
-A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, or the ***unmanaged type constraint*** `unmanaged`.
+A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***not null constraint*** `notnull`, the ***value type constraint*** `struct` or the ***unmanaged type constraint*** `unmanaged`.
```
```suggestion
A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, the ***not null constraint*** `notnull` or the ***unmanaged type constraint*** `unmanaged`.
```

Just to keep the order in the description the same as in the grammar.

standard/classes.md
```diff
@@ -434,12 +433,16 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
 
-A secondary constraint can be a *type_parameter* or *interface_type*.
+A secondary constraint can be a *type_parameter* or *interface_type*, optionally followed by a *nullable_type_attribute*.
```
```suggestion
A secondary constraint can be an *interface_type* or *type_parameter*, optionally followed by a *nullable_type_attribute*.
```

(Again, just for consistency.)

standard/classes.md
jskeet: Do we need to say "a class type optionally followed by a *nullable_type_attribute*"? We do for the secondary constraints, so I'd imagine we should here as well. (Or remove it from there. There's a description in a couple of paragraphs anyway.)

standard/classes.md
```diff
@@ -604,7 +607,9 @@ The ***effective interface set*** of a type parameter `T` is defined as follows
 - If `T` has no *interface_type* constraints but has *type_parameter* constraints, its effective interface set is the union of the effective interface sets of its *type_parameter* constraints.
 - If `T` has both *interface_type* constraints and *type_parameter* constraints, its effective interface set is the union of the set of dynamic erasures of its *interface_type* constraints and the effective interface sets of its *type_parameter* constraints.
 
-A type parameter is *known to be a reference type* if it has the reference type constraint or its effective base class is not `object` or `System.ValueType`.
+A type parameter is *known to be a non-nullable reference type* if it has the non-nullable reference type constraint or its effective base class is not `object` or `System.ValueType`.
```
jskeet: I wonder whether a link to "effective base class" is useful here. It's not immediately clear to me what the relationship between nullability and effective base classes would be. (I suspect this is effectively "known to be a reference type, and assumed to be non-nullable" - I'm not suggesting a terminology change, but we should think about how we could make this clearer, potentially in a note.)

standard/types.md
```diff
@@ -537,11 +542,11 @@ type_arguments
     ;   
 
 type_argument
-    : type
+    : type nullable_type_attribute?
```
jskeet: Just to check, is this only needed so that we can write `Foo<T?>` for a type parameter? Because `string?` and `int?` are already `type`s.

Perhaps we should make `T?` a `type` in the grammar as well, then we don't need `nullable_type_attribute` here?

jskeet (commented):
I think I'm starting to get my head round this again...

standard/types.md
```diff
@@ -6,8 +6,8 @@ The types of the C# language are divided into two main categories: ***reference
 
 ```ANTLR
 type
-    : reference_type
-    | value_type
+    : value_type
+    | reference_type
```
BillWagner: It was not meaningful. Reverting:

```suggestion
    : reference_type
    | value_type
```

standard/types.md
BillWagner: Correct. this is so `string?` can be used. 

I think we already do what you say:  The grammar for NRTs and Nullable Value types specifies them as a `type` followed by the `nullable_type_attribute`.

See the change at line [#79](https://github.com/dotnet/csharpstandard/pull/1178/files#diff-d3ce84f193d1cf106f2447950cca9dc1ed178c85ad941e568d9a9d527e45e522R79)

jskeet: But `type` already matches `string?` because `type` includes `reference_type` as an option, and that includes `nullable_reference_type` as an option.

So I think `Foo<string?>` is already covered without this change, but `Foo<T?>` wouldn't be.

(With this change, `string??` is a valid type argument, because the `type` production would permit `string?` and then there's the optional `nullable_type_argument`.)

BillWagner: Hmm. Are you thinking that something should be added in 8.5 (type parameters) so that a *type_parameter* could be either an *identifier* or *identifier* *nullable_type_attribute*?  That would allow `T` and `T?`.  However, I think that would also allow `string??` when the type argument was `string?` and the type parameter was `T`.

We made need help from Nigel, or a statement in prose about that happens here.

standard/classes.md
BillWagner: Maybe reword this sentence to say "has the *class_type* constraint". That sets its "effective base class" to that type.

jskeet:
As noted by Neal, this is an error:

```csharp
public partial class PartialList : List<string>
{
}

public partial class PartialList : List<string?>
{
}
```

Error: CS0263 Partial declarations of 'PartialList' must not specify different base classes

jskeet:
Discussions:

- Using `T?` when `T` is unconstrained should be forbidden in C# 8. (For C# 9 we'll need to explain some oddities.)
- `where T : class` should be described as `T` being a non-nullable reference type, and `class?` is "any reference type" - although the non-nullable reference constraint only generates a warning. (We're debating how to specify this.)
- We need to explain what `T?` means when `T` itself is nullable

standard/classes.md
```diff
@@ -434,12 +433,16 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
+A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, the ***not null constraint*** `notnull` or the ***unmanaged type constraint*** `unmanaged`.
 
-A secondary constraint can be a *type_parameter* or *interface_type*.
+A secondary constraint can be an *interface_type* or *type_parameter*, optionally followed by a *nullable_type_attribute*.
 
 The reference type constraint specifies that a type argument used for the type parameter shall be a reference type. All class types, interface types, delegate types, array types, and type parameters known to be a reference type (as defined below) satisfy this constraint.
 
+The class type, reference type constraint, and secondary constraints can include the nullable type attribute. The nullability of the type argument need not match the nullability of the type parameter. The compiler may issue a warning if the nullability of the type parameter doesn't match the nullability of the type argument.
```
jskeet: I suspect "type parameter" should be "constraint" here.

jskeet: We're unsure whether `class?` is intended to mean "definitely nullable" or "might be nullable". Current description suggests that using `string` for a type parameter of `T : class?` could issue a warning.

There are three things users might want to specify:

- "T should be a non-nullable reference type"
- "T should be a nullable reference type"
- "T should be any reference type"

Currently we only have two constraints for those three cases, which is far from ideal.

BillWagner: The 2nd isn't actually a constraint. To say "T should be a nullable reference type" you say "T is a non-nullable reference type" and then use `T?` consistently at all uses.

jskeet: Clarification from Mads: `class?` is intended to mean "might be nullable". If you want "definitely nullable" you don't specify it as a constraint, you constrain T to be non-nullable, but use T? where you want to allow a null value.

BillWagner: Also, on line 440, imply that `class` should be non-nullable, but the compiler accepts it and might warn.

Add an example pointing this out.

BillWagner: Also, work out what `T?` means if `T` is the type parameter, and `string?` (or some other nullable reference type) is the type argument.

BillWagner: Also note that `typeof(T?)` is an error if `T` is known to be a reference type?

jskeet: > Also note that `typeof(T?)` is an error if `T` is known to be a reference type?

Yes, and it's an error in 8 if `T` is unconstrained. It's valid *but useless* in 9+ when `T` is unconstrained (`typeof(T)` and `typeof(T?)` always return the same Type). We might want to file a bug about this.

standard/classes.md
jskeet: I think "non-nullable reference type constraint" here should be "reference type constraint without the nullable type attribute".

standard/classes.md
```diff
@@ -604,7 +607,9 @@ The ***effective interface set*** of a type parameter `T` is defined as follows
 - If `T` has no *interface_type* constraints but has *type_parameter* constraints, its effective interface set is the union of the effective interface sets of its *type_parameter* constraints.
 - If `T` has both *interface_type* constraints and *type_parameter* constraints, its effective interface set is the union of the set of dynamic erasures of its *interface_type* constraints and the effective interface sets of its *type_parameter* constraints.
 
-A type parameter is *known to be a reference type* if it has the reference type constraint or its effective base class is not `object` or `System.ValueType`.
+A type parameter is *known to be a non-nullable reference type* if it has the non-nullable reference type constraint or its effective base class is not `object` or `System.ValueType`.
+
+A type parameter is *known to be a reference type* if it has the reference type constraint or it is known to be a non-nullable reference type.
```
jskeet: I think "reference type constraint" should be "reference type constraint including the nullable type attribute".

standard/classes.md
BillWagner: Discussion:  These two sentences need to be reworked.  It's trying to say what an instance of `T` is in terms of reference types, and it's nullability.

Start with "it's known to be a reference type if...". Then, the next sentence is "It's known to be non-nullable reference type if it's known to be a reference type AND ..."

standard/classes.md
```diff
@@ -396,33 +396,32 @@ type_parameter_constraints_clauses
...
-    | primary_constraint ',' secondary_constraints ',' constructor_constraint
     ;
 
 primary_constraint
-    : class_type
-    | 'class'
+    : class_type nullable_type_attribute?
+    | 'class' nullable_type_attribute?
     | 'struct'
+    | 'notnull'
```
BillWagner, line 414: This needs an update. `class_type` can be combined with `notnull`.

jskeet:
We observe that Roslyn allows both of these:

```csharp
public class Foo<T> where T : notnull, Stream
public class Foo<T> where T : notnull, Stream?
```

This PR prohibits both. @BillWagner will file a bug against Roslyn (which we suspect will be closed).

Nigel-Ecma:
@BillWagner – I made some changes in PR #1195 thinking this PR was already merged, oops. I've removed them from 1195 but cannot make direct suggestions here as the changes are on lines this PR has not modified. SO here they are as text blocks.

The changes centred around changing `?` in the text to `*nullable_type_attribute*` – the latter is introduced in this PR so these changes should either be here or in a subsequent PR.

In types.md §8.9.1, first para, line 729 there are a couple of `?`’s, I suggest:

> A *nullable reference type* is denoted by appending a *nullable_type_attribute* (`?`) to a non-nullable reference type. There is no semantic difference between a non-nullable reference type and its corresponding nullable type, both can either be a reference to an object or `null`. The presence or absence of the *nullable_type_attribute* declares whether an expression is intended to permit null values or not. A compiler may provide diagnostics when an expression is not used according to that intent. The null state of an expression is defined in [§8.9.5](types.md#895-nullabilities-and-null-states). An identity conversion exists among a nullable reference type and its corresponding non-nullable reference type ([§10.2.2](conversions.md#1022-identity-conversion)).

I made a few other word changes, e.g. can -> may.

I do think this para needs some work on its clarity, but I’ve no suggestions right now – and that can come in a subsequent PR as its nothing to do with this one *per se*.

For line 738 I suggest:

> The syntactic distinction between a *nullable reference type* and its corresponding *non-nullable reference type* enables a compiler to generate diagnostics. A compiler must allow the *nullable_type_attribute* as defined in [§8.2.1](types.md#821-general). The diagnostics must be limited to warnings. Neither the presence or absence of nullable annotations, nor the state of the nullable context can change the compile time or runtime behavior of a program except for changes in any diagnostic messages generated at compile time.

On line 746 there is another `?` but as it is in reference to a preceding `T?` I think that is fine.

There might be other `?`s which would be better as *nullable_type_attribute* (or nullable type attribute?), I haven’t done an exhaustive search.

(Note: There is a simple typo on line 728, I’ve added PR #1199 to fix it.)

Before I removed the changes from #1195 @KalleOlaviNiemitalo raised a comment on the naming of *nullable_type_attribute* – I've asked them to make the comment in this PR.

standard/types.md
```diff
@@ -76,8 +76,13 @@ delegate_type
     ;
 
 nullable_reference_type
-    : non_nullable_reference_type '?'
+    : non_nullable_reference_type nullable_type_attribute
     ;
+
+nullable_type_attribute
+    : '?'
+    ;
```
KalleOlaviNiemitalo: _Moved from <https://github.com/dotnet/csharpstandard/pull/1195#discussion_r1835287427>_

The term "attribute" has a different meaning in [22 Attributes](https://github.com/dotnet/csharpstandard/blob/94c967e7c4cb49d9c4c1eec887b65793a2368c11/standard/attributes.md#22-attributes) so using it as part of *nullable_type_attribute* feels a bit unfortunate.  Not sure what would be a better term, though.  "Facet" as used in W3C XML Schema Definition Language (XSD) comes to mind but that doesn't feel right either.  Perhaps "specifier" as in *rank_specifier*?

standard/classes.md
BillWagner: fixed in next commit.

standard/classes.md
BillWagner: I believe this paragraph is redundant after making the change on line 610. I removed the paragraph.

standard/classes.md
BillWagner: > > Also note that `typeof(T?)` is an error if `T` is known to be a reference type?
> 
> Yes, and it's an error in 8 if `T` is unconstrained. It's valid _but useless_ in 9+ when `T` is unconstrained (`typeof(T)` and `typeof(T?)` always return the same Type). We might want to file a bug about this.

These changes are added in expressions.md, in section 12.8.12 (The `typeof` operator)

standard/types.md
BillWagner: The grammar has been updated, so this is no longer an issue.

standard/types.md
BillWagner: This no longer applies after the edits. Resolving.

standard/classes.md
BillWagner: Added a sentence to clarify.

standard/classes.md
```diff
@@ -437,13 +433,63 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
...
 
-A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***not null constraint*** `notnull`, the ***value type constraint*** `struct` or the ***unmanaged type constraint*** `unmanaged`.
+A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, the ***not null constraint*** `notnull` or the ***unmanaged type constraint*** `unmanaged`. The class type and the reference type constraint can include the *nullable_type_attribute*.
 
-A secondary constraint can be a *type_parameter* or *interface_type*, optionally followed by a *nullable_type_attribute*.
+A secondary constraint can be an *interface_type* or *type_parameter*, optionally followed by a *nullable_type_attribute*. The presence of the *nullable_type_attribute* indicates that the type argument is allowed to the the nullable reference type that corresponds to a non-nullable reference type that satisfies the constraint.
```
Nigel-Ecma: Doubled “the” typo – ”be the” maybe?

```suggestion
A secondary constraint can be an *interface_type* or *type_parameter*, optionally followed by a *nullable_type_attribute*. The presence of the *nullable_type_attribute* indicates that the type argument is allowed to be the nullable reference type that corresponds to a non-nullable reference type that satisfies the constraint.
```

standard/types.md
BillWagner: At the meeting we decided to use `nullable_type_annotation`.

Nigel will address in #1195

Nigel-Ecma (approved):

PR #1179, opened 2024-09-19
Add status checks to the test runner

BillWagner:
Contributes to #1125

Replace the existing logging code with the utility class that writes to the console, and adds a status check record.

Log failures as status checks, but status is only to the output console.

### DIFF
--- a/tools/ExampleTester/GeneratedExample.cs
+++ b/tools/ExampleTester/GeneratedExample.cs
@@ -5,6 +5,7 @@ using Microsoft.CodeAnalysis.MSBuild;
 using Newtonsoft.Json;
 using System.Reflection;
 using System.Text;
+using Utilities;
 
 namespace ExampleTester;
 
@@ -33,9 +34,9 @@ internal class GeneratedExample
         return new GeneratedExample(directory);
     }
 
-    internal async Task<bool> Test(TesterConfiguration configuration)
+    internal async Task<bool> Test(TesterConfiguration configuration, StatusCheckLogger logger)
     {
-        var outputLines = new List<string> { $"Testing {Metadata.Name} from {Metadata.Source}" };
+        logger.ConsoleOnlyLog(new StatusCheckMessage(Metadata.Source, Metadata.StartLine, Metadata.EndLine, $"Testing {Metadata.Name} from {Metadata.Source}", "ExampleTester"));
 
         // Explicitly do a release build, to avoid implicitly defining DEBUG.
         var properties = new Dictionary<string, string> { { "Configuration", "Release" } };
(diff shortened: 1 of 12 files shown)


tools/Utilities/StatusCheckLogger.cs
```diff
@@ -10,7 +10,7 @@ namespace Utilities;
 /// <param name="Id">The error message ID</param>
 /// <param name="StartLine">The start line (index from 1)</param>
 /// <param name="EndLine">The end line (index from 1)</param>
-public record Diagnostic(string file, int StartLine, int EndLine, string Message, string Id);
+public record StatusCheckMessage(string file, int StartLine, int EndLine, string Message, string Id);
```
BillWagner, line 13: I changed the name of this record type because the tester uses the "Diagnostic" class from the roslyn APIs. We don't use this in too many places, so the global change seemed the right way to go

tools/Utilities/StatusCheckLogger.cs
```diff
@@ -30,7 +30,25 @@ public class StatusCheckLogger(string pathToRoot, string toolName)
...
+
+    /// <summary>
+    /// Log a notice from the status check to the console only
+    /// </summary>
+    /// <param name="d">The diagnostic</param>
+    /// <remarks>
+    /// Log the diagnostic information to console. These will only appear in the console window, not
+    /// as annotations on the changes in the PR. 
+    /// </remarks>
+    public void ConsoleOnlyLog(StatusCheckMessage d)
```
BillWagner: I added this method because it looks like many of the status messages on the tester would be just to the console, rather than annotate the spec with info on every test.

tools/Utilities/StatusCheckLogger.cs
```diff
@@ -30,7 +30,25 @@ public class StatusCheckLogger(string pathToRoot, string toolName)
...
+    {
+        WriteMessageToConsole("", d);
+        annotations.Add(
+            new(FormatPath(d.file),
+            d.StartLine, d.EndLine,
+            CheckAnnotationLevel.Notice, $"{d.Id}::{d.Message}")
+        );
```
BillWagner: This should be removed.

```suggestion
```

tools/Utilities/StatusCheckLogger.cs
```diff
@@ -30,7 +30,20 @@ public class StatusCheckLogger(string pathToRoot, string toolName)
...
+    /// </summary>
+    /// <param name="d">The diagnostic</param>
+    /// <remarks>
+    /// Log the diagnostic information to console. These will only appear in the console window, not
+    /// as annotations on the changes in the PR. 
+    /// </remarks>
+    public void ConsoleOnlyLog(StatusCheckMessage d)
+    {
+        WriteMessageToConsole("", d);
+    }
```
BillWagner:
```suggestion
    public void ConsoleOnlyLog(StatusCheckMessage d) =>
        WriteMessageToConsole("", d);
```

jskeet:
Will review on Tuesday when I'm back at work.

cheshmrangi (approved):

BillWagner:
ping @jskeet for review

jskeet:
Sorry for the delay. Will do it first thing tomorrow UK time, when I'm not in meetings...

cheshmrangi (approved):

tools/ExampleTester/GeneratedExample.cs
```diff
@@ -97,7 +95,7 @@ bool ValidateOutput()
             {
                 if (Metadata.ExpectedOutput != null)
                 {
-                    outputLines.Add("  Output expected, but project has no entry point.");
+                    logger.LogFailure(new StatusCheckMessage(Metadata.Source, Metadata.StartLine, Metadata.EndLine, "  Output expected, but project has no entry point.", "ExampleTester"));
```
jskeet: No need to do it now, but I do wonder whether an overload accepting the parameters directly instead of a StatusCheckMessage would avoid some repetition here. (Not that this code is worth gold-plating...)

jskeet (approved):
This is the sort of PR where I may well have missed something, but I can't see anything right now. We can always fix any problems afterwards :)

tools/ExampleTester/GeneratedExample.cs
BillWagner: Copilot made it easy. I made the change.

PR #1180, opened 2024-09-24
Remove dynamic conversion from explicit reference conversions

jcouv:
@BillWagner for review. Thanks

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -470,7 +470,7 @@ The explicit nullable conversions are those nullable conversions ([§10.6.1](con
 
 The explicit reference conversions are:
 
-- From object and dynamic to any other *reference_type*.
+- From object to any other *reference_type*.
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM.

For @jskeet

The reason this removal is OK:  The section on [implicit dynamic conversions](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/conversions.md#10210-implicit-dynamic-conversions) already has an implicit conversion from `dynamic` to any reference type. There's no need for an additional explicit conversion.

Nigel-Ecma (approved):
This might seem wrong at first, of course there is an explicit conversion from dynamic to a reference type – but that is covered in §10.3.8 *Explicit Dynamic Conversions* and this is just removing the overlap.

jskeet (approved):

PR #1191, opened 2024-10-22
Add the list of attributes that support nullable analysis.

### LINKED ISSUE #1092: [Nullable Reference Types] Specify behavior for nullable analysis attributes
There are a number of attributes read by the compiler that support nullable flow analysis, and provide some additional syntactic rules where those attributes are applied.

The list of attributes, and how that affects the syntax where applied should be normative. How those attributes affect flow analysis may include both normative and informative text.

BillWagner:
Fixes #1092 

Based on the last meeting, list the attributes so that their function is normative, however a compiler need not issues warnings based on these rules.

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -863,7 +863,7 @@ For invocations that occur within declarations of instance constructors, static 
 
 #### §Nullable-Analysis-Attributes-General General
 
-Code compiled with both nullable contexts (§Nullable-Contexts) disabled is null oblivious (§Nullabilities-And-Null-States). That means any reference type variable may be null, but null checks aren't required. Once such code is made nullable-aware, those rules change. Reference type variables should never have the null value, and such variables must be checked against null before being dereferenced.
+Code compiled with both nullable contexts (§8.9.4) disabled is null oblivious (§8.9.5). That means any reference type variable may be null, but null checks aren't required. Once such code is made nullable-aware, those rules change. Reference type variables should never have the null value, and such variables must be checked against null before being dereferenced.
 
 Some APIs have more complex rules for when variables can or can't be null. In these cases, one or more of the nullable-related attributes described below can be used to express those rules. When user code is compiled in a nullable-enabled context, the compiler is required to warn when that code violates those rules. That is, these attributes help define the nullability contract for an API.
 
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -519,13 +519,14 @@ Using the terms defined in [§22.4.2](attributes.md#2242-compilation-of-an-attri
 
 ### 22.5.1 General
 
-A small number of attributes affect the language in some way. These attributes include:
+A number of attributes affect the language in some way. These attributes include:
 
 - `System.AttributeUsageAttribute` ([§22.5.2](attributes.md#2252-the-attributeusage-attribute)), which is used to describe the ways in which an attribute class can be used.
 - `System.Diagnostics.ConditionalAttribute` ([§22.5.3](attributes.md#2253-the-conditional-attribute)), is a multi-use attribute class which is used to define conditional methods and conditional attribute classes. This attribute indicates a condition by testing a conditional compilation symbol.
 - `System.ObsoleteAttribute` ([§22.5.4](attributes.md#2254-the-obsolete-attribute)), which is used to mark a member as obsolete.
 - `System.Runtime.CompilerServices.AsyncMethodBuilderAttribute` ([§22.5.5](attributes.md#2255-the-asyncmethodbuilder-attribute)), which is used to establish a task builder for an async method.
 - `System.Runtime.CompilerServices.CallerLineNumberAttribute` ([§22.5.6.2](attributes.md#22562-the-callerlinenumber-attribute)), `System.Runtime.CompilerServices.CallerFilePathAttribute` ([§22.5.6.3](attributes.md#22563-the-callerfilepath-attribute)), and `System.Runtime.CompilerServices.CallerMemberNameAttribute` ([§22.5.6.4](attributes.md#22564-the-callermembername-attribute)), which are used to supply information about the calling context to optional parameters.
+- Nullable static analysis attributes (§Nullable-Analysis-Attributes).
 
 An execution environment may provide additional implementation-defined attributes that affect the execution of a C# program.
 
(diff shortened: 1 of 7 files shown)


standard/attributes.md
```diff
@@ -519,13 +519,14 @@ Using the terms defined in [§22.4.2](attributes.md#2242-compilation-of-an-attri
...
 - `System.Diagnostics.ConditionalAttribute` ([§22.5.3](attributes.md#2253-the-conditional-attribute)), is a multi-use attribute class which is used to define conditional methods and conditional attribute classes. This attribute indicates a condition by testing a conditional compilation symbol.
 - `System.ObsoleteAttribute` ([§22.5.4](attributes.md#2254-the-obsolete-attribute)), which is used to mark a member as obsolete.
 - `System.Runtime.CompilerServices.AsyncMethodBuilderAttribute` ([§22.5.5](attributes.md#2255-the-asyncmethodbuilder-attribute)), which is used to establish a task builder for an async method.
 - `System.Runtime.CompilerServices.CallerLineNumberAttribute` ([§22.5.6.2](attributes.md#22562-the-callerlinenumber-attribute)), `System.Runtime.CompilerServices.CallerFilePathAttribute` ([§22.5.6.3](attributes.md#22563-the-callerfilepath-attribute)), and `System.Runtime.CompilerServices.CallerMemberNameAttribute` ([§22.5.6.4](attributes.md#22564-the-callermembername-attribute)), which are used to supply information about the calling context to optional parameters.
+- Nullable static analysis attributes (§Nullable-Analysis-Attributes).
```
Nigel-Ecma: This is in a list of “attributes affect the language in some way”, I think they are attributes which *may* affect the *compiler* in some way. Maybe start a new list or just make this a para?

standard/attributes.md
```diff
@@ -858,6 +859,276 @@ For invocations that occur within field or event initializers, the member name u
 
 For invocations that occur within declarations of instance constructors, static constructors, finalizers and operators the member name used is implementation-dependent.
 
+### §Nullable-Analysis-Attributes Code analysis-attributes
+
+#### §Nullable-Analysis-Attributes-General General
+
+Code compiled with both nullable contexts (§8.9.4) disabled is null oblivious (§8.9.5). That means any reference type variable may be null, but null checks aren't required. Once such code is made nullable-aware, those rules change. Reference type variables should never have the null value, and such variables must be checked against null before being dereferenced.
+
+Some APIs have more complex rules for when variables can or can't be null. In these cases, one or more of the nullable-related attributes described below can be used to express those rules. When user code is compiled in a nullable-enabled context, the compiler is required to warn when that code violates those rules. That is, these attributes help define the nullability contract for an API.
```
Nigel-Ecma: “Once such code is made nullable-aware, those rules change” -> “Nullable-aware code follows different rules” – or something like that. Users might write new code which is “null oblivious” and other code which is “nullable-aware” – i.e. for former need not be old and the latter need not have been the former, if that makes sense!

“The complier is required to warn…” – first a compiler is not required to have a null analysis is it? And if it does as the algorithm it uses it implementation defined is it required to enforce the semantics of these attributes or could it just ignore them?

I see the purpose of listing these attributes as (a) enforcing that they are reserved and (b) [probably?] saying that if a compiler chooses to use them it should do so with the meanings listed. Anything more than that is up to the implementation. But maybe we’re going to go further than that?

Nigel-Ecma (commented):
An early review, just comments. In general I’m not sure the must/may language is correct. I haven’t gone through all the examples so no comments there but a glance suggests I'd have similar queries – but examples are informative…

standard/attributes.md
BillWagner: Yes, the next commits updated the language to reflect that.

I'm leaving the examples as instructive (and one more PR is to create more examples on nullability and null states)

standard/attributes.md
```diff
@@ -863,9 +864,7 @@ For invocations that occur within declarations of instance constructors, static
...
-
-Some APIs have more complex rules for when variables can or can't be null. In these cases, one or more of the nullable-related attributes described below can be used to express those rules. When user code is compiled in a nullable-enabled context, the compiler is required to warn when that code violates those rules. That is, these attributes help define the nullability contract for an API.
+The attributes in this section are used to provide additional information to support a compiler that provides nullability and null-state diagnostics (§8.9.5). A compiler isn't required to perform any null-state diagnostics. The presence or absence of these attributes do not affect the language nor the behavior of a program. A compiler that doesn't provide null-state diagnostics shall read and ignore the presence of these attributes. A compiler that provides null-state diagnostics shall use the meaning meaning of the attributes defined in this section.
```
Nigel-Ecma: I think the tense of the last sentence is wrong. As I understand it the aim it to say that *if* a compiler implements an analysis algorithm which tales notice of any of these attributes then it should do so with the meaning defined, it may also choose to ignore any of the attributes that do not suit its algorithm, and presumably may define other implementation specific ones if it wishes.

I read the last sentence as *requiring* any algorithm to take not of all the attributes (considering the previois sentence).

standard/attributes.md
```diff
@@ -1028,13 +1028,13 @@ Specifies that a non-nullable return value may be null.
 
 #### §The-MaybeNullWhen-Attribute The MaybeNullWhen attribute
 
-Specifies that a non-nullable argument may be `null` when the method returns the specified `bool` value.
+Specifies that a non-nullable argument may be `null` when the method returns the specified `bool` value. This is similar to the `MaybeNull` attribute (§The-MaybeNull-Attribute), but include a parameter for the specified return value.
```
Nigel-Ecma: Typo?

```suggestion
Specifies that a non-nullable argument may be `null` when the method returns the specified `bool` value. This is similar to the `MaybeNull` attribute (§The-MaybeNull-Attribute), but includes a parameter for the specified return value.
```

standard/attributes.md
```diff
@@ -1028,13 +1028,13 @@ Specifies that a non-nullable return value may be null.
...
 
 #### §The-MemberNotNull-Attribute The MemberNotNull attribute
 
 Specifies that the given member won't be `null` when the method returns.
 
-> *Example*: The compiler analyzes constructors and field initializers to make sure that all non-nullable reference fields have been initialized before each constructor returns. However, the compiler doesn't track field assignments through all helper methods. The compiler issues a warning when fields aren't initialized directly in the constructor, but rather in a helper method. This warning is suppressed by applying the `MemberNotNull` attribute to a method declaration and specifying the fields that are initialized to a non-null value in the method. For example, consider the following example:
+> *Example*: The compiler may analyze constructors and field initializers to make sure that all non-nullable reference fields have been initialized before each constructor returns. However, the compiler doesn't track field assignments through called helper methods. The `MemberNotNull` attribute specifies the fields that are initialized to a non-null value in that method. For example, consider the following example:
```
Nigel-Ecma: “the compiler doesn’t track” -> “a compiler need not track”?

When I first read “initialized to a non-null value in that method” it took a moment to figure out which method it was referring to.

I think this attribute is intended to help a compiler determine whether a constructor initializes a given field to a non-null value – by being attached to a helper method the constructor calls and to save it analyzing that method. If that is the correct meaning I don’t think it comes across clearly.

Typo on line 1039: Memeber

standard/attributes.md
```diff
@@ -1096,9 +1096,9 @@ Specifies that a nullable return value will never be `null`.
 
 Specifies that a return value isn't `null` if the argument for the specified parameter isn't `null`.
 
-> *Example*: Sometimes the null state of a return value depends on the null state of one or more arguments. Such a method will return a non-null value whenever certain arguments aren't `null`. To correctly annotate these methods, add the `NotNullIfNotNull` attribute. Consider the following method:
+> *Example*: The null state of a return value may depend on the null state of one or more arguments. Such a method will return a non-null value when certain arguments aren't `null`. To correctly annotate these methods, add the `NotNullIfNotNull` attribute. Consider the following method:
```
Nigel-Ecma:
```suggestion
> *Example*: The null state of a return value could depend on the null state of one or more arguments. Such a method may return a non-null value when certain arguments aren't `null`. To assist the compiler with analyzing such methods the `NotNullIfNotNull` attribute may be used. Consider the following method:
```

There first two changes I’m fairly certain about, the last two are bit more dubious but I think the original could do with some re-wording.

Nigel-Ecma (commented):
Just another bunch of comments

standard/attributes.md
```diff
@@ -858,6 +860,275 @@ For invocations that occur within field or event initializers, the member name u
 
 For invocations that occur within declarations of instance constructors, static constructors, finalizers and operators the member name used is implementation-dependent.
 
+### §Nullable-Analysis-Attributes Code analysis-attributes
```
jskeet: Remove the hyphen here?

standard/attributes.md
```diff
@@ -858,6 +860,275 @@ For invocations that occur within field or event initializers, the member name u
...
+
+#### §Nullable-Analysis-Attributes-General General
+
+The attributes in this section are used to provide additional information to support a compiler that provides nullability and null-state diagnostics (§8.9.5). A compiler isn't required to perform any null-state diagnostics. The presence or absence of these attributes do not affect the language nor the behavior of a program. A compiler that doesn't provide null-state diagnostics shall read and ignore the presence of these attributes. A compiler that provides null-state diagnostics shall use the meaning meaning of the attributes defined in this section.
```
jskeet:
```suggestion
The attributes in this section are used to provide additional information to support a compiler that provides nullability and null-state diagnostics (§8.9.5). A compiler isn't required to perform any null-state diagnostics. The presence or absence of these attributes do not affect the language nor the behavior of a program. A compiler that doesn't provide null-state diagnostics shall read and ignore the presence of these attributes. A compiler that provides null-state diagnostics shall use the meaning of the attributes defined in this section.
```

standard/attributes.md
jskeet: How about:

> A compiler that provides null-state diagnostics shall use the meaning defined in this section for any of these attributes which it uses to inform its diagnostics.

standard/attributes.md
```diff
@@ -858,6 +860,275 @@ For invocations that occur within field or event initializers, the member name u
...
+>         set => _comment = value ?? throw new ArgumentNullException(nameof(value),
+>            "Cannot set to null");
+>     }
+>     private string? _comment = default;
+> }
+> ```
+>
+> The get accessor could return the default value of `null`, so the compiler may warn that it must be checked before access. Furthermore, it warns callers that, even though it could be null, callers shouldn't explicitly set it to null. *end example*
+
+#### §The-DoesNotReturn-Attribute The DoesNotReturn attribute
```
jskeet, line 942: This feels only tangentially null-related. Maybe we should specify this one normatively? (Ditto DoesNotReturnIf)

standard/attributes.md
```diff
@@ -858,6 +860,275 @@ For invocations that occur within field or event initializers, the member name u
...
+
+> *Example*: The null state of a return value may depend on the null state of one or more arguments. Such a method will return a non-null value when certain arguments aren't `null`. To correctly annotate these methods, add the `NotNullIfNotNull` attribute. Consider the following method:
+>
+> <!-- Example: {template:"code-in-class-lib", name:"NotNullIfNotNull1Attribute", replaceEllipsis:true, customEllipsisReplacements: ["return \"\";"]} -->
+> ```csharp
+> #nullable enable
+> string GetTopLevelDomainFromFullUrl(string url) { ... }
+> ```
+>
+> If the `url` argument isn't `null`, `null` isn’t returned. When nullable references are enabled, that signature works correctly, provided the API never accepts a null argument. However, if the argument could be null, then return value could also be null. To express that contract correctly, annotate this method as follows:
```
jskeet:
```suggestion
> If the `url` argument isn't `null`, `null` isn’t returned. When nullable references are enabled, that signature works correctly, provided the API never accepts a null argument. However, if the argument could be null, then the return value could also be null. To express that contract correctly, annotate this method as follows:
```

standard/attributes.md
BillWagner: Rewrote that paragraph.

standard/attributes.md
BillWagner: I'll ask compiler folks, but I don't think these attributes affect any other analysis (such as definite assignment, or unreachable code).

See [SharpLab](https://sharplab.io/#v2:EYLgtghglgdgPgAQEwEYCwAoBAGABAlAOgBEoIBzGAewGcAXKAYxsIGEqATAUwEEYIANgE8aUGgG5MCAMz4kuVpgDemAMQwArgIERgArri789XTLnO4zFgNrEqXGgDkqdAEpc6GgE4wAulfMABy8oADcIOgMEABZcAEkwAHkNOjoIAAkuLy4ACgBKXABeAD5cOgALLyoAd1wYLlrHLQF3ADMso0YuAFEADy7AhioYfMkMC0txixl8WIBZHIJsAH5cQLyA3BUpifME5NSMjtHNiYB6M9xnXGqIH1hyXFaqL1wNGGyIRnLdfVxGThcQjA04WC5rXBiXCQITALgAWk02lB5lgdFw+hg5AqRTWhAAMkZseUxrsLCjcODApCaHUXHVmoQKUtcLTCmtSWTJlzIa1cDlaQBCdlIgQbHa7bY8iZojHydksQlYiqcskAXwp4NpUJhcIZyIlMpg6IEsgVBKJKs2GowaqAA)

BillWagner: Verified by roslyn team:  This attribute doesn't affect any other analysis (reachable code, definite assignment), it's only use is to impact null state analysis. I think it should stay here.

I updated the sample to provide one that more likely matches a typical use. It validates an argument as to whether it's null or not.

jskeet: Righto.

standard/attributes.md
```diff
@@ -858,6 +860,275 @@ For invocations that occur within field or event initializers, the member name u
...
+
+#### §The-MaybeNullWhen-Attribute The MaybeNullWhen attribute
+
+Specifies that a non-nullable argument may be `null` when the method returns the specified `bool` value. This is similar to the `MaybeNull` attribute (§The-MaybeNull-Attribute), but include a parameter for the specified return value.
+
+#### §The-MemberNotNull-Attribute The MemberNotNull attribute
+
+Specifies that the given member won't be `null` when the method returns.
+
+> *Example*: The compiler may analyze constructors and field initializers to make sure that all non-nullable reference fields have been initialized before each constructor returns. However, the compiler doesn't track field assignments through called helper methods. The `MemberNotNull` attribute specifies the fields that are initialized to a non-null value in that method. For example, consider the following example:
```
jskeet:
```suggestion
> *Example*: The compiler may analyze constructors and field initializers to make sure that all non-nullable reference fields have been initialized before each constructor returns. However, the compiler doesn't track field assignments through called helper methods. The `MemberNotNull` attribute specifies the fields that are initialized to a non-null value in that method. Consider the following example:
```

(Just to avoid the repetition of "example")

jskeet (approved):
One more suggested change, but basically fine. (Thanks!)

standard/attributes.md
```diff
@@ -1028,15 +1028,15 @@ Specifies that a non-nullable return value may be null.
...
+Specifies that a non-nullable argument may be `null` when the method returns the specified `bool` value. This is similar to the `MaybeNull` attribute (§The-MaybeNull-Attribute), but includes a parameter for the specified return value.
 
 #### §The-MemberNotNull-Attribute The MemberNotNull attribute
 
 Specifies that the given member won't be `null` when the method returns.
 
-> *Example*: The compiler may analyze constructors and field initializers to make sure that all non-nullable reference fields have been initialized before each constructor returns. However, the compiler doesn't track field assignments through called helper methods. The `MemberNotNull` attribute specifies the fields that are initialized to a non-null value in that method. For example, consider the following example:
+> *Example*: A helper method adds the `MemberNotNull` attribute to specify any fields that are assigned to a non-null value in that method. A compiler that analyzes constructors to make sure that all non-nullable reference fields have been initialized uses this attribute to determine that fields have been set in those helper methods. Consider the following example:
```
Nigel-Ecma: Maybe:

```suggestion
> *Example*: A helper method may include the `MemberNotNull` attribute to list any fields that are assigned to a non-null value in that method. A compiler that analyzes constructors to determine whether all non-nullable reference fields have been initialized may then use this attribute to discover which fields have been set by those helper methods. Consider the following example:
```

standard/attributes.md
```diff
@@ -1096,15 +1096,15 @@ Specifies that a nullable return value will never be `null`.
 
 Specifies that a return value isn't `null` if the argument for the specified parameter isn't `null`.
 
-> *Example*: The null state of a return value may depend on the null state of one or more arguments. Such a method will return a non-null value when certain arguments aren't `null`. To correctly annotate these methods, add the `NotNullIfNotNull` attribute. Consider the following method:
+> *Example*: The null state of a return value could depend on the null state of one or more arguments. Such a method always returns a non-null value when certain arguments aren't `null`. To assist the compiler with analyzing such methods the `NotNullIfNotNull` attribute may be used. Consider the following method:
```
Nigel-Ecma: Is the second sentence a necessary consequence of the first? Maybe:

```suggestion
> *Example*: The null state of a return value could depend on the null state of one or more arguments. To assist the compiler’s analysis when a method always returns a non-null value when certain arguments are not `null` the `NotNullIfNotNull` attribute may be used. Consider the following method:
```

Nigel-Ecma (commented):
A couple more comments

standard/attributes.md
```diff
@@ -858,6 +860,275 @@ For invocations that occur within field or event initializers, the member name u
...
+>         // null check not needed.
+>         _field = containedField;
+>     }
+> 
+>     private bool isInitialized = false;
+>     private object _field;
+> }
+> ```
+>
+> The presence of the attribute helps the compiler in a number of ways. First, the compiler can issue a warning if there's a path where the method can exit without throwing an exception. Second, the compiler can consider any code after a call to that method as unreachable, until an appropriate catch clause is found. Third, the unreachable code won't affect any null states. *end example*
```
BillWagner: This should be reworded. The compiler doesn't use this attribute for definite assignment nor reachability. However, it will only supress any nullability.

standard/attributes.md
```diff
@@ -858,6 +860,277 @@ For invocations that occur within field or event initializers, the member name u
...
+>     }
+> 
+>     private bool isInitialized = false;
+>     private object _field;
+> }
+> ```
+>
+> The presence of the attribute helps the compiler in a number of ways. First, the compiler can issue a warning if there's a path where the method can exit without throwing an exception. Second, the compiler can suppress nullable warnings in any code after a call to that method, until an appropriate catch clause is found. Third, the unreachable code won't affect any null states.
+>
+> The compiler can't change reachability (§13.2) or definite assignment (§9.4) analysis based on the presence of this attribute. A compiler can use this attribute only to impact nullability warnings. *end example*
```
gafter:
```suggestion
> The attribute does not change reachability (§13.2) or definite assignment (§9.4) analysis based on the presence of this attribute. It is used only to impact nullability warnings. *end example*
```

standard/attributes.md
```diff
@@ -858,6 +860,277 @@ For invocations that occur within field or event initializers, the member name u
...
+#### §The-MaybeNull-Attribute The MaybeNull attribute
+
+Specifies that a non-nullable return value may be null.
+
+> *Example*: Consider the following generic method:
+>
+> <!-- Example: {template:"code-in-class-lib-without-using", name:"MaybeNull1Attribute", replaceEllipsis:true, customEllipsisReplacements: ["return default;"]} -->
+> ```csharp
+> #nullable enable
+> public T? Find<T>(IEnumerable<T> sequence, Func<T, bool> predicate) { ... }
```
jnm2, line 1017: Should this be valid C# 8? `T?` is only allowed in C# 9+ when there is no class or struct constraint.

> ❌ CS8627 A nullable type parameter must be known to be a value type or non-nullable reference type unless language version '9.0' or greater is used. Consider changing the language version or adding a 'class', 'struct', or type constraint.

BillWagner: The suggestion for this is to use the same declaration as line 1026, but explain that `Find` could return a `null` value.

standard/attributes.md
```diff
@@ -858,6 +860,277 @@ For invocations that occur within field or event initializers, the member name u
...
+> ```
+>
+> If `T` is replaced by `string`, `T?` becomes a nullable annotation. If `T` is replaced by `int`, `T?` becomes an `int?`. When `Find` searches an `IEnumerable<string>`, the default return is `null`, but when `Find` searches an `IEnumerable<int>`, the default return is 0. As such, specifying the return type as `T?` isn’t appropriate. However, adding this attribute solves the problem:
```
jnm2: The example doesn't compile in 8, but in 9 where it does, `T?` does not become `int?`; it stays as `int`. It can be thought of as "defaultable" rather than "nullable" in this context, when unconstrained.

gafter:
```suggestion
> The idea of this code is that if `T` is replaced by `string`, `T?` becomes a nullable annotation. However, this code is not legal because `T` is not constrained to be a reference type. Adding this attribute solves the problem:
```

standard/attributes.md
```diff
@@ -858,6 +860,277 @@ For invocations that occur within field or event initializers, the member name u
...
+
+#### §The-MemberNotNullWhen-Attribute The MemberNotNullWhen attribute
+
+Specifies that the listed member won't be `null` when the method returns the specified `bool` value.
+
+> *Example*: This attribute is like `MemberNotNull` (§The-MemberNotNull-Attribute) except that `MemberNotNullWhen` takes a `bool` argument. `MemberNotNullWhen` is intended for use in situations in which a helper method returns a `bool` indicating whether it initialized fields. *end example*
+
+#### §The-NotNull-Attribute The NotNull attribute
+
+Specifies that a nullable return value will never be `null`.
```
jnm2: Should the postconditional nature be emphasized?
```suggestion
Specifies that a nullable value will never be `null` if the method returns (rather than throwing).
```

PR #1192, opened 2024-10-24
Add examples for nullability and null states

### LINKED ISSUE #1094: [Nullable Reference Types] Provide informative notes and examples on diagnostics
We likely need some informative examples for static null state flow analysis and the diagnostics that might be produced.

As we complete the normative information, we will have a better idea what examples or notes will help readers understand the feature, its benefits, and the diagnostics.

BillWagner:
Fixes #1094 

In the section on nullability and null states, add examples to the notes and text to illustrate possible null analysis results.

### DIFF
--- a/standard/types.md
+++ b/standard/types.md
@@ -863,8 +863,98 @@ The ***default null state*** of an expression is determined by its type, and the
 
 A diagnostic can be produced when a variable ([§9.2.1](variables.md#921-general)) of a non-nullable reference type is initialized or assigned to an expression that is maybe null when that variable is declared in text where the annotation flag is enabled.
 
+> *Example*: Consider the following method where a parameter is nullable and that value is assigned to a non-nullable type:
+>
+> <!-- Example: {template:"code-in-class-lib", name:"NullableAssignment"} -->
+> ```csharp
+> #nullable enable
+> public class C
+> {
+>     public void M(string? p)
+>     {
+>         // Assignment to maybe null
+>         string s = p;
+>     }
+> }
+> ```
+>
+> The compiler may issue a warning where the parameter that might be null is assigned to a variable that should not be null. If the parameter is null-checked before assignment, the compiler won't issue a warning:
+>
+> <!-- Example: {template:"code-in-class-lib", name:"NullChecked"} -->
+> ```csharp
+> #nullable enable
+> public class C
+> {
+>     public void M(string? p)
+>     {
+>         string s = p ?? ""; // null check
+>     }
+> }
+> ```
+>
+> *end example*
+
 The compiler can update the null state of a variable as part of its analysis.
 
-> *Note*: The compiler can treat a property ([§15.7](classes.md#157-properties)) as either a variable with state, or as independent get and set accessors ([§15.7.3](classes.md#1573-accessors)). In other words, a compiler can choose if writing to a property changes the null state of reading the property.
+> *Example*: The compiler may choose to update the state based on any statements in your program:
+>
+> <!-- Example: {template:"code-in-class-lib", name:"UpdateStates"} -->
+> ```csharp
+> #nullable enable
+> public void M(string? p)
+> {
+>     // p is maybe-null
+>     int length = p.Length;
+>
+>     // p is not null.
+>     string s = p;
+> 
+>     if (s != null)
+>     {
+>         int l2 = s.Length;
+>     }
+>     // s is maybe null
+>     int l3 = s.Length;
+> }
+> ```
+>
+> In the previous example, the compiler may decide that after the statement `int length = p.Length;`, the null-state of `p` is not-null. If it were null, that statement would have thrown a `NullReferenceException`. Later in the method, the code checks that `s` is not a null reference. The null-state of `s` can change to maybe null after the null-checked block closes. The compiler can infer that `s` is maybe null because the code was written to assume that it might have been null. *end example*
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+> *Example*: The compiler can treat a property ([§15.7](classes.md#157-properties)) as either a variable with state, or as independent get and set accessors ([§15.7.3](classes.md#1573-accessors)). In other words, a compiler can choose if writing to a property changes the null state of reading the property, or if reading a property changes the null state of that property.
+>
+> <!-- Example: {template:"standalone-console", name:"VolatileFields", inferOutput:true} -->
+> ```csharp
+> class Test
+> {
+>     private string? _field;
+>     public string? DisappearingProperty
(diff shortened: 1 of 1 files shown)


standard/types.md
```diff
@@ -863,8 +863,97 @@ The ***default null state*** of an expression is determined by its type, and the
...
+> *Example*: Consider the following method where a parameter is nullable and that value is assigned to a non-nullable type:
+>
+> <!-- Example: {template:"code-in-class-lib", name:"NullableAssignment"} -->
+> ```csharp
+> #nullable enable
+> public class C
+> {
+>     public void M(string? p)
+>     {
+>         // Assignment to maybe null
```
jskeet: Possibly expand to "Assignment of maybe null value to non-nullable variable"?

standard/types.md
```diff
@@ -863,8 +863,97 @@ The ***default null state*** of an expression is determined by its type, and the
...
+> The compiler may issue a warning where the parameter that might be null is assigned to a variable that should not be null. If the parameter is null-checked before assignment, the compiler won't issue a warning:
+>
+> <!-- Example: {template:"code-in-class-lib", name:"NullChecked"} -->
+> ```csharp
+> #nullable enable
+> public class C
+> {
+>     public void M(string? p)
+>     {
+>         string s = p ?? ""; // null check
```
jskeet: I don't regard null coalescing in the same way as a null *check* - I'd expected to see code such as:

```csharp
if (p is not null)
{
    string s = p;
    // Use s
}
```

standard/types.md
```diff
@@ -863,8 +863,97 @@ The ***default null state*** of an expression is determined by its type, and the
...
+>     // s is maybe null
+>     int l3 = s.Length;
+> }
+> ```
+>
+> In the previous example, the compiler may decide that after the statement `int length = p.Length;`, the null-state of `p` is not-null. If it were null, that statement would have thrown a `NullReferenceException`. Later in the method, the code checks that `s` is not a null reference. The null-state of `s` can change to maybe null after the null-checked block closes. The compiler can infer that `s` is maybe null because the code was written to assume that it might have been null. *end example*
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+> *Example*: The compiler can treat a property ([§15.7](classes.md#157-properties)) as either a variable with state, or as independent get and set accessors ([§15.7.3](classes.md#1573-accessors)). In other words, a compiler can choose if writing to a property changes the null state of reading the property, or if reading a property changes the null state of that property.
```
jskeet: Possibly change each "if" to "whether"?

jskeet (commented):
Supportive of the general approach, just a few suggestions.

standard/types.md
BillWagner: The `is not` pattern isn't available in 8, but I agree with the general principal.

standard/types.md
```diff
@@ -887,7 +887,11 @@ A diagnostic can be produced when a variable ([§9.2.1](variables.md#921-general
 > {
 >     public void M(string? p)
 >     {
->         string s = p ?? ""; // null check
+>         if (p is not null)
```
jskeet: Happy for this to be `!= null` given the lack of `is not null` for C# 8.

standard/types.md
```diff
@@ -863,8 +863,101 @@ The ***default null state*** of an expression is determined by its type, and the
...
+> {
+>     public void M(string? p)
+>     {
+>         // Assignment of maybe null value to non-nullable variable
+>         string s = p;
+>     }
+> }
+> ```
+>
+> The compiler may issue a warning where the parameter that might be null is assigned to a variable that should not be null. If the parameter is null-checked before assignment, the compiler won't issue a warning:
```
jskeet: "won't" is strong. Perhaps "may use that in its nullable state analysis and not issue a warning" or similar?

jskeet (approved):
One aspect that we should probably discuss in the meeting.

standard/types.md
```suggestion
> The compiler may issue a warning where the parameter that might be null is assigned to a variable that should not be null. If the parameter is null-checked before assignment, the compiler may use that in its nullable state analysis and not issue a warning:
```

standard/types.md
```diff
@@ -863,8 +863,101 @@ The ***default null state*** of an expression is determined by its type, and the
...
+>     if (s != null)
+>     {
+>         int l2 = s.Length;
+>     }
+>     // s is maybe null
+>     int l3 = s.Length;
+> }
+> ```
+>
+> In the previous example, the compiler may decide that after the statement `int length = p.Length;`, the null-state of `p` is not-null. If it were null, that statement would have thrown a `NullReferenceException`. Later in the method, the code checks that `s` is not a null reference. The null-state of `s` can change to maybe null after the null-checked block closes. The compiler can infer that `s` is maybe null because the code was written to assume that it might have been null. *end example*
```
gafter:
```suggestion
> In the previous example, the compiler may decide that after the statement `int length = p.Length;`, the null-state of `p` is not-null. If it were null, that statement would have thrown a `NullReferenceException`. This is similar to the behavior if the code had been preceded by `if (p == null) throw NullReferenceException();` except that the code as written may produce a warning, the purpose of which is to warn that an exception may be thrown implicitly.

Later in the method, the code checks that `s` is not a null reference. The null-state of `s` can change to maybe null after the null-checked block closes. The compiler can infer that `s` is maybe null because the code was written to assume that it might have been null. Generally, when the code contains a null check, the compiler may infer that the value might have been null.*end example*
```

standard/types.md
```diff
@@ -863,8 +863,101 @@ The ***default null state*** of an expression is determined by its type, and the
...
+<!-- markdownlint-disable MD028 -->
+
```
BillWagner, line 929: There should be another example that shows how null-checks change the null state. Each of these cause the null state of `s` to be changed to "maybe null":

```csharp
_ = s == null;
_ = s?.Length;
_ = s?[4];
```

standard/types.md
```diff
@@ -863,8 +863,101 @@ The ***default null state*** of an expression is determined by its type, and the
...
+>         var t = new Test();
+>         if (t.DisappearingProperty != null)
+>         {
+>             int len = t.DisappearingProperty.Length;
+>         }
+>     }
+> }
+> ```
+>
+> In the previous example, the backing field for the `DisappearingProperty` is set to null when it is read. However, a compiler may assume that reading a property doesn't change the null state of that expression. *end example*
```
BillWagner, line 964: Another example should include checking the `x.y.z`. decides that `x` and `y` and `z` are all non-null.

BillWagner: Add examples use field-like events, and auto-properties. (So they have a backing field, or a compiler generated backing field)

Issue #1190, opened 2024-10-21
Null forgiving operator grammar issues


jskeet:
The approach to specifying the null forgiving operator to prevent `x!!!` and the like causes some problems, and might be better dealt with as "make it valid in the grammar, but prohibit it in semantic text".

TG2: see Nigel's email about "null_forgiving_expression is either wrong or a lot of work for no gain" for more details.

jskeet:
We agree with the approach. @Nigel-Ecma will create a PR, but may ask for suggestions for how we express "even if parenthesized, it's not valid" (e.g. `((x!))!`).

Nigel-Ecma:
Addressed by #1195

PR #1195, opened 2024-10-31
Null-forgiving operator

Nigel-Ecma:
Null-forgiving operator

- Fix grammar
- Specify semantic restrictions
- Add to precedence table
- Add to MLR group

Addresses issue #1190

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -147,7 +147,7 @@ The precedence of an operator is established by the definition of its associated
 >
 > |  **Subclause**      | **Category**                     | **Operators**                                          |
 > |  -----------------  | -------------------------------  | -------------------------------------------------------|
-> |  [§12.8](expressions.md#128-primary-expressions)              | Primary                          | `x.y` `x?.y` `f(x)` `a[x]` `a?[x]` `x++` `x--` `new` `typeof` `default` `checked` `unchecked` `delegate` `stackalloc`  |
+> |  [§12.8](expressions.md#128-primary-expressions)              | Primary                          | `x.y` `x?.y` `f(x)` `a[x]` `a?[x]` `x++` `x--` `x!` `new` `typeof` `default` `checked` `unchecked` `delegate` `stackalloc`  |
 > |  [§12.9](expressions.md#129-unary-operators)              | Unary                            | `+` `-` `!` `~` `++x` `--x` `(T)x` `await x` |
 > |  [§12.10](expressions.md#1210-arithmetic-operators)              | Multiplicative                   | `*` `/` `%` |
 > |  [§12.10](expressions.md#1210-arithmetic-operators)              | Additive                         | `+` `-` |
@@ -1273,7 +1273,6 @@ Primary expressions include the simplest forms of expressions.
 primary_expression
     : primary_no_array_creation_expression
     | array_creation_expression
-    | null_forgiving_expression
     ;
 
 primary_no_array_creation_expression
@@ -1291,6 +1290,7 @@ primary_no_array_creation_expression
     | base_access
     | post_increment_expression
     | post_decrement_expression
+    | null_forgiving_expression
     | object_creation_expression
     | delegate_creation_expression
     | anonymous_object_creation_expression
@@ -1307,7 +1307,7 @@ primary_no_array_creation_expression
     ;
 ```
 
-> *Note*: These grammar rules are not ANTLR-ready as they are part of a set of mutually left-recursive rules (`primary_expression`, `primary_no_array_creation_expression`, `member_access`, `invocation_expression`, `element_access`, `post_increment_expression`, `post_decrement_expression`, `pointer_member_access` and `pointer_element_access`) which ANTLR does not handle. Standard techniques can be used to transform the grammar to remove the mutual left-recursion. This has not been done as not all parsing strategies require it (e.g. an LALR parser would not) and doing so would obfuscate the structure and description. *end note*
+> *Note*: These grammar rules are not ANTLR-ready as they are part of a set of mutually left-recursive rules (`primary_expression`, `primary_no_array_creation_expression`, `member_access`, `invocation_expression`, `element_access`, `post_increment_expression`, `post_decrement_expression`, `null_forgiving_expression`, `pointer_member_access` and `pointer_element_access`) which ANTLR does not handle. Standard techniques can be used to transform the grammar to remove the mutual left-recursion. This has not been done as not all parsing strategies require it (e.g. an LALR parser would not) and doing so would obfuscate the structure and description. *end note*
 
 *pointer_member_access* ([§23.6.3](unsafe-code.md#2363-pointer-member-access)) and *pointer_element_access* ([§23.6.4](unsafe-code.md#2364-pointer-element-access)) are only available in unsafe code ([§23](unsafe-code.md#23-unsafe-code)).
 
@@ -1824,17 +1824,28 @@ A *null_conditional_projection_initializer* is a restriction of *null_conditiona
 
 ### 12.8.9 Null-forgiving expressions
 
-This operator sets the null state ([§8.9.5](types.md#895-nullabilities-and-null-states)) of the operand to “not null”.
+The null-forgiving operator sets the null state ([§8.9.5](types.md#895-nullabilities-and-null-states)) of the operand to “not null”.
 
 ```ANTLR
 null_forgiving_expression
-    : primary_no_array_creation_expression suppression
+    : primary_expression null_forgiving_operator
     ;
 
-suppression
+null_forgiving_operator
     : '!'
     ;
 ```
+
+The *primary_expression* must not be known to have a value type.
+
+It is an error to apply the null-forgiving operator more than once to the same expression, intervening parenetheses notwithstanding.
+
+> *Example*: the following are all invalid:
+>
+> ```csharp
+> var p = q!!;         // error: cannot apply the null_forgiving_operator more than once
+> var s = ( ( m(t) ! ) )! // error: null_forgiving_operator applied twice to m(t)
+> ```
 
 This operator has no runtime effect; it evaluates to the result of its operand, and that result retains that operand’s classification.
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1843,7 +1843,7 @@ It is an error to apply the null-forgiving operator more than once to the same e
 > *Example*: the following are all invalid:
 >
 > ```csharp
-> var p = q!!;         // error: cannot apply the null_forgiving_operator more than once
+> var p = q!!;            // error: cannot apply the null_forgiving_operator more than once
 > var s = ( ( m(t) ! ) )! // error: null_forgiving_operator applied twice to m(t)
 > ```
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1848,7 +1848,7 @@ It is a compile-time error to apply the null-forgiving operator more than once t
 > *Example*: the following are all invalid:
 >
 > ```csharp
-> var p = q!!;            // error: cannot apply the null_forgiving_operator more than once
+> var p = q!!;            // error: applying null_forgiving_operator more than once
 > var s = ( ( m(t) ! ) )! // error: null_forgiving_operator applied twice to m(t)
 > ```
 >
@@ -1959,8 +1959,9 @@ invocation_expression
     ;
 ```
 
-<!-- [ToDo] C#9’s function pointers are also excluded, as the following restriction is stated
-    in terms of what is included the text will probably be fine but this will need to be confirmed
+<!--
+[ToDo] C#9’s function pointers are also excluded, as the following restriction is stated
+in terms of what is included the text will probably be fine but this will need to be confirmed
 -->
 The *primary_expression* may be a *null_forgiving_expression* if and only if it has a *delegate_type*.
 
@@ -2167,8 +2168,9 @@ null_conditional_invocation_expression
     ;
 ```
 
-<!-- [ToDo] C#9’s function pointers are also excluded, as the following restriction is stated
-    in terms of what is included the text will probably be fine but this will need to be confirmed
+<!--
+[ToDo] C#9’s function pointers are also excluded, as the following restriction is stated
+in terms of what is included the text will probably be fine but this will need to be confirmed
 -->
 The optional *null_forgiving_operator* may be included if and only if the *null_conditional_member_access* or
 *null_conditional_element_access* has a *delegate_type*.
@@ -2256,7 +2258,7 @@ Depending on the context in which it is used, an indexer access causes invocatio
 
 ### 12.8.13 Null Conditional Element Access
 
-A *null_conditional_element_access* consists of a *primary_no_array_creation_expression* followed by the two tokens “`?`” and “`[`”, followed by an *argument_list*, followed by a “`]`” token, followed by zero or more *dependent_access*es any of which can be preceeded by a *null_forgiving_operator*.
+A *null_conditional_element_access* consists of a *primary_no_array_creation_expression* followed by the two tokens “`?`” and “`[`”, followed by an *argument_list*, followed by a “`]`” token, followed by zero or more *dependent_access*es any of which can be preceded by a *null_forgiving_operator*.
 
 ```ANTLR
 null_conditional_element_access
@@ -6774,7 +6776,7 @@ Only the following constructs are permitted in constant expressions:
 - The predefined `+`, `-`, `!` (logical negation) and `~` unary operators.
 - The predefined `+`, `-`, `*`, `/`, `%`, `<<`, `>>`, `&`, `|`, `^`, `&&`, `||`, `==`, `!=`, `<`, `>`, `<=`, and `>=` binary operators.
 - The `?:` conditional operator.
-- The `!` null-forgiving operator ([§12.9.4](expressions.md#1289-null-forgiving-expressions)).
+- The `!` null-forgiving operator ([§12.8.9](expressions.md#1289-null-forgiving-expressions)).
 - `sizeof` expressions, provided the unmanaged-type is one of the types specified in [§23.6.9](unsafe-code.md#2369-the-sizeof-operator) for which `sizeof` returns a constant value.
 - Default value expressions, provided the type is one of the types listed above.
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -182,10 +182,10 @@ All unary and binary operators have predefined implementations. In addition, use
 
 The ***overloadable unary operators*** are:
 
-`+  -  ! `(logical negation only)` ~  ++  --  true  false`
+`+  -  !` (logical negation only) `~  ++  --  true  false`
 
 > *Note*: Although `true` and `false` are not used explicitly in expressions (and therefore are not included in the precedence table in [§12.4.2](expressions.md#1242-operator-precedence-and-associativity)), they are considered operators because they are invoked in several expression contexts: Boolean expressions ([§12.24](expressions.md#1224-boolean-expressions)) and expressions involving the conditional ([§12.18](expressions.md#1218-conditional-operator)) and conditional logical operators ([§12.14](expressions.md#1214-conditional-logical-operators)). *end note*
-
+>
 > *Note*: The null-forgiving operator (postfix `!`, [§12.8.9](expressions.md#1289-null-forgiving-expressions)) is not an overloadable operator. *end note*
 
 The ***overloadable binary operators*** are:
@@ -1827,7 +1827,6 @@ A *null_conditional_projection_initializer* is a restriction of *null_conditiona
 A null-forgiving expression’s value, type, classification ([§12.2](expressions.md#122-expression-classifications))
 and safe-context ([§16.4.12](structs.md#16412-safe-context-constraint)) is the value, type, classification and safe-context of its *primary_expression*.
 
-
 ```ANTLR
 null_forgiving_expression
     : primary_expression null_forgiving_operator
@@ -1878,12 +1877,12 @@ message and informs any ongoing analysis.
 >     person != null && person.Name != null;
 > ```
 >
-> If `IsValid` returns `true`, `p` can safely be dereferenced to access its `Name` property, and the “dereferencing of a possibly null value” warning can be suppressed using `!`. 
+> If `IsValid` returns `true`, `p` can safely be dereferenced to access its `Name` property, and the “dereferencing of a possibly null value” warning can be suppressed using `!`.
 >
 > *end example*
-
+>
 > *Example:* The null-forgiving operator should be used with caution, consider:
-
+>
 > ```csharp
 > #nullable enable
 > int B(int? x)
@@ -1896,6 +1895,8 @@ message and informs any ongoing analysis.
 > Here the null-forgiving operator is applied to a value type and quashes any warning on
 > `x`. However if `x` is `null` at runtime an exception will be thrown as `null` cannot
 > be cast to `int`.
+>
+> *end example*
 
 #### 12.8.9.2 Overriding other null analysis warnings
 
@@ -1926,7 +1927,7 @@ to issue warnings and may also modify its further analysis.
 > Method `M` passes the variable `s`, of type `string`, as `Assign`’s output parameter, the
 > compiler used issues a warning as `s` is not a nullable variable. Given that `Assign`’s
 > second argument cannot be null the null-forgiving operator is used to quash the warning.
-
+>
 > *end example*
 
 **End of conditionally normative text.**
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -185,7 +185,9 @@ The ***overloadable unary operators*** are:
 `+  -  !` (logical negation only) `~  ++  --  true  false`
 
 > *Note*: Although `true` and `false` are not used explicitly in expressions (and therefore are not included in the precedence table in [§12.4.2](expressions.md#1242-operator-precedence-and-associativity)), they are considered operators because they are invoked in several expression contexts: Boolean expressions ([§12.24](expressions.md#1224-boolean-expressions)) and expressions involving the conditional ([§12.18](expressions.md#1218-conditional-operator)) and conditional logical operators ([§12.14](expressions.md#1214-conditional-logical-operators)). *end note*
->
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
 > *Note*: The null-forgiving operator (postfix `!`, [§12.8.9](expressions.md#1289-null-forgiving-expressions)) is not an overloadable operator. *end note*
 
 The ***overloadable binary operators*** are:
@@ -1824,6 +1826,8 @@ A *null_conditional_projection_initializer* is a restriction of *null_conditiona
 
 ### 12.8.9 Null-forgiving expressions
 
+#### 12.8.9.1 General
+
 A null-forgiving expression’s value, type, classification ([§12.2](expressions.md#122-expression-classifications))
 and safe-context ([§16.4.12](structs.md#16412-safe-context-constraint)) is the value, type, classification and safe-context of its *primary_expression*.
 
@@ -1839,9 +1843,18 @@ null_forgiving_operator
 
 *Note*: The postfix null-forgiving and prefix logical negation operators ([§12.9.4](expressions.md#1294-logical-negation-operator)), while represented by the same lexical token (`!`), are distinct. Only the latter may be overriden ([§15.10](classes.md#1510-operators)), the definition of the null-forgiving operator is fixed. *end note*
 
-<!-- [Placeholder] 12.8.9 A -->
-<!-- [Placeholder] 12.8.9 A End -->
-**The remainder of this subclause, including all of its subclauses, is conditionally normative.**
+It is a compile-time error to apply the null-forgiving operator more than once to the same expression, intervening parentheses notwithstanding.
+
+> *Example*: the following are all invalid:
+>
+> ```csharp
+> var p = q!!;            // error: cannot apply the null_forgiving_operator more than once
+> var s = ( ( m(t) ! ) )! // error: null_forgiving_operator applied twice to m(t)
+> ```
+>
+> *end example*
+
+**The remainder of this subclause and the following sibling subclauses are conditionally normative.**
 
 A compiler which performs static null state analysis ([§8.9.5](types.md#895-nullabilities-and-null-states)) must conform to the following specification.
 
@@ -1850,13 +1863,13 @@ The null-forgiving operator is a compile time pseudo-operation that is used to i
 Applying the null-forgiving operator to an expression for which a compiler’s static null state analysis
 does not produce any warnings is not an error.
 
-#### 12.8.9.1 Overriding a *maybe null* determination
+#### 12.8.9.2 Overriding a “maybe null” determination
 
 Under some circumstances a compiler’s static null state analysis may determine that an expression
-has the null state *maybe null* and issue a warning when other information indicates that the
+has the null state *maybe null* and issue a diagnostic warning when other information indicates that the
 expression cannot be null. Applying the null-forgiving operator to such an expression informs the
-compiler’s static null state analysis that the null state is in *not null*; which both prevents the warning
-message and informs any ongoing analysis.
+compiler’s static null state analysis that the null state is in *not null*; which both prevents the diagnostic
+warning and may inform any ongoing analysis.
 
 > *Example*: Consider the following:
 >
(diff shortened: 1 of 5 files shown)


Nigel-Ecma:
Note: The grammar validation will fail until the validator is updated. The validator is a work-in-progress and there is a mutual dependency between that and this PR. The PR for the validator will follow in the not too distant future…
@BillWagner – I suggest holding off merging this even once approved until a validator PR is good to go as well.

jskeet (approved):
Looks good to me, thanks Nigel.

standard/expressions.md
```diff
@@ -1824,18 +1824,29 @@ A *null_conditional_projection_initializer* is a restriction of *null_conditiona
...
 
-suppression
+null_forgiving_operator
     : '!'
     ;
 ```
 
+The *primary_expression* must not be known to have a value type.
+
+It is an error to apply the null-forgiving operator more than once to the same expression, intervening parenetheses notwithstanding.
```
KalleOlaviNiemitalo:
```suggestion
It is an error to apply the null-forgiving operator more than once to the same expression, intervening parentheses notwithstanding.
```

KalleOlaviNiemitalo:
The wording does not seem to explain the effect of the null-forgiving operator in this case:

```csharp
class C {
    static void M() {
        object obj;
        N(out obj); // warning CS8600: Converting null literal or possible null value to non-nullable type.
        N(out obj!); // no warning
    }
    
    static void N(out object? o) { o = null; }
}
```

standard/expressions.md
```diff
@@ -1824,18 +1824,29 @@ A *null_conditional_projection_initializer* is a restriction of *null_conditiona
...
+    : primary_expression null_forgiving_operator
     ;
 
-suppression
+null_forgiving_operator
     : '!'
     ;
 ```
 
+The *primary_expression* must not be known to have a value type.
```
KalleOlaviNiemitalo: Was this restriction relaxed after C# 8?  Here's what I see at SharpLab:

```csharp
class C {
    static void M() {
        (string, string?) t1 = ("", null);
        (string?, string) t2;

        t2 = t1; // warning CS8619: Nullability of reference types in value of type '(string, string?)' doesn't match target type '(string?, string)'.
        t2 = t1!; // no warning, although (string, string?) is a value type
    }

    static void N() {
        (int, int) t1 = (0, 0);
        (int, int) t2;

        t2 = t1!; // no warning, although (int, int) is a value type
    }

    static void O() {
        int t1 = 0;
        int t2;

        t2 = t1!; // no warning, although int is a value type
    }
}
```

KalleOlaviNiemitalo: Roslyn started allowing `e!` on value types in <https://github.com/dotnet/roslyn/pull/32090>, which is included in tag [Visual-Studio-2019-Version-16.0-Preview-2](https://github.com/dotnet/roslyn/releases/tag/Visual-Studio-2019-Version-16.0-Preview-2).  So Visual Studio 2019 version 16.0 already had this change in its C# 8 support.  [Visual Studio 2019 version 16.0 release notes for C#](https://learn.microsoft.com/en-us/visualstudio/releases/2019/release-notes-v16.0#c)

According to <https://github.com/dotnet/roslyn/issues/57142#issuecomment-947193118> on Oct 20, 2021, "the spec" was also changed to allow `!` on value types.  That change was perhaps <https://github.com/dotnet/csharplang/commit/e8ddd37721a7fbb5aff8a027f57d2cb25137b479> in `proposals/csharp-9.0/nullable-reference-types-specification.md`, making it look like `e!` on value types is only allowed starting from C# 9. But AFAIK no C# 8 compiler was released without this feature.

I suppose the committee still has the authority to omit that feature from the C# 8 standard, even if compilers support it.

standard/expressions.md
Nigel-Ecma: Fixed, thanks @KalleOlaviNiemitalo

standard/expressions.md
```diff
@@ -1824,18 +1824,29 @@ A *null_conditional_projection_initializer* is a restriction of *null_conditiona
...
+
```
BillWagner: We should remove this paragraph. I checked with the team, and it was decided that `!` is allowed on a value type. The current compiler even allows it with C# 8 as the chosen version. I don't think even the C# 8 release had this change (as near as I can find in the history).

```suggestion
```

Nigel-Ecma:
@KalleOlaviNiemitalo & @BillWagner – I only came here to fix the grammar! ;-)

Thanks Kalle for the research and further analysis.

It's clear the original spec was never in a released compiler and so we can go straight to allowing value types.

However as Kalle has shown the implemented operator does more than override (aka “forgive”) a compiler’s static analysis determination that a null might be present – it can be a “null*ability* forgiving operator” and cancel any nullability warnings that a compiler might choose to provide.

I’ve moved the PR to Draft while I determine a suitable spec.

KalleOlaviNiemitalo:
Compiler behaviour for the null-forgiving operator in the left operand of an assignment expression looks complex:

```csharp
class C
{
    void M(
        ref object o,
        ref object p)
    {
        // error CS8598
        (o!) = null;

        // warning CS8600, despite ! operator
        (o!, p!) = (1, null);

        // no warning
        (o!, p!) = this;

        // no warning
        Deconstruct(out o!, out p!);
    }
    
    void Deconstruct(
        out object? o,
        out object? p)
    {
        o = null;
        p = null;
    }
}
```

KalleOlaviNiemitalo:
Null-forgiving `!` would make perfect sense in the left operand of a compound assignment operator; but the compiler apparently doesn't allow that.  The new wording should make sure that the standard doesn't allow that either.

```csharp
public class C
{
    public static void M()
    {
        C? c = null;
        
        // error CS8598
        (c!) += new C();
        
        // no warning
        c = c! + new C();
    }
    
    public static C operator +(C c1, C c2)
    {
        return new C();
    }
}
```

KalleOlaviNiemitalo:
warning CS8620 looks unjustified here; the type of the argument `(p, q)` should be `(object p, object? q)`, not `(object? p, object? q)`.  It seems the compiler makes the assignment to the tuple `(p, q)` change the type of `p` to nullable even before the assignment happens.  I suppose, even if it is a bug, it won't be a violation of the C# 8 standard, as compilers are allowed to output extra diagnostics.

```csharp
public class C
{
    public static void M()
    {
        object p = "";
        object? q = null;
        
        // warning CS8600: Converting null literal or possible null value to non-nullable type.
        // warning CS8620: Argument of type '(object? p, object? q)' cannot be used for parameter 't' of type '(object, object?)' in '(object?, object) C.operator +((object, object?) t, C c)' due to differences in the nullability of reference types.
        (p, q) = (p, q) + new C();     
    }
    
    public static (object?, object) operator +((object, object?) t, C c)
    {
        return (null, "");
    }
}
```

The same happens with a named method instead of an operator:

```csharp
public class C
{
    public static void M()
    {
        object p = "";
        object? q = null;
        
        // warning CS8600: Converting null literal or possible null value to non-nullable type.
        // warning CS8620: Argument of type '(object? p, object? q)' cannot be used for parameter 't' of type '(object, object?)' in '(object?, object) C.N((object, object?) t)' due to differences in the nullability of reference types.
        (p, q) = N((p, q));  
    }
    
    public static (object?, object) N((object, object?) t)
    {
        return (null, "");
    }
}
```

standard/expressions.md
```diff
@@ -1824,21 +1824,39 @@ A *null_conditional_projection_initializer* is a restriction of *null_conditiona
...
+null_forgiving_operator
     : '!'
     ;
 ```
 
-This operator has no runtime effect; it evaluates to the result of its operand, and that result retains that operand’s classification.
+*Note*: The postfix null-forgiving and prefix logical negation operators ([§12.9.4](expressions.md#1294-logical-negation-operator)), while represented by the same lexical token (`!`), are distinct. Only the latter may be overriden ([§15.10](classes.md#1510-operators)), the definition of the null-forgiving operator is fixed. *end note*
+
+<!-- [Placeholder] 12.8.9 A -->
+<!-- [Placeholder] 12.8.9 A End -->
```
Nigel-Ecma: ### [Placeholder] 12.8.9

#### First Alternative

A certain compiler produces a compile time error in the following situation.

---
It is a compile-time error to apply the null-forgiving operator more than once to the same expression, intervening parentheses notwithstanding.

> *Example*: the following are all invalid:
>
> ```csharp
> var p = q!!;            // error: cannot apply the null_forgiving_operator more than once
> var s = ( ( m(t) ! ) )! // error: null_forgiving_operator applied twice to m(t)
> ```

---

#### Second Alternative

---
Remove placeholder

---

standard/expressions.md
```diff
@@ -1875,6 +1944,8 @@ invocation_expression
     ;
 ```
 
+<!-- [Placeholder] 12.8.10.1 -->
+<!-- [Placeholder] 12.8.10.1 End -->
```
Nigel-Ecma: ### Placeholder 12.8.10.1

#### First Alternative

A certain compiler produces a compile time error in the following situation.

---
&lt;!-- [ToDo] C#9’s function pointers are also excluded, as the following restriction is stated in terms of what is included the text will probably be fine but this will need to be confirmed
-->
The *primary_expression* may be a *null_forgiving_expression* if and only if it has a *delegate_type*.

---

(The [ToDo] comment is part of the suggestion.)

#### Second Alternative

---
Remove placeholder

---

standard/expressions.md
```diff
@@ -2073,11 +2144,13 @@ Unlike the syntactically equivalent *null_conditional_member_access* or *null_co
...
 null_conditional_invocation_expression
-    : null_conditional_member_access '(' argument_list? ')'
-    | null_conditional_element_access '(' argument_list? ')'
+    : null_conditional_member_access null_forgiving_operator? '(' argument_list? ')'
+    | null_conditional_element_access null_forgiving_operator? '(' argument_list? ')'
     ;
 ```
 
+<!-- [Placeholder] 12.8.11 -->
+<!-- [Placeholder] 12.8.11 End -->
```
Nigel-Ecma: ### Placeholder 12.8.11

#### First Alternative

A certain compiler produces a compiler time error in the following situation.

---
&lt;!-- [ToDo] C#9’s function pointers are also excluded, as the following restriction is stated  in terms of what is included the text will probably be fine but this will need to be confirmed
-->
The optional *null_forgiving_operator* may be included if and only if the *null_conditional_member_access* or
*null_conditional_element_access* has a *delegate_type*.

---

(As before the [ToDo] comment is part of the suggestion.)

#### Second Alternative

---
Remove placeholder

---

standard/expressions.md
```diff
@@ -1824,21 +1824,39 @@ A *null_conditional_projection_initializer* is a restriction of *null_conditiona
 
 ### 12.8.9 Null-forgiving expressions
 
-This operator sets the null state ([§8.9.5](types.md#895-nullabilities-and-null-states)) of the operand to “not null”.
+A null-forgiving expression’s value, type, classification ([§12.2](expressions.md#122-expression-classifications))
```
KalleOlaviNiemitalo: Because there is 12.8.9.1 too, insert a "General" heading.

standard/expressions.md
```diff
@@ -6377,6 +6454,8 @@ assignment_operator
     ;
 ```
 
+<!-- [Placeholder] 12.21.1 -->
+<!-- [Placeholder] 12.21.1 End -->
```
Nigel-Ecma: ### [Placeholder] 12.21.1

#### First Alternative

A certain compiler produces a compile time error in the following situation.

---
&lt;!-- [ToDo] This restriction also applies to null-coalescing assignment, ??= , which is a
     C# 8 feature not yet spec’ed – once it is need to ensure this restriction applies.
-->
The *unary_expression* left operand of an assignment may not be a *null_forgiving_expression*.

---

(As before the [ToDo] comment is part of the suggestion.)

#### Second Alternative

---
Remove placeholder

---

standard/unsafe-code.md
```diff
@@ -544,6 +544,8 @@ addressof_expression
     ;
 ```
 
+<!-- [Placeholder] 23.6.5 -->
+<!-- [Placeholder] 23.6.5 End -->
```
Nigel-Ecma: ### [Placeholder] 23.6.5

#### First Alternative

A certain compiler produces a compile time error in the following situation.

---
The *unary_expression* operand of an address-of operator may not be a *null_forgiving_expression*.

---

#### Second Alternative

---
Remove placeholder

---

standard/expressions.md
```diff
@@ -1824,21 +1824,39 @@ A *null_conditional_projection_initializer* is a restriction of *null_conditiona
...
+does not produce any warnings is not an error.
+
+#### 12.8.9.1 Overriding a *maybe null* determination
 
-The null-forgiving operator is used to declare that an expression not known to be a value type is not null.
+Under some circumstances a compiler’s static null state analysis may determine that an expression
+has the null state *maybe null* and issue a warning when other information indicates that the
+expression cannot be null. Applying the null-forgiving operator to such an expression informs the
+compiler’s static null state analysis that the null state is in *not null*; which both prevents the warning
+message and informs any ongoing analysis.
```
KalleOlaviNiemitalo: "informs any ongoing analysis" apparently does not include propagating the information back to the variable from which the value originated.  [SharpLab](https://sharplab.io/#v2:CYLg1APgAgzABFATHAwnA3gWAFBzw+KAFjgFkAKASwDsAXAfjgEMBKDHfTuANyYCc4AIzgBeOFTosmAQgDcHLnl4CAxqPE1aU2XAD0uuNQD2cAO79qNAOYK8AX1txHsBCQBy5KAEYADI1bsuIo8/ELqnr5Sco6cynBqYhE+2noG5nyW1FaoAMoAHABsPj6ODth2QA===):

```csharp
public class C {
    public void M(int? a) {
        var b = (int)a!;
        var c = (int)a; // no warning
    }
    
    public void N(string? a) {
        var b = (string)a!;
        var c = (string)a; // warning CS8600
    }
}
```

In `N`, despite `a!` suggesting that `a` is known not to be null, the compiler issues a warning for the subsequent `(string)a` expression.

In `M`, this compiler does not warn about `(int)a`; however, this is apparently not caused by the null-forgiving operator in the preceding `(int)a!` expression, but rather by the compiler noticing that, if `a` were null, then `(int)a!` would throw and the subsequent `(int)a` would not be reached.

I did not check whether the specification already spells out that the null state analysis works this way.

Nigel-Ecma (commented):
In previous discussions on this feature it was known to TG2 that there was one situation where a certain compiler produced a compile time error from a use of the null-forgiving operator, rather than the usual warning.

This seemingly diverged from the design goal that the null-forgiving operator only provides information to a compiler’s optional null state static analysis to effect the analysis and its warnings.

Those discussions led to the decision to include the requirement for the error in the Standard.

Since then a certain compiler’s source code has been reviewed and a total of 5 places where that compiler produces compile time errors from usage of the null-forgiving operator found – none of which appear to be specified in any of the feature design documents.

All 5 places, including the originally known one in §12.8.9, are now marked in this PR with placeholding HTML comments.

This review provides possible text for each placeholder if the particular error is to be included in the Standard and required of all implementations, or the choice of simply removing the placeholder.

Please debate away, either on all 5 errors as a whole or on the individual placeholder comments. TG2 will review the discussion before making final determinations.

standard/expressions.md
Nigel-Ecma: I thought about this too but left it as due to the conditionally normative bit on line 1844 (“remainder of this subclause, including all of its subclauses”). This is probably one for @RexJaeschke to wordsmith.

standard/expressions.md
Nigel-Ecma: @KalleOlaviNiemitalo – The specification leaves how the analysis works to the implementation, the null-forgiving operator “informs” the compiler that the user knows something but the compiler does not need to take heed… (well in `e!` the compiler is meant to change the determined null state, but remember that is for that specific `e`).

I’m not overly happy with the particular “informs any ongoing analysis” wording, and I tried and rejected other wordings, but it seems accurate – it only informs not, say, drives… Better wording suggestions welcome!

standard/types.md
```diff
@@ -720,12 +720,12 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
 
 There are two forms of nullability for reference types:
 
 - *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
-- *non-nullable*” A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
+- *non-nullable*: A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.
 
 > *Note:* The types `R` and `R?` are represented by the same underlying type, `R`. A variable of that underlying type can either contain a reference to an object or be the value `null`, which indicates “no reference.” *end note*
```
Nigel-Ecma, line 730: Some small wording corrections were made here; but between there being two distinct types, one just being the other with an *attribute*, the two being represented the same, and both capable of holding *exactly* the same set of values, it is less than clear… Two types? One and an attribute? Open invitation to have a go at wordsmithing!

standard/types.md
```diff
@@ -720,12 +720,12 @@ An *unmanaged_type* is any type that isn’t a *reference_type*, a *type_paramet
...
+A *nullable reference type* is denoted by appending a *nullable_type_attribute* (`?`) to a non-nullable reference type. There is no semantic difference between a non-nullable reference type and its corresponding nullable type, both can either be a reference to an object or `null`. The presence or absence of the *nullable_type_attribute* declares whether an expression is intended to permit null values or not. A compiler may provide diagnostics when an expression is not used according to that intent. The null state of an expression is defined in [§8.9.5](types.md#895-nullabilities-and-null-states). An identity conversion exists among a nullable reference type and its corresponding non-nullable reference type ([§10.2.2](conversions.md#1022-identity-conversion)).
```
KalleOlaviNiemitalo: _Moved to <https://github.com/dotnet/csharpstandard/pull/1178#discussion_r1836917012>_

The term "attribute" has a different meaning in [22 Attributes](https://github.com/dotnet/csharpstandard/blob/94c967e7c4cb49d9c4c1eec887b65793a2368c11/standard/attributes.md#22-attributes) so using it as part of *nullable_type_attribute* feels a bit unfortunate.  Not sure what would be a better term, though.  "Facet" as used in W3C XML Schema Definition Language (XSD) comes to mind but that doesn't feel right either.  Perhaps "specifier" as in *rank_specifier*?

Nigel-Ecma, line 723: Note: *nullable_type_attribute* is not in the current grammar, it is defined in PR #1178

Nigel-Ecma: @KalleOlaviNiemitalo – I forgot to add a note  yesterday that *nullabIe_type_attribute* is introduced in PR #1178; can you move your comment there?

Nigel-Ecma:
General question: Anybody think an example explaining `null!` would be good and maybe also wish to suggest one, and where to include it, if so?

standard/types.md
```diff
@@ -787,7 +787,7 @@ When the nullable context is ***annotations***:
 
 - For any reference type `T`, the annotation `?` in `T?` indicates that `T?` a nullable type, whereas the unannotated `T` is non-nullable.
 - No diagnostic warnings related to nullability are generated.
-- The null-forgiving operator `!` ([§12.8.9](expressions.md#1289-null-forgiving-expressions)) sets the null state of its operand to *not null*.
+- The null-forgiving operator `!` ([§12.8.9](expressions.md#1289-null-forgiving-expressions)) may alter the analyzed null state of its operand and what compile time informative messages are produced.
```
BillWagner: We've used "diagnostic messages" elsewhere to discuss the nullability analysis. I think we should be consistent.

standard/expressions.md
```diff
@@ -182,11 +182,11 @@ All unary and binary operators have predefined implementations. In addition, use
 
 The ***overloadable unary operators*** are:
 
-```csharp
-+  -  !  ~  ++  --  true  false
-```
+`+  -  !` (logical negation only) `~  ++  --  true  false`
 
 > *Note*: Although `true` and `false` are not used explicitly in expressions (and therefore are not included in the precedence table in [§12.4.2](expressions.md#1242-operator-precedence-and-associativity)), they are considered operators because they are invoked in several expression contexts: Boolean expressions ([§12.24](expressions.md#1224-boolean-expressions)) and expressions involving the conditional ([§12.18](expressions.md#1218-conditional-operator)) and conditional logical operators ([§12.14](expressions.md#1214-conditional-logical-operators)). *end note*
+>
```
BillWagner: This will produce the format we want in the final Word output without the markdown lint warnings:

```suggestion
<!-- markdownlint-disable MD028 -->

<!-- markdownlint-enable MD028 -->
```

standard/expressions.md
```diff
@@ -295,15 +295,15 @@ When overload resolution rules ([§12.6.4](expressions.md#1264-overload-resoluti
 
 **This subclause is informative.**
 
-Unary numeric promotion occurs for the operands of the predefined `+`, `–`, and `~` unary operators. Unary numeric promotion simply consists of converting operands of type `sbyte`, `byte`, `short`, `ushort`, or `char` to type `int`. Additionally, for the unary – operator, unary numeric promotion converts operands of type `uint` to type `long`.
+Unary numeric promotion occurs for the operands of the predefined `+`, `-`, and `~` unary operators. Unary numeric promotion simply consists of converting operands of type `sbyte`, `byte`, `short`, `ushort`, or `char` to type `int`. Additionally, for the unary – operator, unary numeric promotion converts operands of type `uint` to type `long`.
```
BillWagner, line 300: Should the 2nd '-' also be the shorter character?

standard/expressions.md
BillWagner: Here I'd prefer the first alternative. Not specifying that would limit later grammar.

standard/expressions.md
```diff
@@ -1824,21 +1824,39 @@ A *null_conditional_projection_initializer* is a restriction of *null_conditiona
...
+does not produce any warnings is not an error.
+
+#### 12.8.9.1 Overriding a *maybe null* determination
 
-The null-forgiving operator is used to declare that an expression not known to be a value type is not null.
+Under some circumstances a compiler’s static null state analysis may determine that an expression
+has the null state *maybe null* and issue a warning when other information indicates that the
+expression cannot be null. Applying the null-forgiving operator to such an expression informs the
+compiler’s static null state analysis that the null state is in *not null*; which both prevents the warning
+message and informs any ongoing analysis.
```
BillWagner: I can't think of better wording, but I can think of weasel words:

```suggestion
message and may inform any ongoing analysis.
```

standard/expressions.md
BillWagner: @Nigel-Ecma @KalleOlaviNiemitalo 

Here, I think the roslyn compiler assumes certain behavior of `Nullable<T>.operator T`, where it throws when the instance is null.

That is by design, to avoid false positives where possible.

BillWagner (commented):
Overall this looks good. I had a few comments.

One other question for @jskeet:  Typically, we've used hard breaks only for paragraphs. I don't know if the hard returns will create problems for the word converter.

standard/expressions.md
Nigel-Ecma: @BillWagner – No that one is an intentional en-dash (it will have been entered directly or as two dashes and smarten will have fixed it up).

standard/expressions.md
Nigel-Ecma: @BillWagner mentions the known limitation argument.

Now I might be playing the devil’s advocate here, or maybe not 😉, but why does it not apply to *prefix* `!` or to operators such as `~`?

One could also ask why it doesn’t apply to negation as well, the following crazy code is valid (***please*** don't write it in real code!):

> ```
> int m = 0;
> int n = - - - --m;
> ```

which sets `m` to `-1` and `n` to `1`.

Given examples like this the apparent harmlessness of `x!!` stands out – it might earn a warning, but syntactic or semantic rules to prevent it?

Of course having the semantic rule as proposed here is also pretty harmless.

standard/expressions.md
Nigel-Ecma: @BillWagner – I think that is better wording and I’ll do it unless others disagree.

standard/expressions.md
Nigel-Ecma: @BillWagner @KalleOlaviNiemitalo – I would put it a bit more forcefully, “the Roslyn compiles *knows*” – and any compiler could. Whether a compiler analyses the bodies of called methods or uses the defined semantics of standard types etc. is part of the implementation specific latitude it has. Here's to the future VS vs. Rider null analysis competition! 🙂

standard/types.md
Nigel-Ecma: @BillWagner – Thanks, I’ll go through and make sure they’re all the same.

Nigel-Ecma: @BillWagner – A few moments later… The Standard actually uses diagnostic to refer to both errors and warnings, as it is key the null analysis produces warnings I’ll go with *diagnostic warnings* (which is already used) everywhere.

standard/expressions.md
KalleOlaviNiemitalo: > why does it not apply to prefix ! or to operators such as ~?

1. Compatibility; cannot disallow `!!e` after it has been allowed in earlier versions of C# (and in C, where it has been used for ensuring that an `int` treated as a Boolean flag is either 0 or 1).
2. Prefix unary `!` and `-` can be overloaded so that `!!e` is not the same as `e` (not even the same type).

Those concerns don't apply to the null-forgiving operator because it's not used in pre-existing code and cannot be overloaded.

standard/expressions.md
```diff
@@ -1859,9 +1877,60 @@ The null-forgiving operator is used to declare that an expression not known to b
...
+
+#### 12.8.9.2 Overriding other null analysis warnings
 
-The null state ([§8.9.5](types.md#895-nullabilities-and-null-states)) of a *null_forgiving_expression* is “not null.”
+In addition to overriding *maybe null* determinations as above there may be other circumstances
+where it is desired to override a compiler’s static null state analysis determination that an
+expression requires one or more warnings. Applying the
+null-forgiving operator to such an expression requests that the compiler
+does not issue any warnings for the expression. In response a compiler may choose not
+to issue warnings and may also modify its further analysis.
```
KalleOlaviNiemitalo, line 1923: OK, this also covers ref assignment.  [SharpLab](https://sharplab.io/#v2:D4AQzABCBMEMIQN4FgBQENUiALBAsgBQCUSamFUAjAAwD8EAzhALwQB2ArgDbcDc5ShgBOAUwBm1GhGGsZEpnwgB6ZRADuAQ2HsAluwDm8AMoAOAGxUAnIKGy2YyYyWqN2vYZMXrtyvflOAIQuauwA9m46+ga+AL6+vuBQeAByJGToQhCEYQBGAFaiAMYALnQANBAAJgCe7JoAtrpFdKQlcoRcvJVd3MQCmXYKhLX1TUWVeYWlpP6OECUhkR5GcGaWNoN+cvOLKmpaUZ5r3ptZMjsKJcH7HBGHK3Fo8ahAA=):

```csharp
public class C {
    public void M() {
        string? s = null;
        ref string r = ref s; // warning CS8619
        r = ref s; // warning CS8619
        r = ref s!; // no warning
    }
    
    public void N() {
        (object?, dynamic?) t = (null, null);
        ref (dynamic, object) r = ref t; // warning CS8619
        r = ref t; // warning CS8619
        r = ref t!; // no warning
    }
}
```

standard/expressions.md
```diff
@@ -1749,12 +1749,12 @@ In a member access of the form `E.I`, if `E` is a single identifier, and if the
 
 A *null_conditional_member_access* is a conditional version of *member_access* ([§12.8.7](expressions.md#1287-member-access)) and it is a binding time error if the result type is `void`. For a null conditional expression where the result type may be `void` see ([§12.8.11](expressions.md#12811-null-conditional-invocation-expression)).
 
-A *null_conditional_member_access* consists of a *primary_expression* followed by the two tokens “`?`” and “`.`”, followed by an *identifier* with an optional *type_argument_list*, followed by zero or more *dependent_access*es.
+A *null_conditional_member_access* consists of a *primary_expression* followed by the two tokens “`?`” and “`.`”, followed by an *identifier* with an optional *type_argument_list*, followed by zero or more *dependent_access*es any of which can be preceeded by a *null_forgiving_operator*.
```
jskeet, line 1754:
```suggestion
A *null_conditional_member_access* consists of a *primary_expression* followed by the two tokens “`?`” and “`.`”, followed by an *identifier* with an optional *type_argument_list*, followed by zero or more *dependent_access*es any of which can be preceded by a *null_forgiving_operator*.
```

standard/expressions.md
jskeet: +1 for the first alternative.

standard/expressions.md
jskeet: I think the conditionally-normative aspect probably doesn't change our numbering requirements. Let's confirm on the call.

standard/expressions.md
jskeet: I think I'd be okay with this being specified with the first alternative. Let's discuss in the call though.

standard/expressions.md
jskeet: Just to check we're thinking about the same thing:

```csharp
ActionHolder? x = null;
x?.Action!(); // This is valid
x?.Dynamic!(); // This is not (CS8598)

record ActionHolder(Action? Action, dynamic? Dynamic);
```

jskeet: Basically I suspect this should go hand-in-hand with "Placeholder 12.8.10.1" - if we go with the first alternative there, we should go with it here too, and vice versa.

standard/expressions.md
```diff
@@ -3360,14 +3433,16 @@ An *anonymous_method_expression* is one of two ways of defining an anonymous fun
 
 ### 12.9.1 General
 
-The `+`, `-`, `!`, `~`, `++`, `--`, cast, and `await` operators are called the unary operators.
+The `+`, `-`, `!` (logical negation [§12.9.4](expressions.md#1294-logical-negation-operator) only), `~`, `++`, `--`, cast, and `await` operators are called the unary operators.
+
+> *Note*: The postfix null-forgiving operator ([§12.8.9](expressions.md#1289-null-forgiving-expressions)), `!`, due to its compile-time and non-overloadable only nature, is excluded from the above list. *end note*
```
jskeet, line 3462: But it's still a unary operator, right? If someone asked "what are the unary operators in C#" it should still be included. We may not want it in `unary_expression`, but I'm nervous about excluding it completely.

Let's talk about this one more.

standard/expressions.md
jskeet: I don't have an opinion on this one yet - we need to talk about it.

standard/unsafe-code.md
jskeet: I'm fine with the first alternative here.

standard/expressions.md
Nigel-Ecma: @jskeet – it is about the text of the conditionally normative bit, if “remainder of this subclause, including all of its subclauses” is *in* the first subclause then it needs to say something along the lines of the the remainder of this subclause, all its sibling subclauses, and any of the subclauses of these… and be readable! Hence why I suggested it is probably one for @RexJaeschke as our master wordsmith.

Nigel-Ecma:
@BillWagner wrote:

> One other question for @jskeet: Typically, we've used hard breaks only for paragraphs. I don't know if the hard returns will create problems for the word converter.

The word convertor errors are not due to hard breaks. One of the subclause titles has emphasised text in it and the convertor doesn’t like that. It will be removed or replaced with quotes in the next commit.

standard/expressions.md
jskeet: @Nigel-Ecma Ah right, thanks for clearing that up. Yes, that makes sense.

standard/unsafe-code.md
jnm2:
```cs
#nullable enable

unsafe
{
    int e = 4;
    _ = &e!; // Error
    _ = &(e!); // Error
    _ = (&e)!; // No error
}
```

jskeet: (Just for the record: I've been completely convinced by removing this, i.e. alternative 2)

standard/expressions.md
```diff
@@ -3360,14 +3433,16 @@ An *anonymous_method_expression* is one of two ways of defining an anonymous fun
 
 ### 12.9.1 General
 
-The `+`, `-`, `!`, `~`, `++`, `--`, cast, and `await` operators are called the unary operators.
+The `+`, `-`, `!` (logical negation [§12.9.4](expressions.md#1294-logical-negation-operator) only), `~`, `++`, `--`, cast, and `await` operators are called the unary operators.
```
gafter, line 3460: Call them unary expressions here?

standard/expressions.md
Nigel-Ecma: Topic moved to #1215

standard/expressions.md
Nigel-Ecma: Fixed

PR #1204, opened 2024-11-13
Update Foreword for V8

RexJaeschke:
I simply replaced the list of V7 features with that expected for V8.


Nigel-Ecma (approved):
Looks good as far as I know.

PR #1206, opened 2024-11-14
conversions.md: indent final 2-entry bullet 1 level

RexJaeschke:
The 3rd-to-last bullet ends in "as follows:" which sure sounds like what follows should be in a subordinate list. And that is indeed what we had in V5. However, I just noticed that in V6 and subsequent versions, that extra indent level has been removed, which I'm thinking was probably an editorial error introduced during the big conversion from Word to md.

Interestingly, I discovered this as I was researching the claim that in V11, the 2nd-to-last bullet had just been added, but when I looked at the spec, it had been there for many years. So either the V11 MS [What's New in V11 text](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-11#improved-method-group-conversion-to-delegate) is wrong, or we jumped the gun a long time ago by adding text that wasn't yet true.


Nigel-Ecma (approved):
👍

BillWagner (approved):
Great catch @RexJaeschke 

I'll :shipit: now

PR #1209, opened 2024-11-17
Grammar rule *parameter_modifier* disallows valid V7 constructs

RexJaeschke:
[I stumbled on this today while looking at how to spec the addition of  `scoped` to a parameter for a future version.]

The relevant grammar for method parameters in [§15.6.2.1](classes.md#15621-general) is, as follows:

```ANTLR
fixed_parameter
    : attributes? parameter_modifier? type identifier default_argument?
    ;

parameter_modifier
    : parameter_mode_modifier
    | 'this'
    ;

parameter_mode_modifier
    : 'ref'
    | 'out'
    | 'in'
    ;
```

Rule *parameter_modifier* allows *parameter_mode_modifier* or `this`, but not both. However, two combinations of these *are* permitted, as follows:

```csharp
public static class E
{
    public static void F1b(this in int p) { }
    public static void F1d(this ref int p) { }
}
```

This possibility *is* supported by the text in [§15.6.10 Extension methods](classes.md#15610-extension-methods), which states:

> When the first parameter of a method includes the `this` modifier, that method is said to be an ***extension method***. Extension methods shall only be declared in non-generic, non-nested static classes. **The first parameter of an extension method is restricted, as follows:**
>
> - It may only be an **input parameter** if it has a value type
> - It may only be a **reference parameter** if it has a value type or has a generic type constrained to struct
> - It shall not be a pointer type.

I propose the following grammar change:

```ANTLR
parameter_modifier
    : parameter_mode_modifier
    | 'this' parameter_mode_modifier?
    ;
```

Now this new grammar also allows `this out`, which according to my compiler is not permitted [CS8328], and that suggests we also add to the extension bullet list above, the following constraint:

> - It shall not be an **output parameter**.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -2130,7 +2130,7 @@ default_argument
 
 parameter_modifier
     : parameter_mode_modifier
-    | 'this'
+    | 'this' parameter_mode_modifier?
     ;
 
 parameter_mode_modifier
@@ -3056,6 +3056,7 @@ When the first parameter of a method includes the `this` modifier, that method i
 
 - It may only be an input parameter if it has a value type
 - It may only be a reference parameter if it has a value type or has a generic type constrained to struct
+- It shall not be an output parameter.
 - It shall not be a pointer type.
 
 > *Example*: The following is an example of a static class that declares two extension methods:
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -2131,6 +2131,7 @@ default_argument
 parameter_modifier
     : parameter_mode_modifier
     | 'this' parameter_mode_modifier?
+    | parameter_mode_modifier? 'this'
     ;
 
 parameter_mode_modifier
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -3055,8 +3055,8 @@ class Customer
 
 When the first parameter of a method includes the `this` modifier, that method is said to be an ***extension method***. Extension methods shall only be declared in non-generic, non-nested static classes. The first parameter of an extension method is restricted, as follows:
 
-- It may only be an input parameter if it has a value type
-- It may only be a reference parameter if it has a value type or has a generic type constrained to struct
+- It may only be an input parameter if it has a value type.
+- It may only be a reference parameter if it has a value type or has a generic type constrained to struct.
 - It shall not be an output parameter.
 - It shall not be a pointer type.
 


RexJaeschke:
Re my second suggestion, I just found the following in [§15.6.2.1](https://github.com/dotnet/csharpstandard/pull/classes.md#15621-general):

> If the parameter is a `struct` type or a type parameter constrained to a `struct`, the `this` modifier may be combined with either the `ref` or `in` modifier, but not the `out` modifier. Extension methods are further described in [§15.6.10](classes.md#15610-extension-methods). 

That said, having the new bullet I proposed clarifies it at the parameter end of the call in the case of extension methods, which are *not* singled out by the above quote.

standard/classes.md
```diff
@@ -2130,7 +2130,7 @@ default_argument
 
 parameter_modifier
     : parameter_mode_modifier
-    | 'this'
+    | 'this' parameter_mode_modifier?
```
Nigel-Ecma, line 2133: The ordering is flexible:

```suggestion
    | 'this' parameter_mode_modifier?
    | parameter_mode_modifier? 'this'
```

Nigel-Ecma (changes requested):
Requires a change to the grammar, apart from that the *changes* are good.

In the preexisting text I did notice that a reference parameter can be of a generic type constrained to a struct, an input parameter cannot be, and there appears to be no explanation why. The answer can be found in the [Readonly references proposal](https://github.com/dotnet/csharplang/blob/0d924a4345b1ffde14c1998b464b6cb7ea54a70f/proposals/csharp-7.2/readonly-ref.md), should something (normative or informative) be added?

jskeet (commented):
Can we rename `parameter_modifier` to `parameter_modifiers`? It doesn't feel like `this ref` is a single parameter modifier. This is purely a stylistic thing, and I'm happy to be overruled.

Nigel-Ecma:
> Can we rename `parameter_modifier` to `parameter_modifiers`? It doesn't feel like `this ref` is a single parameter modifier. This is purely a stylistic thing, and I'm happy to be overruled.

I'll vote against on this one – not because I think the current naming in this area is good, it isn’t really, but because making it plural won’t make it any better so stick with the status quo.

standard/classes.md
RexJaeschke: My test shows that the order is indeed flexible. Given Nigel's proposal, `this` on its own would be ambiguous, but I think ANTLR chooses the lexical-first rule that matches, right, Nigel?

Nigel-Ecma: @RexJaeschke – Yes, ANTLR resolves such ambiguity by picking the first alternative. We have noted elsewhere where the grammar relies on this, however in this case as the two alternatives represent exactly the same thing I don’t think a note is necessary.

KalleOlaviNiemitalo: It was not originally so flexible, see <https://github.com/dotnet/csharplang/blob/8bd52cca31b40d9720a98629aae9a0cd89bb92ec/meetings/2017/LDM-2017-12-04.md>. But I'm not sure which version of C# got that change.

KalleOlaviNiemitalo: The Roslyn change was <https://github.com/dotnet/roslyn/pull/23643>.  According to <https://github.com/dotnet/csharplang/issues/1022>, this change is in C# 7.3.  So OK for the ECMA C# 7 spec.

standard/classes.md
```diff
@@ -3056,6 +3056,7 @@ When the first parameter of a method includes the `this` modifier, that method i
 
 - It may only be an input parameter if it has a value type
 - It may only be a reference parameter if it has a value type or has a generic type constrained to struct
+- It shall not be an output parameter.
```
jskeet, line 3060: While we're here, is it worth being consistent within this bullet list as to whether we end with a period or not?

jskeet (commented):
(Like Nigel, I'm basically happy with this, but I agree with the ordering flexibility.)

BillWagner (commented):
Also, basically happy as well.

jskeet:
Note for the February meeting: this PR has not changed (as far as I can see) since November.

jskeet:
(Closure was accidental.)

RexJaeschke:
closed/reopened to run all checks

RexJaeschke:
@Nigel-Ecma I made no change re your suggestion in https://github.com/dotnet/csharpstandard/pull/1209#pullrequestreview-2446808223. Feel free to add a note in a separate PR.

Nigel-Ecma (approved):
Spun off #1361 following [Rex’s comment](https://github.com/dotnet/csharpstandard/pull/1209#issuecomment-2920585585).

PR #1211, opened 2024-11-20
Remove constraints on stackalloc (support new locations)

gafter:


gafter:
This change reflects the following behavior.

This is illegal:

```
    public void M() {
        var p = stackalloc nint[3];
    }
```

Because a stackalloc expression used as an initializer is of type T* (an unsafe pointer type) which requires an unsafe context. However, these are legal:

```
    public void M() {
        Span<nint> p = stackalloc nint[3];
    }
```

and

```
    public void M() {
        var p = (stackalloc nint[3]);
    }
```

(Note the addition of parens)

gafter:
I don't understand the following behavior:

This is illegal:

```
    public void M() {
        var p = stackalloc nint[3][1];
    }
```

I don't understand why. Perhaps because `nint[3] is parsed as a type. I get the error

	•	error CS0208: Cannot take the address of, get the size of, or declare a pointer to a managed type ('nint[]')
	•	error CS1575: A stackalloc expression requires [] after type

Yet this is legal:

```
    public void M() {
        var p = (stackalloc nint[3])[1];
    }
```

jskeet:
Is this intended to replace or be related to #604? I can't quite get my head round it all.

standard/expressions.md
```diff
@@ -3219,14 +3206,12 @@ When a *stackalloc_initializer* is present:
 
 Each *stackalloc_element_initializer* shall have an implicit conversion to *unmanaged_type* ([§10.2](conversions.md#102-implicit-conversions)). The *stackalloc_element_initializer*s initialize elements in the allocated memory in increasing order, starting with the element at index zero. In the absence of a *stackalloc_initializer*, the content of the newly allocated memory is undefined.
 
-The result of a *stackalloc_expression* is an instance of type `Span<T>`, where `T` is the *unmanaged_type*:
+If a *stackalloc_expression* occurs directly as the initializing expression of a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)), where the *local_variable_type* is either a pointer type ([§23.3](unsafe-code.md#233-pointer-types)) or inferred (`var`), then the result of the *stackalloc_expression* is a pointer of type `T *` ([§23.9](????)). In this case the *stackalloc_expression* must appear in unsafe code. Otherwise the result of a *stackalloc_expression* is an instance of type `Span<T>`, where `T` is the *unmanaged_type*:
```
BillWagner: The tool will fix up the link. (We could make the same change on the other links and the tool will add the correct ones, if you have any doubts about the right anchor.)

```suggestion
If a *stackalloc_expression* occurs directly as the initializing expression of a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)), where the *local_variable_type* is either a pointer type ([§23.3](unsafe-code.md#233-pointer-types)) or inferred (`var`), then the result of the *stackalloc_expression* is a pointer of type `T*` (§23.9). In this case the *stackalloc_expression* must appear in unsafe code. Otherwise the result of a *stackalloc_expression* is an instance of type `Span<T>`, where `T` is the *unmanaged_type*:
```

BillWagner: Nit (mostly for Rex): Should `T *` be `T*` (no space)?

BillWagner (approved):
I think this looks good. We should discuss the general comments.

BillWagner:
> I don't understand the following behavior:
> 
> This is illegal:
> 
> ```
>     public void M() {
>         var p = stackalloc nint[3][1];
>     }
> ```
> 
> [...] Perhaps because `nint[3] is parsed as a type. I get the error

I think that's correct. It's trying to stackalloc an array of 1 where the element is `nint[3]`, which is a managed type.

> Yet this is legal:
> 
> ```
>     public void M() {
>         var p = (stackalloc nint[3])[1];
>     }
> ```

This surprises me as well. I think this should be illegal as well. According to [sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA+ABATARgLABQGAzAATakDCpA3oaQ+WRiqQLIAUAlLfY/6QBuAQyikADqQC8pDgGcALsLABrYQBt1EMKQB2AS10KA2sQC6XYzjMBuPgwC+hB0A==), the type is `ref IntPtr`, which should be pointing to an array of elements. That would seem to mean the compiler took a reference to a managed type.

KalleOlaviNiemitalo:
> ```csharp
> var p = (stackalloc nint[3])[1];
> ```

Because this `stackalloc nint[3]` expression is not the entire initialiser, its type is `Span<nint>` rather than `nint*`.  Then, the `[1]` index means calling the `get` accessor of the `public ref T this[int index] { get; }` indexer of `Span<T>`.  The result of this call has type `ref nint`.  Next, the referenced `nint` value should be loaded from that location, but the compiler emits this load only in the Debug configuration and optimises it out in Release, because it has no side effects.  The C# decompiler used by SharpLab shows a local variable `ref IntPtr reference` being initialised with the returned `ref nint`, but that variable is not the `var p` of the source code.  In the Debug configuration, the decompiler instead shows `IntPtr intPtr`.

standard/expressions.md
Nigel-Ecma: This appears to be a breaking change from v7 – where `stackalloc` is defined to be of type `Span<T>` in a safe context and there is no restriction on using it as the initialiser in an inferred (`var`) declaration.

Nigel-Ecma:
> I don't understand the following behavior:
> 
> This is illegal:
> 
> ```
>     public void M() {
>         var p = stackalloc nint[3][1];
>     }
> ```
> 
> I don't understand why. Perhaps because `nint[3] is parsed as a type.

According to the v7 grammar it is not illegal and parses as an *element_access* of a *stackalloc_expression*. `nint` is an *unmanaged_type*, not a *type*, which excludes array types.

Note that the syntactically very similar `new nint[3][1]` is illegal as it was deemed confusing to allow it in the original C# design (and they syntactic method of making it illegal is the root of why the C# grammar isn’t “ANTLR ready” – but I digress ;-)). Maybe the compiler in question is approaching `stackalloc` as it does `new` and therefore declares it illegal?

The parenthesised version `(stackalloc nint[3])[1]` (and the similiar `new`) is also legal and parses exactly the same as the unparanethised version.

standard/expressions.md
Nigel-Ecma: `T*`

jnm2: The type is inferred to be a pointer here, because the compiler preserves backward compatibility:
```cs
var pointer = stackalloc int[1]; // Infers int*
```
And a span here:
```cs
var span = (stackalloc int[1]); // Infers System.Span<int>
```

jnm2: And, in a safe context, `var x = stackalloc` is still not permitted, because it does continue inferring a pointer type.

standard/unsafe-code.md
```diff
@@ -1049,9 +1049,7 @@ When the outermost containing struct variable of a fixed-size buffer member is a
...
-
-In all other respects the semantics of *local_variable_declaration*s ([§13.6.2](statements.md#1362-local-variable-declarations)) and *stackalloc_expression*s ([§12.8.22](expressions.md#12822-stack-allocation)) in unsafe contexts follow those defined for safe contexts.
+When a *stackalloc_expression* occurs as the initializing expression of a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)), where the *local_variable_type* is either a pointer type ([§23.3](unsafe-code.md#233-pointer-types)) or inferred (`var`), the result of the *stackalloc_expression* is a pointer of type `T *`, where `T` is the *unmanaged_type* of the *stackalloc_expression*. In this case the result is a pointer to be beginning of the allocated block.
```
BillWagner: Same:  `T*` instead of `T *`:  
```suggestion
When a *stackalloc_expression* occurs as the initializing expression of a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)), where the *local_variable_type* is either a pointer type ([§23.3](unsafe-code.md#233-pointer-types)) or inferred (`var`), the result of the *stackalloc_expression* is a pointer of type `T*`, where `T` is the *unmanaged_type* of the *stackalloc_expression*. In this case the result is a pointer to be beginning of the allocated block.
```

MadsTorgersen (approved):

jnm2 (approved):

PR #1218, opened 2024-11-21
Remove attributes that aren't yet available in C# 8

BillWagner:
The `MemberNotNull` and `MemberNotNullWhen` attributes weren't available in C# 8. They were added in C# 9.

PR #1221, opened 2024-11-22
[Version 9.0] Feature support for nullable attributes carried over from V8

BillWagner:
These two attributes were mistakenly added in #1191.  They were removed in #1218

This PR adds them back into the `draft-v9` branch, where they are valid.


RexJaeschke:
Originally, I had these two attributes spec'd for V8, but they have been moved to V9. I just proofed the changes made by this PR, and they look good to me.

However, I seem to recall Bill telling me these should be marked as "conditionally normative." That is, support for them is optional, but if either is included, they must have the syntax/semantics, as specified. Should they be a combined package? That is, if one is implemented then so too should the other?

BillWagner:
This should be ready for review and merge at our next meeting.

RexJaeschke (approved):

BillWagner:
@jskeet Do you want to take a look as well? This is the small PR that reinstates an attribute that was incorrectly added in V8.

jskeet (approved):
Yup, looks fine to me.

PR #1225, opened 2024-12-02
Add more NRT examples

### LINKED ISSUE #1214: Add additional null state analysis examples
The suggestions are in these comments:

- https://github.com/dotnet/csharpstandard/pull/1192/files#r1850962129
- https://github.com/dotnet/csharpstandard/pull/1192/files#r1850965363
- https://github.com/dotnet/csharpstandard/pull/1192/files#r1850966485

BillWagner:
Fixes #1214

Add examples as noted in the referenced comments.

### DIFF
--- a/standard/types.md
+++ b/standard/types.md
@@ -958,7 +958,7 @@ Later in the method, the code checks that `s` is not a null reference. The null-
 > }
 > ```
 
-Both auto-property and field-like event declarations make use of a compiler generated backing field. Null state analysis may assume that assignment to the event or property is an assignment to a compiler generated backing field.
+Both auto-property and field-like event declarations make use of a compiler generated backing field. Null state analysis may infer that assignment to the event or property is an assignment to a compiler generated backing field.
 
 > *Example*: A compiler can determine that writing an auto-property or field-like event writes the corresponding compiler generated backing field. The null state of the property matches that of the backing field.
 >
@@ -1012,7 +1012,7 @@ A compiler can treat a property ([§15.7](classes.md#157-properties)) as either 
 >         var t = new Test();
 >         if (t.DisappearingProperty != null)
 >         {
->             int len = t.DisappearingProperty.Length; // No warning. A compiler can assume state
+>             int len = t.DisappearingProperty.Length; // No warning. A compiler can assume property is stateful
 >         }
 >     }
 > }
@@ -1035,7 +1035,6 @@ A compiler may use any expression that dereferences a variable, property, or eve
 >     {
 >         _ = child.child.child; // Warning. Dereference possible null value
 >         var greatGrandChild = child.child.child; // No warning. 
->
 >     }
 > }
 > ```
--- a/standard/types.md
+++ b/standard/types.md
@@ -958,7 +958,7 @@ Later in the method, the code checks that `s` is not a null reference. The null-
 > }
 > ```
 
-Both auto-property and field-like event declarations make use of a compiler generated backing field. Null state analysis may infer that assignment to the event or property is an assignment to a compiler generated backing field.
+Both auto-property and field-like event declarations make use of a compiler-generated backing field. Null state analysis may infer that assignment to the event or property is an assignment to a compiler generated backing field.
 
 > *Example*: A compiler can determine that writing an auto-property or field-like event writes the corresponding compiler generated backing field. The null state of the property matches that of the backing field.
 >
--- a/standard/types.md
+++ b/standard/types.md
@@ -958,7 +958,7 @@ Later in the method, the code checks that `s` is not a null reference. The null-
 > }
 > ```
 
-Both auto-property and field-like event declarations make use of a compiler generated backing field. Null state may assume that assignment to the event or property as assignment to a compiler generated backing field.
+Both auto-property and field-like event declarations make use of a compiler generated backing field. Null state analysis may assume that assignment to the event or property is an assignment to a compiler generated backing field.
 
 > *Example*: A compiler can determine that writing an auto-property or field-like event writes the corresponding compiler generated backing field. The null state of the property matches that of the backing field.
 >
--- a/standard/types.md
+++ b/standard/types.md
@@ -967,7 +967,7 @@ Both auto-property and field like event declarations make use of a compiler gene
 
 > *Example*: A compiler can determine that writing an auto-property or field like event writes the corresponding compiler generated backing field. The null state of the property matches that of the backing field.
 >
-> <!-- Example: {template:"standalone-console", name:"NullPropertyAnalysis"} -->
+> <!-- Example: {template:"standalone-console", name:"NullPropertyAnalysis", expectedException:"NullReferenceException"} -->
 > ```csharp
 > class Test
 > {
--- a/standard/types.md
+++ b/standard/types.md
@@ -950,8 +950,8 @@ Later in the method, the code checks that `s` is not a null reference. The null-
 >     // s is not null:
 >     int length = s.Length;
 >
->     _ = s = null; // Null check by testing equality. The null state of o is maybe null
->     length = s.Length; // warning, and changes the null state of o to not null
+>     _ = s == null; // Null check by testing equality. The null state of s is maybe null
+>     length = s.Length; // Warning, and changes the null state of o to not null
 >
 >     _ = s?.Length; // The ?. is a null check and changes the null state of o to maybe null
 >     if (s.Length > 4) // Warning. Changes null state of o to not null
@@ -962,9 +962,9 @@ Later in the method, the code checks that `s` is not a null reference. The null-
 > }
 > ```
 
-Both auto-property and field like event declarations make use of a compiler generated backing field. Null state may assume that assignment to the event or property as assignment to a compiler generated backing field.
+Both auto-property and field-like event declarations make use of a compiler generated backing field. Null state may assume that assignment to the event or property as assignment to a compiler generated backing field.
 
-> *Example*: A compiler can determine that writing an auto-property or field like event writes the corresponding compiler generated backing field. The null state of the property matches that of the backing field.
+> *Example*: A compiler can determine that writing an auto-property or field-like event writes the corresponding compiler generated backing field. The null state of the property matches that of the backing field.
 >
 > <!-- Example: {template:"standalone-console", name:"NullPropertyAnalysis", expectedException:"NullReferenceException"} -->
 > ```csharp
@@ -981,7 +981,7 @@ Both auto-property and field like event declarations make use of a compiler gene
 >     static void Main()
 >     {
 >         var t = new Test();
->         int len = t.P.Length; // No warning. null state is not null
+>         int len = t.P.Length; // No warning. Null state is not null.
 >     }
 > }
 > ```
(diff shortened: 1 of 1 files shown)


standard/types.md
```diff
@@ -936,15 +936,64 @@ The compiler can update the null state of a variable as part of its analysis.
...
+>     length = s.Length; // warning, and changes the null state of o to not null
+>
+>     _ = s?.Length; // The ?. is a null check and changes the null state of o to maybe null
+>     if (s.Length > 4) // Warning. Changes null state of o to not null
+>     {
+>        _ = s?[4]; //is a null check and changes the null state of o to maybe null
+>         _ = s.Length; // warning
+>     }
+> }
+> ```
```
Nigel-Ecma, line 959: Should the next line (964) markdownlint-disable be deleted? It isn't paired with an enable.

standard/types.md
```diff
@@ -936,15 +936,64 @@ The compiler can update the null state of a variable as part of its analysis.
...
+>     static void Main()
+>     {
+>         var t = new Test();
+>         int len = t.P.Length; // No warning. null state is not null
+>     }
+> }
+> ```
 >
-> <!-- Example: {template:"standalone-console", name:"NullPropertyAnalysis"} -->
+> In the previous example, the backing field for the `DisappearingProperty` is set to null when it is read. However, a compiler may assume that reading a property doesn’t change the null state of that expression. *end example*
```
Nigel-Ecma: Example above has been replaced, this para describes the old version.

Nigel-Ecma (changes requested):
There is a left-over para, or one that was going to be replaced.

I think “The compiler” here should probably always be ”A compiler” – there are a few of the latter already.

BillWagner:
@Nigel-Ecma @RexJaeschke 

> I think “The compiler” here should probably always be ”A compiler” – there are a few of the latter already.

I did replace instances of "the compiler" as it relates to this PR with "a compiler".  A search in the standard uncovers 68 other instances of "the compiler". Some should not be changed. They are part of a longer phrase like "the compiler generated backing field" for properties and field like events.

Should I change others where the language is similar to those Nigel mentioned in this PR? Should I add it to this PR, or create a separate smaller PR?

RexJaeschke:
@BillWagner Yes, please change them, as appropriate. If there are occurrences in other topics then I think a separate PR would be best, making it a general editorial fix.

Nigel-Ecma:
@BillWagner & @RexJaeschke – I agree with Rex to change the ones you see.

It is easy to go down a rabbit hole here; I started as Bill did and saw each “the” needs to be examined – e.g. as in the backing field example “the” refers to the field and not the compiler. From there I started looking into “automatically” – as in “the compiler automatically…” – and these often should at least be something like “a compiler shall/must/may” – this is a spec not a description after all. Once we get to these we need to decide what is being *required* etc. Rabbit hole! This language I expect in many cases dates from the descriptive material from which the 1st edition was written, it will be a long road and incremental is the way to go.

standard/types.md
```diff
@@ -936,15 +936,63 @@ The compiler can update the null state of a variable as part of its analysis.
...
+>
+> <!-- Example: {template:"code-in-class-lib", name:"NullChecks", expectedWarnings:["CS8600","CS8602", "CS8602", "CS8602"]} -->
+> ```csharp
+> #nullable enable
+> public void M(string s)
+> {
+>     // s is not null:
+>     int length = s.Length;
+>
+>     _ = s = null; // Null check by testing equality. The null state of o is maybe null
```
jskeet: Is this meant to be `s == null`? At the moment this is an assignment...

jskeet: (And what is "o" referred to in the comment? I'm guessing s used to be o - needs to be changed below in multiple places.)

standard/types.md
```diff
@@ -936,15 +936,63 @@ The compiler can update the null state of a variable as part of its analysis.
...
+>     length = s.Length; // warning, and changes the null state of o to not null
+>
+>     _ = s?.Length; // The ?. is a null check and changes the null state of o to maybe null
+>     if (s.Length > 4) // Warning. Changes null state of o to not null
+>     {
+>        _ = s?[4]; //is a null check and changes the null state of o to maybe null
```
jskeet: We're missing a noun at the start of the sentence, and a space after `//`

standard/types.md
```diff
@@ -936,15 +936,63 @@ The compiler can update the null state of a variable as part of its analysis.
...
+>         _ = s.Length; // warning
+>     }
+> }
+> ```
+
+Both auto-property and field like event declarations make use of a compiler generated backing field. Null state may assume that assignment to the event or property as assignment to a compiler generated backing field.
```
```suggestion
Both auto-property and field-like event declarations make use of a compiler generated backing field. Null state may assume that assignment to the event or property as assignment to a compiler generated backing field.
```

standard/types.md
```diff
@@ -936,15 +936,63 @@ The compiler can update the null state of a variable as part of its analysis.
...
+
+> *Example*: A compiler can determine that writing an auto-property or field like event writes the corresponding compiler generated backing field. The null state of the property matches that of the backing field.
```
```suggestion
> *Example*: A compiler can determine that writing an auto-property or field-like event writes the corresponding compiler generated backing field. The null state of the property matches that of the backing field.
```

standard/types.md
jskeet: (I'd personally prefer to hyphenate compiler-generated everywhere as well, but I'm not doing that now. Field-like events are hyphenated consistently though, I believe.)

standard/types.md
```diff
@@ -936,15 +936,63 @@ The compiler can update the null state of a variable as part of its analysis.
...
+> <!-- Example: {template:"code-in-class-lib", name:"NullChecks", expectedWarnings:["CS8600","CS8602", "CS8602", "CS8602"]} -->
+> ```csharp
+> #nullable enable
+> public void M(string s)
+> {
+>     // s is not null:
+>     int length = s.Length;
+>
+>     _ = s = null; // Null check by testing equality. The null state of o is maybe null
+>     length = s.Length; // warning, and changes the null state of o to not null
```
```suggestion
>     length = s.Length; // Warning, and changes the null state of o to not null
```

standard/types.md
```diff
@@ -936,15 +936,63 @@ The compiler can update the null state of a variable as part of its analysis.
...
+>         set;
+>     }
+>
+>     public Test() {} // Warning. "P" not set to a non-null value.
 >
-> <!-- Example: {template:"standalone-console", name:"NullPropertyAnalysis"} -->
+>     static void Main()
+>     {
+>         var t = new Test();
+>         int len = t.P.Length; // No warning. null state is not null
```
```suggestion
>         int len = t.P.Length; // No warning. Null state is not null.
```

standard/types.md
```diff
@@ -936,15 +936,63 @@ The compiler can update the null state of a variable as part of its analysis.
...
+>     }
+> }
+> ```
+>
+> In the previous example, the constructor doesn't set `P` to a not null value, and a compiler issues a warning. There's no warning when the `P` property is accessed, because the type of the property is a non nullable reference type. *end example*
```
jskeet: > and a compiler issues a warning

Issues, or may issue?

standard/types.md
```diff
@@ -976,4 +1024,26 @@ Later in the method, the code checks that `s` is not a null reference. The null-
 >
 > In the previous example, the backing field for the `DisappearingProperty` is set to null when it is read. However, a compiler may assume that reading a property doesn’t change the null state of that expression. *end example*
 
+A compiler may use any expression that dereferences a variable, property, or event to set the null state to not null. If it were null, the dereference expression would have thrown a `NullReferenceException`:
```
jskeet, line 1023: Do we want to add a note about extension methods, where calling an extension method *looks* like it dereferences a variable, but doesn't actually do so?

standard/types.md
```suggestion
>     _ = s == null; // Null check by testing equality. The null state of s is maybe null
```

BillWagner:
> Yes, please change them, as appropriate. If there are occurrences in other topics then I think a separate PR would be best, making it a general editorial fix.

Added in #1126

standard/types.md
```diff
@@ -912,39 +912,83 @@ A diagnostic can be produced when a variable ([§9.2.1](variables.md#921-general
...
+>     _ = s?.Length; // The ?. is a null check and changes the null state of s to maybe null
+>     if (s.Length > 4) // Warning. Changes null state of s to not null
+>     {
+>         _ = s?[4]; // ?[] is a null check and changes the null state of s to maybe null
+>         _ = s.Length; // Warning. s is maybe null
+>     }
+> }
+> ```
+
+Both auto-property and field-like event declarations make use of a compiler generated backing field. Null state may assume that assignment to the event or property as assignment to a compiler generated backing field.
```
```suggestion
Both auto-property and field-like event declarations make use of a compiler generated backing field. Null state analysis may assume that assignment to the event or property is an assignment to a compiler generated backing field.
```

jskeet: (Or change "assume that" to "treat".)

standard/types.md
```diff
@@ -968,12 +1012,34 @@ Later in the method, the code checks that `s` is not a null reference. The null-
 >         var t = new Test();
 >         if (t.DisappearingProperty != null)
 >         {
->             int len = t.DisappearingProperty.Length;
+>             int len = t.DisappearingProperty.Length; // No warning. A compiler can assume state
```
jskeet: I suspect the second sentence *is* as intended, but sounds incomplete to me.

standard/types.md
```diff
@@ -968,12 +1012,34 @@ Later in the method, the code checks that `s` is not a null reference. The null-
...
+>
+> public class C
+> {
+>     private C? child;
+>    
+>     public void M()
+>     {
+>         _ = child.child.child; // Warning. Dereference possible null value
+>         var greatGrandChild = child.child.child; // No warning. 
+>
```
jskeet: Remove blank line?

standard/types.md
BillWagner: reworded that sentence.

standard/types.md
```diff
@@ -912,39 +912,83 @@ A diagnostic can be produced when a variable ([§9.2.1](variables.md#921-general
...
+>     _ = s?.Length; // The ?. is a null check and changes the null state of s to maybe null
+>     if (s.Length > 4) // Warning. Changes null state of s to not null
+>     {
+>         _ = s?[4]; // ?[] is a null check and changes the null state of s to maybe null
+>         _ = s.Length; // Warning. s is maybe null
+>     }
+> }
+> ```
+
+Both auto-property and field-like event declarations make use of a compiler generated backing field. Null state analysis may infer that assignment to the event or property is an assignment to a compiler generated backing field.
```
jnm2: Uber nitpick :D
```suggestion
Both auto-property and field-like event declarations make use of a compiler-generated backing field. Null state analysis may infer that assignment to the event or property is an assignment to a compiler generated backing field.
```

jnm2: Also "auto property" instead of "auto-property" but that one is likely a separate item if it's even palatable to consider.

jnm2 (approved):

jskeet (approved):

Nigel-Ecma (approved):

MadsTorgersen (approved):

jskeet:
Approved 2024-12-11 call.

gafter (approved):

PR #1226, opened 2024-12-04
The compiler -> a compiler

BillWagner:
Search for instances of "the compiler" and where the text describes what a compliant compiler is is required or allowed to do, replace "the compiler" with "a compiler".  Where necessary, include "shall" or "may", as appropriate.

There are still instances of "the compiler" in the standard. Some are in types.md. Those are addressed in #1125.  Others are where the text describes "the compiler generated" backing field, class, or method.


standard/statements.md
```diff
@@ -771,7 +771,7 @@ A `switch` statement is executed as follows:
   - Otherwise, if a `default` label is present, control is transferred to the statement list following the `default` label.
   - Otherwise, control is transferred to the end point of the `switch` statement.
 
-> *Note*: The order in which patterns are matched at runtime is not defined. A compiler is permitted (but not required) to match patterns out of order, and to reuse the results of already matched patterns to compute the result of matching of other patterns. Nevertheless, the compiler is required to determine the lexically first pattern that matches the expression and for which the guard clause is either absent or evaluates to `true`. *end note*
+> *Note*: The order in which patterns are matched at runtime is not defined. A compiler is permitted (but not required) to match patterns out of order, and to reuse the results of already matched patterns to compute the result of matching of other patterns. Nevertheless, a compiler is required to determine the lexically first pattern that matches the expression and for which the guard clause is either absent or evaluateSs to `true`. *end note*
```
RexJaeschke:
```suggestion
> *Note*: The order in which patterns are matched at runtime is not defined. A compiler is permitted (but not required) to match patterns out of order, and to reuse the results of already matched patterns to compute the result of matching of other patterns. Nevertheless, a compiler is required to determine the lexically first pattern that matches the expression and for which the guard clause is either absent or evaluates to `true`. *end note*
```

RexJaeschke (approved):
Bill, I fixed one typo; otherwise, it all looks good.

standard/classes.md
```diff
@@ -1565,7 +1565,7 @@ A constant declaration that declares multiple constants is equivalent to multipl
 >
 > *end example*
 
-Constants are permitted to depend on other constants within the same program as long as the dependencies are not of a circular nature. The compiler automatically arranges to evaluate the constant declarations in the appropriate order.
+Constants are permitted to depend on other constants within the same program as long as the dependencies are not of a circular nature. A compiler automatically arranges to evaluate the constant declarations in the appropriate order.
```
jskeet: This second sentence doesn't sound quite like spec-ese to start with. Presumably it should be more of a *requirement* instead of a statement of what happens.

standard/classes.md
```diff
@@ -996,7 +996,7 @@ The inherited members of a constructed class type are the members of the immedia
...
+An inherited member `M` is considered to be ***available*** if `M` is accessible and there is no other inherited accessible member N that already hides `M`. Implicitly hiding an inherited member is not considered an error, but it does cause a compiler to issue a warning unless the declaration of the derived class member includes a `new` modifier to explicitly indicate that the derived member is intended to hide the base member. If one or more parts of a partial declaration ([§15.2.7](classes.md#1527-partial-declarations)) of a nested type include the `new` modifier, no warning is issued if the nested type hides an available inherited member.
```
jskeet: This feels like "the" would be more appropriate, personally - but I'm quite happy to go with "it's easier to be consistent".

standard/classes.md
```diff
@@ -1583,7 +1583,7 @@ Constants are permitted to depend on other constants within the same program as
 > }
 > ```
 >
-> the compiler first evaluates `A.Y`, then evaluates `B.Z`, and finally evaluates `A.X`, producing the values `10`, `11`, and `12`.
+> a compiler first evaluates `A.Y`, then evaluates `B.Z`, and finally evaluates `A.X`, producing the values `10`, `11`, and `12`.
```
jskeet: Again, perhaps this should be "must first evaluate"? (It sounds odd with "a compiler first evaluates" to me.)

standard/classes.md
```diff
@@ -4119,7 +4119,7 @@ Within the program text of the class or struct that contains the declaration of
 >
 > *end example*
 
-When compiling a field-like event, the compiler automatically creates storage to hold the delegate, and creates accessors for the event that add or remove event handlers to the delegate field. The addition and removal operations are thread safe, and may (but are not required to) be done while holding the lock ([§13.13](statements.md#1313-the-lock-statement)) on the containing object for an instance event, or the `System.Type` object ([§12.8.18](expressions.md#12818-the-typeof-operator)) for a static event.
+When compiling a field-like event, a compiler automatically creates storage to hold the delegate, and creates accessors for the event that add or remove event handlers to the delegate field. The addition and removal operations are thread safe, and may (but are not required to) be done while holding the lock ([§13.13](statements.md#1313-the-lock-statement)) on the containing object for an instance event, or the `System.Type` object ([§12.8.18](expressions.md#12818-the-typeof-operator)) for a static event.
```
jskeet: Possibly "a compiler shall"?

standard/variables.md
```diff
@@ -265,9 +265,9 @@ The following categories of variables are classified as initially unassigned:
 
 #### 9.4.4.1 General
 
-In order to determine that each used variable is definitely assigned, the compiler shall use a process that is equivalent to the one described in this subclause.
+In order to determine that each used variable is definitely assigned, a compiler shall use a process that is equivalent to the one described in this subclause.
 
-The compiler processes the body of each function member that has one or more initially unassigned variables. For each initially unassigned variable *v*, the compiler determines a ***definite-assignment state*** for *v* at each of the following points in the function member:
+A compiler processes the body of each function member that has one or more initially unassigned variables. For each initially unassigned variable *v*, a compiler determines a ***definite-assignment state*** for *v* at each of the following points in the function member:
```
jskeet: These sentences possibly need a bit of work - or *maybe* just use "shall" a lot :)

jskeet (commented):
Perfectly happy with the general thrust, but a few nits.

standard/classes.md
BillWagner: reworked the sentence to permit a compiler to reorder the evaluation of constant declarations

standard/variables.md
BillWagner: Reworded those sentences

standard/classes.md
BillWagner: Consider this rewording:

```suggestion
An inherited member `M` is considered to be ***available*** if `M` is accessible and there is no other inherited accessible member N that already hides `M`. Implicitly hiding an inherited member is not considered an error, but a compiler shall issue a warning unless the declaration of the derived class member includes a `new` modifier to explicitly indicate that the derived member is intended to hide the base member. If one or more parts of a partial declaration ([§15.2.7](classes.md#1527-partial-declarations)) of a nested type include the `new` modifier, no warning is issued if the nested type hides an available inherited member.
```

BillWagner: Or is "can issue a warning" better wording?

standard/classes.md
```diff
@@ -1565,7 +1565,7 @@ A constant declaration that declares multiple constants is equivalent to multipl
 >
 > *end example*
 
-Constants are permitted to depend on other constants within the same program as long as the dependencies are not of a circular nature. The compiler automatically arranges to evaluate the constant declarations in the appropriate order.
+Constants are permitted to depend on other constants within the same program as long as the dependencies are not of a circular nature. A compiler is allowed to evaluate the constant declarations in an appropriate order.
```
Nigel-Ecma: (I can't see how to re-surface @jskeet's comment in “Files Changed”, I've “unresolved” the conversation so its visible in “Conversation”.)

@jskeet asked whether this should be a language requirement.

I think a lot of the original source material *described* what “the compiler” (it was the one and only) did and was never converted into language requirements.

“A compiler is allowed to evaluate the constant declarations in an appropriate order.” implies that it is not required to do so, which is clearly not true. I think any appropriate order was probably meant?

```suggestion
Constants are permitted to depend on other constants within the same program as long as the dependencies are not of a circular nature. A compiler must evaluate the constant declarations in an appropriate order.
```

But are we saying too much? Do compiler writers require this much direction to order the dependencies? Here I might offer:

```suggestion
Constants are permitted to depend on other constants within the same program as long as the dependencies are not of a circular nature.
```

If it is felt that compiler writers need more help to figure this out we could go with something along the lines of:

```suggestion
Constants are permitted to depend on other constants within the same program as long as the dependencies are not of a circular nature. A compiler must evaluate constant declarations in an order which respects the dependencies.
```

or maybe:

```suggestion
Constants are permitted to depend on other constants within the same program as long as the dependencies are not of a circular nature. Constant declarations shall be evaluated in any order which respects any dependencies.
```

I personally lean towards the single sentence option, compiler writers can figure it out.

jskeet (approved):

gafter (approved):

PR #1237, opened 2024-12-15
Treat stackalloc like array creation.

### LINKED ISSUE #1232: Prohibit stackalloc int[3][1]
In #1211 it looks like `stackalloc int[3][1]` should be valid by the standard, but Roslyn complains. We've decided it wouldn't do what is probably expected anyway, so it would be better to prohibit it in the standard.

We haven't yet decided on the best approach for this (e.g. in the grammar or as an additional semantic rule in the text).

gafter:
Fixes #1232

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1305,7 +1305,6 @@ primary_no_array_creation_expression
     | anonymous_method_expression
     | pointer_member_access     // unsafe code support
     | pointer_element_access    // unsafe code support
-    | stackalloc_expression
     ;
 ```
 
@@ -2722,6 +2721,7 @@ array_creation_expression
       array_initializer?
     | 'new' array_type array_initializer
     | 'new' rank_specifier array_initializer
+    | stackalloc_expression
     ;
 ```
 


Nigel-Ecma (changes requested):
Note: I have not reviewed the PR *per se*, just why it is failing grammar validation.

This PR will fail grammar validation as it changes a production that the validator itself alters to resolve MLR.

Before this PR can be committed a manual grammar validation will be required **and** the grammar validator updated (if the later is not done all future PRs will fail validation).

I’ll take responsibility for doing these, till then I'll block merging by checking “Request changes”.

jskeet (approved):
I believe this is a good change, in terms of the grammar giving the right *feeling* for the difference. As Nigel has said, in a parallel universe we might have picked an entirely different approach, but I think this is the best change for where we are today.

jskeet:
(It sounds like @gafter has found more issues though...)

gafter:
No, not more issues. It is just the grammar validation.

jskeet:
Okay, so just to check, is the ball now in @Nigel-Ecma's court to validate the grammar and update the validator?

Nigel-Ecma:
> Okay, so just to check, is the ball now in @Nigel-Ecma's court to validate the grammar and update the validator?

@jskeet – Essentially yes, my [earlier comment](https://github.com/dotnet/csharpstandard/pull/1237#pullrequestreview-2504288404) provides details.

Nigel-Ecma:
This PR is subsumed by *Grammar – Consistency & Cleanup 5* #1322. If #1322 is approved this PR should be closed.

RexJaeschke:
Closing this based on Comment https://github.com/dotnet/csharpstandard/pull/1237#issuecomment-2864542420.

jskeet:
We need to see whether #1322 solved this. We're not sure.

PR #1242, opened 2025-01-01
Nullability: conversion behavior

jnm2:
Maybe we jinxed it after all?

This is my first PR to the repo. Guidance is very helpful. Think of the lines of this PR as a series of questions: how far to go, whether it's worth going here, how do we do things?

### DIFF
--- a/standard/types.md
+++ b/standard/types.md
@@ -1048,7 +1048,7 @@ A compiler may use any expression that dereferences a variable, property, or eve
 
 A compiler that generates diagnostic warnings conforms to these rules.
 
-> *Note:* Differences in top-level or nested nullability annotations in types do not affect whether conversion between the types is permitted, since there is no semantic difference between a non-nullable reference type and its corresponding nullable type ([§8.9.1](types.md#891-general)).
+> *Note:* Differences in top-level or nested nullability annotations in types do not affect whether conversion between the types is permitted, since there is no semantic difference between a non-nullable reference type and its corresponding nullable type ([§8.9.1](types.md#891-general)). *end note*
 
 A compiler may issue a warning when nullability annotations differ between two types, either top-level or nested, when the conversion is narrowing.
 
--- a/standard/types.md
+++ b/standard/types.md
@@ -859,6 +859,8 @@ A compiler is not required to perform any static analysis nor is it required to 
 
 **The remainder of this subclause is conditionally normative.**
 
+#### 8.9.5.1 Flow analysis
+
 A compiler that generates diagnostic warnings conforms to these rules.
 
 Every expression has one of three ***null state***s:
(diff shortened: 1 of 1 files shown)


standard/types.md
```diff
@@ -1042,4 +1044,78 @@ A compiler may use any expression that dereferences a variable, property, or eve
...
+>     public void M4(Action<string> p)
+>     {
+>         Action<string?> v1 = p; // Warning
+>         Action<string?> v2 = p!; // No warning
+>     }
+>
+>     public void M5(string[] p)
+>     {
+>         string?[] v1 = p; // No warning
+>     }
```
jnm2, line 1133: Yes. <https://github.com/dotnet/csharplang/blob/main/meetings/2017/LDM-2017-10-04.md#array-covariance>

We allow `string[]` to be assigned to `object[]`, and so we also allow `string[]` to be assigned to `string?[]`.

standard/types.md
```diff
@@ -1042,4 +1044,78 @@ A compiler may use any expression that dereferences a variable, property, or eve
 >
 > *end example*
 
+#### 8.9.5.2 Type conversions
+
+For the purpose of determining whether a conversion is *permitted*, a compiler must consider every nullable-annotated type to be equivalent to its unannotated version. A compiler may issue warnings if the annotations of the types are not compatible.
```
jnm2: @Nigel-Ecma When aiming to only talk about the bare minimum, we're only thinking about the analysis that produces warnings. But for correctness (is this program allowed or not), do the following concerns already fall out from an existing place in the spec?

- Allowing conversions (not just top-level differences in nullability, but nested ones like `IEnumerable<string>` to `IEnumerable<string?>` or `List<string>` to `IEnumerable<object?>`)
- Not allowing implementation of both `IXyz<string>` and `IXyz<string?>` on the same type

jnm2: In other words, do we speak on whether (and where) `Xyz` and `Xyz?` are the same type or different types?

Nigel-Ecma: > @Nigel-Ecma When aiming to only talk about the bare minimum, we're only thinking about the analysis that produces warnings. But for correctness (is this program allowed or not), do the following concerns already fall out from an existing place in the spec?

Hopefully, I will not be so rash as to say yes as TG2 are but fallible humans 😉

Not withstanding the use of the nomenclature of “nullable reference type” and ”non-nullable reference type” there is in fact only one kind of reference type in C#. The statement in §8.9.1 “There is no semantic difference between a non-nullable reference type and its corresponding nullable type, both can either be a reference to an object or `null`”, along with the definition of `?` as an annotation, is one place (there may be others) this is intended to be conveyed.

A key feature of this design is that if every nullable annotation, null-forgiving operator, nullable analysis related pragma and attribute is erased in a C# program then the result is semantically identical to the original (and should be compiled to the same executable).

> * Allowing conversions (not just top-level differences in nullability, but nested ones like `IEnumerable<string>` to `IEnumerable<string?>` or `List<string>` to `IEnumerable<object?>`)

Following §8.9.1 (and maybe elsewhere) this shouldn’t *need* to be stated, but an informative note might be worthwhile if there isn’t one already.

> * Not allowing implementation of both `IXyz<string>` and `IXyz<string?>` on the same type

Similarly, as you can’t have two implementations of the same type that this is an error shouldn’t *need* to be stated – but an informative note might be worthwhile.

> In other words, do we speak on whether (and where) `Xyz` and `Xyz?` are the same type or different types?

The Standard does, but I suspect like most things it could be improved.

jskeet: Agree with linking back to 8.9.1 every so often - it's good to emphasize.

jnm2:
@BillWagner Would you like to take a look at these added nullability examples since you had been working on previous ones?

standard/types.md
```diff
@@ -859,6 +859,8 @@ A compiler is not required to perform any static analysis nor is it required to
 
 **The remainder of this subclause is conditionally normative.**
 
+#### 8.9.5.1 Flow analysis
```
BillWagner: It's easier if you don't add new section numbers.  The renumbering tool builds a mapping from the existing numbers to any re-numbered sections. That way, when the entire TOC is rebuilt, all xrefs move correctly. Instead, give it a name that starts with `§` so you can add links if you want:

```suggestion
#### §flow-analysis Flow analysis
```

standard/types.md
```diff
@@ -1042,4 +1044,123 @@ A compiler may use any expression that dereferences a variable, property, or eve
 >
 > *end example*
 
+#### 8.9.5.2 Type conversions
```
BillWagner: Same comment:

```suggestion
#### §type-conversions Type conversions
```

standard/types.md
```diff
@@ -1042,4 +1044,123 @@ A compiler may use any expression that dereferences a variable, property, or eve
...
+>     public void M2(string? p)
+>     {
+>         _ = (string)p; // Warning, narrowing
+>         _ = (string)p!; // No warning, suppressed
+>     }
+> }
+> ```
+>
+> *end example*
+
```
BillWagner, line 1077: This Markdown violation occurs often in the spec. Just disable and re-enable the rule. You'll find lots of examples in the standard.

standard/types.md
```diff
@@ -1042,4 +1044,123 @@ A compiler may use any expression that dereferences a variable, property, or eve
...
+>     {
+>         _ = ((string, string))p; // Warning, narrowing
+>         _ = ((string, string))p!; // No warning, suppressed
+>     }
+> }
+> ```
+>
+> *end example*
+
+A compiler may follow rules for interface variance ([§18.2.3.3](interfaces.md#18233-variance-conversion)), delegate variance ([§20.4](delegates.md#204-delegate-compatibility)), and array covariance ([§1.7.6](arrays.md#176-array-covariance)) in determining whether to issue a warning for type conversions.
```
BillWagner: I do think this is 17.6. But, it gives me the chance to explain another feature of the re-numbering tool: You don't need to add the markdown link. The tool will do it for you:

```suggestion
A compiler may follow rules for interface variance ([§18.2.3.3](interfaces.md#18233-variance-conversion)), delegate variance ([§20.4](delegates.md#204-delegate-compatibility)), and array covariance (§17.6) in determining whether to issue a warning for type conversions.
```

BillWagner (approved):
Thanks for adding me to review.

I  looked everything over, and this is all good. The suggestions I have all concern how our tooling works.

standard/types.md
jnm2: I might be misunderstanding. This seems to be rule <https://github.com/DavidAnson/markdownlint/blob/v0.37.4/doc/md028.md>. Scrolling through files and a regex search in all files for `^>.*\r\n\r\n>` didn't turn anything up, so I haven't found any examples to follow.

Looking at what this rule is, it seems to be saving people from creating adjacent block quotes that get unified into a single block quote by some parsers and not others. If that's not a concern in this spot, then it seems that it must not be a concern for the entire repo, and maybe we could ignore MD028 globally with the run of this tool. What do you think?

Nigel-Ecma:
Ii think you need to be looking for ^>.*\r\n> *\r\n in a block quote but not in a contained code block in the block quote (i.e. just be wary of false +ve’s). This sequence occurs at the start & end of this particular code block. As Bill mentions you can disable particular errors in a file, just add HTML comments:

Disable one or more rules by name: 
Enable one or more rules by name: <!-- markdownlint-enable MD001 MD005 —>

standard/types.md
jnm2: Oh, 😆 Of course that search will find nothing, because there are intervening lines that disable and reenable the rule. I got it now.

I'd still be interested in knowing whether we could turn the rule off for the whole repo. Perhaps inserting the HTML comments actually addresses MD028 in its own way :D

jnm2:
CI is nice and fast. Really neat system here!

standard/types.md
```diff
@@ -1042,4 +1044,125 @@ A compiler may use any expression that dereferences a variable, property, or eve
...
+> public class C
+> {
+>     public void M1(string p)
+>     {
+>         _ = (string?)p; // No warning, widening
+>     }
+>
+>     public void M2(string? p)
+>     {
+>         _ = (string)p; // Warning, narrowing
```
Nigel-Ecma, line 1069: Following the resolution of #1309 at today’s meeting you might want to make similar changes to the comments, here and elsewhere, to echo the “may issue” on line 1053 (in #1309 “possible warning” was chosen)

jnm2:
@BillWagner With one review, am I okay to merge this?

standard/types.md
```diff
@@ -1042,4 +1044,125 @@ A compiler may use any expression that dereferences a variable, property, or eve
 >
 > *end example*
 
+#### §type-conversions Type conversions
+
+A compiler that generates diagnostic warnings conforms to these rules.
+
+> *Note:* Differences in top-level or nested nullability annotations in types do not affect whether conversion between the types is permitted, since there is no semantic difference between a non-nullable reference type and its corresponding nullable type ([§8.9.1](types.md#891-general)).
```
BillWagner: I'd missed this earlier, but we should add the "end note" to this note. (I assume the next paragraph is meant to be normative):

```suggestion
> *Note:* Differences in top-level or nested nullability annotations in types do not affect whether conversion between the types is permitted, since there is no semantic difference between a non-nullable reference type and its corresponding nullable type ([§8.9.1](types.md#891-general)). *end note*
```

BillWagner:
> @BillWagner With one review, am I okay to merge this?

I had one final suggestion, and then yes, let's :shipit:

PR #1249, opened 2025-01-09
Update expressions.md

sapph42:
The final sentence of section 12.15 (The null coalescing operator) reads:

Otherwise, `a` and `b` are incompatible, and `a` compile-time error occurs.

The formatting around the word "a" in the third clause is a typo - it is not a reference to an object, it is just the word "a".  The sentence should read:

Otherwise, `a` and `b` are incompatible, and a compile-time error occurs.

PR consists of this change and nothing else.


jskeet (approved):
Thanks, LGTM.

Issue #1246, opened 2025-01-07
The token `notnull`. Is it a keyword? Is it explicitly not a keyword?


BillWagner:
In PR #830, this text was not included in the updates for nullable reference types:

> The token `notnull` is neither a keyword nor a contextual keyword. When it is encountered, it will either:
>
> - Bind to a type named `notnull`
> - Bind to no type, in which case, it is interpreted as the nonnullable type constraint.

Should this text be added to the text for the `notnull` constraint?

#830 also includes a large example starting at [line 503](https://github.com/dotnet/csharpstandard/pull/830/files#diff-2188d7623929000b60debed8c009184e583a35df22469e661825bbdf5f9e295cR503)

Do we want to include that example as well.

Other changes in #830 have been reviewed by me and @RexJaeschke . They are either included, or have been excluded by design.

Nigel-Ecma:
I think by the definition of contextual keyword (§6.4.4) `notnull` is one, just like `unmanaged` is – so add it to §6.4.4

The similarity to `unmanaged` should carry over to how it is described. So in §15.2.5 we already have:

> Because `unmanaged` is not a keyword, in *primary_constraint* the unmanaged constraint is always syntactically ambiguous with *class_type*. For compatibility reasons, if a name lookup ([§12.8.4](expressions.md#1284-simple-names)) of the name `unmanaged` succeeds it is treated as a `class_type`. Otherwise it is treated as the unmanaged constraint.

So for `notnull` we can use:

> Because `notnull` is not a keyword, in *primary_constraint* the notnull constraint is always syntactically ambiguous with *class_type*. For compatibility reasons, if a name lookup ([§12.8.4](expressions.md#1284-simple-names)) of the name `notnull` succeeds it is treated as a `class_type`. Otherwise it is treated as the notnull constraint.

and insert it into §15.2.5.

Note that the description of `nameof` follows the same pattern as well.

jskeet:
@Nigel-Ecma to create a PR along the lines he's outlined.

Nigel-Ecma:
Addressed by PR #1258

PR #1258, opened 2025-01-24
Address issue #1246

Nigel-Ecma:
Adds `notnull` as a contextual keyword and defines how the ambiguity with *class_type* is resolved.

Addresses #1246


jskeet (approved):
Looks good to me, but I wouldn't be surprised to hear there are reasons why we need more :)

BillWagner (approved):
This LGTM as well.

Honestly, I think this is all we need here.

PR #1259, opened 2025-02-03
Reconcile library type name lists

RexJaeschke:
[I consider this an editorial fix, and gave it a label of "formatting."]

Recently, I was looking at the library types we require from CLI ([C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)), those we require since the CLI spec ([C.3](standard/standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)), and the list of fully qualified type names in [C.5](standard-library.md#c5-library-type-abbreviations). There are discrepancies, and this PR addresses them. Specifically:

1. One type (`SystemException`) was missing from C2.
2. Some new types added post-CLI were not in the C.5 list.
1. The attribute types added for the nullable reference types PR were put in the wrong namespace in C.3. I put them in the correct namespace, `System.Diagnostics.CodeAnalysis`.
1. Some types in C.5 had the wrong qualified name.
1. The C.5 list was in alpha order, with types in different namespaces interspersed, which was not a useful order. So I put them in type order within namespace order.

Once this PR has been merged, I'll review V8 (and beyond) Draft PRs (such as Indexes/Ranges) to see which, if any, should also update C.5.

### DIFF
--- a/standard/standard-library.md
+++ b/standard/standard-library.md
@@ -236,6 +236,13 @@ namespace System
         public static string Format(string format, params object[] args);
     }
 
+    public class SystemException : Exception
+    {
+        public SystemException();
+        public SystemException(string? message);
+        public SystemException(string? message, Exception? innerException);
+    }
+
     public abstract class Type : MemberInfo { }
 
     public sealed class TypeInitializationException : Exception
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @RexJaeschke 

Tagging @jskeet to take a look. I don't think the committee as a whole need to weigh in, but want him to check as well.

jskeet (approved):
Yup, I'm fine with all of this.

PR #1260, opened 2025-02-04
update dependency action

BillWagner:
The action that generates the PR when we merge changes is failing because the dependent action for uploading the Word doc is deprecated.

Update it, and configure dependabot to tell us about these updates so this doesn't happen again.


BillWagner:
@jskeet Can you approve? This will make sure the TOC, etc are updated when we merge the PRs Rex made to update some sections.

jskeet (approved):
As it happens I have another repo in a very similar situation :)

PR #1257, opened 2025-01-23
Split the workflow that runs our tools from the workflow that writes the checks.

BillWagner:
Make the initial code to create a check run using a chained workflow.

This improves security. Our automated tools can run in the `pull_request` context, which means it runs in the context of a fork, rather than the context of the base repository. 

That means our actions can't create status checks. That REST API requires write access to the base repository. 

So, run the tools and build the text output for the status check.

That workflow's completion triggers a workflow that writes the status check.

PR #1269, opened 2025-02-10
Separate tasks in our workflows

BillWagner:
The task that writes statuses requires different rights than the task that executes the tools.  So, separate these into separate tasks.

Replace #1257   (This initial version can't be run from a fork).

### DIFF
--- a/.github/workflows/renumber-sections.yaml
+++ b/.github/workflows/renumber-sections.yaml
@@ -29,9 +29,28 @@ jobs:
     - name: Setup .NET 8.0
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 8.0.x
+        dotnet-version: 9.0.x
 
     - name: Run section renumbering dry run
+      id: section-renumber
       run: |
         cd tools
         ./run-section-renumber.sh
+
+    - name: Trigger Workflow
+      uses: actions/github-script@v6
+      with:
+        script: |
+          github.rest.actions.createWorkflowDispatch({
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            workflow_id: 'report-status.yml',
+            ref: '${{ github.head_ref }}',
+            inputs: {
+              "head_sha": '${{ steps.section-renumber.output.head_sha }}',
+              "name": '${{ steps.section-renumber.output.check_name }}',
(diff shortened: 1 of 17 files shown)


tools/StandardAnchorTags/Program.cs
```diff
@@ -61,7 +61,7 @@ static async Task<int> Main(string owner, string repo, bool dryrun =false)
         {
             if ((token is not null) && (headSha is not null))
             {
-                await logger.BuildCheckRunResult(token, owner, repo, headSha);
+                var annotations = await logger.BuildCheckRunResult(token, owner, repo, headSha);
```
jskeet, line 64: Revert this change, as it appears to be unused? (Or emit the annotations somewhere?)

tools/Utilities/StatusCheckLogger.cs
```diff
@@ -182,5 +200,43 @@ public async Task BuildCheckRunResult(string token, string owner, string repo, s
...
             Console.WriteLine(e);
         }
+        try
+        { 
+            var core = _gitHubCoreService.Value;
+            
+            await core.GroupAsync("Writing run outputs", async () => 
+            {
+                await core.SetOutputAsync("check_name", title, JsonCheckRunAnnotationSerializerContext.Default.String);
+
```
jskeet, line 210: Maybe group these into pairs of lines (so remove the blank line between the two check_name-related lines, and then between the two conclusion-related lines, etc.)

jskeet (approved):
Can't say I entirely follow this, but if it works, that's fine by me :)

.github/workflows/renumber-sections.yaml
```diff
@@ -29,9 +29,28 @@ jobs:
     - name: Setup .NET 8.0
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 8.0.x
+        dotnet-version: 9.0.x
```
jskeet, line 32: Let's use 10 now?

jskeet (commented):
@BillWagner It looks like this PR could fairly easily be updated to .NET 10 and merged. I haven't checked the supposed conflicts though.

jskeet:
Closing this PR, after offline discussion with @BillWagner

PR #1271, opened 2025-02-17
Tweak/correct partial types and methods

RexJaeschke:
While looking at the spec for the (V13) feature that adds partial properties and indexers, I had need to review the current spec for partial types and members. When doing so, I came up with some improvements and corrections to that, which this PR addresses. 

Here's what this PR does:

1. Changes w.r.t to uses of can/must/may/shall.
2. Changes the title of 15.2.7, "Partial declarations" to "Partial type declarations" as that section is *only* about partial types and not at all about partial methods.
3. Corrects 15.3 Class members, which currently **forbids** partial methods. As well as fixing the words, I added a partial method to the accompanying example.
4. As the Documentation comments chapter said nothing about the handling of doc comments on partial elements, I added such text. Note than when multiple doc comments on an element are written to the document file, I've made their order unspecified. (They might be processed lexically within the same part, or lexically within multiple parts within in the same compilation unit, or who knows in what order if they span multiple compilation units.)

Note that V9 adds support for explicit accessor modifiers on partial members, and V13 adds support for partial properties and indexers. The edits proposed here will support (and not get in the way of) these future features.


standard/classes.md
```diff
@@ -722,15 +722,15 @@ class_body
     ;
 ```
 
-### 15.2.7 Partial declarations
+### 15.2.7 Partial type declarations
 
-The modifier `partial` is used when defining a class, struct, or interface type in multiple parts. The `partial` modifier is a contextual keyword ([§6.4.4](lexical-structure.md#644-keywords)) and only has special meaning immediately before one of the keywords `class`, `struct`, or `interface`.
+The modifier `partial` is used when defining a class, struct, or interface type in multiple parts. The `partial` modifier is a contextual keyword ([§6.4.4](lexical-structure.md#644-keywords)), which in this context has special meaning immediately before the keywords `class`, `struct`, and `interface`.
```
jskeet: It's also used when defining a partial method... Hmm.

standard/documentation-comments.md
```diff
@@ -61,6 +61,16 @@ Although developers are free to create their own set of tags, a recommended set
 
 Note carefully that the documentation file does not provide full information about the type and members (for example, it does not contain any type information). To get such information about a type or member, the documentation file must be used in conjunction with reflection on the type or member.
 
+A partial type or a partial non-type member (such as a method) can be declared in multiple parts, each of which can be in one or more compilation units, and each of which can have one or more documentation comments. A partial non-type member typically has a “defining partial declaration” and an “implementing partial declaration.”
```
jskeet: Do we have any partial non-type members other than methods at the moment? If not, let's just use "or a partial method" for now.

standard/documentation-comments.md
```diff
@@ -61,6 +61,16 @@ Although developers are free to create their own set of tags, a recommended set
...
+
+For a partial type, the document comments that apply directly to that type, if any, from each of its parts, are all written to the documentation file in some unspecified order.
+
+For a partial non-type member
```
jskeet: Colon at the end of this introduction?

(And again, maybe just partial method.)

standard/documentation-comments.md
```diff
@@ -61,6 +61,16 @@ Although developers are free to create their own set of tags, a recommended set
...
+
+- If a defining partial declaration has no corresponding implementing partial declaration, any documentation comments in that defining partial declaration are ignored (as that declaration will be removed).
+- Otherwise, if the implementing partial declaration has any documentation comments, they are written to the documentation file in some unspecified order, and any documentation comments in the defining partial declaration are ignored.
```
jskeet: Both here and in the comment below, we can remove "in some unspecified order". There's only *one* implementing partial declaration, and only *one* defining partial declaration. (Compare that with partial type declarations, where there really are multiple potential comment sources, so we need the "in some unspecified order" bit.)

standard/classes.md
BillWagner: Maybe a note and a cross-reference to the section on partial members?

standard/classes.md
```diff
@@ -782,7 +782,9 @@ Nested types can be declared in multiple parts by using the `partial` modifier.
 >
 > *end example*
 
-The handling of attributes specified on the type or type parameters of different parts of a partial declaration is discussed in [§22.3](attributes.md#223-attribute-specification).
+The handling of attributes specified on the type or type parameters of different parts of a partial type declaration is discussed in [§22.3](attributes.md#223-attribute-specification).
+
+A partial type may contain partial member declarations ([§15.6.9](classes.md#1569-partial-methods)).
```
BillWagner: See previous comment: Maybe move this paragraph up to line 727?

standard/documentation-comments.md
BillWagner: Other partial members will become important for our future selves, when we get to C# 13.

standard/documentation-comments.md
BillWagner: This might be the right time to adopt the terms used in the C# 13 docs for new style partial members.  For all partial members (including methods) must have an implementation, we've used the terms *declaring declaration* and *implementing declaration*. Adopting these terms made it much easier to describe these members where there must be exactly one implementation and exactly one declaration.

BillWagner (commented):
I think this is close, but some of the comments require discussion

standard/documentation-comments.md
RexJaeschke: It's not an intro. It covers partial types only while the following para starts the list for partial methods.

standard/documentation-comments.md
RexJaeschke: Consider a partial method having a summary tag and a params tag. Should we require they be written out in the same order as they appear in the source? I presumed "No", hence the use of "unspecified order."

jskeet: But why highlight that here? Do we say the same thing about documentation comments for *non*-partial methods? (The word "unspecified" doesn't occur in annex D.2, and "documentation" doesn't occur in classes.md.)

Basically unlike partial types, there's one one source for documentation. Partial types really *are* unusual in terms of doc comments, but partial methods aren't - so I don't think we should be specifying them as if the documentation for partial methods has less of an ordering requirement than documentation for non-partial methods.

standard/classes.md
jskeet: @RexJaeschke to have another go at this paragraph.

standard/classes.md
jskeet: @RexJaeschke to implement

standard/classes.md
```diff
@@ -2969,9 +2973,9 @@ When a method declaration includes a `partial` modifier, that method is said to
...
+There are two kinds of partial method declarations: If the body of the method declaration is a semicolon, the declaration is said to be a ***defining partial method declaration***. If the body is other than a semicolon, the declaration is said to be an ***implementing partial method declaration***. Across the parts of a type declaration, there shall be only one defining partial method declaration with a given signature, and there shall be only one implementing partial method declaration with a given signature. If an implementing partial method declaration is given, a corresponding defining partial method declaration shall exist, and the declarations shall match as specified in the following:
```
jskeet: From Bill: "shall be only one implementing partial method" => "shall be at most one implementing partial method"

standard/classes.md
```diff
@@ -852,7 +854,8 @@ The set of members of a type declared in multiple parts ([§15.2.7](classes.md#1
 > partial class A
 > {
 >     int x;                   // Error, cannot declare x more than once
->
+>     partial void M();        // Ok, M is a partial method
```
jskeet: Use "defining partial method declaration" here and "implementing partial method declaration" below for clarity

standard/documentation-comments.md
jskeet:
```suggestion
For a partial non-type member:
```

standard/documentation-comments.md
jskeet: Resolved for @RexJaeschke  to remove "in some unspecified order" just for methods.

BillWagner:
2025/02/19 meeting:  After Rex addresses remaining comments, this is ready to merge.

standard/classes.md
RexJaeschke: OK; I changed the para, and added a reference to partial methods.

standard/documentation-comments.md
RexJaeschke: @BillWagner. Note that V13 uses *defining declaration*, rather than *declaring declaration*, and I have used the new terms.

RexJaeschke: OK, will remove the ordering for methods.

standard/classes.md
RexJaeschke: Agreed.

standard/classes.md
RexJaeschke: Agreed.

standard/classes.md
RexJaeschke: OK.

RexJaeschke:
@jskeet @BillWagner I think we're good to merge this.

BillWagner (approved):
This LGTM.

jskeet (approved):

PR #1281, opened 2025-03-03
BuildGrammar v2.2.0

Nigel-Ecma:
• Add new MS Rules to provide actions & predicates for a number of C#’s
   non-grammar rules:
    - adjacency requirement for tokens in >> & >>=
    - restrict places declaration expressions can be used – as the grammar for them is very loose
    - type argument list disambiguation
    - report semantic error if null-forgiving is applied to null-forgiving
    - tweak order of choices in a couple of rules so Antlr picks the right one • Samples:
   - Reorganise
   - Spilt the large AllInOneNoPreproccesor-v6 into parts
   - Add new samples for new constructs
   - Switch them all to use the new Rules MS

Note: versioning plan for BuildGrammar vX.Y.Z, increment when:
 X – the BuildGrammar tool or other tools change
 Y – new modification sets added
 Z – samples added
So this is version 2.2.0 as a new MS Rules was added along with new samples


BillWagner (approved):
I went through this in some detail. I admit I got lost in a couple places, but overall it looks good.

What should we be looking for that will run differently once this is merged?

Nigel-Ecma:
@BillWagner – Thanks Bill.

Nothing should obviously run differently, unless we break the grammar moving forward – and it now tests for more things, including for the first time things that are expected not to parse.

Developing it did uncover some issues in the Standard and it will form the basis for tests for a few upcoming issues/PRs to fix those – I'm currently working on §6.2.5 *Grammar Ambiguities* which is no longer correct…

BillWagner:
I would still like @jskeet to take a look, just so he's familiar with this, but I'll merge today to keep our ANTLR testing moving forward.

PR #1284, opened 2025-03-07
[BG 2.3.0] – Grammar checker tooling updates

Nigel-Ecma:
• Fixup to handle `is` properly for v8; `is` will change in v9 and hopefully not require a fixup
• Add support for disambiguation of type_argument_list within a query_expression
• Add (semantically nonsensical) sample to test the above


Nigel-Ecma:
If you’re considering reviewing this you may be daunted by the large number of files impacted, and what is in them, however you may ignore:

- All the files in the “Reference” folders. These are generated by tools from the other files. That gets rid of lots of files with hard to review/unrecognisable text in them :-)
- The “GrammarTestingEnv.tgz” file. This contains the grammar checker binaries and test framework. If this PR passes the grammar checker (it does) then this file is fine. Reviewing what goes into this file is done in another repo.

That leaves you with the `*.cs` files – the tests themselves and the comments in those files stating what they are expected to be recognised as; and the `*.md` files (usually readmes for the tests).

So when you visit the *Files changed* tab you can option-click (or whatever that is in your OS/browser) the first file’s disclosure triangle to collapse them all and then just open those for the `*.cs` & `*.md` files.

BillWagner (approved):
Thanks for the roadmap @Nigel-Ecma 

That did make the reviewing task a lot easier.  

This LGTM.  Once Jon gets a good look, let's :shipit:

jskeet:
I'm still recovering (in terms of time) from a separate urgent project - Bill, unless there's anything specific you'd like me to review, I'm happy for it to just be merged, if you've reviewed it. (I'm hoping to put my time into reviewing the Word converter problem Nigel reported.)

Issue #1283, opened 2025-03-07
Updates to §6.2.5 Grammar ambiguities


Nigel-Ecma:
There are issues with the clause §6.2.5 *Grammar ambiguities* that need addressing:

- It has not kept up as new constructs which need to be, or need not to be, disambiguated
have been added to the language.

- While *how* the ambiguities should be resolved within those constructs has been updated
over time it appears that the translation from source material to the Standard – and the
two can be quite different in approach – has unintentionally resulted in incorrect rules.

The suggestions below have been tested against the current (at the time of writing!)
draft-v8 of the Standard using the grammar checker; the latter has been updated to include
additional test samples and revised support code.

A draft PR should follow shortly.

The “source material” mentioned above, and later, is the usual collection of design and
implementation documents used for the Standard, which include the Roslyn compiler. 

For those who wish to consult Roslyn while reading this, look for the
`ScanTypeArgumentList` method, which can be found in `LanguageParser.cs` and work your
way from there.

> Note: rather than use the `roslyn-Visual-Studio-2019-Version-16.3 (C# 8).zip` go with
`roslyn-Visual-Studio-2019-Version-16.8 (C# 9).zip`. The latter includes a C# v8 change
not included in the former; presumably introduced somewhere in 16.4-16.7.

---

## A. The grammar rules requiring disambiguation

§6.2.5 lists the rules that require disambiguation in three places: right at the start (which
misses one), after the first example, and after the bulleted list. They are:

- *simple_name* (§12.8.4);
- *member_access* (§12.8.7); and
- *pointer_member_access* (§23.6.3).

The common characteristic of these rules is that they all appear in an expression context
and each has one or more alternatives which end in `identifier type_argument_list?`. Being
in an expression context means that the containing expression may continue after the
*type_argument_list* and this may produce an ambiguity between recognising the optional
*type_argument_list* and recognising the ‘`<`’ and ‘`>`’ that would enclose it as
operators within the larger expression. The first example in §6.2.5 covers this.

There are other rules with the same properties which therefore also might require
disambiguation and thus need to be included, these are:

- *base_access* (§12.8.15);
- *null_conditional_member_access* (§12.8.8); and
- *dependent_access* (§12.8.8)

## B. The grammar rules NOT requiring disambiguation

§6.2.5 also contains the list of rules which do not require disambiguation but rather
the *type_argument_list* should always be parsed if present. This list is in a *Note*
just before the examples:

- *namespace_or_type_name* (§7.8)

[...]

PR #1286, opened 2025-03-10
Addresses issue #1283.

Nigel-Ecma:
Updates §6.2.5 as per issue and also changes the descriptive style to the proscriptive requirement of the Standard.

Issuing as draft to allow time for comments on #1283.

Note: In #1283 I promised test cases for *type_argument_list* parsing. These have been incorporated into a grammar testing update, #1284, and can be found in `tools/GrammarTesting/Tests/Parsing/Samples/v8` samples `Type argument list` and `TAL Query expressions`.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4161,7 +4161,7 @@ equality_expression
     ;
 ```
 
-> *Note*: Lookup for the right operand of the `is` operator must first test as a *type*, then as an *expression* which may span multiple tokens. In the case where the operand is an *expreesion*, the pattern expression must have precedence at least as high as *shift_expression*. *end note*
+> *Note*: Lookup for the right operand of the `is` operator must first test as a *type*, then as an *expression* which may span multiple tokens. In the case where the operand is an *expression*, the pattern expression must have precedence at least as high as *shift_expression*. *end note*
 
 The `is` operator is described in [§12.12.12](expressions.md#121212-the-is-operator) and the `as` operator is described in [§12.12.13](expressions.md#121213-the-as-operator).
 
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -52,7 +52,18 @@ Every compilation unit in a C# program shall conform to the *compilation_unit* 
 
 ### 6.2.5 Grammar ambiguities
 
-The productions for *simple_name* ([§12.8.4](expressions.md#1284-simple-names)) and *member_access* ([§12.8.7](expressions.md#1287-member-access)) can give rise to ambiguities in the grammar for expressions.
+The productions for:
+
+- *simple_name* ([§12.8.4](expressions.md#1284-simple-names)),
+- *member_access* ([§12.8.7](expressions.md#1287-member-access)),
+- *null_conditional_member_access* ([§12.8.8](expressions.md#1288-null-conditional-member-access)),
+- *dependent_access* ([§12.8.8](expressions.md#1288-dependent-access)),
+- *base_access* ([§12.8.15](expressions.md#12815-base-access)) and
+- *pointer_member_access* ([§23.6.3](unsafe-code.md#2363-pointer-member-access));
+
+(the “disambiguated productions”) can give rise to ambiguities in the grammar for expressions.
+
+These productions occur in contexts where a value can occur in an expression, and have one or more alternatives that end with the grammar “`identifier type_argument_list?`”. It is the optional *type_argument_list* which results in the possible ambiguity.
 
 > *Example*: The statement:
 >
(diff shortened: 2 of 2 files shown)


Nigel-Ecma:
@jskeet – I thought I’d fixed the bad references but I missed a § in the one that is still causing the fail, simple fix. However I’ve left it with the error as the Word converter is throwing an exception you might wish to look into. If you drop me a line when you’d like me to remove the stray § I'll push it out.

jskeet:
Thanks Nigel - will try to get to that in the next few days.

jskeet:
@Nigel-Ecma: Having looked at the error, that feels like the Word Converter "core" itself "working as intended" - it's reporting the error nicely. The error is coming from `Utilities.StatusCheckLogger.BuildCheckRunResult` which is something @BillWagner knows rather more about than me.

Bill: the job that contains the failure is https://github.com/dotnet/csharpstandard/actions/runs/13755098000/job/38461209510

(I personally don't mind a slightly messy error, so long as it really *does* fail, and the underlying failure is clear - but I expect it could be tidier.)

@Nigel-Ecma: Having captured that information (which Bill can choose to address or not) I think you're fine to remove the stray §.

standard/expressions.md
```diff
@@ -4161,7 +4161,7 @@ equality_expression
     ;
 ```
 
-> *Note*: Lookup for the right operand of the `is` operator must first test as a *type*, then as an *expression* which may span multiple tokens. In the case where the operand is an *expreesion*, the pattern expression must have precedence at least as high as *shift_expression*. *end note*
+> *Note*: Lookup for the right operand of the `is` operator must first test as a *type*, then as an *expression* which may span multiple tokens. In the case where the operand is an *expression*, the pattern expression must have precedence at least as high as *shift_expression*. *end note*
```
jskeet, line 4164: We should consider reverting this change, and doing a global search and replace of expression => expreesion as a far more fun term.

standard/lexical-structure.md
```diff
@@ -52,7 +52,18 @@ Every compilation unit in a C# program shall conform to the *compilation_unit*
...
+The productions for:
+
+- *simple_name* ([§12.8.4](expressions.md#1284-simple-names)),
+- *member_access* ([§12.8.7](expressions.md#1287-member-access)),
+- *null_conditional_member_access* ([§12.8.8](expressions.md#1288-null-conditional-member-access)),
+- *dependent_access* ([§12.8.8](expressions.md#1288-null-conditional-member-access)),
+- *base_access* ([§12.8.15](expressions.md#12815-base-access)) and
+- *pointer_member_access* ([§23.6.3](unsafe-code.md#2363-pointer-member-access));
+
+(the “disambiguated productions”) can give rise to ambiguities in the grammar for expressions.
```
jskeet, line 64:
```suggestion
can give rise to ambiguities in the grammar for expressions. Within this clause, these are known as *disambiguated productions*.
```

(Just as one possible option. I'm not really that bothered.)

standard/lexical-structure.md
```diff
@@ -262,19 +274,19 @@ fragment Input_Character
     // anything but New_Line_Character
     : ~('\u000D' | '\u000A'   | '\u0085' | '\u2028' | '\u2029')
     ;
-    
+
```
jskeet, line 275: Were these space removals deliberate, or just something your editor did? I wonder whether it might be worth removing them from this PR, and then (if we want to) running something to remove them from the whole spec. (My guess is there will be other occurrences.)

jskeet (approved):
I like all the changes here (although as noted we might want to make the whitespace changes separately, and across the board - and then possibly lint for them).

I think @gafter's input would be useful to validate the change as well. What looks reasonable to me can still be entirely incorrect...

standard/expressions.md
BillWagner: I did a global search on GitHub. This is the only instance.

standard/lexical-structure.md
BillWagner: I'm fine either way.

standard/lexical-structure.md
```diff
@@ -65,16 +76,17 @@ The productions for *simple_name* ([§12.8.4](expressions.md#1284-simple-names))
 >
 > *end example*
 
-If a sequence of tokens can be parsed (in context) as a *simple_name* ([§12.8.4](expressions.md#1284-simple-names)), *member_access* ([§12.8.7](expressions.md#1287-member-access)), or *pointer_member_access* ([§23.6.3](unsafe-code.md#2363-pointer-member-access)) ending with a *type_argument_list* ([§8.4.2](types.md#842-type-arguments)), the token immediately following the closing `>` token is examined, to see if it is
+If a sequence of tokens can be parsed, in context, as one of the disambiguated productions
+including an optional *type_argument_list* ([§8.4.2](types.md#842-type-arguments)), then
+the token immediately following the closing `>` token shall be examined and if it is:
```
BillWagner: I think our tool will end up handling this better if we use soft breaks (no semantic change here):

```suggestion
If a sequence of tokens can be parsed, in context, as one of the disambiguated productions including an optional *type_argument_list* ([§8.4.2](types.md#842-type-arguments)), then the token immediately following the closing >` token shall be examined and if it is:
```

standard/lexical-structure.md
BillWagner: (not related to this PR, but this comment)
I'm fine removing them as we find them. There are many places in code and grammar where multiple spaces are intended. I think the sweep will be painful to get right.

BillWagner (approved):
This LGTM once we address open comments (async or in the meeting)

BillWagner:
> @Nigel-Ecma: Having looked at the error, that feels like the Word Converter "core" itself "working as intended" - it's reporting the error nicely. The error is coming from `Utilities.StatusCheckLogger.BuildCheckRunResult` which is something @BillWagner knows rather more about than me.
> 
> Bill: the job that contains the failure is https://github.com/dotnet/csharpstandard/actions/runs/13755098000/job/38461209510
> 
> (I personally don't mind a slightly messy error, so long as it really _does_ fail, and the underlying failure is clear - but I expect it could be tidier.)
> 
The exception from building the check result is what I'm trying to address in #1269. Quick background: When our tools run on a PR, they run in the context of the PR branch, which is often in a fork. That means the tools don't have write access to the base repo. That's a better security practice. 

I've asked for some help to get that working, but no answer yet. If I can't get unstuck, I'll remove the extra logging. We'll still get the output, but it won't be integrated into the files tab.

standard/lexical-structure.md
Nigel-Ecma: I didn’t use *italics* but quoted it as we use those for proper defined terms, the rest is just the style I’m used to, i.e. a shorthand in paren. But otherwise I’m fine either way as well. @RexJaeschke what would you do here?

standard/lexical-structure.md
Nigel-Ecma: It’s usually the editor that puts them in not takes them out in my experience. When I added a check-in action for mdlint to my local git the trailing space checks and a few others came along for the ride. If I only get trailing space warnings I often skip them, but if I’m doing other cleanup I might take them out – 1 RE does the job after all. Of course removing them can sometimes bury the important changes, it’s a balance.

Happy to follow policy either way, but I’ll mark this conversation as resolved as if I tried to put them back in I’d probably break my git (the removal isn’t in the last commit) – I’m practiced at that ;-)

standard/lexical-structure.md
Nigel-Ecma: I’ll join these in the next commit clobbering the stray §.

gafter (approved):

standard/lexical-structure.md
jskeet: I'm happy enough to leave this.

jskeet:
After discussion and raising #1294, we are happy.

PR #1287, opened 2025-03-11
Add support for nullable arrays nullables

Nigel-Ecma:
Adds support for nullable arrays, arrays of nullables, and nullable arrays of nullables.

Review notes:

- Only `arrays.md`, `types.md`, `sample.cs` &` ReadMe.md` need to be considered; all other files are part of the testing machinery. 
- If you find it helpful you can download and view the `sample.svg` file to understand the parse the grammar generates.


standard/types.md
```diff
@@ -83,6 +73,28 @@ nullable_type_annotation
...
+array_reference_type
+    : non_nullable_array_reference_type
+    | nullable_array_reference_type
+    ;
+
+nullable_array_reference_type
+    : non_nullable_array_reference_type nullable_type_annotation
+    ;
+
+non_nullable_array_reference_type
```
jskeet: I find this name confusing. It's about what might be an *element* of an array, right?

standard/arrays.md
```diff
@@ -46,9 +46,18 @@ In effect, the *rank_specifier*s are read from left to right *before* the final
...
 
-At run-time, a value of an array type can be `null` or a reference to an instance of that array type.
+An array type is a reference type and therefore can be a nullable reference type ([§8.9.3](types.md#893-nullable-reference-types)). Likewise the element type of an array can be a nullable reference type or a nullable value type ([§8.3.12](types.md#8312-nullable-value-types)):
 
-> *Note*: Following the rules of [§17.6](arrays.md#176-array-covariance), the value may also be a reference to a covariant array type. *end note*
+- An array type of the form `T[R]` is a non-nullable array with rank `R` and a non-array non-nullable element type `T`.
+- An array type of the form `T[R]?` is a nullable array with rank `R` and a non-array non-nullable element type `T`.
+- An array type of the form `T?[R]` is a non-nullable array with rank `R` and a non-array nullable element type `T`.
+- An array type of the form `T?[R]?` is a nullable array with rank `R` and a non-array nullable element type `T`.
+
```
jskeet, line 55: What happens with arrays of arrays? For example, are all of the following valid - and more importantly, are they all correctly described here?

```csharp
string[][]
string?[][]
string[]?[]
string[][]?
string?[]?[]
string?[][]?
string[]?[]?
string?[]?[]?
```

It looks like they're all allowed, and they do make sense. I suspect it all just drops out in terms of the grammar - but I wonder whether just a single example (maybe `string?[][]?`) would be useful in a note?

standard/arrays.md
```diff
@@ -46,9 +46,18 @@ In effect, the *rank_specifier*s are read from left to right *before* the final
...
+At run-time, a value of an array type can be:
+
+- `null`; or
```
jskeet, line 58: I wonder whether it would be useful to have a note either here or in array creation expressions to warn that code that looks safe still can easily be null-unsafe with arrays:

```csharp
string[] x = new string[5];
int y = x[0].Length;
```

... no warnings, but will fail at run-time.

Nigel-Ecma: @jskeet – Well there were no warnings from the current version of whatever compiler you used, but there is not reason why some compiler couldn’t issue a warning… I suggest you spin this off as a seperate issue.

standard/types.md
Nigel-Ecma: @jskeet – yes it is about the element type. The new array grammar follows the form of the existing non-array grammar with each of the former corresponding with one of the latter:

> *non_array_type* & *type*
> *array_reference_type* & *reference_type*
> *non_nullable_array_reference_type* & *non_nullable_reference_type*
> *nullable_array_reference_type* & *nullable_reference_type*

I did originally consider *non_array_reference_type*, as in:

> *non_array_type* & *type*
> *non_array_reference_type* & *reference_type*
> *non_nullable_non_array_reference_type* & *non_nullable_reference_type*
> *nullable_non_array_reference_type* & *nullable_reference_type*

but went with the shorter option. Happy to add the *non_* in if folk think it helps understanding. Other suggestions can be submitted as well!

jskeet: I think I was most confused by it being called "array type" when it's not about the array so much as the elements of the array. But adding in "element" would make it significantly longer. Will have a mull - and look closely again having confirmed it does mean what I expected it to mean :)

jnm2: I also find "array reference type" very confusing, and would prefer `element` in it regardless of length.

jnm2: Ahh I see what's happening. Given that these types may not appear as an element of an array, I retract my desire for `element` to be in the name.

jnm2: `non_array_non_nullable_reference_type` gets across: "this is the full set of C# types which are not arrays and which are non-nullable reference types." It also gets rid of the confusion about whether this includes arrays.

Nigel-Ecma: Rules now renamed to the longer versions above.

jnm2 (changes requested):
I would prefer to go in the direction of https://github.com/dotnet/csharpstandard/pull/1386, which also adds support for arrays of nullable reference types more broadly, rather creating new names like `non_array_non_nullable_reference_type` and `non_array_nullable_reference_type` which don't sound like core concepts we'd want to be referencing elsewhere.

jskeet:
Added "meeting: discuss" label to consider closing this PR in favor of #1554

PR #1290, opened 2025-03-17
Grammar – Consistency & Cleanup

Nigel-Ecma:
Up to Standard v5 the C# grammar was defined using a custom BNF-like grammar which supported optional and recursion, starting with Standard v6 the ANTLR grammar notation was adopted which added support for repetition. As of now the C# grammar has well almost 70 rules using repetition which has replaced unnecessary recursion.

In general language constructs which are left-associative are often described using left-recursive grammar rules which visualised as trees produce left-deep structures – e.g. this is the case for many arithmetic operators. Similarly right-associative constructs may be described by right-recursive rules which visualised produce right-deep structures. Finally list-like constructs are often described using repetition and visualised produced wide-flater trees – e.g. think argument lists, type parameter lists etc.

In v5 the list-like constructs where often represented by left-recursion as there was no repetition, from v6 such rules have been replaced by those using repetition.

This PR is the first of a number to pickup the stragglers, v5-era list-like structures described using left-recursion. Further a number of these rules have similar “cousins” within the grammar which have long made since made the transition and the dissimilarity in approach due to these straggles make the grammar as a whole less consistent.

As of today there are just 7 stragglers and 1 semi-straggler (it seems to have started transition and faltered).

This first PR handles: *expression_list*, *query_body_clauses*, *type_parameters*, and *variant_type_parameters*

Review Notes:

You only need to consider the changes to the `*.md` files in `standard`; the other files are machine-generated updates to parsing test samples describing the new expected parse for some samples. If you wish to see visualisations of the changes you may download the changed `*.svg` files in the PR and the current versions of them in `draft-v8` and view them side-by-side (multiple monitors helps!)

PR #1291, opened 2025-03-17
Grammar – Consistency & Cleanup 2

Nigel-Ecma:
[Continues on from #1290 Grammar – Consistency & Cleanup]

This PR removes *type_parameter_constraints_clauses* orphaned during the v5 -> v6 transition.

In v5 this rule was present in a number of places, e.g.:

> *class-declaration*:
>    *attributes*<sub>opt</sub> *class-modifiers*<sub>opt</sub> partial<sub>opt</sub> class *identifier* *type-parameter-list*<sub>opt</sub>
>    *class-base*<sub>opt</sub> *type-parameter-constraints-clauses*<sub>opt</sub> *class-body* ;<sub>opt</sub>

Where ever it occurred it was *optional*. The rule itself was defined as:

> *type-parameter-constraints-clauses*:
>    *type-parameter-constraints-clause*
>    *type-parameter-constraints-clauses* *type-parameter-constraints-clause*

which is one or more *type-parameter-constraints-clause* specified using left-recursion.

Combining “optional” with “one or more” produces “zero or more” – which the ANTLR notation provides directly using `*`, so in v6 the clauses using *type-parameter-constraints-clauses* all changed to using *type-parameter-constraints-clause* zero or more times, e.g.:

```ANTLR
class_declaration
    : attributes? class_modifier* 'partial'? 'class' identifier
        type_parameter_list? class_base? type_parameter_constraints_clause*
        class_body ';'?
    ;
```

along with all in-text references to *type_parameter_constraints_clauses* being changed to *type_parameter_constraints_clause*s (trailing 's' not in italics).

Job done…

Well not quite *type_parameter_constraints_clauses* was not deleted but orphaned, and a few of those in-text references were missed and/or the English grammar around them not fixed up.

This PR simply completes the job.

(And there is a little trailing whitespace cleanup done as my check-in “lint” action barfs at them.)

### DIFF
--- a/standard/basic-concepts.md
+++ b/standard/basic-concepts.md
@@ -66,7 +66,7 @@ Declarations in a C# program define the constituent elements of the program. C#
 A declaration defines a name in the ***declaration space*** to which the declaration belongs. It is a compile-time error to have two or more declarations that introduce members with the same name in a declaration space, except in the following cases:
 
 - Two or more namespace declarations with the same name are allowed in the same declaration space. Such namespace declarations are aggregated to form a single logical namespace and share a single declaration space.
-- Declarations in separate programs but in the same namespace declaration space are allowed to share the same name.  
+- Declarations in separate programs but in the same namespace declaration space are allowed to share the same name.
     > *Note*: However, these declarations could introduce ambiguities if included in the same application. *end note*
 - Two or more methods with the same name but distinct signatures are allowed in the same declaration space ([§7.6](basic-concepts.md#76-signatures-and-overloading)).
 - Two or more type declarations with the same name but distinct numbers of type parameters are allowed in the same declaration space ([§7.8.2](basic-concepts.md#782-unqualified-names)).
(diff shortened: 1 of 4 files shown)


jskeet:
Again we've got removal of trailing whitespace... @BillWagner is that something we can lint for? It would be good to remove it in a single PR and then make sure it doesn't creep in again. (I'm fine to keep the changes in this PR for simplicity.)

jskeet (approved):
Again, I'm pretty confident I understand this one...

BillWagner (approved):
This also LGTM.  Let's :shipit: at our next meeting.

BillWagner:
> Again we've got removal of trailing whitespace... @BillWagner is that something we can lint for? It would be good to remove it in a single PR and then make sure it doesn't creep in again. (I'm fine to keep the changes in this PR for simplicity.)

TL;DR;  Instead of doing all the work to automate whitespace removal, I suggest that if we create a PR with a lot of whitespace cleanup, that PR author should recommend reviewers "hide whitespace". 

Here's why:  Markdown lint will flag a single trailing whitespace character. However, 2 whitespace characters creates a [line break](https://www.markdownguide.org/basic-syntax/#line-breaks). I agree with the cited author that it's hard to see, so use a blank line instead. So, removing *all* whitespace likely changes the rendered output.

Instead, let's just keep updating as the tools point out concerns.

Nigel-Ecma:
> Here's why: Markdown lint will flag a single trailing whitespace character. However, 2 whitespace characters creates a [line break](https://www.markdownguide.org/basic-syntax/#line-breaks). I agree with the cited author that it's hard to see, so use a blank line instead. So, removing _all_ whitespace likely changes the rendered output.

As many of you know I find Markdown one of the most beautiful languages ever developed and now I learn the face of it been scared by this carbuncle?!? (/s, just in case)

Across the Standard’s md files there are currently 68 occurences of two or more spaces at the end of lines (the Google flavoured markdown trigger for a hard break). A brief glance at just a few showed they had no effect due to other formatting, but an exhaustive check would be required to see if they all could be clobbered.

If a hard break is required a line can be ended with a backslash, which at least is visible. I suggest we make that a Standard requirement and then look at clobbering all current trailing whitespace and flagging it after that (https://github.com/updownpress/markdown-lint/blob/master/rules/009-no-trailing-spaces.md)

gafter (approved):

jnm2 (approved):

PR #1292, opened 2025-03-19
Remove (s) in favor of just s

### LINKED ISSUE #1274: Can we avoid using `(s)` for things that can be multiple?
Example from #605:

> value(s) of the expression(s)

I think it would be simpler to accept that sometimes when there's a list of things, that might be a list of only one thing.

@jskeet to find all occurrences.

jskeet:
Fixes #1274


standard/expressions.md
```diff
@@ -1187,7 +1187,7 @@ Even though overload resolution of a dynamically bound operation takes place at
 In these cases a limited compile-time check is performed on each member in the known set of function members, to see if it can be known for certain never to be invoked at run-time. For each function member `F` a modified parameter and argument list are constructed:
 
 - First, if `F` is a generic method and type arguments were provided, then those are substituted for the type parameters in the parameter list. However, if type arguments were not provided, no such substitution happens.
-- Then, any parameter whose type is open (i.e., contains a type parameter; see [§8.4.3](types.md#843-open-and-closed-types)) is elided, along with its corresponding parameter(s).
+- Then, any parameter whose type is open (i.e., contains a type parameter; see [§8.4.3](types.md#843-open-and-closed-types)) is elided, along with its corresponding argument.
```
jskeet: This one is worth discussing. I don't think there's such a thing as a "corresponding parameter for a parameter" - the correspondence is between arguments and parameters (unless I've missed something).

Arguably this should be "any argument whose corresponding parameter's type is open is elided, along with that corresponding parameter". I'd be happy to change it to that if folks would prefer it.

BillWagner (approved):
This LGTM @jskeet 

I'm fine with the changes as you've proposed them.

RexJaeschke (approved):

standard/expressions.md
Nigel-Ecma: I don’t think you’ve missed something, but it should be *arguments* – the parameter in question could be a `params` (e.g. `params T[] args`) and so have multiple corresponding arguments.

This would also I think argue against your optional alternate wording.

```suggestion
- Then, any parameter whose type is open (i.e., contains a type parameter; see [§8.4.3](types.md#843-open-and-closed-types)) is elided, along with its corresponding arguments.
```

Nigel-Ecma (dismissed):
I don’t see the *need* to remove the (s)’s – I use it in my own writing when it feels right – but will go with the consensus on this (and try to remember to not add any going forward).

One fix suggested.

standard/expressions.md
jskeet: Yup, I was wondering about that. I've just found that we use "corresponding argument" in various places, which I didn't think we did. Will commit your suggestion now.

jskeet:
Merging this now, as I addressed Nigel's request.

PR #1297, opened 2025-03-19
Draft grammar to permit nullable types as array element types.

gafter:
Proposed replacement for #1287 

@Nigel-Ecma Can you please help me set up the testing for this PR?

### DIFF
--- a/standard/types.md
+++ b/standard/types.md
@@ -64,6 +64,7 @@ non_array_type
     | delegate_type
     | 'dynamic'
     | type_parameter
+    | nullable_reference_type
     | pointer_type      // unsafe code support
     ;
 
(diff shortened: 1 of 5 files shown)


standard/types.md
```diff
@@ -54,7 +54,8 @@ interface_type
     ;
 
 array_type
-    : non_array_type rank_specifier+
+    | non_array_type rank_specifier+
+    | non_array_type ( nullable_type_annotation rank_specifier+ )+
```
jnm2, line 58: Does this allow `int??[]`?

gafter: No, because a nullable type isn't a `non_array_type`.

jnm2: If I'm reading it right, `non_array_type` includes `value_type` which includes `nullable_value_type` which I believe `int?` would be.

Nigel-Ecma (changes requested):
See TG2 email discussion for details.

jskeet:
We want to wait for a meeting where @gafter is present to discuss this as well, but @jnm2 is keen to get his hands dirty with it when he finds time.

jnm2 (changes requested):
> We want to wait for a meeting where @gafter is present to discuss this as well, but @jnm2 is keen to get his hands dirty with it when he finds time.

I made it happen! See https://github.com/dotnet/csharpstandard/pull/1386 for my proposed grammar changes. If we take my PR as the direction, @gafter's other changes in this PR will be valuable to move over.

standard/types.md
BillWagner: I'd agree with @gafter.  `nullable_value_type` is defined as `non_nullable_value_type nullable_type_annotation`. There's no additional condition for a nullable_value_type. So, int?? isn't a valid type.

BillWagner (approved):
Once grammar validation is added, I'm ready to accept this.

standard/types.md
jnm2: `non_array_type` may expand into `value_type`, which may expand into `nullable_value_type`, which expands into `non_nullable_value_type nullable_type_annotation`. Therefore, `non_array_type` may expand into `non_nullable_value_type nullable_type_annotation`. Substituting that:

`non_array_type ( nullable_type_annotation rank_specifier+ )+`

may be

`non_nullable_value_type nullable_type_annotation ( nullable_type_annotation rank_specifier+ )+`

Notice how `nullable_type_annotation` appears twice. The proposed grammar allows `int??[]` as a valid construction.

jskeet:
Closing in favor of #1386 - @gafter if you feel strongly that we should come back to this instead, please reopen and we can discuss.

Issue #1253, opened 2025-01-17
Instance initializers performed in extern instance constructors


jjonescz:
Spec of *instance variable initializers* ([§15.11.3](https://github.com/dotnet/csharpstandard/blob/dcfbe0a482acd93ef41c15dbee431d4a0ad6d6c2/standard/classes.md#15113-instance-variable-initializers)) does not seem to exclude `extern` constructors, i.e., it sounds like `extern` constructors should perform variable initializers as well.

I think the spec should exclude `extern` constructors from the wording - the compiler cannot perform the initializers when an `extern` constructor is used since it doesn't control its code. (Indeed, the roslyn implementation doesn't perform the initializers.)

Note also that if you have a single constructor which is `extern`, then all variable initializers are not emitted anywhere, so they are useless. But there is no warning about that - should one be added to the language? For example:

```cs
class C
{
    int X = 5; // the initializer has no effect on the produced assembly

    extern C();
}
```

Related issue for static extern constructors: https://github.com/dotnet/csharpstandard/issues/158

jskeet:
After discussion... we can't require the extern constructor to do anything in particular; we could potentially warn if there *is* a field initializer in a type which *only* has extern constructors. (There are reasons why you might want to have field initializers *and* extern constructors, if you also have normal constructors.)

We think we don't need to specify such a warning though, as warnings are generally non-normative.

But we agree that 15.11.3 should exclude requirements for extern constructors. I'll see how minimally we can make this change, and try to cover #158 at the same time.

Issue #158, opened 2019-04-01
15.12 Static constructors (when extern)


gafter:
See also https://github.com/dotnet/roslyn/issues/34651

The specification requires an implementation to initialize static fields *before* calling the static constructor.  However, existing implementations compile the field initializations into the body of the generated static constructor.

When the static constructor is itself declared `extern`, who is supposed to run the field initializers?  Existing implementations simply produce no code for those initializations (and produce a warning on the static constructor declaration).  Is this a bug in the compiler?  Iif so, what could it possibly do instead?  Or is this a bug in the (external) body of the static constructor, which should perform the initialization?

The current wording would seem to make the current behavior a compiler bug (that cannot be fixed).

> To initialize a new closed class type, first a new set of static fields (§15.5.2) for that particular closed type is created. Each of the static fields is initialized to its default value (§15.5.5). Next, the static field initializers (§15.5.6.2) are executed for those static fields. Finally, the static constructor is executed.

I propose rewording to say that the static initializers are performed on entry to the static initializer (which is analogous to what we say for instance initializers vs instance constructors).  Alternatively, or in addition, in the paragraph that describes the behavior for an external static constructor should say that it is responsible for executing the field initializers.

PR #1308, opened 2025-04-15
Straw man to address extern constructor requirements

jskeet:
Towards #1253 and #158.

I don't expect this to be mergeable as-is, but I'm hoping that it's enough for us to *get* to something mergeable during the meeting.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -4861,7 +4861,7 @@ An instance constructor initializer cannot access the instance being created. Th
 
 ### 15.11.3 Instance variable initializers
 
-When an instance constructor has no constructor initializer, or it has a constructor initializer of the form `base(...)`, that constructor implicitly performs the initializations specified by the *variable_initializer*s of the instance fields declared in its class. This corresponds to a sequence of assignments that are executed immediately upon entry to the constructor and before the implicit invocation of the direct base class constructor. The variable initializers are executed in the textual order in which they appear in the class declaration ([§15.5.6](classes.md#1556-variable-initializers)).
+When a non-extern instance constructor has no constructor initializer, or it has a constructor initializer of the form `base(...)`, that constructor implicitly performs the initializations specified by the *variable_initializer*s of the instance fields declared in its class. This corresponds to a sequence of assignments that are executed immediately upon entry to the constructor and before the implicit invocation of the direct base class constructor. The variable initializers are executed in the textual order in which they appear in the class declaration ([§15.5.6](classes.md#1556-variable-initializers)).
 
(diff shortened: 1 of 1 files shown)


standard/classes.md
```diff
@@ -4863,6 +4863,8 @@ An instance constructor initializer cannot access the instance being created. Th
 
 When an instance constructor has no constructor initializer, or it has a constructor initializer of the form `base(...)`, that constructor implicitly performs the initializations specified by the *variable_initializer*s of the instance fields declared in its class. This corresponds to a sequence of assignments that are executed immediately upon entry to the constructor and before the implicit invocation of the direct base class constructor. The variable initializers are executed in the textual order in which they appear in the class declaration ([§15.5.6](classes.md#1556-variable-initializers)).
 
+Variable initializers are not required to be executed by extern instance constructors.
```
Nigel-Ecma, line 4866: Is it worth making it clear that the previous para only applies to non-extern constructors in that para? E.g. maybe start it:

> When a non-extern instance constructor…

suitably wordsmithed of course!

standard/classes.md
```diff
@@ -5084,6 +5086,8 @@ If a class contains the `Main` method ([§7.1](basic-concepts.md#71-application-
 
 To initialize a new closed class type, first a new set of static fields ([§15.5.2](classes.md#1552-static-and-instance-fields)) for that particular closed type is created. Each of the static fields is initialized to its default value ([§15.5.5](classes.md#1555-field-initialization)). Next, the static field initializers ([§15.5.6.2](classes.md#15562-static-field-initialization)) are executed for those static fields. Finally, the static constructor is executed.
 
+The requirement for the static constructor to be executed *after* the static field initializers does not apply to external static constructors.
```
Nigel-Ecma: This doesn’t read to me as removing the requirement to execute the field initializers. I think the previous para may need to be changed.

> To initialize a new closed class type, first a new set of static fields ([§15.5.2](classes.md#1552-static-and-instance-fields)) for that particular closed type shall be created. Each of the static fields shall be initialized to its default value ([§15.5.5](classes.md#1555-field-initialization)). Following this:
> - If there is either no static constructor or a non-extern static constructor then:
>   - the static field initializers ([§15.5.6.2](classes.md#15562-static-field-initialization)) shall be executed for those static fields;
>    - then the non-extern static constructor, if any, shall be executed.
> - Otherwise if there is an extern constructor it shall be executed.

That will need some wordsmithing. Yes, I changed from *is* to *shall*, debate away :-)

jnm2 (approved):

BillWagner (approved):
This LGTM.

PR #1309, opened 2025-04-15
Add notnull example

### LINKED ISSUE #1256: Come up with an example-only version of #830
We think all the normative text in #830 is already covered, but we *might* want the example.

@jskeet to come up with a PR which is just the example.

jskeet:
Fixes #1256

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -489,6 +489,39 @@ The ***not null*** constraint specifies that a type argument used for the type p
 
 Because `notnull` is not a keyword, in *primary_constraint* the not null constraint is always syntactically ambiguous with *class_type*. For compatibility reasons, if a name lookup ([§12.8.4](expressions.md#1284-simple-names)) of the name `notnull` succeeds it is treated as a `class_type`. Otherwise it is treated as the not null constraint.
 
+> *Example*: Consider the following:
+>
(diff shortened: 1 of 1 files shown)


Nigel-Ecma (approved):
LGTM

Only suggestion would be to extend the scope of the PR to the previous line (490) where the tense is wrong – it has “it is treated” where it should be some form of *shall* (as this is a spec not a description of compiler behaviour).

standard/classes.md
```diff
@@ -489,6 +489,39 @@ The ***not null*** constraint specifies that a type argument used for the type p
...
+> public class B2<T> where T : C? { }
+> class Test
+> {
+>     static void M()
+>     {
+>         // nonnull constraint allows nonnullable struct type argument
+>         A<int> x1;
+>         // warning: nonnull constraint prohibits nullable struct type argument
+>         A<int?> x2;
+>         // nonnullconstraint allows nonnullable class type argument
```
jnm2:
```suggestion
>         // nonnull constraint allows nonnullable class type argument
```

jnm2 (approved):

standard/classes.md
jskeet: Already done :)

jskeet:
This is a good start, but we should also add examples for `class?` and interaction with type arguments which are themselves type parameters. Will leave #1256 open for that purpose.

Issue #1062, opened 2024-03-27
Example in 18.4.6 is incorrect: calling an ambiguous-looking interface member can work...


jskeet:
We have the following example in 18.4.6 (which isn't tested as it's marked for "needs review"):

```csharp
interface IList
{
    int Count { get; set; }
}

interface ICounter
{
    void Count(int i);
}

interface IListCounter : IList, ICounter {}

class C
{
    void Test(IListCounter x)
    {
        x.Count(1);             // Error
        x.Count = 1;            // Error: cannot assign to a method group
        ((IList)x).Count = 1;   // Ok, invokes IList.Count.set
        ((ICounter)x).Count(1); // Ok, invokes ICounter.Count
    }
}
```

With text:

> In the following code [...] the first two statements cause compile-time errors because the member lookup ([§12.5](expressions.md#125-member-lookup)) of `Count` in `IListCounter` is ambiguous.

However, the first of these statements *works* when running the example tester. It's only the second which fails (CS1656: Cannot assign to 'Count' because it is a 'method group').

Has interface member lookup changed in a way that has made the first statement *become* valid, or has the spec always just been wrong?

jskeet:
Interestingly, if the type of the `Count` property in `IList` is changed to `Action<int> Count { get; set; }`, then it *still* doesn't complain even though `x.Count(1)` could either be a method invocation or a property access followed by a delegate invocation.

KalleOlaviNiemitalo:
FWIW, "Microsoft (R) Visual C# Compiler version 4.8.9037.0 for C# 5" warns about the ambiguity:

```
Class1.cs(18,11): warning CS0467: Ambiguity between method 'ICounter.Count(int)' and non-method 'IList.Count'. Using method group.
Class1.cs(8,10): (Location of symbol related to previous warning)
Class1.cs(3,9): (Location of symbol related to previous warning)
Class1.cs(18,9): error CS1656: Cannot assign to 'Count' because it is a 'method group'
```

Microsoft (R) Visual C# Compiler version 2.10.0.0 (b9fb1610) doesn't.  Roslyn source code has it commented out: <https://github.com/dotnet/roslyn/blob/958f2354c4d83dbb0e7723d0a8079a0dfbc33f25/src/Compilers/CSharp/Portable/Errors/ErrorCode.cs#L321>

[...]

PR #1313, opened 2025-04-17
Option 2 for resolving #1062

jskeet:
This uses two properties instead of a property and a method.

### DIFF
--- a/standard/interfaces.md
+++ b/standard/interfaces.md
@@ -388,7 +388,7 @@ For interfaces that are strictly single-inheritance (each interface in the inher
 > 
 > interface ICounter
 > {
->     void Count(int i);
+>     int Count { get; set; }
 > }
 >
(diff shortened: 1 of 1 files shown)


jnm2 (approved):
I lean toward this one due to the simplicity of demonstrating the concept above it.

BillWagner (approved):
I prefer this option to #1314 as well.

The reason for the ambiguity is clear, without a cross-reference to the method vs. property rule. It's focused on what this example is meant to focus on.

jskeet:
With two votes for this, I reckon if anyone else speaks up in favor, we can just merge this, abandon the other, and mark the issue as closed.

@Nigel-Ecma @gafter @ericlippert thoughts?

Nigel-Ecma (approved):

jskeet:
Closed in favor of #1313.

PR #1314, opened 2025-04-17
Option 1 to resolve #1062

### LINKED ISSUE #1062: Example in 18.4.6 is incorrect: calling an ambiguous-looking interface member can work...

jskeet:
This keeps the "method vs property" aspect - which introduces a little more complexity due to member lookup rules.

### DIFF
--- a/standard/interfaces.md
+++ b/standard/interfaces.md
@@ -397,7 +397,6 @@ For interfaces that are strictly single-inheritance (each interface in the inher
 > {
 >     void Test(IListCounter x)
 >     {
->         x.Count(1);             // Error
 >         x.Count = 1;            // Error
 >         ((IList)x).Count = 1;   // Ok, invokes IList.Count.set
 >         ((ICounter)x).Count(1); // Ok, invokes ICounter.Count
@@ -405,7 +404,9 @@ For interfaces that are strictly single-inheritance (each interface in the inher
 > }
 > ```
 >
-> the first two statements cause compile-time errors because the member lookup ([§12.5](expressions.md#125-member-lookup)) of `Count` in `IListCounter` is ambiguous. As illustrated by the example, the ambiguity is resolved by casting `x` to the appropriate base interface type. Such casts have no run-time costs—they merely consist of viewing the instance as a less derived type at compile-time.
+> the first statement causes a compile-time error because the member lookup ([§12.5](expressions.md#125-member-lookup)) of `Count` in `IListCounter` is ambiguous. As illustrated by the example, the ambiguity is resolved by casting `x` to the appropriate base interface type. Such casts have no run-time costs—they merely consist of viewing the instance as a less derived type at compile-time.
+>
+> The cast to `ICounter` in the final statement isn't strictly required. `x.Count(1);` would be unambiguous due to being an invocation; member lookup excludes non-invocable members (such as the `Count` property in the example) when the member is being invoked.
 >
 > *end example*
 <!-- markdownlint-disable MD028 -->


BillWagner (changes requested):
I selected "request changes" because I prefer the solution in #1313 over this one.

Nigel-Ecma (dismissed):

RexJaeschke:
Closed in favor of #1313.

standard/interfaces.md
```diff
@@ -397,15 +397,16 @@ For interfaces that are strictly single-inheritance (each interface in the inher
...
 >         ((ICounter)x).Count(1); // Ok, invokes ICounter.Count
 >     }
 > }
 > ```
 >
-> the first two statements cause compile-time errors because the member lookup ([§12.5](expressions.md#125-member-lookup)) of `Count` in `IListCounter` is ambiguous. As illustrated by the example, the ambiguity is resolved by casting `x` to the appropriate base interface type. Such casts have no run-time costs—they merely consist of viewing the instance as a less derived type at compile-time.
+> the first statement causes a compile-time error because the member lookup ([§12.5](expressions.md#125-member-lookup)) of `Count` in `IListCounter` is ambiguous. As illustrated by the example, the ambiguity is resolved by casting `x` to the appropriate base interface type. Such casts have no run-time costs—they merely consist of viewing the instance as a less derived type at compile-time.
+>
+> The cast to `ICounter` in the final statement isn't strictly required. `x.Count(1);` would be unambiguous due to being an invocation; member lookup excludes non-invocable members (such as the `Count` property in the example) when the member is being invoked.
```
jnm2, line 409: @jskeet This is interesting. Roslyn docs claim that this member lookup behavior being talked about here is a _deviation_ from the spec: <https://github.com/dotnet/roslyn/blob/main/docs/compilers/CSharp/Deviations%20from%20Standard.md#member-lookup>

I took a look at <https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#125-member-lookup> which it links to, and it does look to me so far as though the spec does not exclude non-invocable members.

jskeet: It's in 12.5.1: https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#1251-general

> A member lookup of a name `N` with `K `type arguments in a type `T` is processed as follows:
> ...
> - Next, if the member is invoked, all non-invocable members are removed from the set.

That comes earlier than the "Finally, having removed hidden members" part quotes in the deviations doc.

I'm not sure whether I'm missing something.

(I think it would be interesting to check the deviations doc in terms of "preference for methods" vs "preference for invocable members" - I *think* a property which returns an `Action` would still count as invocable.)

jnm2: Ah! Thanks for the pointer. The deviation is then correct; the compiler has a preference for methods, and I was conflating that with the invocability point. The method preference deviation can be seen here without invocability being used to narrow anything:

```cs
using System;

default(I).M(); // binds to I1.M (method)
Action y = default(I).M; // binds to I1.M (method)

interface I1 { void M() { } }
interface I2 { Action M => null; }
interface I : I1, I2 { }
```

jskeet: Eek, right - yes, the `Action y` part does indeed look like a deviation. But I think we're okay with where *we've* gone in the standard :)

jnm2: One note for the far future: they did not continue this deviation when resolving extensions in the new extensions feature.

PR #1323, opened 2025-05-07
Unconstrained type parameters and `??`

### LINKED ISSUE #737: Null-Coalescing Changes in V8
@BillWagner 

This issue interacts with PR https://github.com/dotnet/csharpstandard/pull/609, “Add support for Null-coalescing assignment”.

## Background

1.	Historically, compound-assignment operators have been defined in terms of their non-assignment counterparts. For example, in §[11.19.3](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#11193-compound-assignment), “Compound assignment,” the spec says, “An operation of the form x «op»= y is processed by applying binary operator overload resolution (§11.4.5) as if the operation was written x «op» y.” and then goes on to talk about assignment specifics.
2.	The V4 standard introduced the null-coalescing operator, `??` (§[11.14](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#1114-the-null-coalescing-operator) in V7); however, support for `??=` was *not* included.
3.	V8 adds support for `??=`. Using the [MS spec for this feature](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/null-coalescing-assignment.md), Rex created Draft PR 609.
4.	V8 also has an [empty placeholder MS spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/unconstrained-null-coalescing.md) for allowing unconstrained type parameters with `??` and `??=`. Here’s a simple example:

```csharp
private static void Mx<T>(T a, T b)
{
    Console.WriteLine(a ?? b);
    Console.WriteLine(a ??= b);
}
```

## Current Situation

More than a year after he wrote the PR 609 text, Rex recognized that PR and the unconstrained type parameters topic are related, and he is now trying to figure out what the spec should be for allowing unconstrained type parameters. His current thinking is:
1.	This feature addresses both `??` and `??=`.
2.	Rather than treating the addition of `??=` and supporting unconstrained type parameters as separate features, it seems appropriate to combine them into one PR.
3.	The new text re the handling of the unconstrained type parameters should go in the `??` operator section, and only if that impacts `??=` should it be mentioned in the compound-assignment section as well.
4.	In PR 609, the main edits for `??=` went into the “Assignment operators” sections; there were no changes to the `??` operator section itself. In hindsight, we should review the two sections to make sure they work together, and preserve the general notion that the compound-assignment operators be defined in terms of their non-assignment counterparts.
5.	As to the specific changes to support unconstrained type parameters, it seems this would (mostly?) involve the addition of a new, first, step to the bullet list in §[11.14](https://github.com/dotnet/csharpstandard/blob/draft-v7/standard/expressions.md#1114-the-null-coalescing-operator), “The null coalescing operator.”

## Bottom line

Rex needs someone to
1.	Write the words for the new step to support unconstrained type parameters, and to see if they need to overflow into the compound-assignment section.
2.	Review the edits in PR 609 and reconcile them with these new words; that is, to unify the two sets of edits.

BillWagner:
Fixes #737

Add the language to the null-coalescing operator so that `??` can be used with an unconstrained type parameter.

Notes for reviewers:

After looking at the final change, glance at the commit by commit history.

The only change is to remove the phrase "or a reference type" from the first bullet. By doing that, an unconstrained type parameter `T` is allowed. It's not a reference type.

I tested the case Nigel and Kalle described for using `??` with an unconstrained type parameter, and the existing rules produce the behavior they described.

I ran a few other cases to validate, and those worked correctly as well.

The commit-by-commit history shows what I first started to write by adding a bullet for unconstrained type parameters, then spec'ing that behavior. That started me on the path of just removing the restriction. I checked with the compiler team, and they agreed.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4828,8 +4828,7 @@ The type of the expression `a ?? b` depends on which implicit conversions are 
 - Otherwise, if `A` exists and an implicit conversion exists from `b` to `A`, the result type is `A`. At run-time, `a` is first evaluated. If `a` is not `null`, `a` becomes the result. Otherwise, `b` is evaluated and converted to type `A`, and this becomes the result.
 - Otherwise, if `A` exists and is a nullable value type, `b` has a type `B` and an implicit conversion exists from `A₀` to `B`, the result type is `B`. At run-time, `a` is first evaluated. If `a` is not `null`, `a` is unwrapped to type `A₀` and converted to type `B`, and this becomes the result. Otherwise, `b` is evaluated and becomes the result.
 - Otherwise, if `b` has a type `B` and an implicit conversion exists from `a` to `B`, the result type is `B`. At run-time, `a` is first evaluated. If `a` is not `null`, `a` is converted to type `B`, and this becomes the result. Otherwise, `b` is evaluated and becomes the result.
-
-Otherwise, `a` and `b` are incompatible, and a compile-time error occurs.
+- Otherwise, `a` and `b` are incompatible, and a compile-time error occurs.
 
 > *Example*:
 >
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4840,7 +4840,7 @@ The type of the expression `a ?? b` depends on which implicit conversions are 
 > int i = M(19, 23);
 > ```
 >
-> The parameters to `M` are unconstrained. Therefore, the type argument can be a reference type, or nullable value type, as shown in the first call to `M`. The type argument can also be a non-nullable value type, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
+> The type parameter `T` for method `M` is unconstrained. Therefore, the type argument can be a reference type, or nullable value type, as shown in the first call to `M`. The type argument can also be a non-nullable value type, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
 >
 > *end example*
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4841,7 +4841,7 @@ Otherwise, `a` and `b` are incompatible, and a compile-time error occurs.
 > int i = M(19, 23);
 > ```
 >
-> The parameters to `M` are unconstrained. Therefore, the type arguments can be reference types, or nullable value types, as shown in the first call to `M`. The type arguments can also be non-nullable value types, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
+> The parameters to `M` are unconstrained. Therefore, the type argument can be a reference type, or nullable value type, as shown in the first call to `M`. The type argument can also be a non-nullable value type, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
 >
 > *end example*
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4840,7 +4840,7 @@ The type of the expression `a ?? b` depends on which implicit conversions are 
 > int i = M(19, 23);
 > ```
 >
-> The type parameter `T` for method `M` is unconstrained. Therefore, the type argument can be a reference type, or nullable value type, as shown in the first call to `M`. The type argument can also be a non-nullable value type, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
+> The type parameter `T` for method `M` is unconstrained. Therefore, the type argument can be a reference type, or nullable value type, as shown in the first call to `M`. The type argument can also be a non-nullable value type, as shown in the second call to `M`. When the type argument is a non-nullable value type, the value of the expression `a ?? b` is `a`.
 >
 > *end example*
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4822,7 +4822,7 @@ The null coalescing operator is right-associative, meaning that operations are g
 
 The type of the expression `a ?? b` depends on which implicit conversions are available on the operands. In order of preference, the type of `a ?? b` is `A₀`, `A`, or `B`, where `A` is the type of `a` (provided that `a` has a type), `B` is the type of `b`(provided that `b` has a type), and `A₀` is the underlying type of `A` if `A` is a nullable value type, or `A` otherwise. Specifically, `a ?? b` is processed as follows:
 
-- If `A` exists and is not a nullable value type or a reference type, a compile-time error occurs.
+- If `A` exists and is not a nullable value type, a compile-time error occurs.
 - Otherwise, if `A` exists and `b` is a dynamic expression, the result type is `dynamic`. At run-time, `a` is first evaluated. If `a` is not `null`, `a` is converted to `dynamic`, and this becomes the result. Otherwise, `b` is evaluated, and this becomes the result.
 - Otherwise, if `A` exists and is a nullable value type and an implicit conversion exists from `b` to `A₀`, the result type is `A₀`. At run-time, `a` is first evaluated. If `a` is not `null`, `a` is unwrapped to type `A₀`, and this becomes the result. Otherwise, `b` is evaluated and converted to type `A₀`, and this becomes the result.
 - Otherwise, if `A` exists and an implicit conversion exists from `b` to `A`, the result type is `A`. At run-time, `a` is first evaluated. If `a` is not `null`, `a` becomes the result. Otherwise, `b` is evaluated and converted to type `A`, and this becomes the result.
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -4822,7 +4822,7 @@ The null coalescing operator is right-associative, meaning that operations are g
 
 The type of the expression `a ?? b` depends on which implicit conversions are available on the operands. In order of preference, the type of `a ?? b` is `A₀`, `A`, or `B`, where `A` is the type of `a` (provided that `a` has a type), `B` is the type of `b`(provided that `b` has a type), and `A₀` is the underlying type of `A` if `A` is a nullable value type, or `A` otherwise. Specifically, `a ?? b` is processed as follows:
 
-- If `A` exists and is an unmanaged type (§8.8) or a non-nullable value type, a compile-time error occurs.
+- If `A` exists and is an unmanaged type (§8.8) or known to be a non-nullable value type, a compile-time error occurs.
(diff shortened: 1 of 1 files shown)


BillWagner:
related to #609, but this can be merged separately.

standard/expressions.md
```diff
@@ -4822,7 +4822,7 @@ The null coalescing operator is right-associative, meaning that operations are g
 
 The type of the expression `a ?? b` depends on which implicit conversions are available on the operands. In order of preference, the type of `a ?? b` is `A₀`, `A`, or `B`, where `A` is the type of `a` (provided that `a` has a type), `B` is the type of `b`(provided that `b` has a type), and `A₀` is the underlying type of `A` if `A` is a nullable value type, or `A` otherwise. Specifically, `a ?? b` is processed as follows:
 
-- If `A` exists and is not a nullable value type or a reference type, a compile-time error occurs.
+- If `A` exists and is not a nullable value type, a compile-time error occurs.
```
Nigel-Ecma: I’ll review this properly later but for now did you mean:

```suggestion
- If `A` exists and is a non-nullable value type a compile-time error occurs.
```

as the current wording would *exclude* reference types? Should other types be invalid, e.g. pointer types?

jskeet: I share the same confusion as Nigel here. And I don't know offhand whether a type parameter that is constrained (e.g. `T where T : struct`) counts as "a non-nullable value type" or just "known to be a non-nullable value type".

jskeet (commented):
I'm hoping this will be simple to get agreed and merged in the meeting.

standard/expressions.md
BillWagner: Does `T where T : struct` counts as a non-nullable value type.  I think so. Here's a small test:

```csharp
int? number = null;
int? number2 = 4;
IsNullableValueTypeAStruct(number); // Error: CS0453 "The type 'int?' must be a non-nullable value type in order to use it as a parameter 'T' in the generic type or method 'NullCoaelescing.IsNullableValueTypeAStruct<T>(T)'"
NullCoalescingStruct(5, 12);
NullCoalescingStruct(number, number2); // // Error: CS0453

void IsNullableValueTypeAStruct<T>(T a) where T : struct  { }

void NullCoalescingStruct<T>(T a, T b) where T : struct 
{
    T result = a ?? b; // Error: CS0019 "Operator '??' cannot be applied to operands of type 'T' and 'T'"
}
```

BillWagner: I applied Nigel's suggestion and excluded unmanaged types (which includes pointer types). There is some duplication, but it feels more correct.

I unresolved the conversation because of Jon's question.

Nigel-Ecma (dismissed):
Jon’s question regarding whether “is a X type” includes type parameters constrained to X needs to be addressed. Semantically the answer is yes, but that is not clear in the text.

Further that unconstrained type parameters are allowed is maybe not obvious, or maybe that such an unconstrained type parameters can be instantiated with a non-nullable type at runtime and produce no errors (`a ?? b` in this case is `a` as `a` is not `null`).

I think “maybe not obvious” can probably be covered by note/example with x-refs and not require rewording the list.

BillWagner:
> Jon’s question regarding whether “is a X type” includes type parameters constrained to X needs to be addressed. Semantically the answer is yes, but that is not clear in the text.
> 
> Further that unconstrained type parameters are allowed is maybe not obvious, or maybe that such an unconstrained type parameters can be instantiated with a non-nullable type at runtime and produce no errors (`a ?? b` in this case is `a` as `a` is not `null`).
> 
> I think “maybe not obvious” can probably be covered by note/example with x-refs and not require rewording the list.

Updated the text and added an example.

standard/expressions.md
```diff
@@ -4831,6 +4831,20 @@ The type of the expression `a ?? b` depends on which implicit conversions are
...
+> ```csharp
+> T M<T>(T a, T b) => a ?? b;
+>
+> string s = M(null, "text");
+> int i = M(19, 23);
+> ```
+>
+> The parameters to `M` are unconstrained. Therefore, the type arguments can be reference types, or nullable value types, as shown in the first call to `M`. The type arguments can also be non-nullable value types, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
+>
+> *end example*
```
Nigel-Ecma: I don’t think this is there yet.

Bullets 2 thru 5 are all of the form “Otherwise, if `A` exists & <cond>… At runtime…”. This is a compile-time condition followed by the runtime actions. If `A` is an unconstrained type parameter then at compile-time it certainly exists (a type-parameter cannot be bound to a non-type), but what it is is not known until runtime. So this gets you to bullet 6, which is also where you end up if A does not exist (i.e. `a` is `null`), and this is incorrect for the dynamic case – it is correct for `null`.

There needs to be a bullet 5A (i.e. an inserted one), which covers runtime resolution and may be defined either from scratch, may include applying 2-5 at runtime, or whatever is needed.

This example can then refer to the rules and say which result is used in each case.

BTW I think line 4832 should be bullet 7.

Sorry, no time to attempt a starting point for this right now.

standard/expressions.md
```diff
@@ -4831,6 +4831,20 @@ The type of the expression `a ?? b` depends on which implicit conversions are
...
+>
+> <!-- Example: {template:"standalone-console-without-using", name:"DeclarationExpressions1"} -->
+> ```csharp
+> T M<T>(T a, T b) => a ?? b;
+>
+> string s = M(null, "text");
+> int i = M(19, 23);
+> ```
+>
+> The parameters to `M` are unconstrained. Therefore, the type arguments can be reference types, or nullable value types, as shown in the first call to `M`. The type arguments can also be non-nullable value types, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
```
```suggestion
> The parameters to `M` are unconstrained. Therefore, the type argument can be a reference type, or nullable value type, as shown in the first call to `M`. The type argument can also be a non-nullable value type, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
```

(Just because there's only a single type parameter, and therefore a single type argument.)

jskeet: For "the expression `a ?? b` is `a`" should we have "is evaluated equivalently to" or something similar?

Nigel-Ecma: @jskeet – No, if `a` is a non-nullable value type then `a` is non-null (think of the test as `(object)a == null` which for an IL target is a `box` and test) and so `a` is the result – it is not different handling *per se*. What is needed is to revise the bulleted list to cover this, see my earlier comment.

jskeet: My point is that the "is" here is a bit ambiguous. Perhaps "the value of the expression `a ?? b` is the value of `a`" or something like that? Just "is" seems wrong to me.

standard/expressions.md
BillWagner: I did add line 4832 to the bullet list.

The other concerns are ones I'd like to address in a new PR. The condition Nigel describes was already in the standard. A different task for this part makes sense to me.

standard/expressions.md
Nigel-Ecma: @jskeet – I see, your "is evaluated equivalently to" threw me and “is the value of” is a lot better. But that leaves us with the problem that this example is not supported by the text, we could say its aspirational! ;-)

standard/expressions.md
```diff
@@ -4822,14 +4822,27 @@ The null coalescing operator is right-associative, meaning that operations are g
...
+>
+> <!-- Example: {template:"standalone-console-without-using", name:"NullCoalescingGenerics"} -->
+> ```csharp
+> T M<T>(T a, T b) => a ?? b;
+>
+> string s = M(null, "text");
+> int i = M(19, 23);
+> ```
+>
+> The parameters to `M` are unconstrained. Therefore, the type argument can be a reference type, or nullable value type, as shown in the first call to `M`. The type argument can also be a non-nullable value type, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
```
```suggestion
> The type parameter `T` for method `M` is unconstrained. Therefore, the type argument can be a reference type, or nullable value type, as shown in the first call to `M`. The type argument can also be a non-nullable value type, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
```

standard/expressions.md
```diff
@@ -4822,14 +4822,27 @@ The null coalescing operator is right-associative, meaning that operations are g
 
 The type of the expression `a ?? b` depends on which implicit conversions are available on the operands. In order of preference, the type of `a ?? b` is `A₀`, `A`, or `B`, where `A` is the type of `a` (provided that `a` has a type), `B` is the type of `b`(provided that `b` has a type), and `A₀` is the underlying type of `A` if `A` is a nullable value type, or `A` otherwise. Specifically, `a ?? b` is processed as follows:
 
-- If `A` exists and is not a nullable value type or a reference type, a compile-time error occurs.
+- If `A` exists and is an unmanaged type (§8.8) or known to be a non-nullable value type, a compile-time error occurs.
```
jnm2, line 4825: Does "non-nullable value type" not subsume "unmanaged type"?

jnm2: Ah, I missed the pointers conversation.

standard/expressions.md
ericlippert: I concur Bill; if I recall correctly the intention of the C# 2.0 design was that `where T:struct` definitely ensures that `T` is a non-nullable value type.

jnm2 (approved):

standard/expressions.md
```diff
@@ -4822,14 +4822,27 @@ The null coalescing operator is right-associative, meaning that operations are g
...
+>
+> <!-- Example: {template:"standalone-console-without-using", name:"NullCoalescingGenerics"} -->
+> ```csharp
+> T M<T>(T a, T b) => a ?? b;
+>
+> string s = M(null, "text");
+> int i = M(19, 23);
+> ```
+>
+> The type parameter `T` for method `M` is unconstrained. Therefore, the type argument can be a reference type, or nullable value type, as shown in the first call to `M`. The type argument can also be a non-nullable value type, as shown in the second call to `M`. When the type argument is a non-nullable value type, the expression `a ?? b` is `a`.
```
```suggestion
> The type parameter `T` for method `M` is unconstrained. Therefore, the type argument can be a reference type, or nullable value type, as shown in the first call to `M`. The type argument can also be a non-nullable value type, as shown in the second call to `M`. When the type argument is a non-nullable value type, the value of the expression `a ?? b` is `a`.
```

jskeet (approved):

PR #1326, opened 2025-05-09
Add support for unconstrained type parameter annotations

RexJaeschke:
I think the following text might be affected by this PR (however, I have *not* included any changes to it in this initial Draft PR.)

> 15.2.5 Type parameter constraints
>
> The class type, reference type constraint, and secondary constraints can include the nullable type annotation. The presence or absence of this annotation on the type parameter indicates the nullability expectations for the type argument:
>
> - If the constraint does not include the nullable type annotation, the type argument is expected to be a non-nullable reference type. A compiler may issue a warning if the type argument is a nullable reference type.
> - If the constraint includes the nullable type annotation, the constraint is satisfied by both a non-nullable reference type and a nullable reference type.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -407,6 +407,7 @@ primary_constraint
     | 'struct'
     | 'notnull'
     | 'unmanaged'
+    | 'default'
     ;
 
 secondary_constraint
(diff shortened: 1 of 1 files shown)


BillWagner:
Closing in favor of #1470

PR #1335, opened 2025-05-31
Experimenting with grammar for arrays of NRTs

jnm2:


standard/types.md
```diff
@@ -59,10 +59,7 @@ array_type
 
 non_array_type
     : value_type
-    | class_type
-    | interface_type
-    | delegate_type
-    | 'dynamic'
+    | (class_type | interface_type | delegate_type | 'dynamic') nullable_type_annotation?
```
jnm2: Was trying not to end up creating new names like `non_array_non_nullable_reference_type` and `non_array_nullable_reference_type` which don't sound like core concepts we'd want to be referencing elsewhere.

standard/types.md
```diff
@@ -59,10 +59,7 @@ array_type
...
     | type_parameter
```
jnm2: I'm realizing `nullable_reference_type` does not include type parameters unless it's via a mechanism I'm not seeing. We'd also want to be able to `T?[]` where `T` is constrained to a class type.

standard/types.md
```diff
@@ -738,6 +735,8 @@ There are two forms of nullability for reference types:
 
 The syntactic distinction between a *nullable reference type* and its corresponding *non-nullable reference type* enables a compiler to generate diagnostics. A compiler must allow the *nullable_type_annotation* as defined in [§8.2.1](types.md#821-general). The diagnostics must be limited to warnings. Neither the presence or absence of nullable annotations, nor the state of the nullable context can change the compile time or runtime behavior of a program except for changes in any diagnostic messages generated at compile time.
 
+The only exception to this general rule is that adding a nullable annotation can change the syntactic meaning of the type of an array of arrays. Specifically, a type such as `A[][,]` is an array of two-dimensional arrays. That is because the syntactic meaning of the array dimensions in the array type are read from the inside out. However, by adding a nullable annotation between the dimensions, writing `A[]?[,]`, changes the parsing and therefore syntactic meaning of the type. In this case, the element type of the array is `A[]?`. The type is now a two-dimensional array of a nullable array type.
```
Nigel-Ecma: Nooo… The grammar *decides* the parsing, and hopefully the grammar supports the *designed* syntax and semantics. There is no “This is because…”

Something (better phrased) along the lines of “Array types containing nullable annotations after/between/before *(pick the right one)* ranks denote structurally different types to the syntactically similar array type without such annotations; this differs from all other uses of nullable annotations which do not change semantics.” Remember the Standard is not also a Rationale, some languages have published both but C# has not.

jnm2: (this is not new in my draft, this is Gafter's commit from <https://github.com/dotnet/csharpstandard/pull/1297>)

Nigel-Ecma: @jnm2 – Good to know, I'll await your new draft.

tools/GrammarTesting/Tests/Parsing/Samples/v8/Arrays of nullables/sample.cs
```diff
@@ -0,0 +1,36 @@
+class ArraysOfNullable
+{
+   class A { }
+
+   void ArraysOfNullable()
+   {
+      A[] p_1 = new A[5];    // Array of non-nullable A all elements inited to null (oops)
+      A[]? q_1 = null;       // Nullable array of non-nullable A, array inited to null
+      A?[] r_1 = new A[5];   // Array of nullable A all elements inited to null
+      A?[]? s_1 = null;      // Nullable array of nullable A, array inited to null
```
jnm2: @Nigel-Ecma Would you be able to show me how to work with files like this? This is from @gafter's PR and I copied it across, and I'd like to add some new cases when I work with it experimentally.

I don't yet understand what all can be done with this. The other folders appear to have had a tool run on them which generates a graphical representation of the parse. I'd like to know how to run it. Does a PR fail if the generated parse is out of date with the sample.cs file?

PR #1336, opened 2025-05-31
Fix broken example tester workflow on draft-v8, and make tools runnable out of the box

jnm2:
Fixes the following issues:

1. MSBuildLocator isn't needed (https://github.com/dotnet/roslyn/issues/78478#issuecomment-2856390151) and it is failing on my machine where I have only VS Preview installed. I believe MSBuildLocator by default doesn't see preview installations.

2. Add launch settings to the tools so that they are runnable immediately after cloning the repo by pressing F5 in Visual Studio. The launch settings contain the same arguments that are used for the workflow runs.

3. Fix the broken example tester workflow. <https://github.com/dotnet/csharpstandard/pull/1334#pullrequestreview-2886364073> caused it to start failing for the AnonymousFunctionsConv1 example because it pulled in Roslyn v4.14, which no longer reports a CS1661 diagnostic at the same location as a CS1678 error. (Discussed at https://github.com/dotnet/roslyn/pull/75400/files#r1866802220.) The workflow did not run for that PR, so this was not discovered until unrelated PRs started coming through.

4. Add more paths to the example tester workflow trigger to catch issues like the above next time.

Also, this PR adds path filters for the other PR-triggered workflows so that they only run when files change which can currently affect them.

@BillWagner You had [gitignored](https://github.com/dotnet/csharpstandard/pull/1179/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947) the launch profiles; what do you think about reversing this?

### DIFF
--- a/.github/workflows/do-not-merge-label-check.yml
+++ b/.github/workflows/do-not-merge-label-check.yml
@@ -28,7 +28,7 @@ jobs:
 
       - name: 'Check "${{ matrix.label }}" label'
         run: |
-          echo "::notice::Merging permission is diabled for PRs when the '${{ matrix.label }}' label is applied."
+          echo "::notice::Merging permission is disabled for PRs when the '${{ matrix.label }}' label is applied."
 
           if [ "${{ contains(github.event.pull_request.labels.*.name, matrix.label) }}" = "true" ]; then
             echo "::error::Pull request is labeled as '${{ matrix.label }}'. Please remove the label before merging."
--- a/.github/workflows/grammar-validator.yaml
+++ b/.github/workflows/grammar-validator.yaml
@@ -4,6 +4,11 @@ name: ANTLR Grammar validator
 on: 
   pull_request:
     types: [opened, synchronize, reopened]
+    paths:
+      - "standard/**"
+      - "tools/GrammarTesting/**"
+      - "tools/validate-grammar.sh"
+      - ".github/workflows/dependencies/GrammarTestingEnv.tgz"
   workflow_dispatch:
     inputs:
       reason:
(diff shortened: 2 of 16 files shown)


jnm2:
It's possible that removing the MSBuildLocator call will cause the example tester task (not triggered here) to fail with one example changing its expected warnings. I'd like to add to the GitHub Actions filter for this PR to cause the tester to run when the tester code is changed.

jnm2:
Sure enough, it's the removal of MSBuildLocator that causes the AnonymousFunctionsConv1 example to start producing only a single CS1661 error instead in place of two of the CS1661 errors that are being expected for that example.

I'll try to collect compiler versions before and after and see what's up.

jnm2:
The compiler version is the same both before and after the change to stop using MSBuildLocator: `4.14.0-3.25262.10 (8edf7bcd)`

It also turns out that this failure is unrelated to any of my PRs; the `draft-v8` branch is currently broken for the test extractor run. The last PR that was merged showed this same failure: https://github.com/dotnet/csharpstandard/actions/runs/15323346177/job/43111801142

jnm2:
I bisected and found the cause of the failure at https://github.com/dotnet/csharpstandard/pull/1334#pullrequestreview-2886364073. I'll resolve both issues in this PR: the lack of workflow trigger so that the workflow did not run for that PR, and the failure itself.

jnm2:
The failure was due to an intentional change in the Roslyn compiler v4.14, discussed at <https://github.com/dotnet/roslyn/pull/75400/files#r1866802220>. It doesn't make any difference for the purposes of the code example in the standard. I updated the expected diagnostics.

jskeet (approved):
LGTM, but would be good to get @BillWagner to check as well.

BillWagner (approved):
This LGTM as well. Thanks a lot @jnm2 This is really helpful.

Let's :shipit:

jnm2:
Some of the new path filters may need to be reverted on the workflows which are referenced by required checks: https://github.com/dotnet/csharpstandard/pull/1342

PR #1337, opened 2025-05-31
Enable running example tests massively in parallel via `dotnet test` or UI

jnm2:
`dotnet test tools/ExampleTester` now runs and produces the same output as `dotnet run tools/ExampleTester` (which still works), except that it will parallelize the tests and finish in a fraction of the time that Program.Main takes.

You can also click and run a given example in the UI of any .NET test runner, and see red and green states, and see output per test, add a few examples to a test session/playlist for repeated runs, etc. ...these are just standard unit tests.

I've noticed that the GitHub checks are all pretty fast except this one which takes about 9 minutes. We may want to consider switching over the GitHub Actions job to invoke the tests the new way. Even if the build server doesn't parallelize particularly well, simply overlapping the various stages of each test should produce benefits. (Loading csproj, getting compilation, getting diagnostics, emitting, running output.)

### DIFF
--- a/tools/ExampleTester/ExampleTester.csproj
+++ b/tools/ExampleTester/ExampleTester.csproj
@@ -5,6 +5,7 @@
     <TargetFramework>net8.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <GenerateProgramFile>false</GenerateProgramFile>
   </PropertyGroup>
 
   <ItemGroup>
--- a/tools/ExampleTester/ExampleTester.csproj
+++ b/tools/ExampleTester/ExampleTester.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
@@ -11,6 +11,9 @@
     <PackageReference Include="Microsoft.Build.Locator" Version="1.9.1" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.14.0" />
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="4.14.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.14.0" />
+    <PackageReference Include="NUnit" Version="4.3.2" />
+    <PackageReference Include="NUnit3TestAdapter" Version="5.0.0" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta4.22272.1" />
   </ItemGroup>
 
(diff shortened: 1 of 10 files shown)


tools/ExampleTester/ExampleTester.csproj
```diff
@@ -11,6 +11,9 @@
     <PackageReference Include="Microsoft.Build.Locator" Version="1.9.1" />
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.14.0" />
     <PackageReference Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="4.14.0" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.14.0" />
+    <PackageReference Include="NUnit" Version="4.3.2" />
+    <PackageReference Include="NUnit3TestAdapter" Version="5.0.0" />
```
jnm2, line 16: I could not figure out a way to get xUnit to parallelize between individual cases within a theory. Also, the Visual Studio Test Explorer UI for an xUnit theory does not show theory cases as individual nested tests, so they can't be individually run or added to playlists. NUnit shines in this area.

jnm2:
💭 We could consider having these tests read directly from the standard without going through the ExampleExtractor step. That could confer a few benefits such as possibly removing the use of the very slow MSBuildWorkspace, keeping cached Roslyn compilations around in memory that match each project template. This could result in _very_ fast tests. The downside would be that you wouldn't be able to open the .csproj files in an IDE to play around with things, if there is no `tmp` folder.

Edit: I was able to get the very fast runs without disrupting the flow via .csproj files on disk: <https://github.com/dotnet/csharpstandard/pull/1338>

BillWagner (approved):
This LGTM.

I'll also ping @jskeet for review offline. I think we can merge and update this without spending committee time.

jnm2:
@jskeet Would you be willing to take a look?

jskeet:
Yup, I'll look tomorrow.

tools/ExampleTester/GeneratedExample.cs
```diff
@@ -127,46 +131,52 @@ bool ValidateOutput()
                 ? new object[] { Metadata.ExecutionArgs ?? new string[0] }
                 : new object[0];
 
-            var oldOut = Console.Out;
             List<string> actualLines;
             Exception? actualException = null;
-            try
+            lock (ConsoleAccessLock)
```
jskeet: With this lock in place, do we effectively have "parallel build, sequential execution"? That sounds fine to me - I would imagine the build is the main timing bottleneck.

jskeet (approved):
Still coldy, but yes, I don't see problems with this so long as we still actually get consistent results :)

(We're generating everything into separate directories, so parallelizing that part should be fine - and if I understand correctly the actual execution is serial, so that shouldn't interfere.)

tools/ExampleTester/GeneratedExample.cs
jskeet: (We could potentially change this to CodeExecutionLock or something to make it clearer...)

jnm2: Yes, exactly. And thanks for the idea, I'll rename.

PR #1342, opened 2025-06-05
Revert path filters on required checks

jnm2:
Three of the four workflows which gained new path filters in https://github.com/dotnet/csharpstandard/pull/1336 are required checks. It appears that if a workflow is a required check, GitHub will not allow that check to be satisfied when the workflow's filter tells it not to be triggered:

image

### DIFF
--- a/.github/workflows/grammar-validator.yaml
+++ b/.github/workflows/grammar-validator.yaml
@@ -4,11 +4,6 @@ name: ANTLR Grammar validator
 on: 
   pull_request:
     types: [opened, synchronize, reopened]
-    paths:
-      - "standard/**"
-      - "tools/GrammarTesting/**"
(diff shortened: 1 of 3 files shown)


BillWagner:
@jnm2 Yes, that's correct. I've run into that before on other repos in the dotnet org. If a check is *required*. It must run and complete with a successful exit code. If it doesn't run because of a filter, that's considered "not completed", and therefore blocks merging.

A way to keep these reasonably efficient is to use the shell script to quickly exit (successfully) if nothing needs to be done. It's a balance to determine how much work is done to determine how much more work to do....   And, since the container has already been allocated and started, the savings may not be worth the effort and maintainence.

BillWagner (approved):
LGTM

PR #1343, opened 2025-06-09
Add support for variance safety for static interface members

RexJaeschke:
I just discovered this V9 feature spec [here](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/variance-safety-for-static-interface-members.md). To make sure this was actually implemented in V9, I compiled the following code:

```csharp
using System;
using System.Threading.Tasks;

public interface I<out T>
{
    static Task<T> F = Task.FromResult(default(T)); // No problem
    static Task<T> P => Task.FromResult(default(T));   //CS8904 in V8 but OK in V9
    static Task<T> M() => Task.FromResult(default(T));    //CS8904 in V8 but OK in V9
    static event EventHandler<T>? E; // CS8904 in V8 but OK in V9
}
```

The errors produced by V8 do *not* occur in V9.

### DIFF
--- a/standard/interfaces.md
+++ b/standard/interfaces.md
@@ -91,9 +91,9 @@ If a generic interface is declared in multiple parts ([§15.2.3](classes.md#1523
 
 #### 18.2.3.2 Variance safety
 
-The occurrence of variance annotations in the type parameter list of a type restricts the places where types can occur within the type declaration.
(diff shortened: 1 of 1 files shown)


BillWagner:
Closed in favor of #1471

PR #1349, opened 2025-06-13
Remove path filters for required workflows

BillWagner:
It's a good idea to save time and resources to not run those workflows that don't need to run.

However, that introduces a problem for workflows that are required before merging a PR. If the work flow doesn't run, it doesn't succeed.

@jnm2 We need a different way to short-circuit the work and report success when a check isn't necessary.


jnm2:
This duplicates https://github.com/dotnet/csharpstandard/pull/1342

BillWagner:
Closing in favor of #1342.  Thanks @jnm2

PR #1350, opened 2025-06-14
Remove assumptions about corresponding parameters

### LINKED ISSUE #981: Generic type inference assumes same number of arguments and parameters
#975 talks about corresponding parameters in one place, but the surrounding text seems to assume a 1:1 correspondence of arguments and parameters, in the same order. We need to look at everything that's required here - including but not limited to changing the references to Ti within the paragraph changed in #975.

jskeet:
The corresponding parameter for a method argument with index i isn't necessarily i.

An alternative approach would be to effectively consider type inferrence from the perspective of a normalized argument list.

Fixes #981

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -763,7 +763,7 @@ Type inference takes place in phases. Each phase will try to infer type argument
 
 #### 12.6.3.2 The first phase
 
-For each of the method arguments `Eᵢ` with a corresponding parameter with index `j` and type `Tⱼ`:
+For each of the method arguments `Eᵢ` with a corresponding parameter of type `Tⱼ`:
 
 - If `Eᵢ` is an anonymous function, an *explicit parameter type inference* ([§12.6.3.8](expressions.md#12638-explicit-parameter-type-inferences)) is made *from* `Eᵢ` *to* `Tⱼ`.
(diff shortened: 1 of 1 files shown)


standard/expressions.md
```diff
@@ -763,13 +763,13 @@ Type inference takes place in phases. Each phase will try to infer type argument
 
 #### 12.6.3.2 The first phase
 
-For each of the method arguments `Eᵢ`:
+For each of the method arguments `Eᵢ` with a corresponding parameter with index `j` and type `Tⱼ`:
```
jskeet: I don't know whether the "index `j`" part here is needed.

jnm2 (approved):

standard/expressions.md
BillWagner: I think this would work. If that's what you mean, I'd agree:

```suggestion
For each of the method arguments `Eᵢ` with a corresponding parameter of type `Tⱼ`:
```

BillWagner (approved):
This also LGTM @jskeet 

I made one suggestion to make sure I understand your comment.

standard/expressions.md
jskeet: Yes, certainly something like that. Let's see whether others are happy with that suggestion, then we can commit.

jskeet:
It looks like the text here has changed enough that it's possibly better to create a new PR. Will do so and get @BillWagner to review.

jskeet:
Closing in favor of #1407

PR #1359, opened 2025-06-26
Fix tuple assignment and tuple element type inference

BillWagner:
Fixes #1244: type inference from tuple element expressions

I followed the recommended text from [collection expressions](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/collection-expressions.md#type-inference) because we well need them in the future. I did break apart the first phase and the input type inference because tuple elements now needs subscripts as well. We could put them together, but the small `i` and `j` start to look the same.

PR #1366, opened 2025-07-03
Fix tuple assignment

### LINKED ISSUE #1155: Tuple conversions are incorrectly specified
**Describe the bug**

The compiler implements both a tuple type conversion, _and_ a tuple literal conversion from expression. The spec currently only covers that latter conversion. As a concrete example of a feature that explicitly designed for and allowed, but the spec says is illegal:

```cs
short s = 1;

(int, int) t1 = (s, s); // As said in the spec, this is a tuple expression conversion
(short, short) t2 = (s, s);
(int, int) t3 = t2; // By the spec, there shouldn't be a conversion here, but the compiler allows a tuple type conversion from (short, short) to (int, int)
```

Specifically, https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/conversions.md#10213-implicit-tuple-conversions is the only conversions portion of the specification that touches tuples, and it covers `tuple expressions`. These are defined by https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#1286-tuple-expressions. In `t3 = t2`, `t2` is not a tuple expression, but is instead a simple name: https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#1284-simple-names; since we're not dealing tuple expressions, implicit tuple conversions don't apply, and the spec says that no conversion should exist from `t2` to `t3`.

BillWagner:
Fixes #1155

This PR follows the process outlined in https://github.com/dotnet/csharpstandard/issues/1155#issuecomment-2284833989

This replaces the first commit in #1159.  

See https://github.com/dotnet/csharpstandard/pull/1359#pullrequestreview-2973055224 for the status of this PR.

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -345,7 +345,7 @@ In all cases, the rules ensure that a conversion is executed as a boxing convers
 
 ### 10.2.13 Implicit tuple conversions
 
-An implicit conversion exists from a tuple expression `E` to a tuple type `T` if `E` has the same arity as `T` and an implicit conversion exists from each element in `E` to the corresponding element type in `T`. The conversion is performed by creating an instance of `T`’s corresponding `System.ValueTuple<...>` type, and initializing each of its fields in order from left to right by evaluating the corresponding tuple element expression of `E`, converting it to the corresponding element type of `T` using the implicit conversion found, and initializing the field with the result.
+An implicit conversion exists from an expression `E` with a tuple type `S` to a tuple type `T` if `S` has the same arity as `T` and an implicit conversion exists from each element type in `S` to the corresponding element type in `T`.
 
 If an element name in the tuple expression does not match a corresponding element name in the tuple type, a warning shall be issued.
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -33,12 +33,12 @@ An ***instance accessor*** is a property access on an instance, an event access 
 
 ### 12.2.2 Values of expressions
 
-Most of the constructs that involve an expression ultimately require the expression to denote a ***value***. In such cases, if the actual expression denotes a namespace, a type, a method group, or nothing, a compile-time error occurs. However, if the expression denotes a property access, an indexer access, or a variable, the value of the property, indexer, or variable is implicitly substituted:
+Most of the constructs that involve an expression ultimately require the expression to denote a ***value***. In such cases, if the actual expression denotes a namespace, a type, a method group, or nothing, a compile-time error occurs. However, if the expression denotes a property access, an indexer access, a tuple, or a variable, the value of the property, indexer, tuple, or variable is implicitly substituted:
 
 - The value of a variable is simply the value currently stored in the storage location identified by the variable. A variable shall be considered definitely assigned ([§9.4](variables.md#94-definite-assignment)) before its value can be obtained, or otherwise a compile-time error occurs.
 - The value of a property access expression is obtained by invoking the get accessor of the property. If the property has no get accessor, a compile-time error occurs. Otherwise, a function member invocation ([§12.6.6](expressions.md#1266-function-member-invocation)) is performed, and the result of the invocation becomes the value of the property access expression.
 - The value of an indexer access expression is obtained by invoking the get accessor of the indexer. If the indexer has no get accessor, a compile-time error occurs. Otherwise, a function member invocation ([§12.6.6](expressions.md#1266-function-member-invocation)) is performed with the argument list associated with the indexer access expression, and the result of the invocation becomes the value of the indexer access expression.
-- The value of a tuple expression is obtained by applying an implicit tuple conversion ([§10.2.13](conversions.md#10213-implicit-tuple-conversions)) to the type of the tuple expression. It is an error to obtain the value of a tuple expression that does not have a type.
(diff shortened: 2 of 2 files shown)


Nigel-Ecma:
Carried over from #1159:

> My [possible fix](https://github.com/dotnet/csharpstandard/issues/1155#issuecomment-2284833989) to #1155 doesn’t deal with the case of a *tuple_expression* not having a type (RIP null type), and one without a type is not representable directly as a `ValueTuple<…>` as there is no type for one or more of the type params…
>
> A re-think is needed, till then I’ll mark this PR as draft

This PR is for that rethink…

jnm2:
@Nigel-Ecma If a tuple expression doesn't have a type, then I think we don't get to the part where `ValueTuple<...>` comes into play, because of this text:

> The value of a tuple expression is the value obtained by evaluating the tuple expression (§12.8.6). It is an error to obtain the value of a tuple expression that does not have a type.

Thus a tuple expression is only evaluated when it has a type, and thus a tuple expression can be evaluated by constructing the corresponding `ValueTuple<...>` type.

standard/expressions.md
```diff
@@ -1611,7 +1611,7 @@ A tuple expression has a type if and only if each of its element expressions `Ei
 
 A tuple expression is evaluated by evaluating each of its element expressions in order from left to right.
 
-A tuple value can be obtained from a tuple expression by converting it to a tuple type ([§10.2.13](conversions.md#10213-implicit-tuple-conversions)), by reclassifying it as a value ([§12.2.2](expressions.md#1222-values-of-expressions))) or by making it the target of a deconstructing assignment ([§12.21.2](expressions.md#12212-simple-assignment)).
+A tuple value is obtained from a tuple expression by evaluating it and storing the result in corresponding `System.ValueTuple<...>` type, and initializing each of its fields in order from left to right by evaluating the corresponding tuple element expression of `E`, converting it to the corresponding element type of `T` using the implicit conversion found, and initializing the field with the result.
```
jnm2, line 1614: System.ValueTuple isn't mentioned anywhere else in this file. Shouldn't we be a bit more specific? There are wrinkles in how tuples map to the ValueTuple types. For instance, a tuple type `(T1, T2, T3, T4, T5, T6, T7, T8)` must be represented as nested ValueTuples, because the TRest type argument of a tuple must always itself be a nested tuple: `ValueTuple<T1, T2, T3, T4, T5, T6, T7, ValueTuple<T8>>`

standard/expressions.md
```diff
@@ -33,12 +33,12 @@ An ***instance accessor*** is a property access on an instance, an event access
 
 ### 12.2.2 Values of expressions
 
-Most of the constructs that involve an expression ultimately require the expression to denote a ***value***. In such cases, if the actual expression denotes a namespace, a type, a method group, or nothing, a compile-time error occurs. However, if the expression denotes a property access, an indexer access, or a variable, the value of the property, indexer, or variable is implicitly substituted:
+Most of the constructs that involve an expression ultimately require the expression to denote a ***value***. In such cases, if the actual expression denotes a namespace, a type, a method group, or nothing, a compile-time error occurs. However, if the expression denotes a property access, an indexer access, a tuple, or a variable, the value of the property, indexer, tuple, or variable is implicitly substituted:
```
jnm2, line 36: Why do we mention tuple expressions here, but we don't mention other kinds of expressions such as object creation expressions, array creation expressions, lambda expressions, and so on?

standard/expressions.md
jnm2: I see that conversions.md does mention `ValueTuple<...>` a few times, but it doesn't explain the required recursive encoding.

standard/conversions.md
```diff
@@ -345,7 +345,7 @@ In all cases, the rules ensure that a conversion is executed as a boxing convers
 
 ### 10.2.13 Implicit tuple conversions
 
-An implicit conversion exists from a tuple expression `E` to a tuple type `T` if `E` has the same arity as `T` and an implicit conversion exists from each element in `E` to the corresponding element type in `T`. The conversion is performed by creating an instance of `T`’s corresponding `System.ValueTuple<...>` type, and initializing each of its fields in order from left to right by evaluating the corresponding tuple element expression of `E`, converting it to the corresponding element type of `T` using the implicit conversion found, and initializing the field with the result.
+An implicit conversion exists from an expression `E` with a tuple type `S` to a tuple type `T` if `S` has the same arity as `T` and an implicit conversion exists from each element type in `S` to the corresponding element type in `T`.
```
jnm2, line 348: I think I'm following, but just checking. _Is_ there a tuple type `S` for every tuple expression? For instance, does `(default, default)` have a tuple type `S` in:
```cs
(object, object) x = (default, default);
```

standard/expressions.md
jskeet: I vaguely remember that we discussed how far we needed to go with that encoding aspect.

*If* we can avoid referring to `ValueTuple<...>` in this section, it's likely to make our lives easier... even if it means doing a bit more work in 8.3.11 to effectively make that the *only* place that needs to go into that detail. If we can say "logically there's a type with N fields" then we can make this line (in expressions.md) simpler.

jskeet:
@Nigel-Ecma is working on a new PR - he will close this when opening the new one.

jskeet:
@BillWagner Do you think it's worth us closing this PR now?

BillWagner:
> @BillWagner Do you think it's worth us closing this PR now?

Yes, I'll close it now.

PR #1367, opened 2025-07-03
Define behavior when filters throw

### LINKED ISSUE #1086: Documentation for exception thrown within `when` expression
### Type of issue

Missing information

### Description

I could not find good documentation on the expected behavior if an exception is thrown within the `when` expression. I wanted to confirm whether any additional exception handling was necessary in this use case.

Based on a quick test, it appears that the exception is suppressed, and the associated exception handler does not execute. Can the behavior be documented?

### Page URL

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when

### Content source URL

https://github.com/dotnet/docs/blob/main/docs/csharp/language-reference/keywords/when.md

### Document Version Independent Id

15502bbc-5355-2add-0fa8-ae31dc01c122

### Article author

@BillWagner

### Metadata

* ID: 6ef10b29-fbf8-5af9-8f8b-a694a1416e33 
* Service: **dotnet-csharp**
* Sub-service: **lang-reference**

BillWagner:
Define the behavior when an exception_filter expression throws an exception.

The roslyn compiler generates a `filter` block for the exception filter. The runtime produces this behavior for a filter block.

Fixes #1086


BillWagner:
I think we should be able to merge this asynchronously.  

I tagged everyone, but the scope is small enough that a subset should be enough.

One question is if we should say anything about the CLI defined behavior for a `filter` block.

Nigel-Ecma (approved):
LGTM

jnm2 (approved):

jskeet (approved):
I'm surprised by this behavior - it means the "filter exception" itself is completely silent, presumably - but I think it's fine to specify as-is.

KalleOlaviNiemitalo:
StackOverflowException caused by an actual stack overflow is not ignored in the filter.  Does the wording allow this?

```csharp
using System;

class Program
{
#if NETFRAMEWORK
    [System.Runtime.ExceptionServices.HandleProcessCorruptedStateExceptions]
    [System.Security.SecurityCritical]
#endif
    private static void Main()
    {
        try
        {
            try
            {
                Console.WriteLine("Throwing");
                throw new ApplicationException("vacation");
            }
            catch (Exception exception) when (BlowTheStack())
            {
                Console.WriteLine($"Caught inner: {exception}");
            }
        }
        catch (Exception exception)
        {
            Console.WriteLine($"Caught outer: {exception}");
        }
    }

    private static bool BlowTheStack()
    {
        // Make sure it can't be optimized as a tail call.
        return !BlowTheStack();
    }
}
```

jskeet:
@KalleOlaviNiemitalo: I think StackOverflowException (and other "uncatchable" exceptions) count as special cases beyond the scope of the standard - I don't *believe* we specify that it can't be caught "normally" either.

KalleOlaviNiemitalo:
Thread abort on .NET Framework (but neither ThreadAbortException nor Thread.Abort() is in the C# standard):

```csharp
using System;

class Program
{
    private static void Main()
    {
        try
        {
            try
            {
                try
                {
                    Console.WriteLine("Throwing");
                    throw new ApplicationException("vacation");
                }
                catch (Exception exception) when (Abort())
                {
                    Console.WriteLine($"Caught inner: {exception}");
                }
            }
            catch (Exception exception)
            {
                Console.WriteLine($"Caught middle: {exception}");
            }
        }
        catch (Exception exception)
        {
            Console.WriteLine($"Caught outer: {exception}");
        }
    }

    private static bool Abort()
    {
        Console.WriteLine("Aborting");
        System.Threading.Thread.CurrentThread.Abort();
        Console.WriteLine("Aborted");
        return true;
    }
}
```

```
Throwing
Aborting
Caught middle: System.ApplicationException: vacation
   kohteessa Program.Main()
Caught outer: System.Threading.ThreadAbortException: Säiettä oltiin keskeyttämässä.
   kohteessa Program.Main()
```

So:

* In the inner `catch`, it discarded the ThreadAbortException, treated `when (Abort())` as `when (false)`, and did not enter the exception handler.
* In the middle `catch`, it caught the ApplicationException and entered the exception handler.  When exiting the exception handler, it noticed the unacknowledged thread abort and threw ThreadAbortException again.
* In the outer `catch`, it caught the ThreadAbortException and entered the exception handler. When exiting the exception handler, I suppose it threw ThreadAbortException again, but that was not output.

KalleOlaviNiemitalo:
The use of StackOverflowException is in the specification

<https://github.com/dotnet/csharpstandard/blob/1f7e66fdb99c369259132aef5edfec1bc083f6f7/standard/exceptions.md?plain=1#L52>

If this were removed (but the type definition in Annex C were kept), then I think an implementation could document that a stack overflow just terminates the process without throwing StackOverflowException.

jskeet:
> The use of StackOverflowException is in the specification

Its existence and cause is in the specification - but not the fact that it can't be caught. (Or rather, when it's caught it's automatically rethrown - I *think* that's what happens.) It's not clear to me how much detail we want to go into there.

jnm2:
What's not even clear to me is whether the silent swallowing of the exception in the exception filter is a C# thing specifically, or just happens to be a CLR thing which may vary in other runtimes (e.g. when compiled to WebAssembly).

PR #1368, opened 2025-07-03
Add text for `readonly` member

### LINKED ISSUE #928: [C# 8.0] 12.6.6.1 should not require a defensive copy to invoke a `readonly` member
To fix #894, a defensive copy is required to an `in` parameter of a non-`readonly` struct type to invoke a member.

C# 8 adds `readonly` members to non-readonly structs. A defensive copy isn't required to invoke a `readonly` member.

See #927 (desccription)

BillWagner:
Fixes #928

The PR for readonly members in structs did not add text that a temporary copy of `this` wasn't required to invoke a readonly member. (The clause in structs did state that a copy wasn't required, but that left the spec in an inconsistent state.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1217,7 +1217,7 @@ The run-time processing of a function member invocation consists of the followin
   - `M` is invoked.
 - Otherwise, if the type of `E` is a value-type `V`, and `M` is declared or overridden in `V`:
   - `E` is evaluated. If this evaluation causes an exception, then no further steps are executed. For an instance constructor, this evaluation consists of allocating storage (typically from an execution stack) for the new object. In this case `E` is classified as a variable.
-  - If `E` is not classified as a variable, or if `V` is not a readonly struct type ([§16.2.2](structs.md#1622-struct-modifiers)), and `E` is one of:
+  - If `E` is not classified as a variable, or if `V` is not a readonly struct type ([§16.2.2](structs.md#1622-struct-modifiers)) and the declaration of `M` does not include the `readonly` modifier (§16.4.12), and `E` is one of:
     - an input parameter ([§15.6.2.3.2](classes.md#156232-input-parameters)), or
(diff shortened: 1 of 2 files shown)


jnm2 (approved):

standard/expressions.md
```diff
@@ -1217,7 +1217,7 @@ The run-time processing of a function member invocation consists of the followin
   - `M` is invoked.
 - Otherwise, if the type of `E` is a value-type `V`, and `M` is declared or overridden in `V`:
   - `E` is evaluated. If this evaluation causes an exception, then no further steps are executed. For an instance constructor, this evaluation consists of allocating storage (typically from an execution stack) for the new object. In this case `E` is classified as a variable.
-  - If `E` is not classified as a variable, or if `V` is not a readonly struct type ([§16.2.2](structs.md#1622-struct-modifiers)), and `E` is one of:
+  - If `E` is not classified as a variable, or if `V` is not a readonly struct type ([§16.2.2](structs.md#1622-struct-modifiers)) and the declaration of `M` does not include the `readonly` modifier (§16.4.12), and `E` is one of:
```
jskeet: It's a little odd to refer to "a readonly struct type" but then talk about the declaration of M "including the readonly modifier", both in the same sentence. Is this because we can't say "readonly method" as it could be a property or indexer? Would "readonly member" work?

In particular, I think that would be slightly helpful in terms of there being multiple places where the modifier can occur for properties - either at the property declaration itself, or the accessor declaration. If we haven't already defined "readonly member" or "readonly function member" perhaps we could do so, then use it here?

I'd be fine to merge as-is and maybe have a clarity follow-up for this though.

BillWagner: Actually, great catch. Property access is covered under 12.8.7.1. Commit coming with a similar change there. 

I'd thought I was being clever by thinking an accessor could be considered a "function member". Better to put the change in 12.8.7.1 and chose more clear language here.

jskeet (approved):

PR #1386, opened 2025-07-31
Allow arrays of nullable reference types

### LINKED ISSUE #1385: Arrays of nullable references
It looks as though `string?[]` is not allowed by the current grammar, or any `X?[]` when `X` is a class_type, interface_type, delegate_type, 'dynamic', or type_parameter:

https://github.com/dotnet/csharpstandard/blob/286b10a4a7dfe9001cb2634b511ef4c4c36bb562/standard/types.md?plain=1#L56-L68

This extends the problem beyond arrays of nullable arrays, to arrays of nullable references of any kind.

jnm2:
Proposed replacement for @Nigel-Ecma's https://github.com/dotnet/csharpstandard/pull/1287 and @gafter's https://github.com/dotnet/csharpstandard/pull/1297

Will fix #1385

I was trying not to end up creating new names like `non_array_non_nullable_reference_type` and `non_array_nullable_reference_type` which don't sound like core concepts we'd want to be referencing elsewhere. (Cf https://github.com/dotnet/csharpstandard/pull/1287)

TODO:

- [ ] Add samples for the grammer tester. (Getting help from @Nigel-Ecma offline. I would like to create tests that show that `string?[]` and `int[]?[]` were disallowed prior to the grammar changes in this PR.)

### DIFF
--- a/standard/arrays.md
+++ b/standard/arrays.md
@@ -54,9 +54,7 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
 
 The nullable annotation `?` may be placed on an array type, as in `T[R]?`. Such an annotated array type may be used as the element type of another array type, as in `T[R]?[R₂]`.
 
-`T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner *array_type* has ranks `[R₁][R₂]`, read left to right.
-
-> *Note*: This is the sole exception to the general rule that the meaning of a program remains the same when nullable reference types annotations are removed. *end note*
+`T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner *array_type* has ranks `[R₁][R₂]`, read left to right. This is the sole exception to the general rule that the meaning of a program remains the same when nullable reference types annotations are removed.
 
 Every reference type which contains nullable annotations has a corresponding unannotated type with no semantic difference (§8.9.1). The corresponding unannotated type for an array of nullable arrays is a single array type which recursively collects all the ranks of all the nested *array_type*s.
 
--- a/standard/arrays.md
+++ b/standard/arrays.md
@@ -52,9 +52,9 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
 
 ### §arrays-of-nullable-arrays Arrays of nullable arrays
 
-The nullable annotation `?` may be placed on an array type, as in `T[R]?`. Such an array type may be used as the element type of another array type, as in `T[R]?[R₂]`.
+The nullable annotation `?` may be placed on an array type, as in `T[R]?`. Such an annotated array type may be used as the element type of another array type, as in `T[R]?[R₂]`.
 
-The intervening nullable annotation `?` separates the grammar into multiple *array_types*. `T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner *array_type* has ranks `[R₁][R₂]`, read left to right.
+`T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner *array_type* has ranks `[R₁][R₂]`, read left to right.
 
 > *Note*: This is the sole exception to the general rule that the meaning of a program remains the same when nullable reference types annotations are removed. *end note*
 
--- a/standard/arrays.md
+++ b/standard/arrays.md
@@ -58,7 +58,14 @@ The nullable annotation `?` may be placed on an array type, as in `T[R]?`. Such 
 
 Every reference type which contains nullable annotations has a corresponding unannotated type with no semantic difference (§8.9.1). The corresponding unannotated type for an array of nullable arrays is a single array type which recursively collects all the ranks of all the nested *array_type*s.
 
-The unannotated array type of an array of nullable arrays cannot be found by simply removing the nullable annotations `?` from the grammar and reparsing. This is because array ranks are read left to right while nested *array_type* productions are read outside-in, with outer array type ranks to the right, inner array type ranks to the left. Thus, the type `T[R₁][R₂]?[R₃][R₄]` has an unannotated array type of `T[R₃][R₄][R₁][R₂]`. To obtain the unannotated array type of an array of nullable arrays, first take the ranks on the outermost array type in order from left to right, then move to the array type inside the nullable element type and take its ranks in order from left to right. Repeat until the element type is no longer a nullable array type. Then take this remaining element type and place on it all the collected ranks in order from first to last to obtain the unannotated array type.
+The unannotated array type of an array of nullable arrays cannot be found by simply removing the nullable annotations `?` from the grammar and reparsing. This is because array ranks are read left to right while nested *array_type* productions are read outside-in, with outer array type ranks to the right, inner array type ranks to the left. Thus, the type `T[R₁][R₂]?[R₃][R₄]` has an unannotated array type of `T[R₃][R₄][R₁][R₂]`.
+
+To obtain the unannotated array type of an array of nullable arrays, the following steps are followed:
+
+1. Take the ranks on the outermost array type in order from left to right. (From `T[R₁][R₂]?[R₃][R₄]`, take `[R₃]` and then `[R₄]`.)
+2. Move to the array type inside the nullable element type and take its ranks in order from left to right. (From `T[R₁][R₂]`, take `[R₁]` and then `[R₂]`.)
+3. Repeat in this fashion until the element type is no longer a nullable array type. (`T`)
+4. Take this remaining element type and place on it all the collected ranks in order from first to last to obtain the unannotated array type. (On `T`, place `[R₃]`, then `[R₄]`, then `[R₁]`, then `[R₂]`, obtaining the final result of `T[R₃][R₄][R₁][R₂]`.)
 
 > *Example*:
 >
--- a/standard/arrays.md
+++ b/standard/arrays.md
@@ -50,6 +50,33 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
 
 > *Note*: Following the rules of [§17.6](arrays.md#176-array-covariance), the value may also be a reference to a covariant array type. *end note*
 
+### §arrays-of-nullable-arrays Arrays of nullable arrays
+
+The nullable annotation `?` may be placed on an array type, as in `T[R]?`. Such an array type may be used as the element type of another array type, as in `T[R]?[R₂]`.
+
+The intervening nullable annotation (`?`) separates the grammar into multiple *array_types*. `T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner array type has ranks `[R₁][R₂]`, read left to right.
+
+> *Note*: This is the sole exception to the general rule that the meaning of a program remains the same when nullable reference types annotations are removed. *end note*
+
+Every reference type which contains nullable annotations has a corresponding unannotated type with no semantic difference (§8.9.1). The corresponding unannotated type for an array of nullable arrays is a single array type which recursively collects all the ranks of all the nested *array_type*s.
+
+The unannotated array type of an array of nullable arrays cannot be found by simply removing the nullable annotations `?` from the grammar and reparsing. This is because array ranks are read left to right while nested *array_type* productions are read outside-in, with outer array type ranks to the right, inner array type ranks to the left. Thus, the type `T[R₁][R₂]?[R₃][R₄]` has an underlying array type of `T[R₃][R₄][R₁][R₂]`. To obtain the underlying array type of an array of nullable arrays, first take the ranks on the outermost array type in order from left to right, then move to the array type inside the nullable element type and take its ranks in order from left to right. Repeat until the element type is no longer a nullable array type. Then take this remaining element type and place on it all the collected ranks in order from first to last to obtain the unannotated array type.
+
+> *Example*:
+>
+> The following table demonstrates the effect on the underlying array type caused by breaking up array types by inserting nullable annotations:
+>
+> | Annotated                        | Underlying                                |
+> |----------------------------------|-------------------------------------------|
+> | `T?[][,][,,]`                    | `T[][,][,,]` (not intervening, no change) |
+> | `T[][,][,,]?`                    | `T[][,][,,]` (not intervening, no change) |
+> | `T[]?[,]?[,,]`                   | `T[,,][,][]`                              |
+> | `T[]?[,][,,]`                    | `T[,][,,][]`                              |
+> | `T[][,]?[,,]`                    | `T[,,][][,]`                              |
+> | `T[][,]?[,,][,,,]?[,,,,][,,,,,]` | `T[,,,,][,,,,,][,,][,,,][][,]`            |
+>
+> *end example*
+
 ### 17.2.2 The System.Array type
 
 The type `System.Array` is the abstract base type of all array types. An implicit reference conversion ([§10.2.8](conversions.md#1028-implicit-reference-conversions)) exists from any array type to `System.Array` and to any interface type implemented by `System.Array`. An explicit reference conversion ([§10.3.5](conversions.md#1035-explicit-reference-conversions)) exists from `System.Array` and any interface type implemented by `System.Array` to any array type. `System.Array` is not itself an *array_type*. Rather, it is a *class_type* from which all *array_type*s are derived.
--- a/standard/arrays.md
+++ b/standard/arrays.md
@@ -54,19 +54,19 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
 
 The nullable annotation `?` may be placed on an array type, as in `T[R]?`. Such an array type may be used as the element type of another array type, as in `T[R]?[R₂]`.
 
-The intervening nullable annotation (`?`) separates the grammar into multiple *array_types*. `T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner array type has ranks `[R₁][R₂]`, read left to right.
+The intervening nullable annotation `?` separates the grammar into multiple *array_types*. `T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner *array_type* has ranks `[R₁][R₂]`, read left to right.
 
 > *Note*: This is the sole exception to the general rule that the meaning of a program remains the same when nullable reference types annotations are removed. *end note*
 
 Every reference type which contains nullable annotations has a corresponding unannotated type with no semantic difference (§8.9.1). The corresponding unannotated type for an array of nullable arrays is a single array type which recursively collects all the ranks of all the nested *array_type*s.
 
-The unannotated array type of an array of nullable arrays cannot be found by simply removing the nullable annotations `?` from the grammar and reparsing. This is because array ranks are read left to right while nested *array_type* productions are read outside-in, with outer array type ranks to the right, inner array type ranks to the left. Thus, the type `T[R₁][R₂]?[R₃][R₄]` has an underlying array type of `T[R₃][R₄][R₁][R₂]`. To obtain the underlying array type of an array of nullable arrays, first take the ranks on the outermost array type in order from left to right, then move to the array type inside the nullable element type and take its ranks in order from left to right. Repeat until the element type is no longer a nullable array type. Then take this remaining element type and place on it all the collected ranks in order from first to last to obtain the unannotated array type.
+The unannotated array type of an array of nullable arrays cannot be found by simply removing the nullable annotations `?` from the grammar and reparsing. This is because array ranks are read left to right while nested *array_type* productions are read outside-in, with outer array type ranks to the right, inner array type ranks to the left. Thus, the type `T[R₁][R₂]?[R₃][R₄]` has an unannotated array type of `T[R₃][R₄][R₁][R₂]`. To obtain the unannotated array type of an array of nullable arrays, first take the ranks on the outermost array type in order from left to right, then move to the array type inside the nullable element type and take its ranks in order from left to right. Repeat until the element type is no longer a nullable array type. Then take this remaining element type and place on it all the collected ranks in order from first to last to obtain the unannotated array type.
 
 > *Example*:
 >
-> The following table demonstrates the effect on the underlying array type caused by breaking up array types by inserting nullable annotations:
+> The following table demonstrates the effect on the unannotated array type caused by breaking up array types by inserting nullable annotations:
 >
-> | Annotated                        | Underlying                                |
+> | Annotated                        | Unannotated                                |
 > |----------------------------------|-------------------------------------------|
 > | `T?[][,][,,]`                    | `T[][,][,,]` (not intervening, no change) |
 > | `T[][,][,,]?`                    | `T[][,][,,]` (not intervening, no change) |
--- a/standard/arrays.md
+++ b/standard/arrays.md
@@ -66,7 +66,7 @@ The unannotated array type of an array of nullable arrays cannot be found by sim
 >
 > The following table demonstrates the effect on the unannotated array type caused by breaking up array types by inserting nullable annotations:
 >
-> | Annotated                        | Unannotated                                |
+> | Annotated                        | Unannotated                               |
 > |----------------------------------|-------------------------------------------|
 > | `T?[][,][,,]`                    | `T[][,][,,]` (not intervening, no change) |
 > | `T[][,][,,]?`                    | `T[][,][,,]` (not intervening, no change) |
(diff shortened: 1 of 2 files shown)


standard/types.md
```diff
@@ -54,16 +54,13 @@ interface_type
     ;
 
 array_type
-    : non_array_type rank_specifier+
+    : array_type nullable_type_annotation rank_specifier+
+    | non_array_type rank_specifier+
```
jnm2, line 58: This has the property that when parsing `string?[][,]?[,,][,,,]?` we end up with two `array_type` nodes: one being `string?[][,]?[,,][,,,]`, and one being the inner `string?[][,]`.

@Nigel-Ecma It's not mutual left recursion! 😁

standard/types.md
```diff
@@ -54,16 +54,13 @@ interface_type
...
     ;
 
 non_array_type
     : value_type
-    | class_type
-    | interface_type
-    | delegate_type
-    | 'dynamic'
-    | type_parameter
+    | (class_type | interface_type | delegate_type | 'dynamic' | type_parameter) nullable_type_annotation?
```
jnm2: This fixes the issue where `string?[]` and other arrays of reference-typed non-arrays were disallowed by the grammar.

jnm2: Q: Do we specify anywhere that `T?[]` or even just generally `T?` is not allowed when `T` is a type parameter with no class or struct or class type constraint?

Nigel-Ecma: This isn’t the style usually used in the grammar and would be better handled by leaving *non_array_type* untouched and either using `non_array_type nullable_type_annotation?` in the places where it is needed, making a rule for `non_array_type nullable_type_annotation?` and using that rule's name wherever it is needed.

standard/types.md
Nigel-Ecma, line 58: Yes, no MLR, well done ;-)

This change impacts §17.2 which now needs to be revised.

standard/types.md
```diff
@@ -732,7 +729,25 @@ There are two forms of nullability for reference types:
...
+
+A compiler must respect the effect that *nullable_type_annotation* has on the ordering of array rank specifiers. Whereas `A[][,]` is a single *array_type* with two *rank_specifier*s, the presence of a nullable annotation between the rank specifiers in `A[]?[,]` causes it to no longer be a single *array_type*, but rather two: an outer *array_type* with a single *rank_specifier* of `[,]`, and an element type of `A[]?` which is a *nullable_reference_type* containing an inner *array_type* with a single *rank_specifier* of `[]`. Because of this, `A[]?[,]` and `A[,][]` are represented by the same underlying type, while `A[]?[,]` and `A[][,]` are not.
```
Nigel-Ecma: - This is not the place to define the semantics of array types, that is §17.2
- The final sentence doesn’t cover the consequences of having the same underlying, which presumably exist or the statement is pointless. Are there implications for array creation and conversions which will need to addressed in the respective clauses?
- There are “tonal” issues here:
   - A implementation must support the syntax and semantics of the language – rather and “respect the effect of”
   - A token has no power *per se*, the grammar states where a token is allowed and the semantics specifies the meaning of the grammar's phrases/sentences – rather then “the presence of a <token> … causes”
- What does need to go here is a simple statement that nullable annotations can effect the semantics with a cross reference to where that effect is defined (probably just somewhere in §17 at present, but the future may bring others now the “no effect” status has gone)
   - This change impacts at least one other place in the Standard – see #1242 – but there may be more…
   - I am assuming here that you do not intend to use the “conditionally normative” device to maintain the no effect status in the bulk of the Standard. If you do this last point & sub-point need to be considered in that light.

standard/types.md
```diff
@@ -732,7 +729,25 @@ There are two forms of nullability for reference types:
...
+> class C
+> {
+>     void M(string[][,]?[,,][,,,] arrays)
+>     {
+>         string? value = arrays[3, 3, 3][4, 4, 4, 4]?[1][2, 2];
+>     }
+> }
+> ```
+>
+> *end example*
```
Nigel-Ecma: As with the above this needs to be in §17 somewhere. Giving examples of how adding nullable annotations to examples already in §17 could help the understanding, e.g. In §17.2.1 there is:

> *Example*: The type in `T[][,,][,]` is a single-dimensional array of three-dimensional arrays of two-dimensional arrays of `int`. *end example*

In these terms what is `T[]?[,,][,]`, `T[]?[,,][,]` and maybe even `T[]?[,,]?[,]`?

An example demonstrating the differences, in the same terms as above, between `T[][,][,,][,,,][,,,,][,,,,,]` and `T[][,]?[,,][,,,]?[,,,,][,,,,,]` might be helpful (multiple annotations, multiple ranks between & around them)

Nigel-Ecma (changes requested):
Nothing wrong with being concise; however unfortunately some of the conciseness is in the wrong places, and there are concerns this isn’t complete coverage of the feature.

standard/types.md
jnm2: I'd be happy to learn the style.

`non_array_type nullable_type_annotation?` has a problem, in that it allows `int??`. `non_array_type` may be `value_type` which may be `nullable_value_type`.

Nigel-Ecma: @jnm2 – If you look at the existing grammar (`grammar.md`) you will see there are no uses of `( a | … | b ) c`. It is of course quite valid, and I’ve written similar myself before now only to remove them as the audience seems to prefer:

```
x : a
  | … 
  | b
  ;
y : x c
  ;
```

which would also address your long line warning.

BTW Antlr treats a parenthesised group as a rule in its own right, so they're a good way to accidentally introduce MLR, Antlr is “fun” ;-)

standard/types.md
jnm2: Updated!

standard/types.md
jnm2: Updated! I do not intend to use the conditionally-normative device.

standard/types.md
jnm2: Updated!

jnm2:
I have hit my time limit for the week but will try next week to understand the tools to set up grammar tests and run them.

standard/arrays.md
```diff
@@ -50,6 +50,33 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
 
 > *Note*: Following the rules of [§17.6](arrays.md#176-array-covariance), the value may also be a reference to a covariant array type. *end note*
 
+### §arrays-of-nullable-arrays Arrays of nullable arrays
+
+The nullable annotation `?` may be placed on an array type, as in `T[R]?`. Such an array type may be used as the element type of another array type, as in `T[R]?[R₂]`.
+
+The intervening nullable annotation `?` separates the grammar into multiple *array_types*. `T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner *array_type* has ranks `[R₁][R₂]`, read left to right.
```
Nigel-Ecma: Saying nullable annotation “separates the grammar” is incorrect, from a grammar perspective the rule allows an *array_type* to contain another *array_type*.

standard/arrays.md
```diff
@@ -50,6 +50,33 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
...
+
+> *Note*: This is the sole exception to the general rule that the meaning of a program remains the same when nullable reference types annotations are removed. *end note*
```
Nigel-Ecma: This is a normative part of the spec so cannot be a *Note* (which is just informative).

You also need to change wherever this “general rule” is relied upon or mentioned (e.g. §8.9.5.3) because the “exception” invalidates the rule – it can in general never be relied upon.

standard/arrays.md
```diff
@@ -50,6 +50,33 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
...
+
+Every reference type which contains nullable annotations has a corresponding unannotated type with no semantic difference (§8.9.1). The corresponding unannotated type for an array of nullable arrays is a single array type which recursively collects all the ranks of all the nested *array_type*s.
```
Nigel-Ecma, line 59: With array types there is not in general a single “corresponding nullable type” (§8.9.1) – there are multiple, e.g.:

```csharp
static void ArrayEquivalentTypes(int[][,]?[,,]?[,,,]?[,,,,] a1)
{
    int[,,][][,]?[,,,]?[,,,,] a2 = a1;

    int[,,,][,,][][,]?[,,,,] a3_1 = a1;
    int[,,,][,,][][,]?[,,,,] a3_2 = a2;

    int[,,,,][,,,][,,][][,] a4_1 = a1;
    int[,,,,][,,,][,,][][,] a4_2 = a2;
    int[,,,,][,,,][,,][][,] a4_3_1 = a3_1;
    int[,,,,][,,,][,,][][,] a4_3_2 = a3_2;

    int[,,][][,]?[,,,]?[,,,,] b3_1 = a3_1;
    int[,,][][,]?[,,,]?[,,,,] b3_2 = a3_2;
    int[,,][][,]?[,,,]?[,,,,] b4_1 = a4_1;
    int[,,][][,]?[,,,]?[,,,,] b4_2 = a4_2;
    int[,,][][,]?[,,,]?[,,,,] b4_3_1 = a4_3_1;
    int[,,][][,]?[,,,]?[,,,,] b4_3_2 = a4_3_2;

    int[,,,][,,][][,]?[,,,,] c4_1 = a4_1;
    int[,,,][,,][][,]?[,,,,] c4_2 = a4_2;
    int[,,,][,,][][,]?[,,,,] c4_3_1 = a4_3_1;
    int[,,,][,,][][,]?[,,,,] c4_3_2 = a4_3_2;

    // etc...
}
```

This bundle of joy ;-) has four array types that are semantically equivalent to each other, and that’s not the limit by a long shot.

It needs to be specified that any two types (which need not be distinct) selected from a semantically equivalent set are implicitly convertible to each other. (Depending on implementation some, but not all, of the conversions may elicit a nullable warning.)

And while we are here, consider:

```csharp
static void ArrayDifferentTypesSameLiteral()
{
    int[]?[,] a1 =
        {
            {   new int[] { 1, 2 },
                new int[] { 3, 4 }
            }
        };
    int[,][] a2 =
        {
            {   new int[] { 1, 2 },
                new int[] { 3, 4 }
            }
        };
```

Different array types which are semantically equivalent can be init’ed using the same array literal. That will need to be specified.

standard/arrays.md
```diff
@@ -50,6 +50,33 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
...
+
+The unannotated array type of an array of nullable arrays cannot be found by simply removing the nullable annotations `?` from the grammar and reparsing. This is because array ranks are read left to right while nested *array_type* productions are read outside-in, with outer array type ranks to the right, inner array type ranks to the left. Thus, the type `T[R₁][R₂]?[R₃][R₄]` has an unannotated array type of `T[R₃][R₄][R₁][R₂]`. To obtain the unannotated array type of an array of nullable arrays, first take the ranks on the outermost array type in order from left to right, then move to the array type inside the nullable element type and take its ranks in order from left to right. Repeat until the element type is no longer a nullable array type. Then take this remaining element type and place on it all the collected ranks in order from first to last to obtain the unannotated array type.
```
Nigel-Ecma: I’m not yet convinced regarding the algorithm presentation, maybe bullets/numbered steps or pseudo-code would help?

Regardless at minimum I think the example should show the result of each iteration

standard/types.md
```diff
@@ -729,25 +737,7 @@ There are two forms of nullability for reference types:
...
-> {
->     void M(string[][,]?[,,][,,,] arrays)
->     {
->         string? value = arrays[3, 3, 3][4, 4, 4, 4]?[1][2, 2];
->     }
-> }
-> ```
->
-> *end example*
+The syntactic distinction between a *nullable reference type* and its corresponding *non-nullable reference type* enables a compiler to generate diagnostics. A compiler must allow the *nullable_type_annotation* as defined in [§8.2.1](types.md#821-general). The diagnostics must be limited to warnings. Neither the presence or absence of nullable annotations, nor the state of the nullable context can change the compile time or runtime behavior of a program except for changes in any diagnostic messages generated at compile time, with one exception: arrays of nullable arrays are not parsed as a single *array_type*, but rather as multiple nested *array_type*s. The corresponding *non-nullable reference type* of an array of nullable arrays is not the single array type that would be parsed if the nullable annotations were removed; see §arrays-of-nullable-arrays.
```
Nigel-Ecma, line 740: I will surprise everybody ;-) and say “compiler” -> ”implementation”.

Not changing the compile time behavior is no longer correct, add/remove a nullable annotation on an array type without changing all associated index operations will produce compile-time errors.

I’m not sure you can say changing the number of *array_type* productions in the parse is an exception *per se* – the annotations do not change the described array shape in anyway (one might argue that the existing description of arrays is less than clear on the shape, if so adding nullable arrays is the time to fix that).

Nigel-Ecma (changes requested):
Unfortunately I think there is quite a way to go yet.

Overall reading this I get a sense that too much is being placed on the number of occurrences of *array_type*/the change in shape of the array – these are really illusory changes (whether we think they should be or not) due to the erasure of nullability – meaning is not changing, even though adding/removing nullable annotations has the very non-illusory impact of generating compile time errors!

Trying to think of an illustration the best I came up with is language translation:

> The new road avoids the swamp.

In Irish is:

> Seachnaíonn an bóthar nua an portach.

These have the same meaning even if the structure has changed – translate the Irish word-by-word and you get:

> Avoids the road new the swamp.

Arrays of nullable arrays doesn’t change the meaning of the code, what it changes is how the code must be written – e.g. rearrange indices (avoids the road new the swamp) – to achieve the same meaning.

Is this PR getting this across? Should it?

BillWagner:
Overall, I like the approach of #1297 better. I find it easier to follow the concepts. (But that's not a strong preference if others prefer this direction).

Nigel-Ecma:
@BillWagner – unfortunately #1297 has issues, this one starts with a better grammar which binds the nullable annotation the right way – arrays used to have elements of non-array type, now they can have elements of nullable array type (but still not array type), and the grammar models that. There is still some way to go here both on arrays themselves, conversions, and in removing the “nullable annotations do not effect semantics” guarantee (as @jnm2 does not wish to use the “conditionally normative” device as a way out of that). However its not worth pursuing #1297 as an alternative.

standard/arrays.md
jnm2: I thought perhaps that the preceding paragraph _was_ the exception to that rule, and was normative, so the note here seemed redundant to me but was simply being included to drive the point home. However, I'm happy to make it normative if that seems right.

standard/types.md
jnm2: I don't know about taking on the change of compiler -> implementation. "Compiler" is used consistently 31 times in the surrounding context, and the sentence you're commenting on precedes this PR; the term is only shown with a green background in the diff because I tacked some text on to the end of the line.

image

jnm2: > Not changing the compile time behavior is no longer correct, add/remove a nullable annotation on an array type without changing all associated index operations will produce compile-time errors.

The "with one exception" part already addresses this, doesn't it? It is correct that the compile time behavior does not change, with that one exception of adding or removing a nullable annotation on an array type... right?

jnm2: > I’m not sure you can say changing the number of _array_type_ productions in the parse is an exception _per se_ – the annotations do not change the described array shape in anyway (one might argue that the existing description of arrays is less than clear on the shape, if so adding nullable arrays is the time to fix that).

Here is why I think that it is an exception _per se_. The general rule is that the presence or absence of nullable annotations does not change the compile time or runtime behavior. Thus, the removal of `?` that is present in a C# program would generally not affect compile time or runtime behavior. However, the removal of `?` from `T[]?[,]` does affect compile time and runtime behavior. The causal mechanism by which the presence of the `?` has meaning is the parse of the grammar; with the `?`, the parse has two *array_type*s, and without the `?`, the parse has only one *array_type*. The meaning of the resulting compile time behavior and runtime type are defined in terms of the nested structure (if any) of those *array_type*(s) produced by the parse.

standard/arrays.md
jnm2: Updated.

standard/arrays.md
jnm2: Updated.

standard/arrays.md
jnm2: Updated.

jskeet:
Note to myself as much as anything else: I've read through the change, and don't feel competent to express an opinion on it.

jskeet:
Applied the C# 8 milestone, but if it's not ready, we should *at least* consider #1287.

standard/arrays.md
jnm2: I think that both things are already specified when we say there is no semantic difference between an array of nullable arrays and its corresponding unannotated single-array type.

The spec as it is currently drafted in this PR already defines there to be no semantic difference between all of the array types in the examples you share above. If there is no semantic difference, then it's not even a matter of conversion.

Likewise, the existing spec does not explicitly address conversion between `string` and `string?`. It simply relies on "there is no semantic difference." If that is good enough for `string` and `string?`, it's good enough for `string[]?[,]` and `string[,][]`.

The specification only talks about type conversions for the purpose of providing warnings, and so the stuff about type conversions and nullability is conditionally normative. §8.9.5.3 Type conversions:

> A compiler that generates diagnostic warnings conforms to these rules.
>
> > *Note:* Differences in top-level or nested nullability annotations in types do not affect whether conversion between the types is permitted, since there is no semantic difference between a non-nullable reference type and its corresponding nullable type ([§8.9.1](types.md#891-general)). *end note*

jnm2:
@Nigel-Ecma If there is more progress to be made in this PR before our next meeting related to our discussion threads, it is now dependent on those threads continuing.

standard/arrays.md
Nigel-Ecma: I really don't think you can fallback on that – what types are semantically equivalent to each other is not always obvious (think two different arrays of nullable arrays with the same underlying type). And when it comes to array initialisation, starting with the type of the array who does one figure out the shape of the initialiser? Are other shaped initialiser valid?

All you might need is some examples.

standard/types.md
Nigel-Ecma: The “exception” is that adding/removing nullable annotations to array declarations changes the semantics of the array and the order in which array indices must be given when addressing array elements. An implementation which chooses not to implement nullable warnings, and otherwise ignores nullable annotations, must still implement the semantic changes nullable annotations in array declarations produce. The Standard needs to be absolutely clear on this. See overall review comment.

Nigel-Ecma (dismissed):
This PR needs at least two more reviewers to take a thorough look and determine whether what is here is sufficient to clearly define the semantics, what array types are semantically equivalent and what impacts (or not) that has, what needs to be said about array initialisation, what is required about the impact of this feature on the previous “the semantic neutrality” cornerstone NRT; or whatever questions other reviews think are important!

I've made no detailed suggestions as to changes but will mark this as “Request changes” as a hold.

To me this PR does not seem right to me yet but when two or more others are comfortable with it, and the grammar checked over (probably my job), that should do it.

jskeet:
I'd be happy to add a suggestion for the part around "an implementation which chooses not to implement nullable warnings" - which I agree needs to be clear. I'm still not competent to review the rest of the change though.

standard/arrays.md
```diff
@@ -50,6 +50,38 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
...
 
+### §arrays-of-nullable-arrays Arrays of nullable arrays
+
+The nullable annotation `?` may be placed on an array type, as in `T[R]?`. Such an annotated array type may be used as the element type of another array type, as in `T[R]?[R₂]`.
+
+`T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner *array_type* has ranks `[R₁][R₂]`, read left to right. This is the sole exception to the general rule that the meaning of a program remains the same when nullable reference types annotations are removed.
+
+Every reference type which contains nullable annotations has a corresponding unannotated type with no semantic difference (§8.9.1). The corresponding unannotated type for an array of nullable arrays is a single array type which recursively collects all the ranks of all the nested *array_type*s.
```
jskeet, line 59: Reading this again, I don't think I understand at this point what "which recursively collects all the ranks of all the nested *array_type*s" means - I *think* this is just a matter of adding "as described below" but I'm not sure.

standard/arrays.md
jskeet: I think we've previously had some description of "equivalent types" (not just semantically equivalent) and there's some language around conversions. We should make sure we're using that here. @BillWagner may be able to remember; if not I'll see if I can dig it up. (The intention was to avoid having to state parts about implicit reference conversions all over the place.)

standard/types.md
```diff
@@ -54,17 +54,22 @@ interface_type
...
 
 array_type
-    : non_array_type rank_specifier+
+    : array_type nullable_type_annotation rank_specifier+
+    | non_array_type rank_specifier+
     ;
 
 non_array_type
-    : value_type
+    : non_array_non_nullable_type nullable_type_annotation?
```
jskeet, line 62: I'm definitely not competent enough to review this grammar change - @Nigel-Ecma you've not commented on the grammar, and I don't know whether that's because you're happy with it, or whether it's because you wanted to get the prose settled first.

standard/types.md
jskeet: I definitely agree it needs to be really clear - and while I think Joseph's change tries to *make* it clear (not skirting around it) but that there may be alternative formulations which would work better. I'm happy to try that - potentially multiple times - when we're happy with the rest, but it should probably be the last thing we do.

standard/arrays.md
```diff
@@ -50,6 +50,38 @@ At run-time, a value of an array type can be `null` or a reference to an instanc
 
 > *Note*: Following the rules of [§17.6](arrays.md#176-array-covariance), the value may also be a reference to a covariant array type. *end note*
 
+### §arrays-of-nullable-arrays Arrays of nullable arrays
+
+The nullable annotation `?` may be placed on an array type, as in `T[R]?`. Such an annotated array type may be used as the element type of another array type, as in `T[R]?[R₂]`.
+
+`T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner *array_type* has ranks `[R₁][R₂]`, read left to right. This is the sole exception to the general rule that the meaning of a program remains the same when nullable reference types annotations are removed.
```
jskeet, line 57: I'm uneasy about this, because it's so drastic. I think it gets to the right place, but I wonder if there's a way we can be more consistent between nullable and non-nullable array types, such that we end up with the same number of types involved, and you "just" need to parse the text differently to get at them.

As Nigel puts in his comments, if we don't have the shape of arrays-of-arrays adequately specified yet, let's do that now. (I can't work out whether we do or not, but it's certainly pretty terse at the moment.)

Currently it looks like the bulk of the specification of arrays-of-arrays is within the "general" section (17.2.1) - I wonder whether having a new subsection of "arrays of arrays" which bifurcates into "when there are no nullable annotations" and "when there are nullable annotations" parts would be useful. We should make sure it explicitly tackles nullable value types as well.

I'm happy to come up with an alternative PR based on this one if that sounds like a good approach (or at least one worth exploring) to both @jnm2 and @Nigel-Ecma - but I'd like feedback on whether my preference of trying to make the difference in the number of types involved *only* about how you read it, rather than a "deeper" number of types sounds feasible or not before I try.

My expectation is that in order to make sure I fully understand what I'm doing, I'll write a lot of concrete examples (using string and int, with and without nullable annotations), and then wrap explanations around those examples.

If folks would like me to do that, I won't even get to starting it this week, but I could probably have a very rough version available shortly before the next meeting, so we could at least discuss whether it's worth pursuing.

jskeet (commented):
As noted on the line 57 comment, I think it's worth seeing if there's an approach built on this (thanks @jnm2 for this PR as-is, and we may not be able to improve on it anyway) but with a slightly less drastic difference. @Nigel-Ecma and @jnm2 please take a look at my comments to check if they even make sense, and if you'd like me to have a go at another option.

standard/arrays.md
Nigel-Ecma: Warning: *I think* I’ve always got “array type” and *array_type* right, but this is of course self-proofed… So if you think I’ve put the wrong one in anywhere I may well have!

---

This PR has two issues to tackle:

1. That array elements can now be other arrays – but that was always true…
2. That it blows away the NRT guarantee that there is no semantic impact of nullable annotations, now there is.

@jskeet you are talking about (1) and reference §17.2.1. This is the (original!) clause where things can get confusing. It starts with:

> An array type is written as a *non_array_type* followed by one or more *rank_specifiers*.
>
> A *non_array_type* is any type that is not itself an *array_type*.

Here there is both “array type” and “*array_type*”, are they the same?

The usual description of arrays in C# is that the element type is not an array (*non_array_type*) with one or more ranks, and of which have have multiple dimensions. But then the clause continues:

> The element type of an array type is the type that results from deleting the leftmost *rank_specifier*:
>
> - An array type of the form `T[R]` is an array with rank `R` and a non-array element type `T`.
> - An array type of the form `T[R][R₁]...[Rₓ]` is an array with rank `R` and an element type `T[R₁]...[Rₓ]`.

Which (clearly?) defines that the element type of a multi-rank is in fact an array type… So an array type with *x* ranks consists of *x* array types…

To this the PR adds:

> `T[R₁][R₂]?[R₃][R₄]` is not a single *array_type* with four ranks. Rather, it is two *array_type*s, each of which has two ranks. The outer *array_type* has ranks `[R₃][R₄]`, read left to right, and its element type is `T[R₁][R₂]?`. The element type is another *array_type* with a nullable annotation, and this inner *array_type* has ranks `[R₁][R₂]`, read left to right. This is the sole exception to the general rule that the meaning of a program remains the same when nullable reference types annotations are removed.

Which is *still* an array type with a total of *four* array types just grouped in a different order…

What this PR does with the grammar is to produce one which allows an *array_type* to occur as part of another *array_type* – as opposed to an “array type” being part of another *array type*, which was always true…

Confused?

And that is just issue (1).

---

Issue (2) is maybe more *subjective* and has seen two approaches:

- Don’t talk about it – what the NRT changes originally did; and
- Minimise the impact – AKA positive thinking “Sure there is a gaping hole below the water line, but most of the hull is intact!” 😉

As I said, subjective 🙂 This PR tends towards the latter, I don’t think it has the balance right, but that is my subjective opinion.

---

@jskeet – I’m more than happy for you to have a go at it!

standard/arrays.md
jnm2: I think "as described below" would work. I'm also happy to workshop the wording! The term 'recursion' just describes the outside-in traversal of the array types as they are interspersed with `?`s.

standard/arrays.md
jnm2: Whatever we land with, I would just want it to be absolutely clear to any implementer what the meaning of `T[]?[,]` is, since its meaning is not an optional part of the language.

I described it in terms of the spec's existing `array_type` grammar whose element type is never directly an `array_type`, but may now be a `nullable_reference_type` consisting of `array_type nullable_annotation`. If the definition of the term "array type" does not line up with the `array_type` grammar, that does complicate things when this section goes to reference the term. However, if I describe things consistently in terms of `array_type`, this complication seems no worse than it was prior to this PR.

standard/types.md
```diff
@@ -54,17 +54,22 @@ interface_type
     ;
 
 array_type
-    : non_array_type rank_specifier+
+    : array_type nullable_type_annotation rank_specifier+
```
jskeet, line 57: Amusingly, coming back to this now, it explains the rule that I figured out independently (trying not to refer to this PR at all)... basically the rank for any array_type is the `rank_specifier+` part, which has to be either "after the last `?`" or there not be a `?` at all...

standard/arrays.md
jnm2: From meeting: let's not say "sole exception," just "an exception"

jskeet:
Closing in favor of #1554.

PR #1387, opened 2025-07-31
Clarify that readonly fields can only be assigned on the current instance

### LINKED ISSUE #1235: Spec seems to allow readonly fields of instances other than 'this' to be assigned in constructors
<https://github.com/dotnet/csharpstandard/blob/standard-v7/standard/classes.md#1553-readonly-fields>:

> When a *field_declaration* includes a `readonly` modifier, the fields introduced by the declaration are ***readonly fields***. Direct assignments to readonly fields can only occur as part of that declaration or in an instance constructor or static constructor in the same class. (A readonly field can be assigned to multiple times in these contexts.) Specifically, direct assignments to a readonly field are permitted only in the following contexts:
> 
> - In the *variable_declarator* that introduces the field (by including a *variable_initializer* in the declaration).
> - For an instance field, in the instance constructors of the class that contains the field declaration; for a static field, in the static constructor of the class that contains the field declaration. These are also the only contexts in which it is valid to pass a readonly field as an `out` or `ref` parameter.
> 
> Attempting to assign to a readonly field or pass it as an `out` or `ref` parameter in any other context is a compile-time error.

I didn't see how this spec text disallows the following assignment to a readonly field, which occurs in an instance constructor in the same class:

```cs
public class C
{
    private readonly int f;
    
    public C(C other)
    {
        other.f = 42;
    }
}
```

Which in practice ([SharpLab](https://sharplab.io/#v2:D4AQzABCBMEMIFgBQBvZEMQA4CcCWAbgIYAuAphDmUQCYD2AdgDYCeEeDJEAZgNzqYBGcPAAUcCHRIALMjgCUQiGiSY1kmXIB03CAF4IAFmj9VmAL7JzQA==)) gives:

> error CS0191: A readonly field cannot be assigned to (except in a constructor or init-only setter of the type in which the field is defined or a variable initializer)

jnm2:
Fixes #1235

Given that this is a bug in standard-v7 and perhaps earlier, should the PR nonetheless target draft-v8?


standard/classes.md
```diff
@@ -1703,7 +1703,7 @@ As explained in [§15.3.8](classes.md#1538-static-and-instance-members), each in
...
 
 - In the *variable_declarator* that introduces the field (by including a *variable_initializer* in the declaration).
-- For an instance field, in the instance constructors of the class that contains the field declaration; for a static field, in the static constructor of the class that contains the field declaration. These are also the only contexts in which it is valid to pass a readonly field as an output or reference parameter.
+- For an instance field, in the instance constructors of the class that contains the field declaration, and only when the reference to the field either is unqualified or is a member access from the expression `this`; for a static field, in the static constructor of the class that contains the field declaration. These are also the only contexts in which it is valid to pass a readonly field as an output or reference parameter.
```
jnm2: In coming up with this wording, I had in mind that parenthesized expressions may intervene between the member access expression and the this_expression:

```cs
class C
{
    private readonly int f;

    public C()
    {
        (this).f = 32;
    }
}
```

KalleOlaviNiemitalo: I suppose "**in** the instance constructors" does not include local functions in such constructors, but "**within**" would.  It's not entirely clear though.

jnm2: Added wording, what do you think?

standard/classes.md
```diff
@@ -1703,7 +1703,7 @@ As explained in [§15.3.8](classes.md#1538-static-and-instance-members), each in
...
 
 - In the *variable_declarator* that introduces the field (by including a *variable_initializer* in the declaration).
-- For an instance field, in the instance constructors of the class that contains the field declaration; for a static field, in the static constructor of the class that contains the field declaration. These are also the only contexts in which it is valid to pass a readonly field as an output or reference parameter.
+- For an instance field, in the instance constructors of the class that contains the field declaration (excluding local functions, lambdas, and anonymous methods), and only when the reference to the field either is unqualified or is a member access from the expression `this`; for a static field, in the static constructor of the class that contains the field declaration. These are also the only contexts in which it is valid to pass a readonly field as an output or reference parameter.
```
BillWagner: This might be a simpler way to express this:

```suggestion
- For an instance field, in the instance constructors of the class that contains the field declaration (excluding local functions, lambdas, and anonymous methods), and only when the reference to the field represents the receiver instance; for a static field, in the static constructor of the class that contains the field declaration. These are also the only contexts in which it is valid to pass a readonly field as an output or reference parameter.
```

BillWagner (approved):
This looks great @jnm2 

I had one possible wording suggestion. I don't feel strongly about it one way or the other.

standard/classes.md
jnm2: I like where this is going. I'm hung up a bit on the word 'represents.' If the reference to the field is the syntax `this.f`, and the receiver instance is `this`, I don't think of `this.f` as _representing_ `this`.

jnm2: Is this definitive enough?
```suggestion
- For an instance field, in the instance constructors of the class that contains the field declaration (excluding local functions, lambdas, and anonymous methods), and only on the instance being constructed; for a static field, in the static constructor of the class that contains the field declaration. These are also the only contexts in which it is valid to pass a readonly field as an output or reference parameter.
```

standard/classes.md
KalleOlaviNiemitalo: <https://github.com/dotnet/csharpstandard/blob/cec82ab746a8d1f9ae8cc3938c7b3943c19248d1/standard/classes.md?plain=1#L1706>

The term "lambdas" doesn't seem to be used elsewhere so I'd prefer "lambda expressions" here.  Or say "anonymous function" (12.19), which means a lambda expression or an anonymous method.

A lambda expression can be converted to a delegate or to an expression tree.  In the expression-tree case, an assignment to a `readonly` field of `this` would not be immediately harmful, but then the `Compile` method would make a delegate again, so it seems better to reject such assignments right away. (Roslyn tells me "error CS0832: An expression tree may not contain an assignment operator" even when assigning to a non-`readonly` field, but 8.6 and 10.7 do not require the implementation to reject non-`readonly` assignments in expression trees.)

But… should the following be allowed?  Here, the assignment to the readonly static field `x` is lexically within the initializer of `y`, rather than in a static constructor.

```csharp
public class C {
    public static readonly int x = 2;
    public static readonly int y = (x = 3);
}
```

jnm2: By explicitly excluding anonymous functions here, would that not prevent the expression tree subset of anonymous functions from assigning to a readonly field? (Pretending for the moment that assignments were possible in expression trees.)

> But… should the following be allowed? Here, the assignment to the readonly static field x is lexically within the initializer of y, rather than in a static constructor.

Yes, I think so, because this has been the behavior forever. I'll update.

jnm2: Also:
```cs
using System;

class C
{
    private readonly int f;

    public C()
    {
        // CS0192 A readonly field cannot be used as a ref or out value (except in a constructor)
        Expression<Action> x = () => M(ref f);
    }

    private static void M(ref int p) { }
}
```

KalleOlaviNiemitalo: > would that not prevent the expression tree subset of anonymous functions from assigning to a readonly field?

I think that's OK to prevent.

standard/classes.md
BillWagner: That works as well.

standard/classes.md
jnm2: How is the latest wording?

@BillWagner also, in case you want to rereview for the new aspect of these changes.

jnm2:
@BillWagner Can you confirm whether bugs in the v7 or earlier standards should be fixed in the draft-v8 branch only?

BillWagner:
> Can you confirm whether bugs in the v7 or earlier standards should be fixed in the draft-v8 branch only?

Sorry, I thought I'd already wrote this.  Let's target V8 only. We can target v7, and issue an updated v7 spec, but requires a vote of the ECMA general assembly (Rex knows the process well). 

I don't think this issue is impactful enough to go through that process.

jnm2:
That sounds good. There will always be errata, and it won't be that hard to put together that when the spec changed in v8, it was a correction of previous errata rather than a language change.

RexJaeschke:
My guess is that it is a *very* rare thing, if ever, to revise a previous edition of an Ecma standard. I certainly don't recommend it. While our spec edition nominally maps a product edition, we know of small things where that is not the case, and that's OK.

standard/classes.md
KalleOlaviNiemitalo: <https://github.com/dotnet/csharpstandard/blob/e25f08dbe4932ba92e498e7fde3855c976d471c6/standard/classes.md?plain=1#L1706>

What's the rationale for "in the same file and lexically later than the definition of the readonly field"?  At least two C# compilers currently accept:

```csharp
public class C {
    static readonly int i = (j = 4);
    static readonly int j;
}
```

jnm2: Ah. I was confusing this with the order in which static initializers run. Thanks for the catch!

jnm2: @KalleOlaviNiemitalo Another look please! Your presence here is invaluable to me.

PR #1389, opened 2025-08-09
Fix small spec bugs regarding nullable reference type parameters

jnm2:
While working on #1386, I found some spec bugs I could fix.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -442,8 +442,8 @@ The nullability of the type argument need not match the nullability of the type 
 
 > *Note*: To specify that a type argument is a nullable reference type, don’t add the nullable type annotation as a constraint (use `T : class` or `T : BaseClass`), but use `T?` throughout the generic declaration to indicate the corresponding nullable reference type for the type argument. *end note*
 
-<!-- Remove in C# 9, when this is allowed -->
-The nullable type annotation, `?`, can’t be used on an unconstrained type argument.
+<!-- Remove in C# 9, when `?` is allowed on any type parameter. -->
+The nullable type annotation, `?`, can only be used on a type parameter that has the value type constraint, the reference type constraint without the *nullable_type_annotation*, or a class type constraint without the *nullable_type_annotation*.
 
 For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`.
 
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -445,7 +445,8 @@ The nullability of the type argument need not match the nullability of the type 
 <!-- Remove in C# 9, when `?` is allowed on any type parameter. -->
 The nullable type annotation, `?`, can only be used on a type parameter that has the value type constraint, the reference type constraint without the *nullable_type_annotation*, or a class type constraint without the *nullable_type_annotation*.
 
-For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`.
+<!-- Add in C# 9, when `?` is allowed on nullable reference type parameters. -->
+<!-- For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`. -->
 
 > *Example*: The following examples show how the nullability of a type argument impacts the nullability of a declaration of its type parameter:
 >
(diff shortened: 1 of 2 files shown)


standard/types.md
```diff
@@ -27,7 +27,7 @@ For convenience, throughout this specification, some library type names are writ
 
 ### 8.2.1 General
 
-A reference type is a class type, an interface type, an array type, a delegate type, or the `dynamic` type. For each non-nullable reference type, there is a corresponding nullable reference type noted by appending the `?` to the type name.
+A reference type is a class type, an interface type, an array type, a delegate type, the `dynamic` type, or any type parameter that is constrained to be a reference type (that is, any type parameter with the reference type constraint or a class type constraint ([§15.2.5](classes.md#1525-type-parameter-constraints))). For each non-nullable reference type, there is a corresponding nullable reference type noted by appending the `?` to the type name.
```
jnm2, line 30: This borrows the existing language which appears in <https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/types.md#8312-nullable-value-types>:

> Conversely, a ***non-nullable value type*** is any value type other than `System.Nullable<T>` and its shorthand `T?` (for any `T`), plus any type parameter that is constrained to be a non-nullable value type (that is, any type parameter with a value type constraint ([§15.2.5](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#1525-type-parameter-constraints))).

standard/classes.md
```diff
@@ -443,7 +443,7 @@ The nullability of the type argument need not match the nullability of the type
 > *Note*: To specify that a type argument is a nullable reference type, don’t add the nullable type annotation as a constraint (use `T : class` or `T : BaseClass`), but use `T?` throughout the generic declaration to indicate the corresponding nullable reference type for the type argument. *end note*
 
 <!-- Remove in C# 9, when this is allowed -->
-The nullable type annotation, `?`, can’t be used on an unconstrained type argument.
```
jnm2, line 446: Here is an example of a type parameter that is not unconstrained, and thus was not disallowed by the original statement, which is *still* not allowed to use `?`:
```cs
#nullable enable

class C
{
    void M<T>(T? test) where T : System.IDisposable
    {
    }
}
```

Roslyn:
> A nullable type parameter must be known to be a value type or non-nullable reference type unless language version '9.0' or greater is used. Consider changing the language version or adding a 'class', 'struct', or type constraint.

standard/classes.md
```diff
@@ -442,10 +442,11 @@ The nullability of the type argument need not match the nullability of the type
...
 > *Note*: To specify that a type argument is a nullable reference type, don’t add the nullable type annotation as a constraint (use `T : class` or `T : BaseClass`), but use `T?` throughout the generic declaration to indicate the corresponding nullable reference type for the type argument. *end note*
 
-<!-- Remove in C# 9, when this is allowed -->
-The nullable type annotation, `?`, can’t be used on an unconstrained type argument.
+<!-- Remove in C# 9, when `?` is allowed on any type parameter. -->
+The nullable type annotation, `?`, can only be used on a type parameter that has the value type constraint, the reference type constraint, or a class type constraint.
 
-For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`.
+<!-- Add in C# 9, when `?` is allowed on nullable reference type parameters. -->
+<!-- For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`. -->
```
jnm2, line 449: Instances of `T?` do not exist when the type argument is a nullable reference type `C?` in C# 8, e.g.:

```cs
#nullable enable

class C
{
    void M<T>(T? test) where T : class?
    {
    }
}
```

Roslyn:
> A nullable type parameter must be known to be a value type or non-nullable reference type unless language version '9.0' or greater is used. Consider changing the language version or adding a 'class', 'struct', or type constraint.

BillWagner (approved):
These all look good to me. 

I'd like to have another set of eyes on them before merging, or have brief review in our next meeting. 

/cc @jskeet

jnm2:
Oops! Missed bringing this to the meeting since @jskeet didn't have a chance to look at it. I'll add the tag now.

jskeet (approved):

PR #1395, opened 2025-08-27
Minor typos

jnm2:
Spotted one, then decided to walk through with a spell checker.


jskeet (approved):
Typos all look fine - I don't know enough about the files under v6 and v7 to be able to comment on those. @Nigel-Ecma is there anything else we need to do for those?

BillWagner (approved):
This all looks good to me. I'd also like @Nigel-Ecma to take a look for the grammar validation.

RexJaeschke (approved):

Nigel-Ecma (approved):
LGTM as well.

@jskeet & @BillWagner – this PR has fixed a typo in the grammar, this changes the expected parse for a couple of samples – these files are usually machine generated but it looks like they've been hadn't edited, this is fine for csharpstandard but will need to be ported over to antler-testing.

@jnm2 – I'll port over to antler-testing in the next version of that repo, I don't think we need a version just for this typo. Next version is likely to be for PR#681.

PR #1402, opened 2025-09-04
Update more places with respect to ref-valued properties and indexers

### LINKED ISSUE #1371: Property & indexer access not fully updated for ref get accessors
In §12.2.2 we have:

> - The value of a property access expression is obtained by invoking the get accessor of the property. If the property has no get accessor, a compile-time error occurs. Otherwise, a function member invocation ([§12.6.6](expressions.md#1266-function-member-invocation)) is performed, and the result of the invocation becomes the value of the property access expression.
> - The value of an indexer access expression is obtained by invoking the get accessor of the indexer. If the indexer has no get accessor, a compile-time error occurs. Otherwise, a function member invocation ([§12.6.6](expressions.md#1266-function-member-invocation)) is performed with the argument list associated with the indexer access expression, and the result of the invocation becomes the value of the indexer access expression.

Both of these do not allow for a property or indexer with a ref get accessor – in such a case a compile-time error does not occur, a value is the result, etc.

Are ref get accessors classified as something other than property/indexer accesses somewhere?

Property access comes under *member_reference*, indexer under *element_access*, do these handle ref get and correctly classify the expression (§12.2), indeed does classification handle ref valued expressions correctly?

Then there is setting a property or indexer element (§12.21.2), is ref get handled correctly in these cases?

Above questions are **not** claimed to be exhaustive.

The above observations come from adding support for `Index` and `Range` (#1369, draft at time of writing) which has added text such as (§12.8.12.3 *Indexer access* in #1369):

> - Using the best indexer determined at binding-time:
>   - If the indexer access is the target of an assignment, the set accessor or ref get accessor is invoked to assign a new value ([§12.21.2](expressions.md#12212-simple-assignment)).
>   - In all other cases, the get accessor or ref get accesor is invoked to obtain the current value ([§12.2.2](expressions.md#1222-values-of-expressions)).

An ancient (2018 ;-)) issue was raised in csharplang regarding ref get properties, transferred here as #295 years ago (2021 ;-)), and has no milestone…

This surely needs to be sorted for v8.

jnm2:
Fixes #1371

I spread a wide net and put no effort toward deduplicating so far, after seeing the wide variety in types of update. I am open to suggestion on that.


standard/expressions.md
```diff
@@ -359,7 +361,7 @@ In both of the above cases, a cast expression can be used to explicitly convert
 
 A member lookup is the process whereby the meaning of a name in the context of a type is determined. A member lookup can occur as part of evaluating a *simple_name* ([§12.8.4](expressions.md#1284-simple-names)) or a *member_access* ([§12.8.7](expressions.md#1287-member-access)) in an expression. If the *simple_name* or *member_access* occurs as the *primary_expression* of an *invocation_expression* ([§12.8.10.2](expressions.md#128102-method-invocations)), the member is said to be *invoked*.
 
-If a member is a method or event, or if it is a constant, field or property of either a delegate type ([§20](delegates.md#20-delegates)) or the type `dynamic` ([§8.2.4](types.md#824-the-dynamic-type)), then the member is said to be *invocable.*
+If a member is a method or event, or if it is a constant, field or property whose evaluated value is either of a delegate type ([§20](delegates.md#20-delegates)) or the type `dynamic` ([§8.2.4](types.md#824-the-dynamic-type)), then the member is said to be *invocable.*
```
jnm2: This change enables:
```cs
class C
{
    void M()
    {
        Prop();
    }

    private System.Action f;
    public ref System.Action Prop => ref f;
}
```
Since the property type is `ref System.Action` which is not a delegate type, but the property's evaluated value is of type `System.Action`.

standard/expressions.md
```diff
@@ -454,23 +456,23 @@ Once a particular function member has been identified at binding-time, possibly
 >   </tr>
 >   <tr>
 >     <td><code>P = value</code></td>
->     <td>The set accessor of the property <code>P</code> in the containing class or struct is invoked with the argument list <code>(value)</code>. A compile-time error occurs if <code>P</code> is read-only. If <code>P</code> is not <code>static</code>, the instance expression is <code>this</code>.</td>
+>     <td>If <code>P</code> is non-ref-valued, the set accessor of the property <code>P</code> in the containing class or struct is invoked with the argument list <code>(value)</code>. If <code>P</code> is ref-valued, the get accessor of the property <code>P</code> in the containing class or struct is invoked. A compile-time error occurs if <code>P</code> is non-ref-valued and read-only or if <code>P</code> is ref-valued and returns a readonly reference. If <code>P</code> is not <code>static</code>, the instance expression is <code>this</code>.</td>
```
jnm2: I'm not sure about the language "returns" and "readonly reference." Is there established language for this?

standard/expressions.md
```diff
@@ -2557,9 +2559,9 @@ Each *initializer_target* is followed by an equals sign and either an expression
 
 A member initializer that specifies an expression after the equals sign is processed in the same way as an assignment ([§12.21.2](expressions.md#12212-simple-assignment)) to the target.
 
-A member initializer that specifies an object initializer after the equals sign is a ***nested object initializer***, i.e., an initialization of an embedded object. Instead of assigning a new value to the field or property, the assignments in the nested object initializer are treated as assignments to members of the field or property. Nested object initializers cannot be applied to properties with a value type, or to read-only fields with a value type.
+A member initializer that specifies an object initializer after the equals sign is a ***nested object initializer***, i.e., an initialization of an embedded object. Instead of assigning a new value to the field or property, the assignments in the nested object initializer are treated as assignments to members of the field or property. Nested object initializers cannot be applied to properties or indexers with a value type, to ref-valued properties or indexers whose type is a reference to a value type, or to read-only fields with a value type.
```
jnm2: Likely fixes an older spec bug which failed to mention that this rule applies to indexers as well as properties.

standard/expressions.md
```diff
@@ -2557,9 +2559,9 @@ Each *initializer_target* is followed by an equals sign and either an expression
...
 
-A member initializer that specifies a collection initializer after the equals sign is an initialization of an embedded collection. Instead of assigning a new collection to the target field, property, or indexer, the elements given in the initializer are added to the collection referenced by the target. The target shall be of a collection type that satisfies the requirements specified in [§12.8.17.2.3](expressions.md#1281723-collection-initializers).
+A member initializer that specifies a collection initializer after the equals sign is an initialization of an embedded collection. Instead of assigning a new collection to the target field, property, or indexer, the elements given in the initializer are added to the collection referenced by the target. The target shall be of a collection type that satisfies the requirements specified in [§12.8.17.2.3](expressions.md#1281723-collection-initializers). Initialization of an embedded collection cannot be applied to properties or indexers with a value type, to ref-valued properties or indexers whose type is a reference to a value type, or to read-only fields with a value type.
```
jnm2, line 2613: Likely fixes an older spec bug which failed to mention that this rule applies to nested collection initializers as well as nested object initializers.

standard/classes.md
```diff
@@ -3474,8 +3474,8 @@ A set accessor corresponds to a method with a single value parameter of the prop
 Based on the presence or absence of the get and set accessors, a property is classified as follows:
 
 - A property that includes both a get accessor and a set accessor is said to be a ***read-write property***.
-- A property that has only a get accessor is said to be a ***read-only property***. It is a compile-time error for a read-only property to be the target of an assignment.
-- A property that has only a set accessor is said to be a ***write-only property***. Except as the target of an assignment, it is a compile-time error to reference a write-only property in an expression.
+- A property that has only a get accessor is said to be a ***read-only property***. It is a compile-time error for a read-only property to be the target of an assignment unless the property is ref-valued and returns a writeable reference.
+- A property that has only a set accessor is said to be a ***write-only property***. Except as the target of an assignment or as an argument to the `nameof` operator ([§12.8.23](expressions.md#12823-the-nameof-operator)), it is a compile-time error to reference a write-only property in an expression.
```
jnm2: Likely fixes an older spec bug which was disallowing:
```cs
class C
{
    void M()
    {
        _ = nameof(Prop);
    }

    public int Prop { set { } }
}
```

jskeet:
@jnm2 Would you like us to review this and discuss it in the upcoming meeting after the priority PRs? If so, feel free to apply the meeting-discuss label.

jnm2:
Sure. Is applying that label a good automatic step to take when submitting a PR that is ready for review?

jskeet:
> Sure. Is applying that label a good automatic step to take when submitting a PR that is ready for review?

If it's non-trivial, yes. For very small changes, it's fine to just get appropriate folks to review between meetings and merge.

standard/expressions.md
```diff
@@ -2354,7 +2356,7 @@ The binding-time processing of an indexer access of the form `P[A]`, where `P` i
 - If the resulting set of candidate indexers is empty, then no applicable indexers exist, and a binding-time error occurs.
 - The best indexer of the set of candidate indexers is identified using the overload resolution rules of [§12.6.4](expressions.md#1264-overload-resolution). If a single best indexer cannot be identified, the indexer access is ambiguous, and a binding-time error occurs.
 - The accessors of the best indexer are checked:
-  - If the indexer access is the target of an assignment then the indexer shall have a set or ref get accessor, otherwise, a binding-time error occurs;
+  - If the indexer access is the target of an assignment then the indexer either shall have a set accessor or shall return a non-readonly ref, otherwise, a binding-time error occurs;
```
jnm2: Fixes a spec issue where it  would be valid to assign to a `ref readonly` indexer. (That is, an indexer whose returned ref is a readonly ref. I'm not talking about a `readonly ref` indexer, which is an indexer which is a readonly struct member.)

standard/expressions.md
```diff
@@ -27,7 +27,9 @@ For expressions which occur as subexpressions of larger expressions, with the no
...
+A non-ref-valued property access or non-ref-valued indexer access is always reclassified as a value by performing an invocation of the get accessor or the set accessor. The particular accessor is determined by the context of the property or indexer access: If the access is the target of an assignment, the set accessor is invoked to assign a new value ([§12.22.2](expressions.md#12222-simple-assignment)). Otherwise, the get accessor is invoked to obtain the current value ([§12.2.2](expressions.md#1222-values-of-expressions)).
+
+A ref-valued property access or non-ref-valued indexer access is always reclassified as a variable by performing an invocation of the get accessor ([§15.7.3](classes.md#1573-accessors)).
```
BillWagner: Wouldn't this invoke a setter for a ref-valued property?

`A.P = ref someVariable;`

The above would need to take that into account.

standard/expressions.md
BillWagner: I think this is fine. Because you're using the context of "invoking the get accessor", that is a method invocation, so "returns" is reasonable. 15.6.1 defines *ref_kind*, which defines a readonly reference.

standard/expressions.md
BillWagner: Probably in classes on indexers.

BillWagner (commented):
This is good progress. I'd also like to see more xrefs to15.7.1 where "non-ref-valued" and "ref-valued" properties are defined.

standard/expressions.md
jnm2: Ref-valued properties can't have setters (https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs8147).

standard/expressions.md
jnm2: Sorry, I'm not sure what you're saying. Is there an open question on the change here or another place to update?

standard/expressions.md
```suggestion
A ref-valued property access or ref-valued indexer access is always reclassified as a variable by performing an invocation of the get accessor ([§15.7.3](classes.md#1573-accessors)).
```

jnm2:
@BillWagner Thanks! The xrefs have been added.

standard/classes.md
```diff
@@ -3480,8 +3480,8 @@ A set accessor corresponds to a method with a single value parameter of the prop
...
```
```suggestion
- A property that has only a set accessor is said to be a ***write-only property***. Except as the target of an assignment or as the operand of the `nameof` operator ([§12.8.23](expressions.md#12823-the-nameof-operator)), it is a compile-time error to reference a write-only property in an expression.
```

I'd initially used "within" instead of "as", but it's invalid to further dereference the write-only property - so `nameof(Test.X.Y)` is invalid if `X` is a write-only property, but `nameof(Test.X)` is okay.

standard/expressions.md
```diff
@@ -36,8 +38,8 @@ An ***instance accessor*** is a property access on an instance, an event access
...
-- The value of an indexer access expression is obtained by invoking the get accessor of the indexer. If the indexer has no get accessor, a compile-time error occurs. Otherwise, a function member invocation ([§12.6.6](expressions.md#1266-function-member-invocation)) is performed with the argument list associated with the indexer access expression, and the result of the invocation becomes the value of the indexer access expression.
+- The value of a property access expression is obtained by invoking the get accessor of the property. If the property has no get accessor, a compile-time error occurs. Otherwise, a function member invocation ([§12.6.6](expressions.md#1266-function-member-invocation)) is performed. For non-ref-valued properties (§15.7.1), the result of the invocation becomes the value of the property access expression. For ref-valued properties, the result of the invocation is a variable and the value of the variable becomes the value of the property access expression.
```
jskeet, line 41: I'm not sure about "and the value of the variable becomes the value of the property access expression". We've classified it *as* a variable, not a value... Can we not just leave it as something like "and that variable is the result of evaluating the get accessor"? I may very well be wrong on this, and haven't checked other places yet.

Same applies to the line below, of course.

standard/expressions.md
```diff
@@ -454,23 +456,23 @@ Once a particular function member has been identified at binding-time, possibly
...
 >   </tr>
 >   <tr>
 >     <td><code>e.P = value</code></td>
->     <td>The set accessor of the property <code>P</code> in the class, struct, or interface given by the type of <code>E</code> is invoked with the instance expression <code>e</code> and the argument list <code>(value)</code>. A binding-time error occurs if <code>P</code> is <code>static</code> or if <code>P</code> is read-only.</td>
+>     <td>If <code>P</code> is non-ref-valued (§15.7.1), the set accessor of the property <code>P</code> in the class, struct, or interface given by the type of <code>E</code> is invoked with the instance expression <code>e</code> and the argument list <code>(value)</code>. If <code>P</code> is ref-valued, the get accessor of the property <code>P</code> in the containing class, struct, or interface given by the type of <code>E</code> is invoked is invoked with the instance expression <code>e</code>. A binding-time error occurs if <code>P</code> is <code>static</code>, if <code>P</code> is non-ref-valued and read-only, or if <code>P</code> is ref-valued and returns a readonly reference.</td>
```
jskeet: For the ref-valued case, I think we still need to say "and then an assignment is performed to the resulting variable" or something like that, don't we? Otherwise we're saying the get accessor is invoked - and that's all. (There's no reference to `value`.)

standard/expressions.md
```diff
@@ -504,7 +506,7 @@ Once a particular function member has been identified at binding-time, possibly
 >   </tr>
 >   <tr>
 >     <td><code>e[x, y] = value</code></td>
->     <td>Overload resolution is applied to select the best indexer in the class, struct, or interface given by the type of <code>e</code>. The set accessor of the indexer is invoked with the instance expression <code>e</code> and the argument list <code>(x, y, value)</code>. A binding-time error occurs if the indexer is read-only.
+>     <td>Overload resolution is applied to select the best indexer in the class, struct, or interface given by the type of <code>e</code>. If the indexer is non-ref-valued (§15.9.1), the set accessor of the indexer is invoked with the instance expression <code>e</code> and the argument list <code>(x, y, value)</code>. If the indexer is ref-valued, the get accessor of the indexer is invoked with the instance expression <code>e</code> and the argument list <code>(x, y)</code>. A binding-time error occurs if the indexer is non-ref valued and read-only or if the indexer is ref-valued and returns a readonly reference.
```
jskeet: Same as above, I think.

jskeet (commented):
I'm generally very much in favour of this - just a few nits.

standard/expressions.md
BillWagner: Sorry. I was saying the update for `ref readonly` should be in the section on indexers.

standard/expressions.md
jskeet: Nigel: "the variable reference becomes the result of the property access expression"

Or maybe "the variable becomes" - please check for use of "variable reference" elsewhere.

jskeet: See also #324 and clause 9.5.

standard/expressions.md
jnm2: I read <https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#1591-general> but I don't see anything describing assignments to indexers.

jnm2: Sorry, that was the wrong link. FIxed!

standard/expressions.md
jnm2: The spec currently uses `variable reference` 33 times. For instance, <https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/arrays.md#174-array-element-access>:

> The result of an array access is a variable reference ([§9.5](variables.md#95-variable-references)) to the array element selected by the indices.

"Variable references" (https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#95-variable-references) defines "variable reference" as the _expression_ that refers to the variable, rather than the _value_ of, or the _result of evaluating_, the expression. However, the array access spec uses "variable reference" to mean the result, so I think I'll maintain consistency with the array access spec. If #324 comes to some additional clarity that would change the array access spec, then it will also pick up this section.

jnm2: We even have a section named "Variable references" (https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#95-variable-references) and a following section "Atomicity of variable references" (https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/variables.md#96-atomicity-of-variable-references).

jnm2: I'm going to back up a second. I'm not sure we can say that the value is a variable or variable reference. This section starts with this bullet point:

> - The value of a variable is simply the value currently stored in the storage location identified by the variable. A variable shall be considered definitely assigned ([§9.4](variables.md#94-definite-assignment)) before its value can be obtained, or otherwise a compile-time error occurs.

That bullet point doesn't say "the value of a variable is a variable" or "variable reference." It says the value is the contents of the variable, not the variable itself. So if the value of the expression `myVariable` is not a variable, why would the value of the expression `obj[p]` be a variable?

jnm2: (I took this section to be _the_ section which explains how a usage of `myVariable` ends up being a read of the contents of the variable. So e.g. if you see `M(myVariable);`, this section describes how it is that we're obtaining the value of that variable rather than e.g. passing a variable itself. Wouldn't the same hold for `M(obj[p]);`?)

standard/classes.md
```diff
@@ -3480,8 +3480,8 @@ A set accessor corresponds to a method with a single value parameter of the prop
 Based on the presence or absence of the get and set accessors, a property is classified as follows:
 
 - A property that includes both a get accessor and a set accessor is said to be a ***read-write property***.
-- A property that has only a get accessor is said to be a ***read-only property***. It is a compile-time error for a read-only property to be the target of an assignment.
-- A property that has only a set accessor is said to be a ***write-only property***. Except as the target of an assignment, it is a compile-time error to reference a write-only property in an expression.
+- A property that has only a get accessor is said to be a ***read-only property***. It is a compile-time error for a read-only property to be the target of an assignment unless the property is ref-valued and returns a writeable reference.
```
Nigel-Ecma, line 3485: The added text equates to stating that it is an error to write to a read-only property unless it isn't a read-only property…

That doesn’t make sense.

A property is read-only if:

- it is a non-ref property and only has a get accessor; or
- it is a ref property and marked readonly

C# started with one kind of property and one kind of indexer, now it has two – ref properties and ref indexers which behave more like array indexing – which though they don’t have an indexer *per se* have an indexing operation whose return is classified as a variable aka. a variable reference just like ref properties and ref indexers.

standard/expressions.md
```diff
@@ -27,7 +27,9 @@ For expressions which occur as subexpressions of larger expressions, with the no
...
+A non-ref-valued property access (§15.7.1) or non-ref-valued indexer access (§15.9.1) is always reclassified as a value by performing an invocation of the get accessor or the set accessor. The particular accessor is determined by the context of the property or indexer access: If the access is the target of an assignment, the set accessor is invoked to assign a new value ([§12.23.2](expressions.md#12232-simple-assignment)). Otherwise, the get accessor is invoked to obtain the current value ([§12.2.2](expressions.md#1222-values-of-expressions)).
+
+A ref-valued property access or ref-valued indexer access is always reclassified as a variable by performing an invocation of the get accessor ([§15.7.3](classes.md#1573-accessors)).
```
Nigel-Ecma, line 32: The original here states that a non-ref property/indexer access is reclassified as value and goes on to say what that value is for get, “the current value”, but not for set. Maybe its missing something along the lines of “the set accessor is invoked to assign a new value *and that new value also becomes the value of the access*”? Set accessors do not have a return value *per se* but something like `a = b.property = c;` is valid.

The added para for ref properties/indexers does not continue with how that variable is read/written/neither based on being the target of an assignment/ref/neither. Maybe it should?

standard/expressions.md
Nigel-Ecma: @jskeet & @jnm2: The value of a ref variable would indeed appear to be a variable reference, *but* unless it is the target of a `ref` – the “do not de-reference operator” :-) – that variable reference is *invisibly* deferenced to read or write a value.

The change made here is confusing as the first sentence states what the value is, and then the added text says that value is actually not the value if it is a ref property.

I suggest this is another case where the two kinds of property and indexer (non-ref & ref) need to be specified separately as they do not have the same semantics.

(I'll add an overall review comment on this.)

standard/expressions.md
```diff
@@ -359,7 +361,7 @@ A ***lifted operator*** permits predefined and user-defined operators that opera
 
 A member lookup is the process whereby the meaning of a name in the context of a type is determined. A member lookup can occur as part of evaluating a *simple_name* ([§12.8.4](expressions.md#1284-simple-names)) or a *member_access* ([§12.8.7](expressions.md#1287-member-access)) in an expression. If the *simple_name* or *member_access* occurs as the *primary_expression* of an *invocation_expression* ([§12.8.10.2](expressions.md#128102-method-invocations)), the member is said to be *invoked*.
 
-If a member is a method or event, or if it is a constant, field or property of either a delegate type ([§21](delegates.md#21-delegates)) or the type `dynamic` ([§8.2.4](types.md#824-the-dynamic-type)), then the member is said to be *invocable.*
+If a member is a method or event, or if it is a constant, field or property whose evaluated value is either of a delegate type ([§21](delegates.md#21-delegates)) or the type `dynamic` ([§8.2.4](types.md#824-the-dynamic-type)), then the member is said to be *invocable.*
```
Nigel-Ecma, line 364: The current text of the Standard it is not clear that `ref` is part of the type *per se*. If it isn't this change is not required, if it is it would be better to word it along the line of “a delegate, or ref delegate, type” than bring the “evaluated value” into it.

See overall review comment for more on what ref is.

standard/expressions.md
```diff
@@ -454,23 +456,23 @@ Once a particular function member has been identified at binding-time, possibly
 >   </tr>
 >   <tr>
 >     <td><code>P = value</code></td>
->     <td>The set accessor of the property <code>P</code> in the containing class or struct is invoked with the argument list <code>(value)</code>. A compile-time error occurs if <code>P</code> is read-only. If <code>P</code> is not <code>static</code>, the instance expression is <code>this</code>.</td>
+>     <td>If <code>P</code> is non-ref-valued (§15.7.1), the set accessor of the property <code>P</code> in the containing class or struct is invoked with the argument list <code>(value)</code>. If <code>P</code> is ref-valued, the get accessor of the property <code>P</code> in the containing class or struct is invoked. A compile-time error occurs if <code>P</code> is non-ref-valued and read-only or if <code>P</code> is ref-valued and returns a readonly reference. If <code>P</code> is not <code>static</code>, the instance expression is <code>this</code>.</td>
```
Nigel-Ecma, line 459: How ref get accessor (see the next para) is used to set the value is not complete..

Both the grammar and the text currently distinguish between three accessors: *get accessor*, *set accessor*, and *ref get accessor*. E.g. in §12.8.12.4 Indexer access we have:

> - If the indexer access is the target of an assignment then the indexer shall have a set or ref get accessor, otherwise, a binding-time error occurs;

Here the “ref get” is needed, just “get” would be wrong.

The change here (and possibly in other existing or changed text) refers to just get in relation to a ref property (`If <code>P</code> is ref-valued, the get accessor of the property`.

I think consistency is required, either it is the get accessor of a ref property/indexer, or the ref get accessor. I would go with the latter, but if the former then existing text and grammar rule names and/or their associated text will need rewording.

standard/expressions.md
```diff
@@ -454,23 +456,23 @@ Once a particular function member has been identified at binding-time, possibly
...
 >   </tr>  
 >   <tr>
 >     <td><code>T.P</code></td>
 >     <td>The get accessor of the property <code>P</code> in the class or struct <code>T</code> is invoked. A compile-time error occurs if <code>P</code> is not <code>static</code> or if <code>P</code> is write-only.</td>
 >   </tr>  
 >   <tr>
 >     <td><code>T.P = value</code></td>
->     <td>The set accessor of the property <code>P</code> in the class or struct <code>T</code> is invoked with the argument list <code>(value)</code>. A compile-time error occurs if <code>P</code> is not <code>static</code> or if <code>P</code> is read-only.</td>
+>     <td>If <code>P</code> is non-ref-valued (§15.7.1), the set accessor of the property <code>P</code> in the class or struct <code>T</code> is invoked with the argument list <code>(value)</code>. If <code>P</code> is ref-valued, the get accessor of the property <code>P</code> in the containing class or struct is invoked. A compile-time error occurs if <code>P</code> is not <code>static</code>, if <code>P</code> is non-ref-valued and read-only, or if <code>P</code> is ref-valued and returns a readonly reference.</td>
```
Nigel-Ecma, line 467: See comment on line 459.

I am stopping individual change review at this point for now.

Nigel-Ecma (changes requested):
Some changes are required for both correctness and consistency, there might be more changes in general required. In some of the comments I promised an overall review comment…

One issue that came up is whether `ref` is part of the type, which in turn raises whether the semantics of ref assignment are fully covered. Consider the fragment:

> ```csharp
> int p = 42;
> ref int q = ref p;
> ref int r = ref q;
> r = 24;
> ```

What is the semantics of the third line? If ref is part of the type then would not the result of `ref q` have type `ref ref int`, or do refs collapse, or …? What does `r` reference, `q` as the RHS seems to say, or `p`? Are the answers clear in the Standard?

For comparison consider the following *unsafe* fragment which does exactly the same as the above:

> ```csharp
> int p = 42;
> int* q = &p;
> int* r = q;
> *r = 24;
> ```

Note that lines 2 & 3 differ in a way they don’t in the safe fragment; so `ref` on the same safe lines is doing something different in each case – is that clear in the Standard? In the unsafe version it is clear that `r` is set to the value of `q`, which is a reference (pointer) to `p`.

Further the unsafe line 4 has an explicit indirection/dereference which is omitted in safe code. Is when such auto-deferencing occurs always clear in the Standard? Does it need to be? Some of the text seems to suggest the term “alias” for a ref variable, which might provide an alternative to using de-referencing, should that be used and does it need to be clearer in the Standard?

Now try printing out the addresses of `p`, `q` & `r` in both fragments (you need to this using unsafe code), or otherwise determine the addresses, are the results what you expect? Is it specified in the Standard? Does it need to be?

And the question brought up in some of the comments: what is the name of the accessor of a ref property/indexer: its *ref get accessor* as indicated by the grammar and in some places in the test; or just its *get accessor*, with added qualification that it belongs to a ref property/indexer if the context requires it? Here I will go for “ref get” as it is just simpler and conveys what is needed without more verbiage.

In summary:

- Is ref part of the type?
- Are the semantics of `ref variable-reference` fully specified?
- Are the semantics of auto-deferencing fully specified? Do they need to be? Or should the language use “alias”?
- What is the name of the ref get accessor?

jskeet:
On the "is ref part of the type" - I don't believe it is. I think I *used* to think of it that way, but I now think of it more as "just how the storage location for the variable is decided" (with restrictions about what the RHS can be, where there is an RHS). (After all, we've had `ref` parameters for a long time, and I don't think we've deemed those to be a different type.)

@Nigel-Ecma Does that make things easier or harder, do you think?

BillWagner:
@Nigel-Ecma @jskeet 

I want to dive into this fragment from Nigel's comment:

```csharp
int p = 42;
ref int q = ref p;
ref int r = ref q;
r = 24;
```

One key is that `= ref` is an operator: The ref assignment. That is how the standard rationalizes the differences between the safe and unsafe variations:

- `ref int q` and `ref int r` declare variables that are *reference_variable* types: They must refer to some other location.
- `= ref` assigns the left side to *refer* to the storage location used by the right side. If the right side is a *reference_variable*, the left side refers to the same location (pointer-to-pointer assignment in the unsafe example). If the right side is a variable, the left side refers to the location of the right side (pointer-to-address-of assignment in the unsafe example).

BillWagner:
Re-opening because I clicked the wrong button earlier today...

Nigel-Ecma:
The “`= ref`” device @BillWagner mentions wasn’t carried over to “`? ref : ref`” which sticks to using “`?:`” and “If `ref` is (not) present” in its description. In the description of `?:` while the two forms are covered in different sets of bullets both sets just use the language “`x`/`y` is evaluated”.

§9.7.1 covers the reference variables and the referent, covering the semantics for reading & writing. However it does not give the semantics for when it a reference variables is the target of a `ref` (as in parameter passing, or just all of the following if your mental model has `ref` as the lowest possible priority unary operator), `= ref`, `? ref`, `: ref`, `return ref`, and any I’ve missed!

standard/classes.md
Nigel-Ecma: In:

>```csharp
>   (_, Prop) = (42, 24);
>```

`Prop` is not the target of an assignment *per se*, but it is assigned to…

Maybe language which tackles it the other way around by saying it cannot be read (which also removes the need to call out `nameof`), and cannot be written for get-only?

standard/classes.md
jnm2: As discussed, I will investigate getting rid of the terms ***read-only property*** and ***write-only property*** across the spec and see what that change would amount to. Since the newish `readonly` modifier may be applied to a property and it has an entirely different meaning (a property could be settable and still readonly), we would find it confusing to continue using the term ***read-only property*** to refer to what assignments are allowed.

jskeet:
@jnm2 Would it be appropriate to remove the meeting discuss label as this hasn't changed since the last meeting?

jnm2:
Sure, I'll reapply if I get to it today and it's ready.

standard/expressions.md
jnm2: @jskeet I'd like input on this during the meeting actually. Is the property type `ref System.Action`- does the language have a concept of `ref` types, or would we want it to? Possibly this line it was accurate as it was previously.

standard/classes.md
jnm2: I like this idea. How do we judge whether the spec needs to say "it cannot be read" and let spec readers form logical conclusions about all the ways that reading can happen, versus specifying the exact list of constructs that must be disallowed?

standard/expressions.md
jnm2: @BillWagner Sorry, I wasn't following what to do here, could you break it down for me?

standard/expressions.md
jskeet: Yup, let's talk in the meeting. I *think* that's a ref-returning property rather than a property of type `ref Action` but I could be wrong...

jnm2: Discussed in meeting, the original text is okay.

Nigel-Ecma:
I added the _meeting: priority_ label as a transfer from #1371, but checking now seems this has not yet been updated since the last discussion on it so maybe the label is wrong…

110326: Fixed my error, removed the label.

Nigel-Ecma:
Not to make this more complicated ;-) but I *think* the grammar is wrong (and has been for ages).

We now have *get*, *set* and *ref get* accessors; the first two share *accessor_body* and the latter uses *ref_accessor_body*. I think the former maybe should be split into:

```ANTLR
get_accessor_body
    : block
    | '=>' expression ';'
    | ';' 
    ;

set_accessor_body
    : block
    | '=>' statement_expression ';'
    | ';' 
    ;
```

I don’t think the restriction to *statement_expression* is currently specified, but I may be wrong…

So for the three distinct kinds the return is classified as: get => value, set => nothing, ref get => variable reference.

---

If you are wondering a certain compiler will issue an error if `=>` is not followed by a valid *statement_expression* with the text “Only assignment, call, increment, decrement, await, and new object expressions can be used as a statement” – yup *statement*, oops. (The list given equates to *statement_expression* and are all expressions.

standard/expressions.md
Nigel-Ecma: “set accessor or shall return a non-readonly ref” reads as mixing two things: the first is about the existence of an accessor, the second about the value of the operation which *indirectly* is specifying something about an accessor…

Maybe something along the lines of “set accessor or a ref get accessor which returns a non-readonly ref”?

PR #1407, opened 2025-09-04
Remove assumptions about corresponding parameters

### LINKED ISSUE #981: Generic type inference assumes same number of arguments and parameters

jskeet:
This replaces #1350. It's simpler due to #1359 centralizing the tie between an argument and the corresponding parameter's type.

Fixes #981

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -763,7 +763,7 @@ Type inference takes place in phases. Each phase will try to infer type argument
 
 #### 12.6.3.2 The first phase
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM. Let's :shipit:

PR #1408, opened 2025-09-04
Fix definite assignment in || operator

### LINKED ISSUE #1070: [C# language specification 7th edition] Incorrect rule for definite-assignment state for || expressions
I downloaded the C# language specification 7th edition via https://ecma-international.org/publications-and-standards/standards/ecma-334/. In chapter 9.4.4.27 about definite-assignment state of || expressions, it was written

> The definite-assignment state of v before expr_second is definitely assigned if and only if the state of
v after expr_first is either definitely assigned or “definitely assigned after **true** expression”.
Otherwise, it is not definitely assigned.

However, i believe that is incorrect, the correct one should be

> The definite-assignment state of v before expr_second is definitely assigned if and only if the state of
v after expr_first is either definitely assigned or “definitely assigned after **false** expression”.
Otherwise, it is not definitely assigned.

Because if the former is correct, then v can be definitely-assigned when it shouldn't be, consider the following situation:

1. The definite-assignment state of v after expr_first is definitely assigned after true expression
2. expr_first is false
3. expr_second is evaluated and the definite-assignment state of v before expr_second is definitely-assigned, however because number 1 and 2, then v is not actually assigned.

From the reason above, i believe that the former is incorrect and the latter is the correct one

jskeet:
Fixes #1070 - I *think* this is all that's required.

### DIFF
--- a/standard/variables.md
+++ b/standard/variables.md
@@ -786,7 +786,7 @@ For an expression *expr* of the form:
 ```
 
 - The definite-assignment state of *v* before *expr_first* is the same as the definite-assignment state of *v* before *expr*.
-- The definite-assignment state of *v* before *expr_second* is definitely assigned if and only if the state of *v* after *expr_first* is either definitely assigned or “definitely assigned after true expression”. Otherwise, it is not definitely assigned.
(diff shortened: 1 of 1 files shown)


Nigel-Ecma (approved):
LGTM

gafter (approved):

gafter (approved):
I withdraw this comment. The issue was resolved previously.

Issue #1404, opened 2025-09-04
The name of the unary ^ operator


Nigel-Ecma:
The source material for the unary `^` operator, which converts an `int` value to a from-end `System.Index` value, names it “hat” and that name has been adopted in the Standard.

This choice has generated [discussion](https://github.com/dotnet/csharpstandard/pull/1369/files#r2294479312).

This PR spins that discussion off, please discuss, and if the consensus is for change produce a PR.

Nigel-Ecma:
Calling the operator “hat” is effectively a *fait accompli* – the use of it in design documents has spread to articles on the internet discussing slicing – and there is little to gain from changing it, though I will admit to considering doing so when specifying the feature!

My vote: leave it as “hat”

jskeet:
While I certainly take Nigel's point, my preference would still be to call it the "from end operator" (or "index-from-end operator") but acknowledge the use of the term "hat" for searchability.

My reasoning:

- If we just went by what articles on the internet call things, we'd call the ?: conditional operator "the ternary operator".
- All other operators are named after their meaning rather their appearance (I think)
- We already have a binary ^ operator which we don't call "hat". "When is a hat not a hat? When it's got two operands..." doesn't feel great to me.

Obviously I'll go with the majority, but thought I'd at least set out my reasons...

jnm2:
My vote is for "the from-end operator" because it is the operator that produces a from-end index. This ties nicely with our existing terminology around indexes. Jon's last point about "the binary hat operator" really hits home for me as well.

Nigel-Ecma:
Considering the suggestions so far; the Standard (currently) using the term *from-end index* when discussing such indices; [this link](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#index-from-end-operator-); and a meeting suggestion that folk searching for “C# hat” should get a hit from the Standard then maybe:

> index from-end operator (sometimes called “hat”)
>
> *(parenthetical on first use only)*
 
might be the happy middle?

jskeet:
I haven't checked where we're at in terms of the actual merged spec at the moment, but we should:

- Agree on the name (Nigel's last comment looks popular)
- Make sure that's implemented

jskeet:
Agreed to do this as per https://github.com/dotnet/csharpstandard/issues/1404#issuecomment-3260299801 - Rex will implement

Nigel-Ecma:
Resolved by #1437

PR #1437, opened 2025-10-23
Tweak terminology re the index from-end (hat) operator

RexJaeschke:
This PR completes Rex's action item from the TG2 call on 2025-10-22 re resolving Issue https://github.com/dotnet/csharpstandard/issues/1404.

### DIFF
--- a/standard/ranges.md
+++ b/standard/ranges.md
@@ -88,7 +88,7 @@ The required members for a type to qualify as a sequence or sliceable may be inh
 
 ## 18.2 The Index type
 
-The `System.Index` type represents an *abstract* index which is either a *from-start index* or a *from-end index*.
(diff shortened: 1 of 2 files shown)


standard/ranges.md
```diff
@@ -88,7 +88,7 @@ The required members for a type to qualify as a sequence or sliceable may be inh
 
 ## 18.2 The Index type
 
-The `System.Index` type represents an *abstract* index which is either a *from-start index* or a *from-end index*.
+The `System.Index` type represents an *abstract* index which is either a from-start index or a from-end index.
```
jskeet, line 91: Not sure we necessarily need to remove the italics here - I'd have thought the italics around "abstract" are less useful than emphasizing that it can be from the start or from the end.

jskeet (approved):
Definitely worth waiting for Nigel's review too.

PR #1441, opened 2025-10-27
Minor tweaks to v8

RexJaeschke:
1. Updated the Foreword 
2. Fixed most lines longer than 80 chars, as reported by the Word converter, but left a few that only go a *few* chars past and that do *not* actually wrap!
3. Modify numerous term definitions by using singular instead of plural forms (sometimes by simply moving the trailing "s" outside the three-star terminator, and avoid having a leading uppercase letter just because that first word starts a sentence (by changing "Using directives ..." to "A using directive ..." for example).


BillWagner (approved):
This LGTM @RexJaeschke 

I'll :shipit: now.

PR #1442, opened 2025-10-27
“Review” of PR #1437

Nigel-Ecma:
Apologies, came here to write a review and the boat has sailed.

@jskeet –I don’t see why ranges.md@89 was changed at all and I wouldn’t even have looked at the line for #1437. The *abstract* is for emphasis, i.e. that it is abstract! Maybe @RexJaeschke has an English reason to remove the other italics, but to me it delineates the two different kinds of index clearly.

Looking at the line now I think the “is” would be better as “represents”, so I’ve made that change.

Over to you two to judge: accept/modify/reject the changed line – no version is wrong *per se*!

### DIFF
--- a/standard/ranges.md
+++ b/standard/ranges.md
@@ -86,7 +86,7 @@ The required members for a type to qualify as a sequence or sliceable may be inh
 
 ## 18.2 The Index type
 
-The `System.Index` type represents an *abstract* index which is either a from-start index or a from-end index.
+The `System.Index` type represents an *abstract* index which represents either a *from-start index* or a *from-end index*.
 
 ```csharp
     public readonly struct Index : IEquatable<Index>
--- a/standard/ranges.md
+++ b/standard/ranges.md
@@ -86,7 +86,7 @@ The required members for a type to qualify as a sequence or sliceable may be inh
 
 ## 18.2 The Index type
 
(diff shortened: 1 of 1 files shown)


jskeet (approved):
I'm fine with this; up to @RexJaeschke. (I'd still personally not emphasize the abstract - just its presence in the sentence is a sufficient highlight IMO - but I'm definitely not going to argue about it any further.)

BillWagner (approved):
This also looks good to me. As Jon, I'll wait for Rex's input as well.

RexJaeschke:
I'm OK with restoring the italics and the change to "represents." However, we definitely don't want "from-start index" and "from-end index" set in italics as they are *not* grammar rule names, just uses of previously defined terms.

standard/ranges.md
```diff
@@ -86,7 +86,7 @@ The required members for a type to qualify as a sequence or sliceable may be inh
 
 ## 18.2 The Index type
 
-The `System.Index` type represents an *abstract* index which is either a from-start index or a from-end index.
+The `System.Index` type represents an *abstract* index which represents either a *from-start index* or a *from-end index*.
```
Nigel-Ecma: Based on feedback from @RexJaeschke 

```suggestion
The `System.Index` type represents an *abstract* index which represents either a from-start index or a from-end index.
```

PR #1447, opened 2025-10-31
Update the introduction of open and closed types.

Nigel-Ecma:
Alternatives for the first line of §8.4.3, the starting point is the **temporary** reinstatement of the line as it was before #1441 was merged. See comments on the line.

### DIFF
--- a/standard/types.md
+++ b/standard/types.md
@@ -547,7 +547,7 @@ Each type argument shall satisfy any constraints on the corresponding type param
 
 ### 8.4.3 Open and closed types
 
-Any type can be classified as either an ***open type*** or a ***closed type***. An open type is a type that involves type parameters. More specifically:
+All types can be classified as either ***open types*** or ***closed types***. An open type is a type that involves type parameters. More specifically:
 
 - A type parameter defines an open type.
 - An array type is an open type if and only if its element type is an open type.
--- a/standard/types.md
+++ b/standard/types.md
@@ -547,7 +547,7 @@ Each type argument shall satisfy any constraints on the corresponding type param
 
 ### 8.4.3 Open and closed types
 
-All types can be classified as either ***open types*** or ***closed types***. An open type is a type that involves type parameters. More specifically:
+A type is either an ***open type*** or a ***closed type***. An open type is a type that involves type parameters. More specifically:
 
 - A type parameter defines an open type.
 - An array type is an open type if and only if its element type is an open type.


standard/types.md
```diff
@@ -547,7 +547,7 @@ Each type argument shall satisfy any constraints on the corresponding type param
 
 ### 8.4.3 Open and closed types
 
-Any type can be classified as either an ***open type*** or a ***closed type***. An open type is a type that involves type parameters. More specifically:
+All types can be classified as either ***open types*** or ***closed types***. An open type is a type that involves type parameters. More specifically:
```
Nigel-Ecma: The “can be” here should be something like “are” – we are defining things here. #1441 changes this line to:

```suggestion
Any type can be classified as either an ***open type*** or a ***closed type***. An open type is a type that involves type parameters. More specifically:
```

This still says “can be” so I think it is wrong. It also reads wrong to me, but that is subjective. I’m going to make some alternative suggestions, each in a seperate comment. Please 👍/👎🏻 as you see fit, including this one, and whomever merges can OK the preferred choice, or we don't merge and the #1441 version and “can be” stays.

Nigel-Ecma:
```suggestion
A type is classified as an ***open type*** or a ***closed type***. An open type is a type that involves type parameters. More specifically:
```

Nigel-Ecma: Is “Every” better than “A”?

```suggestion
Every type is classified as an ***open type*** or a ***closed type***. An open type is a type that involves type parameters. More specifically:
```

Nigel-Ecma: Is “classified” really required?

```suggestion
A type is either an ***open type*** or a ***closed type***. An open type is a type that involves type parameters. More specifically:
```

Nigel-Ecma: The last one feels awkward, this seems better but loses ***defined terms***, @RexJaeschke feel free to correct my use of *italics* in lieu!

```suggestion
Every type is either **open** or **closed**. An open type is a type that involves type parameters. More specifically:
```

Nigel-Ecma: Well those are my suggestions, all to remove “can be” – types *are* open/closed. Feel free to add your own of course :-)

Nigel-Ecma (commented):
I think “can be” is wrong, but I’m flexible on the wording beyond that – I’ve made a few suggestions but feel free to propose something better.

standard/types.md
BillWagner: I'm OK with most of the alternatives.

BillWagner (approved):
This LGTM once others weigh in as well.

standard/types.md
RexJaeschke: I prefer "A type is either an ***open type*** or a ***closed type***. " for 2 reasons:

1. When defining a term it's best to use the singular.
2. I prefer "type" as part of the term being defined, as without it the terms ***open*** and ***closed*** are just adjectives with no context.

Nigel-Ecma: With @RexJaeschke’s 👍 I think we have a winner (3 votes counting Rex), unless there are some late votes…

BillWagner:
closing and reopening for a fresh build to check #1477

BillWagner:
Adding the meeting priority label to make sure we discuss and take action in the next meeting.

jskeet (approved):

PR #1448, opened 2025-11-01
Small correction after #873 merge

Nigel-Ecma:
#873 changed `!x` to `!` on the second row of the table in §12.4.2; this PR reverts that change.

The `x` is required to indicate that this is the prefix logical negation operator, the first row of the table contains `x!` for the postfix null-forgiving operator


BillWagner (approved):
Thanks @Nigel-Ecma 

I'll merge this now.

PR #1449, opened 2025-11-03
Merge V8 features into draft-v9

BillWagner:
DON'T SQUASH MERGE THIS. WE WANT TO PRESERVE ALL PRs FROM V8

This only contains a merge commit from the latest `draft-v8` into the `draft-v9` branch.


BillWagner:
@RexJaeschke 

This is the precursor to creating the updated branches and PRs for the V9 features before our next meeting.

The only change is the merge, so while the diff is large, I'd only focus on status checks.

jskeet (approved):
Approving based on the comment suggesting to focus on the status checks, and the fact that those checks are green.

PR #1452, opened 2025-11-06
[Version 9.0] Feature support for `init` accessors

BillWagner:
This PR contains the work for `init` accessors in C# 9.

The commits from #978 were squashed to one commit in this branch.

There are a number of comments that haven't been addressed on #978:

- https://github.com/dotnet/csharpstandard/pull/978/files#r2326611994
- https://github.com/dotnet/csharpstandard/pull/978/files#r1378416566
- https://github.com/dotnet/csharpstandard/pull/978/files#r2326613914
- 

Notes:

- The ANTLR grammar validation fails for one test where the validation is designed to fail – it is testing that right shift is not erroneously recognised – and the checker passes the test if the error message is what it expected. Antlr writes a long informative error message listing all the tokens it might have expected to find at the failing point of the code and that list now has `init` in it.
- The renumbering action fails because this feature depends on the grammar in "records".


standard/basic-concepts.md
```diff
@@ -81,7 +81,7 @@ There are several different types of declaration spaces, as described in the fol
...
+- Each method declaration, property declaration, property accessor declaration, indexer declaration, indexer accessor declaration, operator declaration, instance constructor declaration, anonymous function, and local function creates a new declaration space called a ***local variable declaration space***. Names are introduced into this declaration space through parameters (*fixed_parameter*s and *parameter_array*s) and *type_parameter*s. The set and init accessor for a property or an indexer introduces the name `value` as a parameter. The body of the function member, anonymous function, or local function, if any, is considered to be nested within the local variable declaration space. When a local variable declaration space and a nested local variable declaration space contain elements with the same name, within the scope of the nested local name, the outer local name is hidden ([§7.7.1](basic-concepts.md#771-general)) by the nested local name.
```
RexJaeschke:
```suggestion
- Each method declaration, property declaration, property accessor declaration, indexer declaration, indexer accessor declaration, operator declaration, instance constructor declaration, anonymous function, and local function creates a new declaration space called a ***local variable declaration space***. Names are introduced into this declaration space through parameters (*fixed_parameter*s and *parameter_array*s) and *type_parameter*s. The set and init accessor for a property or an indexer introduce the name `value` as a parameter. The body of the function member, anonymous function, or local function, if any, is considered to be nested within the local variable declaration space. When a local variable declaration space and a nested local variable declaration space contain elements with the same name, within the scope of the nested local name, the outer local name is hidden ([§7.7.1](basic-concepts.md#771-general)) by the nested local name.
```

standard/attributes.md
```diff
@@ -105,9 +105,9 @@ class X : Attribute { ... }
 
 ### 23.2.3 Positional and named parameters
 
-Attribute classes can have ***positional parameter***s and ***named parameter***s. Each public instance constructor for an attribute class defines a valid sequence of positional parameters for that attribute class. Each non-static public read-write field and property for an attribute class defines a named parameter for the attribute class. For a property to define a named parameter, that property shall have both a public get accessor and a public set accessor.
+Attribute classes can have ***positional parameter***s and ***named parameter***s. Each public instance constructor for an attribute class defines a valid sequence of positional parameters for that attribute class. Each non-static public read-write field and non-static public read-write or read-init property for an attribute class defines a named parameter for the attribute class. For a property to define a named parameter, that property shall have both a public get accessor and a public set accessor.
```
RexJaeschke:
```suggestion
Attribute classes can have ***positional parameter***s and ***named parameter***s. Each public instance constructor for an attribute class defines a valid sequence of positional parameters for that attribute class. Each non-static public read-write field and non-static public read-write or read-init property for an attribute class defines a named parameter for the attribute class. For a property to define a named parameter, that property shall have both a public get accessor and a public set or init accessor.
```

standard/attributes.md
```diff
@@ -465,7 +465,7 @@ The compilation of an *attribute* with attribute class `T`, *positional_argumen
 - If `C` does not have public accessibility, then a compile-time error occurs.
 - For each *named_argument* `Arg` in `N`:
   - Let `Name` be the *identifier* of the *named_argument* `Arg`.
-  - `Name` shall identify a non-static read-write public field or property on `T`. If `T` has no such field or property, then a compile-time error occurs.
+  - `Name` shall identify a non-static read-write public field or a non-static public read-write or read-init property on `T`. If `T` has no such field or property, then a compile-time error occurs.
```
RexJaeschke:
```suggestion
  - `Name` shall identify a non-static read-write public field or a public non-static read-write or read-init property on `T`. If `T` has no such field or property, then a compile-time error occurs.
```

standard/expressions.md
```diff
@@ -450,27 +450,27 @@ Once a particular function member has been identified at binding-time, possibly
 >   <tr>
 >     <td rowspan="6">Property access</td>
 >     <td><code>P</code></td>
->     <td>The get accessor of the property <code>P</code> in the containing class or struct is invoked. A compile-time error occurs if <code>P</code> is write-only. If <code>P</code> is not <code>static</code>, the instance expression is <code>this</code>.</td>
+>     <td>The get accessor of the property <code>P</code> in the containing class or struct is invoked. A compile-time error occurs if <code>P</code> is write-only or init-only. If <code>P</code> is not <code>static</code>, the instance expression is <code>this</code>.</td>
```
RexJaeschke, line 453: Should "containing class or struct" be extended to include interface? Here and a few other table rows below?

standard/classes.md
```diff
@@ -3373,14 +3377,18 @@ The differences between static and instance members are discussed further in [§
 
 ### 15.7.3 Accessors
 
-*Note*: This subclause applies to both properties ([§15.7](classes.md#157-properties)) and indexers ([§15.9](classes.md#159-indexers)). The subclause is written in terms of properties, when reading for indexers substitute indexer/indexers for property/properties and consult the list of differences between properties and indexers given in [§15.9.2](classes.md#1592-indexer-and-property-differences). *end note*
+#### §accessors-general General
+
+*Note*: This subclause and its sibling subclauses apply to both properties ([§15.7](classes.md#157-properties)) and indexers ([§15.9](classes.md#159-indexers)). The clause is written in terms of properties, when reading for indexers substitute indexer/indexers for property/properties and consult the list of differences between properties and indexers given in [§15.9.2](classes.md#1592-indexer-and-property-differences). *end note*
```
RexJaeschke:
```suggestion
*Note*: This subclause and its sibling subclauses apply to both properties ([§15.7](classes.md#157-properties)) and indexers ([§15.9](classes.md#159-indexers)). The text is written in terms of properties; when reading for indexers substitute indexer/indexers for property/properties and consult the list of differences between properties and indexers given in [§15.9.2](classes.md#1592-indexer-and-property-differences). *end note*
```

standard/classes.md
```diff
@@ -3528,6 +3546,34 @@ Based on the presence or absence of the get and set accessors, a property is cla
 >
 > *end example*
 
+For more information about set and get accessors, see §set-and-get-accessors.
+
+> *Example*: Consider the following, immutable type, which has auto-implemented properties:
+>
+> <!-- Example: {template:"standalone-lib", name:"AccessorsInit1"} -->
+> <!-- FIX: create and add file PointStructWithInit.cs. -->
```
RexJaeschke:
```suggestion
```

BillWagner:
Thanks!

jskeet:

standard/classes.md
```diff
@@ -1395,7 +1395,9 @@ T get_P();
 void set_P(T value);
 ```
 
-Both signatures are reserved, even if the property is read-only or write-only.
+Both signatures are reserved, even if the property has only one accessor.
+
+A set accessor and init accessor have the same signature; however, that for the init accessor also has an implementation-defined form of annotation to distinguish it from a set accessor.
```
Nigel-Ecma, line 1400: I think this is description of a certain implementation, there appears to be no language level reason to *require* this. (Also there seems to be no CLR reason to do this; the CLR allows extra methods [accessors in C#] in a property. It is a particular C# implementation’s choice.)

It is also counter to the C# description of signature, which includes the *name*.

The language restriction is that there cannot be both a `set` and an `init` accessor.

Recommendation: remove


```suggestion
```

standard/classes.md
```diff
@@ -1395,7 +1395,9 @@ T get_P();
 void set_P(T value);
 ```
 
-Both signatures are reserved, even if the property is read-only or write-only.
+Both signatures are reserved, even if the property has only one accessor.
```
Nigel-Ecma: What does “reserved” here mean? While the CLR allows user-defined additions to `get` & `set` there is no mechanism to do this in C#. Is this needed?

standard/classes.md
```diff
@@ -1460,7 +1462,9 @@ T get_Item(L);
 void set_Item(L, T value);
 ```
 
-Both signatures are reserved, even if the indexer is read-only or write-only.
+Both signatures are reserved, even if the indexer has only one accessor.
```
Nigel-Ecma, line 1465: See comment for properties

standard/classes.md
```diff
@@ -1460,7 +1462,9 @@ T get_Item(L);
...
+
+A set accessor and init accessor have the same signature; however, that for the init accessor also has an implementation-defined form of annotation to distinguish it from a set accessor.
```
Nigel-Ecma, line 1467: See comment for properties

standard/classes.md
```diff
@@ -3421,15 +3433,15 @@ ref_accessor_body
     ;
 ```
 
-The *accessor_declarations* consist of a *get_accessor_declaration*, a *set_accessor_declaration*, or both. Each accessor declaration consists of optional attributes, an optional *accessor_modifier*, the token `get` or `set`, followed by an *accessor_body*.
+The *accessor_declarations* consist of either a *get_accessor_declaration*, optionally with a *set_accessor_declaration* or an *init_accessor_declaration*, or a *set_accessor_declaration* or *init_accessor_declaration* optionally with a *get_accessor_declaration*.  Each accessor declaration consists of optional *attributes*, an optional *accessor_modifier*, the token `get`, `init`, or `set`, followed by an *accessor_body*.
 
 For a ref-valued property the *ref_get_accessor_declaration* consists optional attributes, an optional *accessor_modifier*, the token `get`, followed by an *ref_accessor_body*.
```
Nigel-Ecma, line 3438: Missing word? Also an -> a


```suggestion
For a ref-valued property the *ref_get_accessor_declaration* consists of optional attributes, an optional *accessor_modifier*, the token `get`, followed by a *ref_accessor_body*.
```

standard/classes.md
```diff
@@ -3421,15 +3433,15 @@ ref_accessor_body
...
 
 The use of *accessor_modifier*s is governed by the following restrictions:
 
 - An *accessor_modifier* shall not be used in an explicit interface member implementation.
 - The *accessor_modifier* `readonly` is permitted only in a *property_declaration* or *indexer_declaration* that is contained directly by a *struct_declaration* ([§16.4.11](structs.md#16411-properties), [§16.4.13](structs.md#16413-indexers)).
-- For a property or indexer that has no `override` modifier, an *accessor_modifier* is permitted only if the property or indexer has both a get and set accessor, and then is permitted only on one of those accessors.
+- For a property or indexer that has no `override` modifier, an *accessor_modifier* is permitted only if the property or indexer has both a get and set or init accessor, and then is permitted only on one of those accessors.
```
Nigel-Ecma, line 3444: One for the grammarians, could not this be parsed (get and set) or init?

If so and it needs fixing check everywhere init has been added!

Update: asked a non-techie, they parsed it (get and set) or init – so with a sample of 1 we have 4000 DJT%™ for and before or ;-)

standard/classes.md
```diff
@@ -3479,11 +3491,17 @@ The body of a get accessor for a ref-valued property shall conform to the rules
...
+An init accessor corresponds to a method with a single value parameter of the property type and a `void` return type. The implicit parameter of an init accessor is always named `value`. Only during the construction phase of an object (§init-accessors), and if an init accessor exists, when a property is referenced as the target of an assignment ([§12.21](expressions.md#1223-assignment-operators)), or as the operand of `++` or `–-` ([§12.8.16](expressions.md#12816-postfix-increment-and-decrement-operators), [§12.9.7](expressions.md#1297-prefix-increment-and-decrement-operators)), the init accessor is invoked with an argument that provides the new value ([§12.23.2](expressions.md#12232-simple-assignment)). The body of an init accessor shall conform to the rules for `void` methods described in [§15.6.11](classes.md#15611-method-body). In particular, return statements in the init accessor body are not permitted to specify an expression. Since an init accessor implicitly has a parameter named `value`, it is a compile-time error for a local variable or constant declaration in an init accessor to have that name.
```
Nigel-Ecma, line 3494: The repetition obscures the important difference. I suggest something along the lines of:


```suggestion
An init accessor is similar to a set accessor, and is used in the same situations, except that it is restricted to be only invokable during the construction phase of an object (§init-accessors).
```

The error cases are already covered lines 3501 & 3504.

standard/classes.md
```diff
@@ -3531,6 +3549,33 @@ Based on the presence or absence of the get and set accessors, a property is cla
...
+> var p = new Point() { X = 42, Y = 13 };
+> ```
+>
+> *end example*
+
+For more information about init accessors, see §init-accessors.
+
+#### §set-and-get-accessors Set and get accessors
+
 The get and set accessors of a property are not distinct members, and it is not possible to declare the accessors of a property separately.
```
Nigel-Ecma, line 3580: Thus must apply to init properties as well. If we wish to keep the added  “Set and get accessors” header to match the also added “Init accessors” then maybe switch these two lines and add init, something like:


```suggestion
The get, set and init accessors of a property are not distinct members, and it is not possible to declare the accessors of a property separately.

#### §set-and-get-accessors Set and get accessors

```

That still leaves me with the question “distinct members” of what?

Clearly they are distinct members of the property (even if only notionally for set/init in a well known implementation – semantically they are distinct). Accessors are also defined seperately in braces (`{ get … set … }` which echoes are methods are cleared in a class.

So what is this trying to say? Is it really necessary at all?

standard/classes.md
```diff
@@ -3731,11 +3776,197 @@ Properties can be used to delay initialization of a resource until the moment it
...
+>     }
+> }
+> ```
+>
+> *end example*
+
+At the point an init accessor is invoked, the instance is known to be in the construction phase. Hence an init accessor may take the following actions in addition to what a set accessor can do:
+
+1. Call other init accessors available through `this` or `base`
+1. Assign `readonly` fields declared on the same type through `this`
```
Nigel-Ecma, line 3848: What is this “instance”, it seems to pop up out of nowhere. Just sketching, hence not a suggestion block…

> At the point a property’s init accessor is invoked, the property’s containing/owning type instance is known to be in the construction phase. Hence an init accessor may take the following actions in addition to what a set accessor can do:
> 1. Set the value of read-init/init-only properties accessible through `this` or `base`
> 1. Assign `readonly` fields declared on the same type through `this`

This also changes the wording “Call other init accessors” as code doesn’t directly call an accessor, unlike methods are directly called – the accessor is invoked as the consequence of some other action (assignment, `++`, etc.). This might be a wider issue though!

standard/classes.md
```diff
@@ -3731,11 +3776,197 @@ Properties can be used to delay initialization of a resource until the moment it
...
+
+class Init
+{
+    void M<T>() where T : IPerson, new()
+    {
+        var local = new T()
+        {
+            Name = "Jared"
+        };
+        local.Name = "Jraed"; // Error
```
Nigel-Ecma, line 3934: Intentionally 2 errors (assignment & spelling) or not? If intentional it is fine.

standard/classes.md
```diff
@@ -3942,14 +4173,15 @@ A property declaration that includes both the `abstract` and `override` modifier
...
+The override declaration and the overridden base property are required to have the same declared accessibility. In other words, an override declaration shall not change the accessibility of the base property. However, if the overridden base property is protected internal and it is declared in a different assembly than the assembly containing the override declaration then the override declaration’s declared accessibility shall be protected. If the inherited property has only a single accessor (i.e., if the inherited property is read-only, ninit-only, or write-only), the overriding property shall include only that accessor. If the inherited property includes two accessors (i.e., if the inherited property is read-write or read-init), the overriding property can include either a single accessor or both accessors. There shall be an identity conversion between the type of the overriding and the inherited property.
```
Nigel-Ecma, line 4176: ninit-only typo:


```suggestion
The override declaration and the overridden base property are required to have the same declared accessibility. In other words, an override declaration shall not change the accessibility of the base property. However, if the overridden base property is protected internal and it is declared in a different assembly than the assembly containing the override declaration then the override declaration’s declared accessibility shall be protected. If the inherited property has only a single accessor (i.e., if the inherited property is read-only, init-only, or write-only), the overriding property shall include only that accessor. If the inherited property includes two accessors (i.e., if the inherited property is read-write or read-init), the overriding property can include either a single accessor or both accessors. There shall be an identity conversion between the type of the overriding and the inherited property.
```

standard/classes.md
```diff
@@ -3942,14 +4173,15 @@ A property declaration that includes both the `abstract` and `override` modifier
...
 
 An overriding property declaration may include the `sealed` modifier. Use of this modifier prevents a derived class from further overriding the property. The accessors of a sealed property are also sealed.
 
 Except for differences in declaration and invocation syntax, virtual, sealed, override, and abstract accessors behave exactly like virtual, sealed, override and abstract methods. Specifically, the rules described in [§15.6.4](classes.md#1564-virtual-methods), [§15.6.5](classes.md#1565-override-methods), [§15.6.6](classes.md#1566-sealed-methods), and [§15.6.7](classes.md#1567-abstract-methods) apply as if accessors were methods of a corresponding form:
 
 - A get accessor corresponds to a parameterless method with a return value of the property type and the same modifiers as the containing property.
 - A set accessor corresponds to a method with a single value parameter of the property type, a void return type, and the same modifiers as the containing property.
+- An init accessor corresponds to a method with a single value parameter of the property type, a `void` return type, and the same modifiers as the containing property.
```
Nigel-Ecma, line 4184: Following the spirit of the suggestion on 3494, why not:


```suggestion
- A set or init accessor corresponds to a method with a single value parameter of the property type, a `void` return type, and the same modifiers as the containing property.
```

I tending to think that init & set accessors should be described as differing only in *when* they can be invoked, and not be listing out the (identical) other features.

standard/classes.md
```diff
@@ -4034,6 +4266,31 @@ When a property is declared as an override, any overridden accessors shall be ac
 >
 > *end example*
 
+When an init accessor appears in a virtual property, all overrides for it shall also be marked `init`. Likewise, it is not possible to override a set accessor with an init accessor.
```
Nigel-Ecma, line 4269: Is it stated elsewhere that all overrides for set & get also need to be “marked” set & get respectively?

I’m unsure whether the second sentence is saying something useful, or is maybe coming from a particular implementation translating an init into a set with flag. If it is the latter then an earlier suggestion has been made to remove this implementation specific detail, and if that is accepted then the text here might need to be reworded as well. (I.e. if the Standard never surfaces the implementation specific detail that an init is a set why would anyone think you can override one with the other?)

Nigel-Ecma (changes requested):
This is a partial review, stopped after reviewing `classes.md`.

I think changes are required.

standard/classes.md
jskeet, line 1398: We think this is okay - it's about the *method* signatures rather than the property accessor names being reserved.

standard/classes.md
jskeet: We could possibly add a note that `init_P` is *not* reserved. (But it doesn't need to be normative text.)

PR #1454, opened 2025-11-06
[Version 9.0] Feature support for top level statements

BillWagner:
This PR contains the work for Top-level statements in C# 9.

The commits from #980 were squashed to the first commit on this branch.

The ANTLR grammar is failing. I admit I'm not sure why. There are no grammar changes in this PR.


RexJaeschke:
@BillWagner There *is* one grammar change: the extension of *compilation_unit* to allow `statement_list*` before `namespace_member_declaration*`.

standard/basic-concepts.md
```diff
@@ -602,8 +771,8 @@ The ***scope*** of a name is the region of program text within which it is possi
...
+- The scope of a name defined by an *extern_alias_directive* ([§14.4](namespaces.md#144-extern-alias-directives)) extends over the *using_directive*s, *global_attributes*, *statement_list*s, and *namespace_member_declaration*s of its immediately containing *compilation_unit* or *namespace_body*. An *extern_alias_directive* does not contribute any new members to the underlying declaration space. In other words, an *extern_alias_directive* is not transitive, but, rather, affects only the *compilation_unit* or *namespace_body* in which it occurs.
+- The scope of a name defined or imported by a *using_directive* ([§14.5](namespaces.md#145-using-directives)) extends over the *global_attributes*, *statement_list*s, and *namespace_member_declaration*s of the *compilation_unit* or *namespace_body* in which the *using_directive* occurs. A *using_directive* may make zero or more namespace or type names available within a particular *compilation_unit* or *namespace_body*, but does not contribute any new members to the underlying declaration space. In other words, a *using_directive* is not transitive but rather affects only the *compilation_unit* or *namespace_body* in which it occurs.
```
RexJaeschke, line 777: I just made these two edits (after working on the V10 feature "File-Scoped Namespaces," which further changes these two bullet items). Now that top-level statements (*statement_list*s) are permitted immediately before *namespace_member_declaration*s, we need to make sure those statements are in the scope of these two directives.

BillWagner:
Thanks!

jskeet:

standard/basic-concepts.md
Nigel-Ecma: The commit I just pushed fixes *compilation_unit* to include zero or one *statement_list*, I missed these additions were plural - they should be singular.

```suggestion
- The scope of a name defined by an *extern_alias_directive* ([§14.4](namespaces.md#144-extern-alias-directives)) extends over the *using_directive*s, *global_attributes*, *statement_list*, and *namespace_member_declaration*s of its immediately containing *compilation_unit* or *namespace_body*. An *extern_alias_directive* does not contribute any new members to the underlying declaration space. In other words, an *extern_alias_directive* is not transitive, but, rather, affects only the *compilation_unit* or *namespace_body* in which it occurs.
- The scope of a name defined or imported by a *using_directive* ([§14.5](namespaces.md#145-using-directives)) extends over the *global_attributes*, *statement_list*, and *namespace_member_declaration*s of the *compilation_unit* or *namespace_body* in which the *using_directive* occurs. A *using_directive* may make zero or more namespace or type names available within a particular *compilation_unit* or *namespace_body*, but does not contribute any new members to the underlying declaration space. In other words, a *using_directive* is not transitive but rather affects only the *compilation_unit* or *namespace_body* in which it occurs.
```

standard/basic-concepts.md
```diff
@@ -1,59 +1,134 @@
...
+
+A method qualifies as a canditate entry point by satisfying the following requirements:
 
 - It shall have the name `Main`.
 - It shall be `static`.
 - It shall not be generic.
 - It shall be declared in a non-generic type. If the type declaring the method is a nested type, none of its enclosing types may be generic.
-- It may have the `async` modifier provided the method’s return type is `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`.
 - The return type shall be `void`, `int`, `System.Threading.Tasks.Task`, or `System.Threading.Tasks.Task<int>`.
+- It may have the `async` modifier only if the method’s return type is `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`.
```
jskeet, line 37:
```suggestion
- If the `async` modifier is present, the method’s return type shall be `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`.
```

(Just to avoid "may".)

standard/basic-concepts.md
```diff
@@ -1,59 +1,134 @@
...
 - It shall not be a partial method ([§15.6.9](classes.md#1569-partial-methods)) without an implementation.
 - The parameter list shall either be empty, or have a single value parameter of type `string[]`.
```
jskeet, line 39: How about `string?[]`, `string[]?` or `string?[]?` parameter types?

standard/basic-concepts.md
```diff
@@ -1,59 +1,134 @@
...
+The ***termination status code***, the purpose of which is to allow communication of success or failure to the host environment, is determined based on the result from the entry point invocation (§entry-point-invocation).
+
+### §named-entry-points Named entry points
+
+A method qualifies as a canditate entry point by satisfying the following requirements:
 
 - It shall have the name `Main`.
 - It shall be `static`.
 - It shall not be generic.
 - It shall be declared in a non-generic type. If the type declaring the method is a nested type, none of its enclosing types may be generic.
```
jskeet, line 35:
```suggestion
- It shall be declared in a non-generic type. If the type declaring the method is a nested type, all of its enclosing types shall be non-generic.
```

(Just to keep things consistent.)

standard/basic-concepts.md
```diff
@@ -1,59 +1,134 @@
...
+A candidate entry point specified by an external mechanism takes precedence in entry point selection (§entry-point-selection).
+
+### §entry-point-selection Entry point selection
+
+It is a compile-time error if there are no candidate entry points.
+
+Otherwise the entry point is selected from the candidates:
+
+- If one of the candidates is specified by an external mechanism (§externally-defined-entry-point) it is selected as the entry point;
+- Otherwise, iff one of the candidates is defined by top-level statements (§using-top-level-statements) it is selected as the entry point;
```
jskeet, line 105:
```suggestion
- Otherwise, if exactly one of the candidates is defined by top-level statements (§using-top-level-statements) it is selected as the entry point;
```

(Just to avoid expanding "iff".)

standard/basic-concepts.md
```diff
@@ -1,59 +1,134 @@
...
+
+A candidate entry point specified by an external mechanism takes precedence in entry point selection (§entry-point-selection).
+
+### §entry-point-selection Entry point selection
+
+It is a compile-time error if there are no candidate entry points.
+
+Otherwise the entry point is selected from the candidates:
+
+- If one of the candidates is specified by an external mechanism (§externally-defined-entry-point) it is selected as the entry point;
```
jskeet, line 104: We should sense-check this - is it really okay to have top-level statements and have them ignored due to specifying an entry point type separately? (And do we want to *mandate* that that's okay?)

standard/basic-concepts.md
```diff
@@ -1,59 +1,134 @@
...
+- Otherwise:
+  - If any of the candidates have a return type of `int` or `void` then any candidates having a return type of `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>` are removed from the candidate pool.
+  - If there is a single remaining candidate it is selected as the entry point.
+  - Otherwise the entry point cannot be determined and a compile-time error shall be reported.
+
+The entry point selected at compile-time is invoked at runtime as part of application startup (§entry-point-invocation).
+
+### §entry-point-invocation Entry point invocation
+
+If the entry point declares a parameter, then the implementation shall as the initial value of that parameter provide a non-null reference to a string array. This array shall consist of non-null references to zero or more strings, called ***application parameter***s, which are given implementation-defined values by the host environment prior to application startup.
```
jskeet, line 115: Should we rename these to "application arguments"? That would be more consistent with parameter/argument elsewhere? (I only noticed due to the note below.)

standard/basic-concepts.md
```diff
@@ -1,59 +1,134 @@
...
+
+> *Note*: On systems supporting a command line, application parameters correspond to what are generally known as command-line arguments. *end note*
+
+The application startup and termination process is semantically equivalent to the following steps:
+
+- An application run is started by either:
```
jskeet, line 121: We've got "application run" twice here, but we haven't defined it. We could just talk about an application being started...

standard/basic-concepts.md
```diff
@@ -1,59 +1,134 @@
...
+  - Invoking ([§12.8.10](expressions.md#12810-invocation-expressions)) the entry-point method, if its return type is `void` or `int`; or
+  - Awaiting ([§12.9.9](expressions.md#1299-await-expressions)) the result of invoking the entry-point method, if its return type is a `Task` type.
+  - In either case if the entry point requires an argument the application parameter array is supplied as its value.
```
jskeet, line 124:
```suggestion

In either case if the entry point requires an argument the application parameter array is supplied as its value.
```

standard/basic-concepts.md
```diff
@@ -1,59 +1,134 @@
...
+
+> *Note*: Invoking the entry-point method will cause the static constructor, if any, of the enclosing type to be executed first ([§15.12](classes.md#1512-static-constructors), [§16.4.10](structs.md#16410-static-constructors)). *end note*
+
+- The application is terminated
+  - If the run results in an `int` value it serves as the termination status code;
+  - Otherwise, if the run results in no return value the termination status code is `0`;
+  - Otherwise, if the run terminates due to an exception ([§22.4](exceptions.md#224-how-exceptions-are-handled)), the exit code is implementation-defined. Additionally, the implementation may provide alternative APIs for specifying the exit code.
```
jskeet, line 131:
```suggestion
  - Otherwise, if the run terminates due to an exception ([§22.4](exceptions.md#224-how-exceptions-are-handled)), the exit code is implementation-defined.

Additionally, the implementation may provide alternative APIs for specifying the exit code.
```

(It feels like that "additionally" isn't really part of the "otherwise".)

standard/basic-concepts.md
```diff
@@ -710,6 +785,14 @@ Within the scope of a local variable, it is a compile-time error to refer to the
 >
 > *end note*
 
+As described in §using-top-level-statements, top-level source tokens are enclosed by the generated entry-point method.
+
+For the purpose of simple-name evaluation, once the global namespace is reached, first, an attempt is made to evaluate the name within the generated entry-point method and only if this attempt fails is the evaluation within the global namespace declaration performed.
```
jskeet, line 790: This feels odd to me, but it does look correct. I'd expected the code below to compile, but no, it fails with an error of "Cannot use local variable or local function 'SimpleName' declared in a top-level statement in this context."

Progam.cs:

```csharp
int SimpleName = 10;
OtherClass.OtherMethod(SimpleName);
```

OtherClasses.cs:

```csharp
class SimpleName {}

class OtherClass
{
    public static void OtherMethod(int x) => Console.WriteLine(nameof(SimpleName));
}
```

jskeet: (Although using `typeof` instead of `nameof` works... I haven't chased down whether that's a contradiction or not; I haven't figured out whether `typeof` uses simple name evaluation.)

standard/basic-concepts.md
Nigel-Ecma: I’ll put it in next commit.

standard/basic-concepts.md
Nigel-Ecma: I haven’t checked. However the normal expectation is the entry point is called at startup and is passed the argument array – which is specified to be a non-null array of non-null strings and so an entry-point method can be written with that in mind.

standard/basic-concepts.md
Nigel-Ecma: I’ll put it in next commit.

standard/basic-concepts.md
Nigel-Ecma: I knew someone would comment on the iff ;-)

I’ll put it in next commit.

standard/basic-concepts.md
Nigel-Ecma: I did test current behaviour of an implementation.

Regardless it does make sense to me:

- top-level statements are just a shorthand for an entry-point method, they're not otherwise special for a user's perspective.

- command line arguments are often the way to override some default; here it is either resolving an ambiguity, thus avoiding an error, or overriding the selection made by the resolution rules.

Maybe the question to ask is why do the rules favour top-level statements over `Main` methods?

The answer may be that as the name used for a top-level statements entry point is implementation defined it is harder to pick use a command line argument – the current rules allow any of the methods to be picked, and if you didn't spot it I wrote the loose enough such that the external mechanism could select a method not automatically recognised as an entry-point. I haven’t tested with the CLI on my system supports it, I know the IDE doesn’t (it provide a menu of recognised entry-points) – but there seems to be no reason to forbid it.

standard/basic-concepts.md
Nigel-Ecma: A caller passes arguments, the callee has parameters. So the CLI passes application arguments, the entry-point gets application parameters. I just did a quick check and the C Standard has the names them this way as well. So I think application parameters is correct here. Maybe the Note could be written better?

standard/basic-concepts.md
Nigel-Ecma: I think “run” is used with its common (computing) meaning. In §3 we have:

> - **application** – assembly with an entry point
> - **program** – one or more compilation units that are presented to the compiler and are run or executed by an execution environment

that is the first use of “run”, is line 9 of this file. I have just gone with this use of “run”.

Reading the above definitions now I think the Standard might have more issues with “application” (84 occurrences) vs. “program” (186 occurrences) – is this section about application startup or program startup? Maybe we can leave that for another day…

standard/basic-concepts.md
Nigel-Ecma: I actually considered:

>The application startup and termination process is semantically equivalent to the following steps:
>
>- An application run is started by either:
>   - Invoking ([§12.8.10](expressions.md#12810-invocation-expressions)) the entry-point method, if its return type is `void` or `int`; or
>   - Awaiting ([§12.9.9](expressions.md#1299-await-expressions)) the result of invoking the entry-point method, if its return type is a `Task` type.
>- In either case if the entry point requires an argument the application parameter array is supplied as its value.

You’re suggesting taking it right out of the list:

>The application startup and termination process is semantically equivalent to the following steps:
>
>- An application run is started by either:
>   - Invoking ([§12.8.10](expressions.md#12810-invocation-expressions)) the entry-point method, if its return type is `void` or `int`; or
>   - Awaiting ([§12.9.9](expressions.md#1299-await-expressions)) the result of invoking the entry-point method, if its return type is a `Task` type.
>
>In either case if the entry point requires an argument the application parameter array is supplied as its value.

I dithered over what is in the commit and the first one above wishing to keep them connected; in yours I think the last bullet/line is too disconnected.

Would you prefer the first over the commit?

standard/basic-concepts.md
Nigel-Ecma: The statement appears to allow APIs that replace the three default options, but not how that interacts with entry-point return (e.g. an implementation could effectively return from the entry-point method on calling an “`exit(code)`” API). Maybe keep it in the list? 

>- The application is terminated
>   - If the run results in an `int` value it serves as the termination status code;
>   - Otherwise, if the run results in no return value the termination status code is `0`;
>   - Otherwise, if the run terminates due to an exception ([§22.4](exceptions.md#224-how-exceptions-are-handled)), the exit code is implementation-defined.
>- Additionally, the implementation may provide alternative APIs for specifying the exit code.

standard/basic-concepts.md
Nigel-Ecma: I’ll add this to the next commit

standard/basic-concepts.md
Nigel-Ecma: Note: This wording was in the received text, I just left it. What it does **not** say is that if the attempt to evaluate it in the generated method succeeds an error must be produced, should it?

I think it is odd as well.

On the one hand it makes no sense to say the top-level statements are wrapped in a method and then insert that method’s body before the global namespace.

However on the other hand those top-level statements *visually appear* to be at global level and so should be accessible, so recognising them and reporting an error could be deemed better UI.

It is said people get odder as they age, programming languages certainly do.

I would suggest adding a *Note* to suggest why this is done (either my guess above or some other explanation) @jskeet do you agree? Fancy writing one?

jskeet: @BillWagner to check with Mads and Fred about this. (Perhaps this is a Roslyn-specific value-add?)

jskeet:
Other than the noted oddity, we think this is ready for "final tweaks and final review".

standard/basic-concepts.md
BillWagner: The reason for this decision is to leave open the design space for top-level members.  See dotnet/csharplang#9803 and [the proposed feature spec](https://github.com/dotnet/csharplang/blob/main/proposals/top-level-members.md).

By binding to the "global" symbol instead of the symbols in other files, it ensures that the future feature (if implemented) won't break existing code with top-level statements and any associated local functions. The local functions may be promoted to top-level functions, and might be accessible in this case.

Nigel-Ecma: Taking this offline for the moment, something doesn’t appear to add up…

PR #1451, opened 2025-11-05
Update test templates for V9

BillWagner:
Update the language version in all our testing templates for C# 9.

Required for testing PRs for V9 features.

PR #1456, opened 2025-11-06
Tweak the MaybeNull attribute test examples

RexJaeschke:
[Addresses the 2nd test issue in PR #1451.]

@BillWagner I couldn't come up with a failing-code replacement for MaybeNull1Attribute, but other sibling attributes don't show an error case that is resolved by the presence of an attribute, so I mimicked them.

Note that I wrote "the compiler ***might*** generate a warning if the method could return `null`." In the test case, the return value of `default` is seen by the compiler, so it knows a null is being returned. I couldn't come up with a replacement for the return statement that didn't explicitly return null. So I couldn't say for certain that it ***would*** generate a warning.

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -982,15 +982,6 @@ Specifies that a non-nullable return value may be null.
 
 > *Example*: Consider the following generic method:
 >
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM @RexJaeschke 

I'll :shipit: now.

PR #1457, opened 2025-11-07
[Version 9.0] Feature support for native sized integers (`nint`, and `nuint`)

BillWagner:
I wrote the bulk of this spec some years ago (based on the [V9 MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/native-integers.md)), and was happy with it. Then several years later, when I started looking at V11 features, I discovered that this topic was revisited (see [V11 MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/numeric-intptr.md)). And I got confused as to which things should go in the V9 standard and which in the V11. Hopefully, I have it right now as far as V9 goes, but **anyone reviewing this PR would do well to at least be conversant with how things will be changed in V11, to make sure the V9 spec is complete/correct.** [Outside GitHub I have a Word document showing the expected changes for V11; ping me to get a copy (which Bill and Mads also have).] Rex

The commits from #1060 were squashed into one commit in this PR.

There are several comments on #1060 that need to be addressed:

- https://github.com/dotnet/csharpstandard/pull/1060#discussion_r2085211124
- https://github.com/dotnet/csharpstandard/pull/1060#discussion_r2085219962
- https://github.com/dotnet/csharpstandard/pull/1060#discussion_r2085239142
- https://github.com/dotnet/csharpstandard/pull/1060#discussion_r2085614999
- https://github.com/dotnet/csharpstandard/pull/1060#discussion_r2101572606


standard/types.md
```diff
@@ -271,7 +273,7 @@ A struct type is a value type that can declare constants, fields, methods, prope
 
 ### 8.3.5 Simple types
 
-C# provides a set of predefined `struct` types called the simple types. The simple types are identified through keywords, but these keywords are simply aliases for predefined `struct` types in the `System` namespace, as described in the table below.
```
RexJaeschke, line 279: I struck the first sentence as this is already stated in 8.3.1.

RexJaeschke:
Can someone please check if changes should be made to [§12.6.4.7](expressions.md#12647-better-conversion-target).

standard/expressions.md
```diff
@@ -2274,17 +2278,13 @@ In this case the compile-time type of the *element_access* depends on the compil
 
 If the *primary_expression* of an *element_access* is:
 
-- a value of an array type, the *element_access* is an array access ([§12.8.12.2](expressions.md#128122-array-access));
-- a value of `string` type, the *element_access* is a string access ([§12.8.12.3](expressions.md#128123-string-access));
-- otherwise, the *primary_expression* shall be a variable or value of a class, struct, or interface type that has one or more indexer members, in which case the *element_access* is an indexer access ([§12.8.12.4](expressions.md#128124-indexer-access)).
-
```
RexJaeschke, line 2256: These lines need to be restored. I have no idea how/why they were deleted.

standard/conversions.md
```diff
@@ -115,22 +115,28 @@ All identity conversions are symmetric. If an identity conversion exists from `T
 
 In most cases, an identity conversion has no effect at runtime. However, since floating point operations may be performed at higher precision than prescribed by their type ([§8.3.7](types.md#837-floating-point-types)), assignment of their results may result in a loss of precision, and explicit casts are guaranteed to reduce precision to what is prescribed by the type ([§12.9.8](expressions.md#1298-cast-expressions)).
 
+There is an identity conversion between `nint` and `System.IntPtr`, and between `nuint` and `System.UIntPtr`.
+
+For the compound types array, nullable type, constructed type, and tuple, there is an identity conversion between native integers ([§8.3.6]( types.md#836-integral-types)) and their underlying types.
```
RexJaeschke:
```suggestion
For the compound types array, nullable type, constructed type, and tuple, there is an identity conversion between native integers ([§8.3.6](types.md#836-integral-types)) and their underlying types.
```

standard/types.md
```diff
@@ -313,16 +317,78 @@ Because a simple type aliases a struct type, every simple type has members.
...
 
+Although `nint` and `nuint` shall be represented by the types `System.IntPtr` and `System.UIntPtr`, respectively, `nint` and `nuint` are *not* aliases for those types. As such, not all members of the corresponding `System` types are defined for `nint` and `nuint`. Instead, the compiler shall make available additional conversions and operations for the types `System.IntPtr` and `System.UIntPtr` when used in the context of native integer types.
+
+Consider the following:
+
+<!-- Example: {template:"standalone-console-without-using", name:"SimpleTypes2", expectedErrors:["CS0266"], ignoredWarnings:["CS0219"]} -->
+```csharp
+nint a1 = 1;           // OK
+System.IntPtr a2 = 1;  // Error: no implicit conversion
+```
```
RexJaeschke:
```suggestion
```

standard/types.md
```diff
@@ -313,16 +317,78 @@ Because a simple type aliases a struct type, every simple type has members.
...
+
+While the implementation provides operations and conversions for `nint` and `nuint` that are appropriate for integer types, those operations and conversions are not available on the `System` type counterparts. Similarly,
```
RexJaeschke:
```suggestion
While the implementation provides operations and conversions for `nint` and `nuint` that are appropriate for integer types, those operations and conversions are not available on the `System` type counterparts. For example,
```

standard/types.md
```diff
@@ -313,16 +317,72 @@ Because a simple type aliases a struct type, every simple type has members.
 >
 > *end note*.
 
+Although `nint` and `nuint` shall be represented by the types `System.IntPtr` and `System.UIntPtr`, respectively, `nint` and `nuint` are *not* aliases for those types. As such, not all members of the corresponding `System` types are defined for `nint` and `nuint`. Instead, the compiler shall make available additional conversions and operations for the types `System.IntPtr` and `System.UIntPtr` when used in the context of native integer types.
+
+Consider the following:
+
```
RexJaeschke:
```suggestion
```

BillWagner:
Retaining draft status for this meeting. 

If there's time, I'd like to have everyone get a first look at this before the January meeting.

There's an interesting time discussion: The changes to the feature will remove quite a bit of language from the standard. Once the keywords `nint` and `nuint` are synonyms for `System.IntPtr` and `System.UIntPtr`, the text for overloading, overriding and hiding goes away. The implicit conversions also go away. The built-in operator discussion changes: the built-in operators become the members of those `System` types.

Do we want full fidelity in both versions? If not, how much hand-waving for temporary C# 9 behavior do we allow?

standard/conversions.md
```diff
@@ -115,22 +115,28 @@ All identity conversions are symmetric. If an identity conversion exists from `T
 
 In most cases, an identity conversion has no effect at runtime. However, since floating point operations may be performed at higher precision than prescribed by their type ([§8.3.7](types.md#837-floating-point-types)), assignment of their results may result in a loss of precision, and explicit casts are guaranteed to reduce precision to what is prescribed by the type ([§12.9.8](expressions.md#1298-cast-expressions)).
 
+There is an identity conversion between `nint` and `System.IntPtr`, and between `nuint` and `System.UIntPtr`.
+
+For the compound types array, nullable type, constructed type, and tuple, there is an identity conversion between native integers ([§8.3.6](types.md#836-integral-types)) and their underlying types.
```
jskeet: I'm not sure what this means - could you put an example as a comment, and then we can figure out whether we should include it in the text?

standard/types.md
```diff
@@ -313,16 +319,72 @@ Because a simple type aliases a struct type, every simple type has members.
 >
 > *end note*.
 
+Although `nint` and `nuint` shall be represented by the types `System.IntPtr` and `System.UIntPtr`, respectively, `nint` and `nuint` are *not* aliases for those types. As such, not all members of the corresponding `System` types are defined for `nint` and `nuint`. Instead, the compiler shall make available additional conversions, unary operators (§12.9) and arithmetic operators (§12.12) for the types `System.IntPtr` and `System.UIntPtr` when used in the context of native integer types.
```
jskeet: We've said that there's an identity conversion between `nint` and `IntPtr` - but do we say elsewhere that `IntPtr` *must* be used to represent `nint` values? It may well be somewhere, and I've missed it...

standard/types.md
```diff
@@ -313,16 +319,72 @@ Because a simple type aliases a struct type, every simple type has members.
...
+public static readonly IntPtr Zero; // use 0 instead
+public static int Size { get; }     // use sizeof() instead
+public static IntPtr Add(IntPtr pointer, int offset);
+public static IntPtr Subtract(IntPtr pointer, int offset);
+public int ToInt32();
+public long ToInt64();
+public void* ToPointer();
+```
+
+The remaining members of `System.IntPtr` and `System.UIntPtr` are implicitly included in `nint` and `nuint`. These are:
```
jskeet: "These are" sounds like we're prohibiting the introduction of other members, which we've previously said would be okay for an implementation to do. Let's discuss what we want to require here.

standard/types.md
jskeet: (Line 365 of this file states it, effectively - but given that that is after this paragraph, presumably it's somewhere else as well...)

jskeet: (Not sure what "shall be represented" even means in the language spec...)

jskeet:
Next step: @BillWagner to see whether it would be possible to spec this in a way that looks forward to C# 11 more, defining them as aliases but specifying the exceptions in one place (e.g. no bitwise shift on IntPtr), and then removing that section for C# 11.

jnm2:
> There's an interesting time discussion: The changes to the feature will remove quite a bit of language from the standard. Once the keywords `nint` and `nuint` are synonyms for `System.IntPtr` and `System.UIntPtr`, the text for overloading, overriding and hiding goes away. The implicit conversions also go away. The built-in operator discussion changes: the built-in operators become the members of those `System` types.
> 
> Do we want full fidelity in both versions? If not, how much hand-waving for temporary C# 9 behavior do we allow?

I favor full fidelity. I'm not sure we can call them aliases of the same type and place exceptions on them; wouldn't that effectively be making them separate types?

Something we didn't discuss in the meeting is that the C# 11 feature is only available on some runtimes. At the top of the [C# 11 spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/numeric-intptr.md):

> The `System.Runtime.CompilerServices.RuntimeFeature.NumericIntPtr` runtime feature flag triggers this new behavior.

So it seems that even C# 11 still maintains `nint` as a separate type from `IntPtr` if the runtime doesn't support unifying. I'm getting more details on what exactly the runtime needs to be providing for the language feature to work; maybe just the definition of new members on (U)IntPtr? Maybe behavioral changes (see the [Breaking Changes](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/numeric-intptr.md#various-considerations) section)?

What will the C# spec do in 11 then? Keep the feature "bimodal" forever, or assume that C# 11 _requires_ .NET 7?

Update: I've raised the same question, but a more time-pressing version of it, at <https://github.com/dotnet/csharpstandard/issues/1532>

standard/arrays.md
```diff
@@ -112,7 +112,7 @@ Elements of arrays created by *array_creation_expression*s are always initialize
 
 ## 17.4 Array element access
 
-Array elements are accessed using the *array access* variant of *element_access* expressions ([§12.8.12.2](expressions.md#128122-array-access)) of the form `A[I₁, I₂, ..., Iₓ]`, where `A` is an expression of an array type and each `Iₑ` is an expression of type `int`, `uint`, `long`, `ulong`, or can be implicitly converted to one or more of these types. The result of an array access is a variable reference ([§9.5](variables.md#95-variable-references)) to the array element selected by the indices.
+Array elements are accessed using the *array access* variant of *element_access* expressions ([§12.8.12.2](expressions.md#128122-array-access)) of the form `A[I₁, I₂, ..., Iₓ]`, where `A` is an expression of an array type and each `Iₑ` is an expression of type `int`, `uint`, `nint`, `nuint`, `long`, `ulong`, or can be implicitly converted to one or more of these types. The result of an array access is a variable reference ([§9.5](variables.md#95-variable-references)) to the array element selected by the indices.
```
jnm2, line 157: What was the reason to have listed `int` and `uint` in the first place? They fall under "can be implicitly converted to `long`/`ulong`."

standard/conversions.md
jnm2: I think this means:
```cs
IntPtr[] array1 = new nint[1];
nint[] array2 = new IntPtr[1];
IntPtr? nullable1 = new nint?();
nint? nullable2 = new IntPtr?();
List<IntPtr> list1 = new List<nint>();
List<nint> list2 = new List<IntPtr>();
(nint, IntPtr) tuple = default((IntPtr, nint));
```

standard/conversions.md
```diff
@@ -115,22 +115,28 @@ All identity conversions are symmetric. If an identity conversion exists from `T
...
-- From `short` to `int`, `long`, `float`, `double`, or `decimal`.
-- From `ushort` to `int`, `uint`, `long`, `ulong`, `float`, `double`, or `decimal`.
-- From `int` to `long`, `float`, `double`, or `decimal`.
-- From `uint` to `long`, `ulong`, `float`, `double`, or `decimal`.
+- From `sbyte` to `short`, `int`, `nint`, `long`, `float`, `double`, or `decimal`.
+- From `byte` to `short`, `ushort`, `int`, `uint`, `nint`, `nuint`, `long`, `ulong`, `float`, `double`, or `decimal`.
+- From `short` to `int`, `nint`, `long`, `float`, `double`, or `decimal`.
+- From `ushort` to `int`, `uint`, `nint`, `nuint`, `long`, `ulong`, `float`, `double`, or `decimal`.
+- From `int` to `nint`, `long`, `float`, `double`, or `decimal`.
+- From `uint` to `long`, `nuint`, `ulong`, `float`, `double`, or `decimal`.
```
jnm2: Elsewhere, the sizes of the integers only get wider as the list goes on:
```suggestion
- From `uint` to `nuint`, `long`, `ulong`, `float`, `double`, or `decimal`.
```

standard/conversions.md
```diff
@@ -315,9 +321,11 @@ This implicit conversion seemingly violates the advice in the beginning of [§10
 
 An implicit constant expression conversion permits the following conversions:
 
-- A *constant_expression* ([§12.25](expressions.md#1225-constant-expressions)) of type `int` can be converted to type `sbyte`, `byte`, `short`, `ushort`, `uint`, or `ulong`, provided the value of the *constant_expression* is within the range of the destination type.
+- A *constant_expression* ([§12.25](expressions.md#1225-constant-expressions)) of type `int` can be converted to type `sbyte`, `byte`, `short`, `ushort`, `uint`, `nint`, `nuint`, or `ulong`, provided the value of the *constant_expression* is within the range of the destination type.
```
jnm2: I'm not sure `nint` would be mentioned here because it's not a special case of "the constant value is in a range which permits it." Rather, _every_ `int` can be converted to `nint`.
```suggestion
- A *constant_expression* ([§12.25](expressions.md#1225-constant-expressions)) of type `int` can be converted to type `sbyte`, `byte`, `short`, `ushort`, `uint`, `nuint`, or `ulong`, provided the value of the *constant_expression* is within the range of the destination type.
```

standard/conversions.md
```diff
@@ -413,18 +421,20 @@ The explicit conversions that are not implicit conversions are conversions that
...
-- From `short` to `sbyte`, `byte`, `ushort`, `uint`, `ulong`, or `char`.
+- From `short` to `sbyte`, `byte`, `ushort`, `uint`, `nuint`, `ulong`, or `char`.
 - From `ushort` to `sbyte`, `byte`, `short`, or `char`.
-- From `int` to `sbyte`, `byte`, `short`, `ushort`, `uint`, `ulong`, or `char`.
-- From `uint` to `sbyte`, `byte`, `short`, `ushort`, `int`, or `char`.
-- From `long` to `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `ulong`, or `char`.
-- From `ulong` to `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, or `char`.
+- From `int` to `sbyte`, `byte`, `short`, `ushort`, `uint`, `nuint`, `ulong`, or `char`.
+- From `uint` to `sbyte`, `byte`, `short`, `ushort`, `int`, `nint`, or `char`.
+- From `nint` to `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `nuint`, `long`, `ulong`, or `char`.
```
jnm2: There is already an implicit numeric conversion from `nint` to `long`
```suggestion
- From `nint` to `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `nuint`, `ulong`, or `char`.
```

standard/conversions.md
```diff
@@ -413,18 +421,20 @@ The explicit conversions that are not implicit conversions are conversions that
...
+- From `nuint` to `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `nint`, or `char`.
```
jnm2: A `nuint` has the same size as `ulong` on 64-bit machines and thus might contain values too large for a `long`.

```suggestion
- From `nuint` to `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `nint`, `long`, or `char`.
```

standard/conversions.md
```diff
@@ -749,6 +759,12 @@ Evaluation of a nullable conversion based on an underlying conversion from `S`
 - If the nullable conversion is from `S` to `T?`, the conversion is evaluated as the underlying conversion from `S` to `T` followed by a wrapping from `T` to `T?`.
 - If the nullable conversion is from `S?` to `T`, the conversion is evaluated as an unwrapping from `S?` to `S` followed by the underlying conversion from `S` to `T`.
 
+Conversion from `A` to `Nullable<B>` is:
+
+- an implicit nullable conversion if there is an identity conversion or implicit conversion from `A` to `B`;
+- an explicit nullable conversion if there is an explicit conversion from `A` to `B`;
+- otherwise, invalid.
```
jnm2: Instead of adding this new section and the new letters A and B, what if we took this section above:
```cs
- An implicit or explicit conversion from `S?` to `T?`
- An implicit or explicit conversion from `S` to `T?`
- An explicit conversion from `S?` to `T`.
```

And augmented it thus:
```cs
- A conversion from `S?` to `T?`, implicit if the conversion from `S` to `T` is implicit, otherwise explicit
- A conversion from `S` to `T?`, implicit if the conversion from `S` to `T` is implicit, otherwise explicit
- A conversion from `S?` to `T`, always explicit
```

I think "invalid" is implied; when no conversion exists from `S` to `T`, we will not have gotten anywhere in this section.

standard/expressions.md
```diff
@@ -297,7 +297,7 @@ When overload resolution rules ([§12.6.4](expressions.md#1264-overload-resoluti
 
 **This subclause is informative.**
 
-Unary numeric promotion occurs for the operands of the predefined `+`, `-`, and `~` unary operators. Unary numeric promotion simply consists of converting operands of type `sbyte`, `byte`, `short`, `ushort`, or `char` to type `int`. Additionally, for the unary – operator, unary numeric promotion converts operands of type `uint` to type `long`.
+Unary numeric promotion occurs for the operands of the predefined `+`, `–`, and `~` unary operators. Unary numeric promotion simply consists of converting operands of type `sbyte`, `byte`, `short`, `ushort`, or `char` to type `int`. Additionally, for the unary – operator, unary numeric promotion converts operands of type `uint` or `nint` to type `long`.
```
jnm2: Is the en dash used when referring to the operator, despite the fact that the syntax for the operator must use a hyphen?

standard/expressions.md
```diff
@@ -310,10 +310,12 @@ Binary numeric promotion occurs for the operands of the predefined `+`, `-`, `*`
...
 - Otherwise, if either operand is of type `float`, the other operand is converted to type `float`.
-- Otherwise, if either operand is of type `ulong`, the other operand is converted to type `ulong`, or a binding-time error occurs if the other operand is of `type sbyte`, `short`, `int`, or `long`.
+- Otherwise, if either operand is of type `ulong`, the other operand is converted to type `ulong`, or a binding-time error occurs if the other operand is of type `sbyte`, `short`, `int`, `nint`, or `long`.
+- Otherwise, if either operand is of type `nuint`, the other operand is converted to type `nuint`, or a binding-time error occurs if the other operand is of type `sbyte`, `short`, `int`, `nint`, or `long`.
 - Otherwise, if either operand is of type `long`, the other operand is converted to type `long`.
-- Otherwise, if either operand is of type `uint` and the other operand is of type `sbyte`, `short`, or `int`, both operands are converted to type `long`.
+- Otherwise, if either operand is of type `uint` and the other operand is of type `sbyte`, `short`, `nint`, or `int`, both operands are converted to type `long`.
```
jnm2: To keep it increasing in size:
```suggestion
- Otherwise, if either operand is of type `uint` and the other operand is of type `sbyte`, `short`, `int`, or `nint`, both operands are converted to type `long`.
```

standard/expressions.md
```diff
@@ -310,10 +310,12 @@ Binary numeric promotion occurs for the operands of the predefined `+`, `-`, `*`
 - If either operand is of type `decimal`, the other operand is converted to type `decimal`, or a binding-time error occurs if the other operand is of type `float` or `double`.
 - Otherwise, if either operand is of type `double`, the other operand is converted to type `double`.
 - Otherwise, if either operand is of type `float`, the other operand is converted to type `float`.
-- Otherwise, if either operand is of type `ulong`, the other operand is converted to type `ulong`, or a binding-time error occurs if the other operand is of `type sbyte`, `short`, `int`, or `long`.
+- Otherwise, if either operand is of type `ulong`, the other operand is converted to type `ulong`, or a binding-time error occurs if the other operand is of type `sbyte`, `short`, `int`, `nint`, or `long`.
+- Otherwise, if either operand is of type `nuint`, the other operand is converted to type `nuint`, or a binding-time error occurs if the other operand is of type `sbyte`, `short`, `int`, `nint`, or `long`.
```
jnm2, line 314: Could we simplify "or a binding-time error occurs" with the idea that the other operand is only converted if an implicit numeric conversion exists, otherwise a binding-time error occurs? Otherwise it seems like we're maintaining a copy of the explicit numeric conversions list (i.e. the inverse of the implicit numeric conversions list).

standard/expressions.md
```diff
@@ -1136,10 +1138,10 @@ Given two types `T₁` and `T₂`, `T₁` is a ***better conversion target*** th
 - `T₁` is `«TaskType»<S₁>`([§15.14.1](classes.md#15141-general)), `T₂` is `«TaskType»<S₂>`, and `S₁` is a better conversion target than `S₂`
 - `T₁` is `«TaskType»<S₁>`([§15.14.1](classes.md#15141-general)), `T₂` is `«TaskType»<S₂>`, and `T₁` is more specialized than `T₂`
 - `T₁` is `S₁` or `S₁?` where `S₁` is a signed integral type, and `T₂` is `S₂` or `S₂?` where `S₂` is an unsigned integral type. Specifically:
-  - `S₁` is `sbyte` and `S₂` is `byte`, `ushort`, `uint`, or `ulong`
-  - `S₁` is `short` and `S₂` is `ushort`, `uint`, or `ulong`
-  - `S₁` is `int` and `S₂` is `uint`, or `ulong`
-  - `S₁` is `long` and `S₂` is `ulong`
+  - `S₁` is `sbyte` and `S₂` is `byte`, `ushort`, `uint`, `nuint`, or `ulong`
+  - `S₁` is `short` and `S₂` is `ushort`, `uint`, `nuint`, or `ulong`
+  - `S₁` is `int` and `S₂` is `uint`, `nuint`, `nuint`, or `ulong`
```
```suggestion
  - `S₁` is `int` and `S₂` is `uint`, `nuint`, or `ulong`
```

standard/expressions.md
```diff
@@ -1136,10 +1138,10 @@ Given two types `T₁` and `T₂`, `T₁` is a ***better conversion target*** th
...
+  - `S₁` is `long` and `S₂` is `nuint`, or `ulong`
```
```suggestion
  - `S₁` is `long` and `S₂` is `nuint` or `ulong`
```

standard/expressions.md
```diff
@@ -2591,7 +2593,9 @@ The run-time processing of an *object_creation_expression* of the form new `T(A)
   - An instance of type `T` is created by allocating a temporary local variable. Since an instance constructor of a *struct_type* is required to definitely assign a value to each field of the instance being created, no initialization of the temporary variable is necessary.
   - The instance constructor is invoked according to the rules of function member invocation ([§12.6.6](expressions.md#1266-function-member-invocation)). A reference to the newly allocated instance is automatically passed to the instance constructor and the instance can be accessed from within that constructor as this.
 
-##### 12.8.17.2.2 Object initializers
+`new nint()` is equivalent to `(nint)0`, and `new nuint()` is equivalent to `(nuint)0`.
+
+#### 12.8.17.2.2 Object initializers
```
```suggestion
##### 12.8.17.2.2 Object initializers
```

standard/expressions.md
```diff
@@ -2284,7 +2286,7 @@ For an array access the *argument_list* shall not contain named arguments or by-
 
 The number of expressions in the *argument_list* shall be the same as the rank of the *array_type*, and each expression shall be:
 
-- of type `int`, `uint`, `long`, or `ulong`; or
+- of type `int`, `uint`, `nint`, `nuint`, `long`, or `ulong`; or
 - for single rank array access only, of type `Index` or `Range`; or
 - be implicitly convertible to one or more of the above types.
```
jnm2: For consistency with the prior `or`s:
```suggestion
- implicitly convertible to one or more of the above types.
```

standard/conversions.md
```diff
@@ -315,9 +321,11 @@ This implicit conversion seemingly violates the advice in the beginning of [§10
 
 An implicit constant expression conversion permits the following conversions:
 
-- A *constant_expression* ([§12.25](expressions.md#1225-constant-expressions)) of type `int` can be converted to type `sbyte`, `byte`, `short`, `ushort`, `uint`, or `ulong`, provided the value of the *constant_expression* is within the range of the destination type.
+- A *constant_expression* ([§12.25](expressions.md#1225-constant-expressions)) of type `int` can be converted to type `sbyte`, `byte`, `short`, `ushort`, `uint`, `nint`, `nuint`, or `ulong`, provided the value of the *constant_expression* is within the range of the destination type.
 - A *constant_expression* of type `long` can be converted to type `ulong`, provided the value of the *constant_expression* is not negative.
 
+The range for constants of `nint` is the same range as `int`. The range of constants for `nuint` is the same range as `uint`.
```
jnm2: 12.25 Constant expressions now already states this, so should it be a note linking there?

> A *constant_expression* of type `nint` shall have a value in the range \[`int.MinValue`,`int.MaxValue`\]. A *constant_expression* of type `nuint` shall have a value in the range \[`uint.MinValue`,`uint.MaxValue`\].

standard/expressions.md
```diff
@@ -7023,11 +7029,13 @@ constant_expression
     ;
 ```
 
-A constant expression shall either have the value `null` or one of the following types:
+A constant expression may be either a value type or a reference type. If a constant expression has a value type, that type shall be one of the following: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `nint`, `nuint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type. If a constant expression has a reference type, the expression shall:
+
+- have a value of type `string`;
+- have a value of `null`; or
+- be a default value expression ([§12.8.21](expressions.md#12821-default-value-expressions)) of reference type.
```
jnm2, line 7207: This change no longer permits a constant expression of `null` of a nullable value type, which would make the following code invalid:

```cs
class C
{
    void M(int? x = null) { }
}
```

<https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/classes.md#15621-general>:

> An input parameter may have a *default_argument*. The *expression* in a *default_argument* shall be one of the following:
> 
> - a *constant_expression*
> - an expression of the form `new S()` where `S` is a value type
> - an expression of the form `default(S)` where `S` is a value type

standard/expressions.md
```diff
@@ -7077,6 +7085,16 @@ Whenever an expression fulfills the requirements listed above, the expression is
...
 
+The compile-time evaluation of a *constant_expression* shall:
+
+- treat all values of type `nint` as `System.Int32`;
+- treat all values of type `nuint` as `System.UInt32`;
+- and otherwise use the same evaluation rules as for run-time non-constant expressions.
+
+If any `nint`/`nuint` values are not representable as `Int32`/`UInt32`, or the run-time evaluation of the whole expression would throw an exception, then a compile-time error shall be produced.
+
+> *Note*: These rules mean that an expression involving native integers which is superficially valid as a *constant_expression* may only be valid as an *expression* evaluated at runtime using the full implementation-defined native integer precision. *end note*
```
jnm2: I'd love to see an example!

standard/expressions.md
```diff
@@ -7077,6 +7085,16 @@ Whenever an expression fulfills the requirements listed above, the expression is
 
 The compile-time evaluation of constant expressions uses the same rules as run-time evaluation of non-constant expressions, except that where run-time evaluation would have thrown an exception, compile-time evaluation causes a compile-time error to occur.
 
+The compile-time evaluation of a *constant_expression* shall:
+
+- treat all values of type `nint` as `System.Int32`;
+- treat all values of type `nuint` as `System.UInt32`;
+- and otherwise use the same evaluation rules as for run-time non-constant expressions.
+
+If any `nint`/`nuint` values are not representable as `Int32`/`UInt32`, or the run-time evaluation of the whole expression would throw an exception, then a compile-time error shall be produced.
```
jnm2, line 7265: Aren't all `nint`/`nuint` values within a constant expression representable as `Int32`/`UInt32`, given this line above?

> A *constant_expression* of type `nint` shall have a value in the range \[`int.MinValue`,`int.MaxValue`\]. A *constant_expression* of type `nuint` shall have a value in the range \[`uint.MinValue`,`uint.MaxValue`\].

standard/types.md
```diff
@@ -313,16 +319,72 @@ Because a simple type aliases a struct type, every simple type has members.
...
+public override int GetHashCode();
+public override string ToString();
+public string ToString(string format);
+```
+
+Interfaces implemented by `System.IntPtr` and `System.UIntPtr` are implicitly included in `nint` and `nuint`, with occurrences of the underlying types replaced by the corresponding native integer types. For example, if `IntPtr` implements `ISerializable`, `IEquatable<IntPtr>`, and `IComparable<IntPtr>`, then `nint` implements `ISerializable`, `IEquatable<nint>`, and `IComparable<nint>`.
+
+`typeof(nint)` is `typeof(System.IntPtr)`, and `typeof(nuint)` is `typeof(System.UIntPtr)`.
+
+<!-- The following three paragraphs can be deleted for C# 10 -->
```
jnm2:
```suggestion
<!-- The following three paragraphs can be deleted for C# 11 -->
```

standard/unsafe-code.md
```diff
@@ -616,6 +616,8 @@ T* operator –(T* x, ulong y);
 long operator –(T* x, T* y);
 ```
 
+There are no predefined operators for pointer addition or subtraction with native integer ([§8.3.6]( types.md#836-integral-types)) offsets. Instead, `nint` and `nuint` values shall be promoted to `long` and `ulong`, respectively, with pointer arithmetic using the predefined operators for those types.
```
jnm2:
```suggestion
There are no predefined operators for pointer addition or subtraction with native integer ([§8.3.6](types.md#836-integral-types)) offsets. Instead, `nint` and `nuint` values shall be promoted to `long` and `ulong`, respectively, with pointer arithmetic using the predefined operators for those types.
```

jnm2 (changes requested):

standard/types.md
Nigel-Ecma: I think we *might* be able to remove a lot of the distinction, but more work required for us to be sure. What I think we might be seeing here are two different things:

1. The *language* adding contextual keywords in C#9, and moving to keywords in C#11
2. An *implementation* changing how it provides the language features in its C#11 release compared to how it provided them in its C#9 & C#10 releases.

If this is correct then (1) is for the Standard and (2) is not.

Nigel-Ecma:
(Quoting as this is coming some way down the conversation)

> > There's an interesting time discussion: The changes to the feature will remove quite a bit of language from the standard. Once the keywords `nint` and `nuint` are synonyms for `System.IntPtr` and `System.UIntPtr`, the text for overloading, overriding and hiding goes away. The implicit conversions also go away. The built-in operator discussion changes: the built-in operators become the members of those `System` types.
> > Do we want full fidelity in both versions? If not, how much hand-waving for temporary C# 9 behavior do we allow?
> 
> I favor full fidelity. I'm not sure we can call them aliases of the same type and place exceptions on them; wouldn't that effectively be making them separate types?
> 
> Something we didn't discuss in the meeting is that the C# 11 feature is only available on some runtimes. At the top of the [C# 11 spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/numeric-intptr.md):
> 
> > The `System.Runtime.CompilerServices.RuntimeFeature.NumericIntPtr` runtime feature flag triggers this new behavior.
> 
> So it seems that even C# 11 still maintains `nint` as a separate type from `IntPtr` if the runtime doesn't support unifying. I'm getting more details on what exactly the runtime needs to be providing for the language feature to work; maybe just the definition of new members on (U)IntPtr? Maybe behavioral changes (see the [Breaking Changes](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/numeric-intptr.md#various-considerations) section)?
> 
> What will the C# spec do in 11 then? Keep the feature "bimodal" forever, or assume that C# 11 _requires_ .NET 7?
> 
> Update: I've raised the same question, but a more time-pressing version of it, at #1532

See my reply [above](https://github.com/dotnet/csharpstandard/pull/1457#discussion_r2734136562), I think this might be a question not of “fidelity” but of implementation-specific material sneaking into this PR.

We are certainly **_not_** going to be saying in the Standard that C#11 requires .NET 7 as C# does not require any version of .NET at all.

standard/arrays.md
BillWagner: Completeness for one. And, to make it clear that in those cases, no conversion is required.

standard/conversions.md
BillWagner: This is one case where I think we want to leave it. While true that every `int` in the .NET runtime fits, the reason for `nint` is to have a *machine native integer*. An embedded processor may not be a 32 bit (or larger) processor.

standard/conversions.md
BillWagner: Joseph is correct.

It turns out that this requirement was missing in general for conversions. For example, the following works

```csharp
object[] arr = new object[] { m, m2 };
dynamic[] darr = arr;
```

I added text for an identity conversion between array types.

standard/expressions.md
BillWagner: No, that must have snuck in in another PR.

standard/expressions.md
BillWagner: Leaving for discussion. I don't remember why this was initially added for the other numeric types.

standard/expressions.md
BillWagner: This is saying two different things, I think: First that any value on the RHS of a constant expression must be in valid range. The second is saying that the compiler must evaluate the full expression at compile time and store the value. If the result of that compile time evaluation is out of range, the compiler emits an error.

standard/expressions.md
BillWagner: Added in the next commit. (Also changed note to an example)

standard/types.md
BillWagner: I removed the list. I think it's valid just to say "everything else"

standard/types.md
BillWagner: That implementation change likely has impact on the standard.

Until there was runtime support for `checked` and `unchecked` user defined operators, `System.IntPtr` and `System.UIntPtr` couldn't support the correct behavior for numeric types. We need to discuss how we want to handle that requirement.

standard/types.md
```diff
@@ -313,16 +319,65 @@ Because a simple type aliases a struct type, every simple type has members.
 >
 > *end note*.
 
+Although `nint` and `nuint` shall be represented by the types `System.IntPtr` and `System.UIntPtr`, respectively, `nint` and `nuint` are *not* aliases for those types. As such, not all members of the corresponding `System` types are defined for `nint` and `nuint`. Instead, the compiler shall make available additional conversions, unary operators (§12.9) and arithmetic operators (§12.12) for the types `System.IntPtr` and `System.UIntPtr` when used in the context of native integer types.
+
+While the language provides operations and conversions for `nint` and `nuint` that are appropriate for integer types, those operations and conversions are not available on the `System` type counterparts. For example,
```
Nigel-Ecma: These few lines give rise to a few questions:

> Although `nint` and `nuint` shall be represented by the types `System.IntPtr` and `System.UIntPtr`, respectively, `nint` and `nuint` are *not* aliases for those types. As such, not all members of the corresponding `System` types are defined for `nint` and `nuint`.

This seems to say there are members defined on `IntPtr` & `UIntPtr` that are explicitly not accessible via objects/values of type `nint` & `nuint`.

So is the deviation of `nint` & `nuint` from the other numeric types the restriction that not all members of the `System.X` type used for them may be accessed whether those members?

Remember under the conformance rules an implementation may provide types with additional members which are then accessible, are these native types avoiding that by not being “aliases“?

> Instead, the compiler shall make available additional conversions, unary operators (§12.9) and arithmetic operators (§12.12) for the types `System.IntPtr` and `System.UIntPtr` when used in the context of native integer types.

First, and obviously, the “compiler” shall do not such thing – the *language* defines the operators, e.g. §12.10.5 (v7):

> The predefined addition operators are listed below. For numeric and enumeration types, the predefined addition operators compute the sum of the two operands.

Where in the Standard does it ever state that these predefined operators are members of the “aliased” `System.X` type?

What difference is being asserted here that the “compiler” is making available that the language doesn’t for the other numeric types?

From a language perspective there are very few required members of any of numeric types. Take a look at `System.Int` and in §C no members are listed, look at the adopted spec from the CLI and it contains: 2 fields (min & max values), 2 compare methods, 2 equality methods, 1 hash code method, 4 parsing methods, and 4 conversions to string. What isn’t there is more notable than what is, e.g. not a single operator.

> While the language provides operations and conversions for `nint` and `nuint` that are appropriate for integer types, those operations and conversions are not available on the `System` type counterparts.

So how does that differ from `System.Int` above, it also has no operators…

Now today .NET’s `System.Int` does, for example, implement `IAdditionOperators<TSelf,TOther,TResult>` which provides and addition operator but it is not in the C# Standard and use Roslyn to add to integers and it does not call it – nothing has changed, the language defines the operators, the implementation implements them using the features of its target architecture (for Roslyn compiling C#9 the CLR’s `add` for both `int` and `nint`).

So what is *different* about `nint` and `nuint` that needs to go into the Standard (v9 & v11)? At present this PR has no native int operators listed in §12.10.5 (v7) which seems rather key, but does have stuff about the compiler making additional things available which seems entirely irrelevant.

Looking at the source material linked at the top of the PR it is clear it is heavily implementation specific, going as far as listing CLR instructions.

All the implementation specific stuff picked up from the source material needs to be elided, and the core of the C# languages changes found in what is left – and once that is done I suspect the differences between C#9 & C#11 will be much reduced.

Nigel-Ecma (changes requested):
A lot of changes seem to be required, but what they are will require some work. See the one comment under this review.

standard/conversions.md
```diff
@@ -83,6 +83,7 @@ An identity conversion converts from any type to the same type or a type that is
 - Between `object` and `dynamic`.
 - Between all tuple types with the same arity, and the corresponding constructed `ValueTuple<...>` type, when an identity conversion exists between each pair of corresponding element types.
 - Between types constructed from the same generic type where there exists an identity conversion between each corresponding type argument.
+- Between a array types containing elements of type `T` and `S`, such as `T[]` and `S[]`, where the rank of the two arrays is the same and there is an identity conversion between `T` and `S`.
```
KalleOlaviNiemitalo: Extra "a".

```suggestion
- Between array types containing elements of type `T` and `S`, such as `T[]` and `S[]`, where the rank of the two arrays is the same and there is an identity conversion between `T` and `S`.
```

(CLR also has array types with known bounds, but C# doesn't, so no need to add a statement on whether the bounds need to match.)

standard/conversions.md
Nigel-Ecma: I think @jnm2 is correct here, there seems to be an implicit statement in the source material that the size of `nint` is at least 32 bits. For example in the section on constants and based them of `int`. It might be nice if `nint` was 8-bits on a 8-bit microcontroller, but is seems not.

standard/expressions.md
```diff
@@ -2591,6 +2593,8 @@ The run-time processing of an *object_creation_expression* of the form new `T(A)
   - An instance of type `T` is created by allocating a temporary local variable. Since an instance constructor of a *struct_type* is required to definitely assign a value to each field of the instance being created, no initialization of the temporary variable is necessary.
   - The instance constructor is invoked according to the rules of function member invocation ([§12.6.6](expressions.md#1266-function-member-invocation)). A reference to the newly allocated instance is automatically passed to the instance constructor and the instance can be accessed from within that constructor as this.
 
+`new nint()` is equivalent to `(nint)0`, and `new nuint()` is equivalent to `(nuint)0`.
+
```
Nigel-Ecma: The same statement is not made for the other type (e.g. `new int()` is equivalent to `(int)0`). There is zero difference here between the `nint` & `nuint` and the other types adding this statement suggests there is so it, so it should go.

```suggestion
```

standard/expressions.md
```diff
@@ -4035,7 +4041,7 @@ The predefined division operators are listed below. The operators all compute th
 
   The division rounds the result towards zero. Thus the absolute value of the result is the largest possible integer that is less than or equal to the absolute value of the quotient of the two operands. The result is zero or positive when the two operands have the same sign and zero or negative when the two operands have opposite signs.
```
Nigel-Ecma, line 4028: I’ll comment here as it is the closest I’ve got to any of the operators, and I’ll add a no-change change request so it (maybe) gets marked as a change.

The `nint` & `nuint` operators all seem to be MIA. There are in the source material for C#9, and appear to be in VS 2026 with language target C#9 on .NET 6 (pre 7 as that I think is when the underlying types were changed).

For division here the operators would be:

```csharp
nint operator /(nint left, nint right);
nuint operator /(nuint left, nuint right);
```

and the same for most (if not all) the unary & binary operators the other numeric types support. There really does seem to be very little difference between the natural ints and the other numeric types.

```suggestion
  The division rounds the result towards zero. Thus the absolute value of the result is the largest possible integer that is less than or equal to the absolute value of the quotient of the two operands. The result is zero or positive when the two operands have the same sign and zero or negative when the two operands have opposite signs.
```

standard/types.md
```diff
@@ -289,7 +295,7 @@ C# provides a set of predefined `struct` types called the simple types. The simp
   `bool`    |   `System.Boolean`
   `decimal` |   `System.Decimal`
 
-Because a simple type aliases a struct type, every simple type has members.
+Every simple type has members. Each simple type that is an alias for a predefined struct type, has that struct type’s members.
 
 > *Example*: `int` has the members declared in `System.Int32` and the members inherited from `System.Object`, and the following statements are permitted:
```
Nigel-Ecma: This is possibly misleading, the Standard requires no members on `System.Int32`, but does require a few on `System.Object`. Maybe it should be something like:

```suggestion
Every simple type has members. Each simple type that is an alias for a predefined struct type, has that struct type’s members.

> *Example*: `int` has any implementation-specific members declared in `System.Int32` and the required and implementation-specific members inherited from `System.Object`, and the following statements are permitted:
```

But is something like this *too* accurate?

standard/types.md
```diff
@@ -313,16 +319,65 @@ Because a simple type aliases a struct type, every simple type has members.
 >
 > *end note*.
 
+Although `nint` and `nuint` shall be represented by the types `System.IntPtr` and `System.UIntPtr`, respectively, `nint` and `nuint` are *not* aliases for those types. As such, not all members of the corresponding `System` types are defined for `nint` and `nuint`. Instead, the compiler shall make available additional conversions, unary operators (§12.9) and arithmetic operators (§12.12) for the types `System.IntPtr` and `System.UIntPtr` when used in the context of native integer types.
```
Nigel-Ecma: As covered in this [comment](https://github.com/dotnet/csharpstandard/pull/1457/files#r2743902094) this seems to be largely incorrect from a Standard viewpoint. The only methods hat are required by the Standard are `Object`’s `Equals`, `GetHashCode`, `GetType` and `ToString`.

So this might be something like:

```suggestion
Although `nint` and `nuint` shall be represented by the types `System.IntPtr` and `System.UIntPtr`, respectively, `nint` and `nuint` are *not* aliases for those types. As such, the only members directly accessible for `nint` and `nuint` are the required methods of `Object` (§insert).
```

That, at minimum, needs wordsmithing!

standard/types.md
```diff
@@ -313,16 +319,65 @@ Because a simple type aliases a struct type, every simple type has members.
...
+M((nint)1);
+
+static void M(dynamic d)
+{
+    var v = d >> 2; // RuntimeBinderException: '>>' cannot be applied to operands
+                    // of type System.IntPtr/System.UIntPtr and int
+}
+```
+
+The only constructor for `nint` or `nuint` is the parameter-less constructor.
```
Nigel-Ecma: Does the Standard ever mention the constructor for `int` and the other numeric types? I suspect this in an implementation-specific detail.

```suggestion
```

standard/types.md
```diff
@@ -313,16 +319,65 @@ Because a simple type aliases a struct type, every simple type has members.
...
+public static IntPtr Add(IntPtr pointer, int offset);
+public static IntPtr Subtract(IntPtr pointer, int offset);
+public int ToInt32();
+public long ToInt64();
+public void* ToPointer();
+```
+
+The remaining members of `System.IntPtr` and `System.UIntPtr` are implicitly included in `nint` and `nuint`.
+
+Interfaces implemented by `System.IntPtr` and `System.UIntPtr` are implicitly included in `nint` and `nuint`, with occurrences of the underlying types replaced by the corresponding native integer types. For example, if `IntPtr` implements `ISerializable`, `IEquatable<IntPtr>`, and `IComparable<IntPtr>`, then `nint` implements `ISerializable`, `IEquatable<nint>`, and `IComparable<nint>`.
```
Nigel-Ecma: See this [comment](https://github.com/dotnet/csharpstandard/pull/1457/files#r2752405643) above. All the mentioned methods are implementation-specific.

Line 356 maybe could be finessed to say that *if* the underlying types have implementation-specific interface implementations then they are mapped to the non-alias contextual keywords… But that would only be needed until C#11 so maybe its a point not worth making?

So based on the referenced comment and the above para about line 356 this may be deleted (no change request as it has an embedded code block and I've no idea how to make a change suggestion including one of those!).

standard/types.md
BillWagner: With the reworking of this section, do you want this added, or does the new version address your concerns here?

standard/types.md
BillWagner: This was removed.

standard/types.md
BillWagner: I made some slight word choice changes here.

standard/conversions.md
Nigel-Ecma: Line 121 is redundant (and a little confusing) as it is all covered in lines 81-86 above. In can just go.

```suggestion
```

standard/conversions.md
Nigel-Ecma: I will update my last comment: @jnm2 *is* correct here. The native types are by definition either 32-bit or 64-bit (see types.md lines 354 & 355). So @jnm2’s change suggestion above should be followed.

standard/conversions.md
```diff
@@ -315,9 +322,13 @@ This implicit conversion seemingly violates the advice in the beginning of [§10
 
 An implicit constant expression conversion permits the following conversions:
 
-- A *constant_expression* ([§12.25](expressions.md#1225-constant-expressions)) of type `int` can be converted to type `sbyte`, `byte`, `short`, `ushort`, `uint`, or `ulong`, provided the value of the *constant_expression* is within the range of the destination type.
+- A *constant_expression* ([§12.25](expressions.md#1225-constant-expressions)) of type `int` can be converted to type `sbyte`, `byte`, `short`, `ushort`, `uint`, `nint`, `nuint`, or `ulong`, provided the value of the *constant_expression* is within the range of the destination type.
 - A *constant_expression* of type `long` can be converted to type `ulong`, provided the value of the *constant_expression* is not negative.
 
+The range for constants of type `nint` is the same range as `int`, and the range for constants of type `nuint` is the same range as `uint` ([§12.25](expressions.md#1225-constant-expressions)).
+
+> *Note*: This is a consequence of most common implementations being at least 32-bit ([§8.3.6](types.md#836-integral-types)). *end note*
```
Nigel-Ecma: It’s not about the most common implementations, the types are defined to be `int`/`uint` or `long`/`ulong` and no other options. Possible rewording:

```suggestion
> *Note*: This is a consequence of `nint`/`nuint` being the same size as, or larger than, `int`/`uint` ([§8.3.6](types.md#836-integral-types)). *end note*
```

standard/conversions.md
Nigel-Ecma: I agree in principle with @jnm2 above but the wording of his suggestion does not read well this non-american as it is missing small words 😉. If there is consensus that 749-751 are not clear then I’ll offer instead:

```CSharp
- A matching implicit or explicit conversion from `S?` to `T?`
- A matching implicit or explicit conversion from `S` to `T?`
- An explicit conversion from `S?` to `T`.
```

Either way the added lines are redundant:

```suggestion
```

standard/enums.md
```diff
@@ -44,7 +44,7 @@ enum_body
...
+Each enum type has a corresponding integral type called the ***underlying type*** of the enum type. This underlying type shall be able to represent all the enumerator values defined in the enumeration. If the *enum_base* is present, it explicitly declares the underlying type. The underlying type shall be one of the *integral types* ([§8.3.6](types.md#836-integral-types)) other than `nint`, `nuint`, and `char`. The underlying type may be specified either by an `integral_type` ([§8.3.5](types.md#835-simple-types)), or an `integral_type_name`. The `integral_type_name` is resolved in the same way as `type_name` ([§7.8.1](basic-concepts.md#781-general)), including taking any using directives ([§14.5](namespaces.md#145-using-directives)) into account.
 
 > *Note*: The `char` type cannot be used as an underlying type, either by keyword or via an `integral_type_name`. *end note*
```
Nigel-Ecma: Should this *Note* also now cover `nint` & `nuint`? I think the note borderline as it the text:

> The underlying type shall be one of the *integral types* ([§8.3.6](types.md#836-integral-types)) other than `nint`, `nuint`, and `char`. The underlying type may be specified either by an `integral_type` ([§8.3.5](types.md#835-simple-types)), or an `integral_type_name`.

seems fairly clear, so dropping rather than updating would also be fine.

standard/expressions.md
```diff
@@ -297,7 +297,7 @@ When overload resolution rules ([§12.6.4](expressions.md#1264-overload-resoluti
 
 **This subclause is informative.**
 
-Unary numeric promotion occurs for the operands of the predefined `+`, `-`, and `~` unary operators. Unary numeric promotion simply consists of converting operands of type `sbyte`, `byte`, `short`, `ushort`, or `char` to type `int`. Additionally, for the unary – operator, unary numeric promotion converts operands of type `uint` to type `long`.
+Unary numeric promotion occurs for the operands of the predefined `+`, `-`, and `~` unary operators. Unary numeric promotion simply consists of converting operands of type `sbyte`, `byte`, `short`, `ushort`, or `char` to type `int`. Additionally, for the unary - operator, unary numeric promotion converts operands of type `uint` or `nint` to type `long`.
```
Nigel-Ecma: Missing back ticks:

```suggestion
Unary numeric promotion occurs for the operands of the predefined `+`, `-`, and `~` unary operators. Unary numeric promotion simply consists of converting operands of type `sbyte`, `byte`, `short`, `ushort`, or `char` to type `int`. Additionally, for the unary `-` operator, unary numeric promotion converts operands of type `uint` or `nint` to type `long`.
```

standard/expressions.md
Nigel-Ecma: @jnm2: I think it can probably be simplified along the lines you suggest and avoid the lists. Maybe use letters as typical, e.g. let the two operands have types `S` & `T`. If `S` and `T` are the same type no promotion is required. If `S` implicitly converts to `T` then the first arg is promoted, if `T` to `S` then the second. Does the case if `S` & `T` are both implicitly convertible to `R` cover the case on line 316?

I’ll leave the wordsmithing to you (or others) 🙂

standard/expressions.md
```diff
@@ -7023,11 +7049,13 @@ constant_expression
...
+A constant expression may be either a value type or a reference type. If a constant expression has a value type, that type shall be one of the following: `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `nint`, `nuint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool,` or any enumeration type. If a constant expression has a reference type, the expression shall:
 
-- `sbyte`, `byte`, `short`, `ushort`, `int`, `uint`, `long`, `ulong`, `char`, `float`, `double`, `decimal`, `bool`, `string`;
-- an enumeration type; or
-- a default value expression ([§12.8.21](expressions.md#12821-default-value-expressions)) for a reference type.
+- have a value of type `string`;
+- have a value of `null`; or
+- be a default value expression ([§12.8.21](expressions.md#12821-default-value-expressions)) of reference type.
+
+A *constant_expression* of type `nint` shall have a value in the range \[`int.MinValue`,`int.MaxValue`\]. A *constant_expression* of type `nuint` shall have a value in the range \[`uint.MinValue`,`uint.MaxValue`\].
```
Nigel-Ecma: `MinValue` & `MaxValue` are not required by C# (not listed in §C), though they are used in 4 informative examples. Having avoided them so far we’d need a good reason to add them now.

Note: my MD viewer does not require backslash-bracket for a bracket, if GitHub does not as well then the backslashes are typos – I’ve left them as is in the suggestion.

Note: the number values are written out longhand in §8.3.6 but in §12.9.3 they are written using powers of 2. I’ve followed the former here but I’d be quite happy with the latter.

```suggestion
A *constant_expression* of type `nint` shall have a value in the range \[-2147483648, 2147483647\]. A *constant_expression* of type `nuint` shall have a value in the range \[0, 4294967295\].
```

standard/types.md
```diff
@@ -313,16 +319,41 @@ Because a simple type aliases a struct type, every simple type has members.
...
+
+<!-- C# 11: The following three paragraphs about overriding/hiding equivalence can be removed
+     when nint/nuint become true aliases for IntPtr/UIntPtr. -->
+
+`nint` and `System.IntPtr`, and `nuint` and `System.UIntPtr`, are considered equivalent for overriding, hiding, and implementing.
+
+Overloads cannot differ by `nint` and `System.IntPtr`, or by `nuint` and `System.UIntPtr`, alone. However, overrides and implementations may differ by `nint` and `System.IntPtr`, or by `nuint` and `System.UIntPtr`, alone.
+
+Methods hide other methods that differ by `nint` and `System.IntPtr`, or by `nuint` and `System.UIntPtr`, alone.
+
```
Nigel-Ecma: The difference between being *represented by* and being *an alias of* turn out to be very minimal and a lot of this section is not required. By just defining these differences we get:

```suggestion
<!-- C# 11: In C# 11, nint and nuint become true aliases for System.IntPtr and System.UIntPtr. The following paragraphs describing the non-alias relationship should be updated or removed. -->

The types `nint` and `nuint` are represented by the types `System.IntPtr` and `System.UIntPtr`, respectively, and are *not* aliases for these types. In this context being *represented by* means:

- The only members directly accessible for `nint` and `nuint` are the required methods of `Object` ([§C.2](standard-library.md#c2-standard-library-types-defined-in-isoiec-23271)). Any other members of System.IntPtr` and `System.UIntPtr` may be accessed via those types.
- Operations performed through `dynamic` binding on `System.IntPtr` and `System.UIntPtr` values do not have access to the `nint` and `nuint` operators.
- In all other respects `nint` and `nuint` behave as if they are aliases of `System.IntPtr` and `System.UIntPtr`.

```

standard/types.md
```diff
@@ -313,16 +319,41 @@ Because a simple type aliases a struct type, every simple type has members.
...
 - The `short` type represents signed 16-bit integers with values from `-32768` to `32767`, inclusive.
 - The `ushort` type represents unsigned 16-bit integers with values from `0` to `65535`, inclusive.
 - The `int` type represents signed 32-bit integers with values from `-2147483648` to `2147483647`, inclusive.
 - The `uint` type represents unsigned 32-bit integers with values from `0` to `4294967295`, inclusive.
+- The `nint` type represents a ***native signed integer*** whose size and value range are implementation-defined, but which shall be either that of `int` or `long`.
+- The `nuint` type represents a ***native unsigned integer*** whose size and value range are implementation-defined, but which shall be either that of `uint` or `ulong`. The size of a native unsigned integer shall be the same as that of a native signed integer.
+  > *Note*: Unlike the other integral types `nint` and `nuint` do not have `const` fields called `MinValue` and `MaxValue`. As a practical consideration, implementations are expected to be at least 32-bit, so the value range of `nint` includes at least that of `int`, and the value range of `nuint` includes at least that of `uint`. *end note*
```
Nigel-Ecma: `MinValue` & `MaxValue` are not part of the Standard, implementations also require `nint` & `nuint` to be 32-bit or 64-bit. Just delete the *Note*.

```suggestion
 ```

Nigel-Ecma (changes requested):
This is hopefully getting close. It is pleasing to see that the differences between being *aliases of* & *represented by* are really minimal so the step to C#11 will be small.

standard/enums.md
BillWagner: removing the note in the next commit.

BillWagner:
Thanks!

PR #1458, opened 2025-11-07
[Version 9.0] Feature support for record classes

BillWagner:
All commits from #983 have been squashed and added to this PR.

This PR / feature makes numerous changes to the grammar. We still need corresponding updates to the test and validation suite.

The following notes are carried over for additional work needed:

A. I put the new subclause "With expression" prior to "Arithmetic operators", which once the V8 features "Indices and Ranges" and "Pattern matching" have been merged, should immediately follow "Range operator" and "Switch expression." Make sure these are all in the correct place. 12.4.2 Operator precedence and associativity will also need to be revised accordingly.

B. New subclause §rec-class-prtmem Printing members mentions a method `System.Runtime.CompilerServices.RuntimeHelpers.EnsureSufficientExecutionStack()`. It's quite possible that this is an implementation-specific detail, in which case, we'll need to find words to make it abstract, and possibly update the Portability annex accordingly.


standard/expressions.md
```diff
@@ -3840,6 +3840,49 @@ At run-time, the expression `await t` is evaluated as follows:
...
+## §with-expressions With expressions
+
+A *with_expression* allows for ***non-destructive mutation*** by making a new record class instance that is a copy of an existing record class instance, optionally with specified properties and fields modified.
+
+```ANTLR
+with_expression
+    : switch_expression
+    | switch_expression 'with' '{' member_initializer_list? '}'
+    ;
+
```
RexJaeschke:
```suggestion
```

standard/expressions.md
```diff
@@ -3840,6 +3840,49 @@ At run-time, the expression `await t` is evaluated as follows:
 
 An awaiter’s implementation of the interface methods `INotifyCompletion.OnCompleted` and `ICriticalNotifyCompletion.UnsafeOnCompleted` should cause the delegate `r` to be invoked at most once. Otherwise, the behavior of the enclosing async function is undefined.
 
+## §with-expressions With expressions
```
RexJaeschke, line 3825: We need to add "with" as a new row to the table in the (earlier) section "Operator precedence and associativity." And as we have historically organized the sections in this clause in descending precedence order, we might need to move this new section up or down, accordingly.

standard/classes.md
```diff
@@ -14,18 +14,35 @@ A *class_declaration* is a *type_declaration* ([§14.7](namespaces.md#147-type-d
...
-        type_parameter_list? class_base? type_parameter_constraints_clause*
-        class_body ';'?
+    : attributes? class_modifier* 'partial'? class_tag identifier
+        type_parameter_list? parameter_list? class_base?
+        type_parameter_constraints_clause* class_body
+    ;
+
+class_tag
+    : 'class'
+    | 'record'
```
RexJaeschke: FYI: In V9, `record` means `record class` but we can't say that explicitly. In V10, this is permitted and we'll replace these 2  rules with `'record'? 'class'`.

standard/classes.md
```diff
@@ -14,18 +14,35 @@ A *class_declaration* is a *type_declaration* ([§14.7](namespaces.md#147-type-d
 
 ```ANTLR
```
RexJaeschke, line 15: **Heads up grammarians!** As you will see below, I have a generous syntax for a class in which not all combinations are permitted, and those combinations are described in text as constraints. (For example, a record class can have a *delimited_parameter_list* while a non-record class cannot. And for a record class, *class_modifier* shall not be `static`.) Yes, we could try to break this up into separate rules for record classes and non-record classes to eliminate some (but maybe not all) of those constraints. However, there will be tweaks to this grammar in future versions; specifically:

* V10 allows `record class` as well as `record`.
* V10 supports `record struct`.
* V12 allows a primary constructor (which uses *delimited_parameter_list*) on non-record classes and structs.

which means that extra work done now to make the grammar rules more complete likely will be (at least partially) undone in future revs.

standard/classes.md
```diff
@@ -748,14 +773,19 @@ The *class_body* of a class defines the members of that class.
...
 class_body
-    : '{' class_member_declaration* '}'
+    : '{' class_member_declaration* '}' ';'?
+    | ';'
     ;
 ```
 
+For a record class, the *class_body*s `{}`, `{};`, and `;` are equivalent. They all indicate that the only members are those synthesized by the compiler (§synth-members).
```
RexJaeschke:
```suggestion
A non-record class shall not have a *class_body* of `;`.

For a record class, the *class_body*s `{}`, `{};`, and `;` are equivalent. They all indicate that the only members are those synthesized by the compiler (§synth-members).
```

standard/classes.md
```diff
@@ -748,14 +773,19 @@ The *class_body* of a class defines the members of that class.
...
+
 ### 15.2.7 Partial type declarations
 
 The modifier `partial` is used when defining a class, struct, or interface type in multiple parts. The `partial` modifier is a contextual keyword ([§6.4.4](lexical-structure.md#644-keywords)) and has special meaning immediately before the keywords `class`, `struct`, and `interface`. (A partial type may contain partial method declarations ([§15.6.9](classes.md#1569-partial-methods)).
 
+A non-record class shall not have a *class_body* of `;`.
+
```
RexJaeschke:
```suggestion
```

RexJaeschke: This is in the wrong section, and had been moved to the class body section.

standard/classes.md
```diff
@@ -748,14 +773,19 @@ The *class_body* of a class defines the members of that class.
...
+    : '{' class_member_declaration* '}' ';'?
+    | ';'
     ;
 ```
 
+For a record class, the *class_body*s `{}`, `{};`, and `;` are equivalent. They all indicate that the only members are those synthesized by the compiler (§synth-members).
+
 ### 15.2.7 Partial type declarations
 
 The modifier `partial` is used when defining a class, struct, or interface type in multiple parts. The `partial` modifier is a contextual keyword ([§6.4.4](lexical-structure.md#644-keywords)) and has special meaning immediately before the keywords `class`, `struct`, and `interface`. (A partial type may contain partial method declarations ([§15.6.9](classes.md#1569-partial-methods)).
```
RexJaeschke:
```suggestion
The modifier `partial` is used when defining a class, struct, or interface type in multiple parts. The `partial` modifier is a contextual keyword ([§6.4.4](lexical-structure.md#644-keywords)) and has special meaning in a *class_declaration*, a *struct_declaration*, or an *interface_declaration*. (A partial type may contain partial method declarations ([§15.6.9](classes.md#1569-partial-methods)).
```

standard/classes.md
```diff
@@ -5642,3 +5715,471 @@ An enumerable object may implement more interfaces than those specified above.
...
+```csharp
+public virtual bool Equals(R? other);
+```
+
+The synthesized `Equals(R?)` returns `true` if and only if each of the following are `true`:
+
+- `other` is not `null`, and
+- For each instance field `fieldN` in the record class type that is not inherited, the value of `System.Collections.Generic.EqualityComparer<TN>.Default.Equals(fieldN, other.fieldN)` where `TN` is the field type, and
+- If there is a base record class type, the value of `base.Equals(other)` (a non-virtual call to `public virtual bool Equals(Base? other)`); otherwise
+the value of `EqualityContract == other.EqualityContract`.
```
RexJaeschke:
```suggestion
- If there is a base record class type, the value of `base.Equals(other)` (a non-virtual call to `public virtual bool Equals(Base? other)`); otherwise the value of `EqualityContract == other.EqualityContract`.
```

standard/classes.md
```diff
@@ -5642,3 +5715,471 @@ An enumerable object may implement more interfaces than those specified above.
...
+
+The method may be declared explicitly. It is an error if the explicit declaration doesn't allow overriding it in a derived type and the record class type is not `sealed`. It is an error if either the synthesized, or the explicitly declared, method doesn't override `object.GetHashCode()` (for example, due to shadowing in intermediate base types).
+
+A warning shall be issued if one of `Equals(R?)` and `GetHashCode()` is explicitly declared, but the other is not.
+
+The synthesized override of `GetHashCode()` returns an `int` result of combining the following values:
+
+- For each instance field `fieldN` in the record class type that is not inherited, the value of `System.Collections.Generic.EqualityComparer<TN>.Default.GetHashCode(fieldN)` where `TN` is the field type, and
+- If there is a base record class type, the value of `base.GetHashCode()`; otherwise
+the value of `System.Collections.Generic.EqualityComparer<System.Type>.Default.GetHashCode(EqualityContract)`.
```
RexJaeschke:
```suggestion
- If there is a base record class type, the value of `base.GetHashCode()`; otherwise the value of `System.Collections.Generic.EqualityComparer<System.Type>.Default.GetHashCode(EqualityContract)`.
```

standard/classes.md
```diff
@@ -5642,3 +5715,471 @@ An enumerable object may implement more interfaces than those specified above.
...
+bool PrintMembers(System.Text.StringBuilder builder);
+```
+
+The method is `private` if the record class type is sealed. Otherwise, the method is virtual and protected.
+
+The ***printable members of a class*** are the instance public field and readable property members.
+
+The method performs the following tasks:
+
+1. Calls the method `System.Runtime.CompilerServices.RuntimeHelpers.EnsureSufficientExecutionStack()` if that method is present and the record class has printable members.
```
RexJaeschke, line 5960: Do we really want to expose this type in the spec?

standard/classes.md
```diff
@@ -2200,6 +2234,10 @@ The parameters of a method, if any, are declared by the method’s *parameter_li
 
 ```ANTLR
 parameter_list
```
RexJaeschke:
```suggestion
delimited_parameter_list
```

standard/classes.md
```diff
@@ -2200,6 +2234,10 @@ The parameters of a method, if any, are declared by the method’s *parameter_li
...
+    : '(' formal_parameter_list? ')'
+    ;
+
+formal_parameter_list
```
RexJaeschke:
```suggestion
parameter_list
```

standard/classes.md
```diff
@@ -2200,6 +2234,10 @@ The parameters of a method, if any, are declared by the method’s *parameter_li
 
 ```ANTLR
 parameter_list
+    : '(' formal_parameter_list? ')'
```
RexJaeschke:
```suggestion
    : '(' parameter_list? ')'
```

standard/classes.md
```diff
@@ -14,18 +14,35 @@ A *class_declaration* is a *type_declaration* ([§14.7](namespaces.md#147-type-d
 
 ```ANTLR
 class_declaration
-    : attributes? class_modifier* 'partial'? 'class' identifier
-        type_parameter_list? class_base? type_parameter_constraints_clause*
-        class_body ';'?
+    : attributes? class_modifier* 'partial'? class_tag identifier
+        type_parameter_list? parameter_list? class_base?
```
RexJaeschke:
```suggestion
        type_parameter_list? delimited_parameter_list? class_base?
```

standard/classes.md
```diff
@@ -14,18 +14,35 @@ A *class_declaration* is a *type_declaration* ([§14.7](namespaces.md#147-type-d
...
+A *class_declaration* consists of an optional set of *attributes* ([§23](attributes.md#23-attributes)), followed by an optional set of *class_modifier*s ([§15.2.2](classes.md#1522-class-modifiers)), followed by an optional `partial` modifier ([§15.2.7](classes.md#1527-partial-type-declarations)), followed by a *class_tag* and an *identifier* that names the class, followed by an optional *type_parameter_list* ([§15.2.3](classes.md#1523-type-parameters)), followed by an optional *parameter_list* ([§15.6.2.1]( classes.md#15621-general)), followed by an optional *class_base* specification ([§15.2.4](classes.md#1524-class-base-specification)), followed by an optional set of *type_parameter_constraints_clause*s ([§15.2.5](classes.md#1525-type-parameter-constraints)), followed by a *class_body* ([§15.2.6](classes.md#1526-class-body)), optionally followed by a semicolon.
```
RexJaeschke:
```suggestion
A *class_declaration* consists of an optional set of *attributes* ([§23](attributes.md#23-attributes)), followed by an optional set of *class_modifier*s ([§15.2.2](classes.md#1522-class-modifiers)), followed by an optional `partial` modifier ([§15.2.7](classes.md#1527-partial-type-declarations)), followed by a *class_tag* and an *identifier* that names the class, followed by an optional *type_parameter_list* ([§15.2.3](classes.md#1523-type-parameters)), followed by an optional *delimited_parameter_list* ([§15.6.2.1]( classes.md#15621-general)), followed by an optional *class_base* specification ([§15.2.4](classes.md#1524-class-base-specification)), followed by an optional set of *type_parameter_constraints_clause*s ([§15.2.5](classes.md#1525-type-parameter-constraints)), followed by a *class_body* ([§15.2.6](classes.md#1526-class-body)), optionally followed by a semicolon.
```

standard/classes.md
```diff
@@ -14,18 +14,35 @@ A *class_declaration* is a *type_declaration* ([§14.7](namespaces.md#147-type-d
...
+
+For a record class, *class_modifier* shall not be `static`.
+
+*parameter_list* shall not be present in a non-record class.
+
+A *class_declaration* having a *parameter_list* declares a ***positional record class***.
+
+At most only one partial type declaration of a partial record class may provide a *parameter_list*.
+
+Parameters in *parameter_list* shall not have `ref`, `out` or `this` modifiers; however, `in` and `params` modifiers are permitted.
```
RexJaeschke:
```suggestion
*delimited_parameter_list* shall not be present in a non-record class.

A *class_declaration* having a *delimited_parameter_list* declares a ***positional record class***.

At most only one partial type declaration of a partial record class may provide a *delimited_parameter_list*.

Parameters in *delimited_parameter_list* shall not have `ref`, `out` or `this` modifiers; however, `in` and `params` modifiers are permitted.
```

standard/classes.md
```diff
@@ -172,16 +189,24 @@ A class declaration may include a *class_base* specification, which defines the
...
     ;
 
 interface_type_list
     : interface_type (',' interface_type)*
     ;
 ```
 
+A record class may not inherit from a non-record class other than `object`, and a non-record class may not inherit from a record class.
+
+It is an error for *class_base* to have a *base_argument_list* if the corresponding *class_declaration* does not contain a *parameter_list*.
```
RexJaeschke:
```suggestion
It is an error for *class_base* to have a *base_argument_list* if the corresponding *class_declaration* does not contain a *delimited_parameter_list*.
```

standard/classes.md
RexJaeschke: I recently finished the spec for V10’s record structs (see Draft PR #[1556](https://github.com/dotnet/csharpstandard/pull/1556)), in which I was able to quite easily split the top-level grammar into non-record struct and record struct rules *without* the need for textual constraints. Since then, I have completed a review of this V9 spec for record classes and I believe the same approach I used for structs can be used here. However, rather than making those changes to this PR, after discussing this privately with Nigel, I'm proposing them in this comment.
 
Here’s what I have for the revised grammar (which mirrors what I have done for V10’s record structs):
 
```ANTLR
class_declaration
    : non_record_class_declaration
    | record_class_declaration
    ;
 
non_record_class_declaration
    : attributes? class_modifier* 'partial'? 'class' identifier
        type_parameter_list? class_base? type_parameter_constraints_clause*
        class_body ';'?
    ;
 
record_class_declaration
    : attributes? class_modifier* 'partial'? 'record' identifier 
      type_parameter_list? delimited_parameter_list? class_base? 
      type_parameter_constraints_clause* record_class_body
    ;
 
class_body
    : '{' class_member_declaration* '}'
    ;
 
record_class_body
    : class_body ';'?
    | ';'
    ;
```

standard/classes.md
RexJaeschke: This change has been reversed.

standard/classes.md
RexJaeschke: I have changed the grammar along the lines I proposed earlier in this comment thread, so there are now separate rules for record and non-record class.

RexJaeschke:
Based on Issue #1683, I have updated the grammar for *class_declaration*.

Re my choice of grammar rule name prefixes, “non-record” and “record,” for a reader starting at 15.1, which says nothing about records, you might think we should *not* use the word “record” here. I’m OK with that. ***As a strawman, how about we call the two class flavors “plain class” and “record class?”*** (I think “class” and “record class” is *not* a clear enough distinction, as “class” might mean any class or just a non-record class.)

I currently use the two-favor distinction in new definitions, as follows:

> A *non_record_class_declaration* declares a ***non-record class***, while a *record_class_declaration* declares a ***record class***.

This might then become: 

> A *plain_class_declaration* declares a ***plain class***, while a *record_class_declaration* declares a ***record class***.

I’m thinking that all existing uses of “class” will apply to both flavors, so no distinction will be needed for them.

For now, I’ll use “non-record” and “record.”

BillWagner:
Thanks!

RexJaeschke:
After researching and writing-up "Primary constructors" as a general feature in v12 for classes and structs *outside* of record variants, I have made a lot of tweaks to this spec (and to the V10 record struct spec) to ease the transition to that v12 spec.

jskeet:

Nigel-Ecma (changes requested):
This is a review looking at a high-level, the details semantics of records has not been reviewed.

## Issue: The grammar changes break all the existing tests.

Often when adding *new* constructs to the language, records in this case, the existing tests all pass. In short this is down to the tests checking the shape of the tree, adding a new branch kind to a list of branch choices doesn’t change that (think adding a new statement kind to the list of statement kinds). However if where existing branches fit is changed then exisiting tests will fail. There is nothing wrong with this *per se*, but it should only be done if there is a good reason.

This proposal changes the location where classes fit into the overall grammar (by adding a branch node above them), there are a *lot* of classes in C# programs, and the result is all the tests fail. If there was a strong reason for this change then the expected parse for all tests could be updated (not hard, a little tedious).

However in this case the change is unnecessary:

- *class_declaration* & *class_body* can be restored to their previous form;
- *record_class_body* needs an optional ending semicolon, as it has in the source material;
- *type_declaration* needs *record_class_body* added; and
- *non_record_class_declaration* is consigned to the bin (yay)

```ANTLR
class_declaration
    : attributes? class_modifier* 'partial'? 'class' identifier
        type_parameter_list? class_base? type_parameter_constraints_clause*
        class_body ';'?
    ;

class_body
    : '{' class_member_declaration* '}'
    ;

record_class_body
    : class_body ';'?
    | ';'
    ;

type_declaration
    : class_declaration
    | record_class_declaration
    | struct_declaration
    | interface_declaration
    | enum_declaration
    | delegate_declaration
    ;
```

All tests pass, including a new beta one for records (not yet released).

These changes do not fix all the concerns with the grammar for records, in particular there is unneeded sharing of rules – reduces the number of grammar rules but adds prose rules to split the overlaid rules up again (e.g. the source material has *record_base* and *class_base*, this proposal overlays them and adds prose rules), that’s not a win. (Sometimes overlaying makes sense, e.g. using *class_body* in *record_body*, no added prose rules required).

However these issues can be addressed in later reviews.

## non-record class – no no no!

Sorry but this hurt my eyes and ears. Classes have been a central feature of C# of around a quarter of a century, changing them to be called non-record classes is a poor idea with no saving grace. Call them classes and remove any added *non_record_class* bits from rule names (note the grammar changes up consigned *non_record_class_declaration* to the bin).

I would also call records *record*, not *record class*, which wouldn’t prevent v10 adding *record struct*; it is just adding words for no real benefit.

The source material states:

> Record types are reference types, similar to a class declaration.

Not that a record declaration is a class declaration, and uses rules such as *record_declaration*, *record_base* & *record_body*, in v10 it adds *record_struct_X*.

## TL;DR

- Restore the grammar rules listed at the top – don’t break things unnecessarily. And it “fixes” all the tests 🙂.
- No “non-record class”, its a “class” with a quarter century of use.
- Recommend dropping *non_record_* prefix in grammar rule names (sometimes the source material gets it right 😉)

TTFN

PR #1459, opened 2025-11-10
[Version 9.0] Feature support for function pointers

BillWagner:
All commits from #984 have been added.

PR #984 has three conversations that haven't been addressed:

- https://github.com/dotnet/csharpstandard/pull/984#pullrequestreview-1713886670
- https://github.com/dotnet/csharpstandard/pull/984#issuecomment-1793687122
- https://github.com/dotnet/csharpstandard/pull/984#issuecomment-2913138473

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -819,7 +819,7 @@ An *output type inference* is made *from* an expression `E` *to* a type `T` in
 - If `E` is a tuple literal with arity `N` and elements `Eᵢ`, and `T` is a tuple type with arity `N` with corresponding element types `Tₑ` or `T` is a nullable value type `T0?` and `T0` is a tuple type with arity `N` that has a corresponding element type `Tₑ`, then for each `Eᵢ` an output type inference is made from `Eᵢ` to `Tₑ`.
 - If `E` is an anonymous function with inferred return type `U` ([§12.6.3.14](expressions.md#126314-inferred-return-type)) and `T` is a delegate type or expression tree type with return type `Tₓ`, then a *lower-bound inference* ([§12.6.3.11](expressions.md#126311-lower-bound-inferences)) is made *from* `U` *to* `Tₓ`.
 - Otherwise, if `E` is a method group and `T` is a delegate type or expression tree type with parameter types `T₁...Tᵥ` and return type `Tₓ`, and overload resolution of `E` with the types `T₁...Tᵥ` yields a single method with return type `U`, then a *lower-bound inference* is made *from* `U` *to* `Tₓ`.
-- If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then with parameter types `T1..Tk` and return type `Tb`, and overload resolution of `E` with the types `T1..Tk` yields a single method with return type `U`, then a *lower-bound inference* is made from `U` to `Tb`.
+- If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then with parameter types `T₁..Tₖ` and return type `Tₓ`, and overload resolution of `E` with the types `T₁..Tₖ` yields a single method with return type `U`, then a *lower-bound inference* is made from `U` to `Tₓ`.
   > *Note*: This is only applicable in unsafe code. *end note*
 - Otherwise, if `E` is an expression with type `U`, then a *lower-bound inference* is made *from* `U` *to* `T`.
 - Otherwise, no inferences are made.
@@ -853,18 +853,18 @@ A *lower-bound inference from* a type `U` *to* a type `V` is made as follows:
   - `V` is an array type `V₁[...]`and `U` is an array type `U₁[...]`of the same rank
   - `V` is one of `IEnumerable<V₁>`, `ICollection<V₁>`, `IReadOnlyList<V₁>>`, `IReadOnlyCollection<V₁>` or `IList<V₁>` and `U` is a single-dimensional array type `U₁[]`
   - `V` is a constructed `class`, `struct`, `interface` or `delegate` type `C<V₁...Vₑ>` and there is a unique type `C<U₁...Uₑ>` such that `U` (or, if `U` is a type `parameter`, its effective base class or any member of its effective interface set) is identical to, `inherits` from (directly or indirectly), or implements (directly or indirectly) `C<U₁...Uₑ>`.
-  - `V` is a function pointer type (§function-pointers) `delegate*<V2..Vk, V1>` and there is a function pointer type `delegate*<U2..Uk, U1>` such that `U` is identical to `delegate*<U2..Uk, U1>`, and the calling convention of `V` is identical to `U`, and the refness of `Vi` is identical to `Ui`.
+  - `V` is a function pointer type (§function-pointers) `delegate*<V₂..Vₖ, V₁>` and there is a function pointer type `delegate*<U₂..Uₖ, U₁>` such that `U` is identical to `delegate*<U₂..Uₖ, U₁>`, and the calling convention of `V` is identical to `U`, and the refness of `Vᵢ` is identical to `Uᵢ`.
     > *Note*: This is only applicable in unsafe code. *end note*
   - (The “uniqueness” restriction means that in the case interface `C<T>{} class U: C<X>, C<Y>{}`, then no inference is made when inferring from `U` to `C<T>` because `U₁` could be `X` or `Y`.)
   If any of these cases apply then an inference is made from each `Uᵢ` to the corresponding `Vᵢ` as follows:
-  - If `Uᵢ` is not known to be a reference type then an *exact inference* is made; or alternatively, if `U` is not a function pointer type and `Ui` is not known to be a reference type, or if `U` is a function pointer type and `Ui` is not known to be a function pointer type or a reference type, then an exact inference is made
+  - If `Uᵢ` is not known to be a reference type then an *exact inference* is made; or alternatively, if `U` is not a function pointer type and `Uᵢ` is not known to be a reference type, or if `U` is a function pointer type and `Uᵢ` is not known to be a function pointer type or a reference type, then an exact inference is made
     > *Note*: This is only applicable in unsafe code. *end note*
   - Otherwise, if `U` is an array type then a *lower-bound inference* is made
   - Otherwise, if `V` is `C<V₁...Vₑ>` then inference depends on the `i-th` type parameter of `C`:
     - If it is covariant then a *lower-bound inference* is made.
     - If it is contravariant then an *upper-bound inference* is made.
     - If it is invariant then an *exact inference* is made.
-  - Otherwise, if `V` is `delegate*<V2..Vk, V1>` then inference depends on the i-th parameter of `delegate*<V2..Vk, V1>`:
+  - Otherwise, if `V` is `delegate*<V₂..Vₖ, V₁>` then inference depends on the i-th parameter of `delegate*<V₂..Vₖ, V₁>`:
     - If V1:
       - If the return is by value, then a lower-bound inference is made.
       - If the return is by reference, then an exact inference is made.
@@ -884,18 +884,18 @@ An *upper-bound inference from* a type `U` *to* a type `V` is made as follows:
   - `U` is one of `IEnumerable<Uₑ>`, `ICollection<Uₑ>`, `IReadOnlyList<Uₑ>`, `IReadOnlyCollection<Uₑ>` or `IList<Uₑ>` and `V` is a single-dimensional array type `Vₑ[]`
   - `U` is the type `U1?` and `V` is the type `V1?`
   - `U` is constructed class, struct, interface or delegate type `C<U₁...Uₑ>` and `V` is a `class, struct, interface` or `delegate` type which is `identical` to, `inherits` from (directly or indirectly), or implements (directly or indirectly) a unique type `C<V₁...Vₑ>`
-  - `U` is a function pointer type (§function-pointers) then `delegate*<U2..Uk, U1>` and `V` is a function pointer type which is identical to `delegate*<V2..Vk, V1>`, and the calling convention of `U` is identical to `V`, and the refness of `Ui` is identical to `Vi`.
+  - `U` is a function pointer type (§function-pointers) then `delegate*<U₂..Uₖ, U₁>` and `V` is a function pointer type which is identical to `delegate*<V₂..Vₖ, V₁>`, and the calling convention of `U` is identical to `V`, and the refness of `Uᵢ` is identical to `Vᵢ`.
     > *Note*: This is only applicable in unsafe code. *end note*
   - (The “uniqueness” restriction means that given an interface `C<T>{} class V<Z>: C<X<Z>>, C<Y<Z>>{}`, then no inference is made when inferring from `C<U₁>` to `V<Q>`. Inferences are not made from `U₁` to either `X<Q>` or `Y<Q>`.)
   If any of these cases apply then an inference is made from each `Uᵢ` to the corresponding `Vᵢ` as follows:
-  - If `U` is not a function pointer type and `Ui` is not known to be a reference type, or if `U` is a function pointer type and `Ui` is not known to be a function pointer type or a reference type, then an *exact inference* is made
+  - If `U` is not a function pointer type and `Uᵢ` is not known to be a reference type, or if `U` is a function pointer type and `Uᵢ` is not known to be a function pointer type or a reference type, then an *exact inference* is made
     > *Note*: Function-pointer type-related text is only applicable in unsafe code. *end note*
   - Otherwise, if `V` is an array type then an *upper-bound inference* is made
   - Otherwise, if `U` is `C<U₁...Uₑ>` then inference depends on the `i-th` type parameter of `C`:
     - If it is covariant then an *upper-bound inference* is made.
     - If it is contravariant then a *lower-bound inference* is made.
     - If it is invariant then an *exact inference* is made.
-- Otherwise, if `U` is `delegate*<U2..Uk, U1>` then inference depends on the i-th parameter of `delegate*<U2..Uk, U1>`:
+- Otherwise, if `U` is `delegate*<U₂..Uₖ, U₁>` then inference depends on the i-th parameter of `delegate*<U₂..Uₖ, U₁>`:
   - If `U1`:
     - If the return is by value, then an upper-bound inference is made.
     - If the return is by reference, then an exact inference is made.
@@ -1152,7 +1152,7 @@ Given an implicit conversion `C₁` that converts from an expression `E` to a ty
 
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` ([§12.6.4.6](expressions.md#12646-exactly-matching-expression))
 - `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§12.6.4.7](expressions.md#12647-better-conversion-target))
-  - `V` is a function pointer type `delegate*<V2..Vk, V1>` and `U` is a function pointer type `delegate*<U2..Uk, U1>`, and the calling convention of `V` is identical to `U`, and the refness of `Vi` is identical to `Ui`.
+  - `V` is a function pointer type `delegate*<V₂..Vₖ, V₁>` and `U` is a function pointer type `delegate*<U₂..Uₖ, U₁>`, and the calling convention of `V` is identical to `U`, and the refness of `Vᵢ` is identical to `Uᵢ`.
     > *Note*: This is only applicable in unsafe code. *end note*
 - `E` is a method group ([§12.2](expressions.md#122-expression-classifications)), `T₁` is compatible ([§21.4](delegates.md#214-delegate-compatibility)) with the single best method from the method group for conversion `C₁`, and `T₂` is not compatible with the single best method from the method group for conversion `C₂`
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -782,9 +782,17 @@ The second phase proceeds as follows:
 
 If `E` is a method group or implicitly typed anonymous function and `T` is a delegate type or expression tree type then all the parameter types of `T` are *input types of* `E` *with type* `T`.
 
+If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then all the parameter types of `T` are input types of `E` with type `T`.
+
+> *Note*: This is only applicable in unsafe code. *end note*
+
 #### 12.6.3.5 Output types
 
 If `E` is a method group or an anonymous function and `T` is a delegate type or expression tree type then the return type of `T` is an *output type of* `E` *with type* `T`.
+
+If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then the return type of `T` is an output type of `E` with type `T`.
+
+> *Note*: This is only applicable in unsafe code. *end note*
 
 #### 12.6.3.6 Dependence
 
@@ -811,6 +819,8 @@ An *output type inference* is made *from* an expression `E` *to* a type `T` in
 - If `E` is a tuple literal with arity `N` and elements `Eᵢ`, and `T` is a tuple type with arity `N` with corresponding element types `Tₑ` or `T` is a nullable value type `T0?` and `T0` is a tuple type with arity `N` that has a corresponding element type `Tₑ`, then for each `Eᵢ` an output type inference is made from `Eᵢ` to `Tₑ`.
 - If `E` is an anonymous function with inferred return type `U` ([§12.6.3.14](expressions.md#126314-inferred-return-type)) and `T` is a delegate type or expression tree type with return type `Tₓ`, then a *lower-bound inference* ([§12.6.3.11](expressions.md#126311-lower-bound-inferences)) is made *from* `U` *to* `Tₓ`.
 - Otherwise, if `E` is a method group and `T` is a delegate type or expression tree type with parameter types `T₁...Tᵥ` and return type `Tₓ`, and overload resolution of `E` with the types `T₁...Tᵥ` yields a single method with return type `U`, then a *lower-bound inference* is made *from* `U` *to* `Tₓ`.
+- If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then with parameter types `T1..Tk` and return type `Tb`, and overload resolution of `E` with the types `T1..Tk` yields a single method with return type `U`, then a *lower-bound inference* is made from `U` to `Tb`.
+  > *Note*: This is only applicable in unsafe code. *end note*
 - Otherwise, if `E` is an expression with type `U`, then a *lower-bound inference* is made *from* `U` *to* `T`.
 - Otherwise, no inferences are made.
 
@@ -843,15 +853,25 @@ A *lower-bound inference from* a type `U` *to* a type `V` is made as follows:
   - `V` is an array type `V₁[...]`and `U` is an array type `U₁[...]`of the same rank
   - `V` is one of `IEnumerable<V₁>`, `ICollection<V₁>`, `IReadOnlyList<V₁>>`, `IReadOnlyCollection<V₁>` or `IList<V₁>` and `U` is a single-dimensional array type `U₁[]`
   - `V` is a constructed `class`, `struct`, `interface` or `delegate` type `C<V₁...Vₑ>` and there is a unique type `C<U₁...Uₑ>` such that `U` (or, if `U` is a type `parameter`, its effective base class or any member of its effective interface set) is identical to, `inherits` from (directly or indirectly), or implements (directly or indirectly) `C<U₁...Uₑ>`.
+  - `V` is a function pointer type (§function-pointers) `delegate*<V2..Vk, V1>` and there is a function pointer type `delegate*<U2..Uk, U1>` such that `U` is identical to `delegate*<U2..Uk, U1>`, and the calling convention of `V` is identical to `U`, and the refness of `Vi` is identical to `Ui`.
+    > *Note*: This is only applicable in unsafe code. *end note*
   - (The “uniqueness” restriction means that in the case interface `C<T>{} class U: C<X>, C<Y>{}`, then no inference is made when inferring from `U` to `C<T>` because `U₁` could be `X` or `Y`.)
-
   If any of these cases apply then an inference is made from each `Uᵢ` to the corresponding `Vᵢ` as follows:
-  - If `Uᵢ` is not known to be a reference type then an *exact inference* is made
+  - If `Uᵢ` is not known to be a reference type then an *exact inference* is made; or alternatively, if `U` is not a function pointer type and `Ui` is not known to be a reference type, or if `U` is a function pointer type and `Ui` is not known to be a function pointer type or a reference type, then an exact inference is made
+    > *Note*: This is only applicable in unsafe code. *end note*
   - Otherwise, if `U` is an array type then a *lower-bound inference* is made
   - Otherwise, if `V` is `C<V₁...Vₑ>` then inference depends on the `i-th` type parameter of `C`:
     - If it is covariant then a *lower-bound inference* is made.
     - If it is contravariant then an *upper-bound inference* is made.
     - If it is invariant then an *exact inference* is made.
+  - Otherwise, if `V` is `delegate*<V2..Vk, V1>` then inference depends on the i-th parameter of `delegate*<V2..Vk, V1>`:
+    - If V1:
+      - If the return is by value, then a lower-bound inference is made.
+      - If the return is by reference, then an exact inference is made.
+    - If V2..Vk:
+      - If the parameter is by value, then an upper-bound inference is made.
+      - If the parameter is by reference, then an exact inference is made.
+    > *Note*: This is only applicable in unsafe code. *end note*
 - Otherwise, no inferences are made.
 
 #### 12.6.3.12 Upper-bound inferences
@@ -864,15 +884,25 @@ An *upper-bound inference from* a type `U` *to* a type `V` is made as follows:
   - `U` is one of `IEnumerable<Uₑ>`, `ICollection<Uₑ>`, `IReadOnlyList<Uₑ>`, `IReadOnlyCollection<Uₑ>` or `IList<Uₑ>` and `V` is a single-dimensional array type `Vₑ[]`
   - `U` is the type `U1?` and `V` is the type `V1?`
   - `U` is constructed class, struct, interface or delegate type `C<U₁...Uₑ>` and `V` is a `class, struct, interface` or `delegate` type which is `identical` to, `inherits` from (directly or indirectly), or implements (directly or indirectly) a unique type `C<V₁...Vₑ>`
+  - `U` is a function pointer type (§function-pointers) then `delegate*<U2..Uk, U1>` and `V` is a function pointer type which is identical to `delegate*<V2..Vk, V1>`, and the calling convention of `U` is identical to `V`, and the refness of `Ui` is identical to `Vi`.
+    > *Note*: This is only applicable in unsafe code. *end note*
   - (The “uniqueness” restriction means that given an interface `C<T>{} class V<Z>: C<X<Z>>, C<Y<Z>>{}`, then no inference is made when inferring from `C<U₁>` to `V<Q>`. Inferences are not made from `U₁` to either `X<Q>` or `Y<Q>`.)
-
   If any of these cases apply then an inference is made from each `Uᵢ` to the corresponding `Vᵢ` as follows:
-  - If `Uᵢ` is not known to be a reference type then an *exact inference* is made
+  - If `U` is not a function pointer type and `Ui` is not known to be a reference type, or if `U` is a function pointer type and `Ui` is not known to be a function pointer type or a reference type, then an *exact inference* is made
+    > *Note*: Function-pointer type-related text is only applicable in unsafe code. *end note*
   - Otherwise, if `V` is an array type then an *upper-bound inference* is made
   - Otherwise, if `U` is `C<U₁...Uₑ>` then inference depends on the `i-th` type parameter of `C`:
     - If it is covariant then an *upper-bound inference* is made.
     - If it is contravariant then a *lower-bound inference* is made.
     - If it is invariant then an *exact inference* is made.
+- Otherwise, if `U` is `delegate*<U2..Uk, U1>` then inference depends on the i-th parameter of `delegate*<U2..Uk, U1>`:
+  - If `U1`:
+    - If the return is by value, then an upper-bound inference is made.
+    - If the return is by reference, then an exact inference is made.
+  - If `U2`..`Uk`:
+    - If the parameter is by value, then a lower-bound inference is made.
+    - If the parameter is by reference, then an exact inference is made.
+  > *Note*: This is only applicable in unsafe code. *end note*
 - Otherwise, no inferences are made.
 
 #### 12.6.3.13 Fixing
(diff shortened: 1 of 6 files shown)


standard/expressions.md
```diff
@@ -811,6 +819,8 @@ An *output type inference* is made *from* an expression `E` *to* a type `T` in
...
 - If `E` is an anonymous function with inferred return type `U` ([§12.6.3.14](expressions.md#126314-inferred-return-type)) and `T` is a delegate type or expression tree type with return type `Tₓ`, then a *lower-bound inference* ([§12.6.3.11](expressions.md#126311-lower-bound-inferences)) is made *from* `U` *to* `Tₓ`.
 - Otherwise, if `E` is a method group and `T` is a delegate type or expression tree type with parameter types `T₁...Tᵥ` and return type `Tₓ`, and overload resolution of `E` with the types `T₁...Tᵥ` yields a single method with return type `U`, then a *lower-bound inference* is made *from* `U` *to* `Tₓ`.
+- If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then with parameter types `T1..Tk` and return type `Tb`, and overload resolution of `E` with the types `T1..Tk` yields a single method with return type `U`, then a *lower-bound inference* is made from `U` to `Tb`.
```
```suggestion
- If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then with parameter types `T₁..Tₖ` and return type `Tₓ`, and overload resolution of `E` with the types `T₁..Tₖ` yields a single method with return type `U`, then a *lower-bound inference* is made from `U` to `Tₓ`.
```

(Note: changing "b" to "x" as there's no Unicode subscript for b.)

standard/expressions.md
```diff
@@ -843,15 +853,25 @@ A *lower-bound inference from* a type `U` *to* a type `V` is made as follows:
   - `V` is an array type `V₁[...]`and `U` is an array type `U₁[...]`of the same rank
   - `V` is one of `IEnumerable<V₁>`, `ICollection<V₁>`, `IReadOnlyList<V₁>>`, `IReadOnlyCollection<V₁>` or `IList<V₁>` and `U` is a single-dimensional array type `U₁[]`
   - `V` is a constructed `class`, `struct`, `interface` or `delegate` type `C<V₁...Vₑ>` and there is a unique type `C<U₁...Uₑ>` such that `U` (or, if `U` is a type `parameter`, its effective base class or any member of its effective interface set) is identical to, `inherits` from (directly or indirectly), or implements (directly or indirectly) `C<U₁...Uₑ>`.
+  - `V` is a function pointer type (§function-pointers) `delegate*<V2..Vk, V1>` and there is a function pointer type `delegate*<U2..Uk, U1>` such that `U` is identical to `delegate*<U2..Uk, U1>`, and the calling convention of `V` is identical to `U`, and the refness of `Vi` is identical to `Ui`.
```
```suggestion
  - `V` is a function pointer type (§function-pointers) `delegate*<V₂..Vₖ, V₁>` and there is a function pointer type `delegate*<U₂..Uₖ, U₁>` such that `U` is identical to `delegate*<U₂..Uₖ, U₁>`, and the calling convention of `V` is identical to `U`, and the refness of `Vᵢ` is identical to `Uᵢ`.
```

jskeet: For later polish: we should check whether we've got a better term than "refness".

standard/expressions.md
```diff
@@ -843,15 +853,25 @@ A *lower-bound inference from* a type `U` *to* a type `V` is made as follows:
...
+    > *Note*: This is only applicable in unsafe code. *end note*
   - (The “uniqueness” restriction means that in the case interface `C<T>{} class U: C<X>, C<Y>{}`, then no inference is made when inferring from `U` to `C<T>` because `U₁` could be `X` or `Y`.)
-
   If any of these cases apply then an inference is made from each `Uᵢ` to the corresponding `Vᵢ` as follows:
-  - If `Uᵢ` is not known to be a reference type then an *exact inference* is made
+  - If `Uᵢ` is not known to be a reference type then an *exact inference* is made; or alternatively, if `U` is not a function pointer type and `Ui` is not known to be a reference type, or if `U` is a function pointer type and `Ui` is not known to be a function pointer type or a reference type, then an exact inference is made
```
```suggestion
  - If `Uᵢ` is not known to be a reference type then an *exact inference* is made; or alternatively, if `U` is not a function pointer type and `Uᵢ` is not known to be a reference type, or if `U` is a function pointer type and `Uᵢ` is not known to be a function pointer type or a reference type, then an exact inference is made
```

jskeet: I think we'll want to reword this bullet - I know I find it confusing at the moment...

standard/expressions.md
```diff
@@ -843,15 +853,25 @@ A *lower-bound inference from* a type `U` *to* a type `V` is made as follows:
...
+    > *Note*: This is only applicable in unsafe code. *end note*
   - Otherwise, if `U` is an array type then a *lower-bound inference* is made
   - Otherwise, if `V` is `C<V₁...Vₑ>` then inference depends on the `i-th` type parameter of `C`:
     - If it is covariant then a *lower-bound inference* is made.
     - If it is contravariant then an *upper-bound inference* is made.
     - If it is invariant then an *exact inference* is made.
+  - Otherwise, if `V` is `delegate*<V2..Vk, V1>` then inference depends on the i-th parameter of `delegate*<V2..Vk, V1>`:
```
```suggestion
  - Otherwise, if `V` is `delegate*<V₂..Vₖ, V₁>` then inference depends on the i-th parameter of `delegate*<V₂..Vₖ, V₁>`:
```

standard/expressions.md
```diff
@@ -843,15 +853,25 @@ A *lower-bound inference from* a type `U` *to* a type `V` is made as follows:
...
+    - If V1:
```
jskeet, line 868: I don't understand these bullets... Perhaps "For the return type V1" and "For the parameter types V2..VK"?

standard/expressions.md
```diff
@@ -864,15 +884,25 @@ An *upper-bound inference from* a type `U` *to* a type `V` is made as follows:
   - `U` is one of `IEnumerable<Uₑ>`, `ICollection<Uₑ>`, `IReadOnlyList<Uₑ>`, `IReadOnlyCollection<Uₑ>` or `IList<Uₑ>` and `V` is a single-dimensional array type `Vₑ[]`
   - `U` is the type `U1?` and `V` is the type `V1?`
   - `U` is constructed class, struct, interface or delegate type `C<U₁...Uₑ>` and `V` is a `class, struct, interface` or `delegate` type which is `identical` to, `inherits` from (directly or indirectly), or implements (directly or indirectly) a unique type `C<V₁...Vₑ>`
+  - `U` is a function pointer type (§function-pointers) then `delegate*<U2..Uk, U1>` and `V` is a function pointer type which is identical to `delegate*<V2..Vk, V1>`, and the calling convention of `U` is identical to `V`, and the refness of `Ui` is identical to `Vi`.
```
```suggestion
  - `U` is a function pointer type (§function-pointers) then `delegate*<U₂..Uₖ, U₁>` and `V` is a function pointer type which is identical to `delegate*<V₂..Vₖ, V₁>`, and the calling convention of `U` is identical to `V`, and the refness of `Uᵢ` is identical to `Vᵢ`.
```

standard/expressions.md
```diff
@@ -864,15 +884,25 @@ An *upper-bound inference from* a type `U` *to* a type `V` is made as follows:
...
+    > *Note*: This is only applicable in unsafe code. *end note*
   - (The “uniqueness” restriction means that given an interface `C<T>{} class V<Z>: C<X<Z>>, C<Y<Z>>{}`, then no inference is made when inferring from `C<U₁>` to `V<Q>`. Inferences are not made from `U₁` to either `X<Q>` or `Y<Q>`.)
-
   If any of these cases apply then an inference is made from each `Uᵢ` to the corresponding `Vᵢ` as follows:
-  - If `Uᵢ` is not known to be a reference type then an *exact inference* is made
+  - If `U` is not a function pointer type and `Ui` is not known to be a reference type, or if `U` is a function pointer type and `Ui` is not known to be a function pointer type or a reference type, then an *exact inference* is made
```
```suggestion
  - If `U` is not a function pointer type and `Uᵢ` is not known to be a reference type, or if `U` is a function pointer type and `Uᵢ` is not known to be a function pointer type or a reference type, then an *exact inference* is made
```

standard/expressions.md
```diff
@@ -864,15 +884,25 @@ An *upper-bound inference from* a type `U` *to* a type `V` is made as follows:
...
+    > *Note*: Function-pointer type-related text is only applicable in unsafe code. *end note*
   - Otherwise, if `V` is an array type then an *upper-bound inference* is made
   - Otherwise, if `U` is `C<U₁...Uₑ>` then inference depends on the `i-th` type parameter of `C`:
     - If it is covariant then an *upper-bound inference* is made.
     - If it is contravariant then a *lower-bound inference* is made.
     - If it is invariant then an *exact inference* is made.
+- Otherwise, if `U` is `delegate*<U2..Uk, U1>` then inference depends on the i-th parameter of `delegate*<U2..Uk, U1>`:
```
```suggestion
- Otherwise, if `U` is `delegate*<U₂..Uₖ, U₁>` then inference depends on the i-th parameter of `delegate*<U₂..Uₖ, U₁>`:
```

standard/expressions.md
```diff
@@ -864,15 +884,25 @@ An *upper-bound inference from* a type `U` *to* a type `V` is made as follows:
...
+  - If `U1`:
```
jskeet, line 899: Same as earlier, I think we can do better than this.

standard/expressions.md
```diff
@@ -1098,6 +1128,10 @@ In case the parameter type sequences `{P₁, P₂, ..., Pᵥ}` and `{Q₁, Q₂
 - If for at least one parameter `Mᵥ` uses the ***better parameter-passing choice*** ([§12.6.4.4](expressions.md#12644-better-parameter-passing-mode)) than the corresponding parameter in `Mₓ` and none of the parameters in `Mₓ` use the better parameter-passing choice than `Mᵥ`, `Mᵥ` is better than `Mₓ`.
 - Otherwise, no function member is better.
 
+A `delegate*` is more specific than `void*`.
```
jskeet, line 1131: This feels a little out of place at the moment - is this for parameters, return types, anything else? (If it's parameters, then it should be in the "if Mv has more specific parameter types" bit.

standard/expressions.md
```diff
@@ -1118,6 +1152,8 @@ Given an implicit conversion `C₁` that converts from an expression `E` to a ty
 
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` ([§12.6.4.6](expressions.md#12646-exactly-matching-expression))
 - `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§12.6.4.7](expressions.md#12647-better-conversion-target))
+  - `V` is a function pointer type `delegate*<V2..Vk, V1>` and `U` is a function pointer type `delegate*<U2..Uk, U1>`, and the calling convention of `V` is identical to `U`, and the refness of `Vi` is identical to `Ui`.
```
```suggestion
  - `V` is a function pointer type `delegate*<V₂..Vₖ, V₁>` and `U` is a function pointer type `delegate*<U₂..Uₖ, U₁>`, and the calling convention of `V` is identical to `U`, and the refness of `Vᵢ` is identical to `Uᵢ`.
```

standard/expressions.md
jskeet, line 1996: We'll want to remove this TODO - it's not clear to me whether the text additions later are sufficient.

standard/expressions.md
```diff
@@ -1977,6 +2013,25 @@ The result of evaluating an *invocation_expression* is classified as follows:
...
+
+- It is an error to directly call that method from C#. Instead, one can obtain a function pointer (§function-pointers) to that method and then invoke the method via that pointer.
+- It is an error for that method to have a parameter or return type that is not an `unmanaged_type` ([§8.8](types.md#88-unmanaged-types)).
+- It is an error for that method to have type parameters, even if those type parameters are constrained to `unmanaged`.
+- It is an error for that method to be in a generic type.
+- It is an error to convert that method to a delegate type.
+
+For a function pointer invocation, the *primary_expression* of the *invocation_expression* shall be a value of a *funcptr_type*. Furthermore, considering the method being pointed to to be a function member with the same parameter list as the *funcptr_type*, the *funcptr_type* shall be applicable ([§12.6.4.2](expressions.md#12642-applicable-function-member)) with respect to the *argument_list* of the *invocation_expression*.
```
jskeet, line 2026: Let's try to avoid the stuttering in this part:

> Furthermore, considering the method being pointed to to be a function member with the same parameter list as the *funcptr_type*

standard/expressions.md
```diff
@@ -1977,6 +2013,25 @@ The result of evaluating an *invocation_expression* is classified as follows:
...
+
+The run-time processing of a function pointer invocation of the form `F(A)`, where `F` is a *primary_expression* of a *funcptr_type* and `A` is an optional *argument_list*, consists of the following steps:
+
+- `F` is evaluated. If this evaluation causes an exception, no further steps are executed.
+- The argument list `A` is evaluated. If this evaluation causes an exception, no further steps are executed.
+- The value of `F` is checked to be valid. If that value is `null`, an implementation-defined exception is thrown, and no further steps are executed.
```
jskeet, line 2032: Does "valid" just mean "non-null"? If so, let's change it to say that - otherwise, let's add more detail about what "valid" means.

standard/unsafe-code.md
```diff
@@ -115,39 +115,96 @@ When the `unsafe` modifier is used on a partial type declaration ([§15.2.7](cla
...
+Unlike references (values of reference types), pointers are not tracked by the garbage collector—the garbage collector has no knowledge of pointers and the data or static methods to which they point. For this reason a pointer is not permitted to point to a reference or to a struct that contains references, and the referent type of a pointer shall be an *unmanaged_type*. Pointer types themselves are unmanaged types, so a pointer type may be used as the referent type for another pointer type.
+
+The intuitive rule for mixing pointers and references is that referents of references (objects) are permitted to contain pointers, but referents of pointers are not permitted to contain references.
+
+For a given implementation, all pointer types shall have the same size and representation. A null pointer value shall be represented by all-bits-zero.
```
jskeet, line 140: Do we need this to be in the spec?

standard/unsafe-code.md
```diff
@@ -115,39 +115,96 @@ When the `unsafe` modifier is used on a partial type declaration ([§15.2.7](cla
...
+The *dynamic erasure* of a type `E*` is the pointer type with referent type of the dynamic erasure of `E`.
+
+An expression with a pointer type cannot be used to provide the value in a *member_declarator* within an *anonymous_object_creation_expression* ([§12.8.17.3](expressions.md#128173-anonymous-object-creation-expressions)).
+
+The default value ([§9.3](variables.md#93-default-values)) for any pointer type is `null`.
+
+A method can return a value of some type, and that type can be a pointer.
 
-> *Example*: Some examples of pointer types are given in the table below:
+> *Example*: When given a pointer to a contiguous sequence of `int`s, that sequence’s element count, and some other `int` value, the following method returns the address of that value in that sequence, if a match occurs; otherwise it returns `null`:
```
jskeet:
```suggestion
> *Example*: When given a pointer to a contiguous sequence of `int`s, that sequence’s element count, and some other `int` value, the following method returns the address of the first occurrence of that value in that sequence, if a match occurs; otherwise it returns `null`:
```

standard/unsafe-code.md
```diff
@@ -115,39 +115,96 @@ When the `unsafe` modifier is used on a partial type declaration ([§15.2.7](cla
...
+
+The intuitive rule for mixing pointers and references is that referents of references (objects) are permitted to contain pointers, but referents of pointers are not permitted to contain references.
+
+For a given implementation, all pointer types shall have the same size and representation. A null pointer value shall be represented by all-bits-zero.
+
+Pointer types are a separate category of types. Unlike reference types and value types, pointer types do not inherit from `object` and no conversions exist between pointer types and `object`. In particular, boxing and unboxing ([§8.3.13](types.md#8313-boxing-and-unboxing)) are not supported for pointers. However, conversions are permitted between different pointer types and between pointer types and the integral types. This is described in [§24.5](unsafe-code.md#245-pointer-conversions).
```
jskeet, line 142: Note: this is moved text rather than new text (including the restrictions and permissions below).

standard/unsafe-code.md
```diff
@@ -115,39 +115,96 @@ When the `unsafe` modifier is used on a partial type declaration ([§15.2.7](cla
...
+
+```ANTLR
+dataptr_type
+    : value_type ('*')+
+    | funcptr_type ('*')+
+    | voidptr_type ('*')+
+    ;
+```
+
+A *dataptr_type* is written as an *unmanaged_type* ([§8.8](types.md#88-unmanaged-types)), *funcptr_type*, or *voidptr_type*, followed by one or more `*` tokens.
```
jskeet, line 192: Hmm... that doesn't seem to be what's described above. (`value_type` instead of `unmanaged_type`)

standard/unsafe-code.md
```diff
@@ -115,39 +115,96 @@ When the `unsafe` modifier is used on a partial type declaration ([§15.2.7](cla
...
+>     }
+>     return null;
+> }
+> ```
+>
+> *end example*
+
+### §data-pointers Data pointers
+
+A ***data pointer*** is a pointer capable of containing the address of a variable having *value_type* ([§8.3.1](types.md#831-general)), *funcptr_type* (§function-pointers), or *voidptr_type* (§void-pointers).
```
jskeet, line 182: Not sure what "having" means here.

standard/unsafe-code.md
```diff
@@ -115,39 +115,96 @@ When the `unsafe` modifier is used on a partial type declaration ([§15.2.7](cla
...
+>         }
+>         ++pi;
+>     }
+>     return null;
+> }
+> ```
+>
+> *end example*
+
+### §data-pointers Data pointers
```
jskeet, line 180: I'd expected to see "data pointers" as separate from "function pointers" here, but it looks like it includes function pointers...

jskeet: (Or is it only "pointers to function pointers"? Maybe...)

standard/unsafe-code.md
```diff
@@ -115,39 +115,96 @@ When the `unsafe` modifier is used on a partial type declaration ([§15.2.7](cla
...
 >
 > **Example**   | **Description**
 > --------- | -----------
 > `byte*`   | Pointer to `byte`
-> `char*`   | Pointer to `char`
-> `int**`   | Pointer to pointer to `int`
 > `int*[]`  | Single-dimensional array of pointers to `int`
-> `void*`   | Pointer to unknown type
+> `char**`  | Pointer to pointer to `char`
+> `delegate*<void>*`   | Pointer to a pointer to a static method having no parameters and a `void` return type
```
jskeet, line 201: Maybe add an example with parameter types and a return type.

standard/unsafe-code.md
jskeet: I think it's probably worth clarifying this...

standard/unsafe-code.md
```diff
@@ -156,106 +213,198 @@ For a given implementation, all pointer types shall have the same size and repre
...
 > }
 > ```
 >
 > *end note*
 
-A method can return a value of some type, and that type can be a pointer.
+In an unsafe context, several constructs are available for operating on data pointers:
 
-> *Example*: When given a pointer to a contiguous sequence of `int`s, that sequence’s element count, and some other `int` value, the following method returns the address of that value in that sequence, if a match occurs; otherwise it returns `null`:
+- The unary `*` operator may be used to perform pointer indirection ([§24.6.2](unsafe-code.md#2462-pointer-indirection)).
```
jskeet, line 263: Again, this is text which has been moved.

standard/unsafe-code.md
```diff
@@ -156,106 +213,198 @@ For a given implementation, all pointer types shall have the same size and repre
...
+funcptr_parameter_list
+    : (funcptr_parameter ',')*
+    ;
+
+funcptr_parameter
+    : parameter_mode_modifier? type
+    ;
+
+funcptr_return_type
+    : ref_kind? return_type
```
jskeet, line 305: This would appear to allow `ref void` as a return type. We probably need to prohibit that.

standard/unsafe-code.md
```diff
@@ -156,106 +213,198 @@ For a given implementation, all pointer types shall have the same size and repre
...
 >     }
->     return null;
 > }
 > ```
 >
-> *end example*
+> Given that the function pointers in the array point to methods with no parameters, `&Log` takes the address of the `Log` method having no parameters. *end example*
 
-In an unsafe context, several constructs are available for operating on pointers:
+*unmanaged_calling_convention* supports a small number of predefined conventions (`Cdecl`, `Stdcall`, `Thiscall`, and `Fastcall`, all of which are contextual keywords), which may be used standalone or as an *identifier* in an *unmanaged_calling_convention* identifier list. Other implementation-defined conventions are permitted, and multiple conventions can be combined by using an *identifier* list, possibly containing one or more of these predefined conventions. Lookup and processing for identifiers in this list is done in an implementation-defined manner.
```
jskeet, line 355: It's probably worth discussing in a meeting whether we really need the predefined conventions to be included in the standard. If we have to make sure that implementation-defined conventions "work" (in terms of still being contextual keywords etc) then do we actually need these?

standard/unsafe-code.md
```diff
@@ -156,106 +213,198 @@ For a given implementation, all pointer types shall have the same size and repre
...
+    ;
+```
+
+Just as a method has a signature ([§7.6](basic-concepts.md#76-signatures-and-overloading)), a function pointer type has a signature for the method type to which it may point. That signature includes the calling convention.
+
+The non-null value of a function pointer having type `T` represents the address of a method having a signature compatible with type `T`.
+
+If no *calling_convention_specifier* is provided, the default is `managed`, which results in the execution environment’s default mechanism being used. Specific unmanaged conventions can be specified using *unmanaged_calling_convention* whose tokens are mapped to implementation-defined names having implementation-defined semantics. The set of valid combinations of these tokens is implementation-defined.
+
+> *Note*: The *calling_convention_specifier* allows a potentially more efficient calling mechanism to be chosen, or for methods written in languages other than C# to be called. *end note*.
```
jskeet, line 315: It's unclear what's meant by multiple calling conventions being specified - are they alternatives, additional constraints, or is that implementation-specific?

standard/unsafe-code.md
```diff
@@ -156,106 +213,198 @@ For a given implementation, all pointer types shall have the same size and repre
...
+> both cases use the identifier-list grammar rule. *end example*
+
+Custom attributes cannot be applied to a *funcptr_type* or to any of its elements.
+
+A parameter of type *funcptr_type* shall not be marked as `params` ([§15.6.2.1](classes.md#15621-general)).
+
+In an unsafe context, the following constructs are available for operating on function pointers:
+
+- The `&` operator may be used to obtain the address of a static method ([§24.6.5](unsafe-code.md#2465-the-address-of-operator))
+- The `==`, `!=`, `<`, `>`, `<=`, and `=>` operators may be used to compare pointers ([§24.6.8](unsafe-code.md#2468-pointer-comparison)).
```
jskeet:
```suggestion
- The `==`, `!=`, `<`, `>`, `<=`, and `>=` operators may be used to compare pointers ([§24.6.8](unsafe-code.md#2468-pointer-comparison)).
```

standard/unsafe-code.md
```diff
@@ -156,106 +213,198 @@ For a given implementation, all pointer types shall have the same size and repre
...
+- The invocation_expression operator, `()`, may be used to call the method being pointed to ([§12.8.9.1](expressions.md#12891-general)).
```
jskeet, line 378: Not sure whether `invocation_expression` should be in italics, code, or should just not have the underscore.

standard/unsafe-code.md
```diff
@@ -156,106 +213,198 @@ For a given implementation, all pointer types shall have the same size and repre
...
+> <!-- Example: {template:"standalone-lib-without-using", name:"FunctionPointers2", ignoredWarnings:["CS0169"]} -->
+> ```csharp
+> unsafe class C
+> {
+>     delegate* unmanaged[SuppressGCTransition]<int, int> fpx;
+>     delegate* unmanaged[Stdcall, SuppressGCTransition]<int, int> fpy;
+> }
+> ```
+>
+> both cases use the identifier-list grammar rule. *end example*
```
jskeet, line 368: I don't think there is such a grammar rule. Not sure what's intended here.

standard/unsafe-code.md
```diff
@@ -156,106 +213,198 @@ For a given implementation, all pointer types shall have the same size and repre
...
+
+A ***void pointer*** is a pointer capable of containing the value of a data pointer or a function pointer.
+
+```ANTLR
+voidptr_type
+    : 'void' '*'
+    ;
+```
+
+A *voidptr_type* is written as the keyword `void` followed by thye `*` token.
```
jskeet:
```suggestion
A *voidptr_type* is written as the keyword `void` followed by the `*` token.
```

standard/unsafe-code.md
```diff
@@ -156,106 +213,198 @@ For a given implementation, all pointer types shall have the same size and repre
...
+
+> *Example*: Some examples of void-pointer types are given in the table below:
```
jskeet, line 392: I'm not sure whether the array is really a void pointer type... it's not *itself* unknown, and it doesn't follow the rules above. Maybe this should be "some examples of the use of void pointers" are given in the table below"?

standard/unsafe-code.md
```diff
@@ -276,7 +425,14 @@ A static field is classified as a moveable variable. Also, a by-reference parame
...
-- From the `null` literal ([§6.4.5.7](lexical-structure.md#6457-the-null-literal)) to any *pointer_type*.
+- From *null_literal* ([§6.4.5.7](lexical-structure.md#6457-the-null-literal)) to any *pointer_type*.
+- From *funcptr_type* `F0` to *funcptr_type* `F1`, provided all of the following are true:
+  - `F0` and `F1` have the same number of parameters, and each parameter `D0n` in `F0` has the same by-reference parameter modifiers as the corresponding parameter `D1n` in `F1`.
+  - For each value parameter, an identity conversion, implicit reference conversion, or implicit pointer conversion exists from the parameter type in `F0` to the corresponding parameter type in `F1`.
+  - For each by-reference parameter, the parameter type in `F0` is the same as the corresponding parameter type in `F1`.
+  - If the return type is by value, an identity, implicit reference, or implicit pointer conversion exists from the return type of `F1` to the return type of `F0`.
+  - If the return type is by reference, the return type and `ref` modifiers of `F1` are the same as the return type and `ref` modifiers of `F0`.
+  - The calling convention of `F0` is the same as the calling convention of `F1`.
```
jskeet:
```suggestion
- From *funcptr_type* `U` to *funcptr_type* `V`, provided all of the following are true:
  - `U` and `V` have the same number of parameters, and each parameter `Uᵢ` in `U` has the same by-reference parameter modifiers as the corresponding parameter `Vᵢ` in `V`.
  - For each value parameter, an identity conversion, implicit reference conversion, or implicit pointer conversion exists from the parameter type in `U` to the corresponding parameter type in `V`.
  - For each by-reference parameter, the parameter type in `U` is the same as the corresponding parameter type in `V`.
  - If the return type is by value, an identity, implicit reference, or implicit pointer conversion exists from the return type of `U` to the return type of `V`.
  - If the return type is by reference, the return type and `ref` modifiers of `V` are the same as the return type and `ref` modifiers of `U`.
  - The calling convention of `U` is the same as the calling convention of `V`.
```

Note: this is a more comprehensive rewriting than normal, as having `F0` and `F1` seems unconventional compared with `U` and `V`.

standard/unsafe-code.md
```diff
@@ -400,13 +558,15 @@ pointer_indirection_expression
     ;
 ```
 
-The unary `*` operator denotes pointer indirection and is used to obtain the variable to which a pointer points. The result of evaluating `*P`, where `P` is an expression of a pointer type `T*`, is a variable of type `T`. It is a compile-time error to apply the unary `*` operator to an expression of type `void*` or to an expression that is not of a pointer type.
+The unary `*` operator denotes pointer indirection and is used to obtain the variable to which a data pointer points. The result of evaluating `*P`, where `P` is an expression of a pointer type `T*`, is a variable of type `T`. It is a compile-time error to apply the unary `*` operator to an operand having type *funcptr_type* or *voidptr_type* or an expression that is not a pointer type.
```
jskeet:
```suggestion
The unary `*` operator denotes pointer indirection and is used to obtain the variable to which a data pointer points. The result of evaluating `*P`, where `P` is an expression of a data pointer type `T*`, is a variable of type `T`. It is a compile-time error to apply the unary `*` operator to an operand having type *funcptr_type* or *voidptr_type* or an expression that is not a pointer type.
```

standard/unsafe-code.md
```diff
@@ -486,8 +646,6 @@ pointer_element_access
 
 When recognising a *primary_expression* if both the *element_access* and *pointer_element_access* ([§24.6.4](unsafe-code.md#2464-pointer-element-access)) alternatives are applicable then the latter shall be chosen if the embedded *primary_expression* is of pointer type ([§24.3](unsafe-code.md#243-pointer-types)).
 
-In a pointer element access of the form `P[E]`, `P` shall be an expression of a pointer type other than `void*`, and `E` shall be an expression that can be implicitly converted to `int`, `uint`, `long`, or `ulong`.
```
jskeet, line 489: Not sure whether this has been moved or just removed - or why...

standard/unsafe-code.md
```diff
@@ -583,22 +743,44 @@ The `&` operator does not require its argument to be definitely assigned, but fo
...
+- For each value parameter, an identity conversion, implicit reference conversion, or implicit pointer conversion exists from the parameter type in `M` to the corresponding parameter type in `F`.
+- For each by-reference parameter, the parameter type in `M` is the same as the corresponding parameter type in `F`.
+- If the return type is by value, an identity, implicit reference, or implicit pointer conversion exists from the return type of `F` to the return type of `M`.
+- If the return type is by reference, the return type and `ref` modifiers of `F` are the same as the return type and `ref` modifiers of `M`.
+- The calling convention of `M` is the same as the calling convention of `F`.
+- `M` is a static method.
+
+An implicit conversion exists from a *unary_expression* whose target is a method group `E`, to a compatible function pointer type `F` if `E` contains at least one method that is applicable in its normal form to an argument list constructed by use of the parameter types and modifiers of `F`, as described in the following:
```
jskeet:
```suggestion
An implicit conversion exists from a *unary_expression* whose target is a method group `E`, to a compatible function pointer type `F` if `E` contains at least one method that is applicable in its normal form ([§12.6.4.2](expressions.md#12642-applicable-function-member)) to an argument list constructed by use of the parameter types and modifiers of `F`, as described in the following:
```

standard/unsafe-code.md
```diff
@@ -583,22 +743,44 @@ The `&` operator does not require its argument to be definitely assigned, but fo
...
+- The selected method `M` shall be compatible (as defined above) with the function pointer type `F`. Otherwise, a compile-time error occurs.
+- The result of the conversion is a function pointer of type `F`.
+
 ### 24.6.6 Pointer increment and decrement
 
-In an unsafe context, the `++` and `--` operators ([§12.8.16](expressions.md#12816-postfix-increment-and-decrement-operators) and [§12.9.7](expressions.md#1297-prefix-increment-and-decrement-operators)) can be applied to pointer variables of all types except `void*`. Thus, for every pointer type `T*`, the following operators are implicitly defined:
+In an unsafe context, the `++` and `--` operators ([§12.8.16](expressions.md#12816-postfix-increment-and-decrement-operators) and [§12.9.7](expressions.md#1297-prefix-increment-and-decrement-operators)) can be applied to pointer variables of all types It is a compile-time error for these operators to be applied to variables of type *funcptr_type* or *voidptr_type*. Thus, for every data pointer type `T*`, the following operators are implicitly defined:
```
jskeet:
```suggestion
In an unsafe context, the `++` and `--` operators ([§12.8.16](expressions.md#12816-postfix-increment-and-decrement-operators) and [§12.9.7](expressions.md#1297-prefix-increment-and-decrement-operators)) can be applied to data pointer variables of all types. It is a compile-time error for these operators to be applied to variables of type *funcptr_type* or *voidptr_type*. Thus, for every data pointer type `T*`, the following operators are implicitly defined:
```

standard/unsafe-code.md
```diff
@@ -616,9 +798,9 @@ T* operator –(T* x, ulong y);
 long operator –(T* x, T* y);
 ```
 
-Given an expression `P` of a pointer type `T*` and an expression `N` of type `int`, `uint`, `long`, or `ulong`, the expressions `P + N` and `N + P` compute the pointer value of type `T*` that results from adding `N * sizeof(T)` to the address given by `P`. Likewise, the expression `P – N` computes the pointer value of type `T*` that results from subtracting `N * sizeof(T)` from the address given by `P`.
+Given an expression `P` of a data pointer type `T*` and an expression `N` of type `int`, `uint`, `long`, or `ulong`, the expressions `P + N` and `N + P` compute the pointer value of type `T*` that results from adding `N * sizeof(T)` to the address given by `P`. Likewise, the expression `P – N` computes the pointer value of type `T*` that results from subtracting `N * sizeof(T)` from the address given by `P`.
```
jskeet, line 801: I'm tempted to change these to `p` and `n` to be more consistent, given that they're not types. Feedback welcome. (Ditto P and Q below.)

standard/unsafe-code.md
```diff
@@ -649,11 +831,11 @@ Given two expressions, `P` and `Q`, of a pointer type `T*`, the expression `P 
 >
 > *end example*
 
-If a pointer arithmetic operation overflows the domain of the pointer type, the result is truncated in an implementation-defined fashion, but no exceptions are produced.
+If a pointer arithmetic operation overflows the domain of the pointer type, the result is truncated in an implementation-defined fashion, and no exception is required.
```
jskeet:
```suggestion
If a pointer arithmetic operation overflows the domain of the pointer type, the result is implementation-defined, and no exception is required.
```

This is more consistent with the part at the end of 24.6.6.

standard/expressions.md
BillWagner: I don't think we do. I think we should.

standard/expressions.md
BillWagner: I think this bullet is trying to make the point that a function pointer (`delegate*`) is classified as a *pointer_type*. So, an exact isn't needed for a function pointer, the rules are detailed in the unsafe code clause.

standard/unsafe-code.md
jskeet: (This echoes Kalle's comment in #984.)

standard/expressions.md
```diff
@@ -782,10 +782,18 @@ The second phase proceeds as follows:
 
 If `E` is a method group or implicitly typed anonymous function and `T` is a delegate type or expression tree type then all the parameter types of `T` are *input types of* `E` *with type* `T`.
 
+If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then all the parameter types of `T` are input types of `E` with type `T`.
```
jskeet, line 785:
```suggestion
If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then all the parameter types of `T` are *input types of* `E` *with type* `T`.
```

standard/expressions.md
```diff
@@ -782,10 +782,18 @@ The second phase proceeds as follows:
...
+
+> *Note*: This is only applicable in unsafe code. *end note*
+
 #### 12.6.3.5 Output types
 
 If `E` is a method group or an anonymous function and `T` is a delegate type or expression tree type then the return type of `T` is an *output type of* `E` *with type* `T`.
 
+If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then the return type of `T` is an output type of `E` with type `T`.
```
jskeet, line 793:
```suggestion
If `E` is an address-of method group and `T` is a function pointer type (§function-pointers) then the return type of `T` is an *output type of* `E` *with type* `T`.
```

standard/expressions.md
jskeet:
```suggestion
For a function pointer invocation, the *primary_expression* of the *invocation_expression* shall be a value of a *funcptr_type*. Furthermore, considering the method being pointed to as a function member with the same parameter list as the *funcptr_type*, the *funcptr_type* shall be applicable ([§12.6.4.2](expressions.md#12642-applicable-function-member)) with respect to the *argument_list* of the *invocation_expression*.
```

(I think this keeps the same sense without stuttering.)

standard/lexical-structure.md
```diff
@@ -604,12 +604,13 @@ A ***contextual keyword*** is an identifier-like sequence of characters that has
 
 ```ANTLR
 contextual_keyword
-    : 'add'     | 'alias'      | 'ascending' | 'async'     | 'await'
-    | 'by'      | 'descending' | 'dynamic'   | 'equals'    | 'from'
-    | 'get'     | 'global'     | 'group'     | 'into'      | 'join'
-    | 'let'     | 'nameof'     | 'notnull'   | 'on'        | 'orderby'
-    | 'partial' | 'remove'     | 'select'    | 'set'       | 'unmanaged'
-    | 'value'   | 'var'        | 'when'      | 'where'     | 'yield'
+    : 'add'      | 'alias'      | 'ascending'  | 'async'     | 'await'
```
jskeet, line 607: Note: there's a later comment suggesting that we might not want Fastcall, Stdcall, Thiscall etc... in which case we'd need to edit this bit.

standard/portability-issues.md
```diff
@@ -11,10 +11,11 @@ This annex collects some information about portability that appears in this spec
 The behavior is undefined in the following circumstances:
 
 1. The behavior of the enclosing async function when an awaiter’s implementation of the interface methods `INotifyCompletion.OnCompleted` and `ICriticalNotifyCompletion.UnsafeOnCompleted` does not cause the resumption delegate to be invoked at most once ([§12.9.9.4](expressions.md#12994-run-time-evaluation-of-await-expressions)).
-1. Passing pointers as reference or output parameters ([§24.3](unsafe-code.md#243-pointer-types)).
+1. Passing pointers as `ref` or `out` parameters (§data-pointers).
```
jskeet, line 14: Should this actually be "passing data pointers" or does it also apply to function pointers?

standard/portability-issues.md
```diff
@@ -43,13 +44,16 @@ A conforming implementation is required to document its choice of behavior in ea
 1. The impact of thread termination when a thread has no handler for an exception, and the thread is itself terminated. ([§13.10.6](statements.md#13106-the-throw-statement))
 1. The mechanism by which linkage to an external method is achieved. ([§15.6.8](classes.md#1568-external-methods))
 1. The impact of thread termination when no matching `catch` clause is found for an exception and the code that initially started that thread is reached. ([§22.4](exceptions.md#224-how-exceptions-are-handled)).
+1. The token name mapping and semantics of unmanaged calling conventions beyond those required by this specification, and the set of valid combinations of those tokens (§function-pointers).
```
jskeet, line 47: This specification doesn't actually provide any semantics for Stdcall etc as far as I can see...

standard/unsafe-code.md
```diff
@@ -115,39 +115,96 @@ When the `unsafe` modifier is used on a partial type declaration ([§15.2.7](cla
 
 ## 24.3 Pointer types
 
-In an unsafe context, a *type* ([§8.1](types.md#81-general)) can be a *pointer_type* as well as a *value_type*, a *reference_type*, or a *type_parameter*. In an unsafe context a *pointer_type* may also be the element type of an array ([§17](arrays.md#17-arrays)). A *pointer_type* may also be used in a typeof expression ([§12.8.18](expressions.md#12818-the-typeof-operator)) outside of an unsafe context (as such usage is not unsafe).
+### §pointer-types-general General
 
-A *pointer_type* is written as a *value_type* that is an unmanaged type ([§8.8](types.md#88-unmanaged-types)) or the keyword `void`, followed by one or more `*` tokens:
+A ***pointer*** is a variable that is capable of containing the address of a variable or static method, referred to as that pointer's target. A pointer with value `null` is a ***null pointer***, and does not currently point to a variable or static method. The act of attempting to access the target of a pointer is called ***dereferencing*** ([§24.6.2](unsafe-code.md#2462-pointer-indirection) and [§24.6.4](unsafe-code.md#2464-pointer-element-access)).
```
jskeet, line 120: Is a pointer actually a variable? I'd expect it to be a value. This may be academic...

standard/unsafe-code.md
```diff
@@ -115,39 +115,96 @@ When the `unsafe` modifier is used on a partial type declaration ([§15.2.7](cla
...
-    : value_type ('*')+
-    | 'void' ('*')+
+    : dataptr_type
+    | funcptr_type
+    | voidptr_type
     ;
 ```
 
-The type specified before the `*` in a pointer type is called the ***referent type*** of the pointer type. It represents the type of the variable to which a value of the pointer type points.
+The type of the target of a pointer type is called the ***referent type*** of the pointer type. It represents the type of the variable to which a value of the pointer type points.
```
jskeet, line 132: "the type of the variable" feels like it misses out function pointers. Not sure what we'll want to do with that.

jskeet (commented):
I think this is probably ready for other folks to have a look at, but as noted I've got several comments. I'm not marking it as "meeting: discuss" as I don't think we should focus on it for September, but possibly in the October meeting.

I think the overall *shape* is probably okay, but there are some important decisions to make (e.g. calling conventions) and various nitpicks and tweaks required.

PR #1460, opened 2025-11-10
[Version 9.0] Feature support for pattern-matching enhancements

BillWagner:
This includes all commits from #1026 

There are a number of conversations that haven't been addressed from #1026:

- https://github.com/dotnet/csharpstandard/pull/1026#issuecomment-2533295419 (overall review)
- https://github.com/dotnet/csharpstandard/pull/1026#discussion_r1879122401
- https://github.com/dotnet/csharpstandard/pull/1026#discussion_r1881291819
- https://github.com/dotnet/csharpstandard/pull/1026#discussion_r1879128109
- https://github.com/dotnet/csharpstandard/pull/1026#discussion_r1879134901


BillWagner:
The [proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/patterns3.md) has several open issues. We need to ensure this PR matches the final decision on those questions.

I've looked through the LDM notes, and here's my initial analysis:

> Are `and`, `or`, and `not` some kind of contextual keyword? 

Yes, `and`, `or` and `not` are contextual keywords in a pattern. Even if a type or identifier is in scope, it's a pattern keyword. when used as a pattern. However, you can write this (confusing though it is):

```csharp
public class or { }

or or = new or();

if (or is or)
{ }
```

> Semantic type for relational operators:  We have proposed that when the input type is already a comparable primitive, that is the type of the comparison. However, when the input is not a comparable primitive, we treat the relational as including an implicit type test to the type of the constant on the right-hand-side of the relational.  If the programmer intends to support more than one input type, that must be done explicitly:

That is the decision. Those rules are in this PR.

Most of the other open questions in the speclet have resolutions documented. ~~I haven't yet verified that this PR includes those decisions.~~ Now verified, or added in the 8th commit.

The only remaining issue is the diagnostic rules for relational and logical patterns.

BillWagner:
Thanks!

BillWagner:
I attempted to rebase this PR onto the updated `draft-v9` after the latest TC49-TG2 committee meeting but encountered conflicts that need author resolution. Please rebase locally onto `draft-v9` and resolve them. Thanks!

PR #1461, opened 2025-11-11
[Version 9.0] Feature support for Static anonymous functions 

BillWagner:
This pulls in all the commits from #988 

There is one conversation on #988 that hasn't been addressed yet:

- https://github.com/dotnet/csharpstandard/pull/988#discussion_r1390496561


standard/expressions.md
```diff
@@ -5316,6 +5321,12 @@ anonymous_function_body
     ;
 ```
 
+If the modifier `static` is present, the anonymous function cannot capture state from the enclosing scope.
+
+A non-`static` local function or non-`static` anonymous function can capture state from an enclosing `static` anonymous function, but cannot capture state outside the enclosing static anonymous function.
+
+Removing the `static` modifier from an anonymous function in a valid program does not change the meaning of the program.
```
BillWagner: The unresolved comment https://github.com/dotnet/csharpstandard/pull/988#discussion_r1390496561 is regarding delegate equality for `static` anonymous functions. This PR doesn't state that `static` anonymous functions are equal (or the same instance). Roslyn doesn't guarantee that equality. 

As such, the standard is fine. (An implementation could do either, as there's no requirement.)  Do we want a note that clarifies this?

BillWagner:
@jskeet This is ready for first review at our next committee meeting.

standard/expressions.md
jskeet: Hmmm... without this, an implementation could reuse a delegate for a static anonymous function, but not reuse a delegate for a non-static one... but then removing the static modifier *does* change the meaning of the program. I'd be tempted to remove this line, or caveat it somewhat.

standard/expressions.md
```diff
@@ -5366,10 +5377,12 @@ A *block* body of an anonymous function is always reachable ([§13.2](statements
 > x => { return x + 1; }                 // Implicitly typed, block body
 > (int x) => x + 1                       // Explicitly typed, expression body
 > (int x) => { return x + 1; }           // Explicitly typed, block body
+> static (int x) => x + 1                // static anonymous function, explicitly typed, expression body
+> static (int x) => { return x + 1; }    // static anonymous function, explicitly typed, block body
 > (x, y) => x * y                        // Multiple parameters
 > () => Console.WriteLine()              // No parameters
 > async (t1,t2) => await t1 + await t2   // Async
-> delegate (int x) { return x + 1; }     // Anonymous method expression
```
jskeet, line 5352: Any reason for deleting this one rather than just adding a new one?

jskeet (commented):
Seems mostly fine.

standard/expressions.md
```diff
@@ -5366,10 +5377,12 @@ A *block* body of an anonymous function is always reachable ([§13.2](statements
...
+> static delegate (int x) { return x + 1; } // Anonymous method expression
```
BillWagner:
```suggestion
> delegate (int x) { return x + 1; } // Anonymous method expression
> static delegate (int x) { return x + 1; } // static anonymous method expression
```

standard/expressions.md
BillWagner: I just edited the original one. Adding a second one.

standard/expressions.md
BillWagner: Converted to a note.

BillWagner:
Thanks!

PR #1462, opened 2025-11-11
[Version 9.0] Feature support for covariant return types

RexJaeschke:
This PR is my formal version of the corresponding [MS spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/covariant-returns.md).

The following issues are *not* yet handled by this PR and need to be addressed:

1. [12.8.12.4 Indexer access](https://github.com/dotnet/csharpstandard/blob/v9-covariant-return-types/standard/expressions.md#128124-indexer-access): The MS spec states, “A similar change should be made in §12.8.12.4 Indexer access.” However, no proposed edits were given.
2. [15.7.6 Virtual, sealed, override, and abstract accessors](https://github.com/dotnet/csharpstandard/blob/v9-covariant-return-types/standard/classes.md#1576-virtual-sealed-override-and-abstract-accessors): The MS spec states, “We similarly permit override properties and indexers in interfaces as specified for classes in §15.7.6 Virtual, sealed, override, and abstract accessors.” However, no proposed edits were given.
3. [19 Interfaces](https://github.com/dotnet/csharpstandard/blob/v9-covariant-return-types/standard/interfaces.md#19-interfaces): The MS spec states, “We will need a rule that an explicit interface implementation must declare a return type no less derived than the return type declared in any override in its base interfaces.” However, no proposed edits were given.
4. [19.6.5 interface mapping](https://github.com/dotnet/csharpstandard/blob/v9-covariant-return-types/standard/interfaces.md#1965-interface-mapping): The MS spec states, “This is technically a breaking change, … Due to this breaking change, we might consider not supporting covariant return types on implicit implementations.” We need to check if it’s supported.


standard/interfaces.md
```diff
@@ -1099,10 +1101,12 @@ Members of a constructed interface type are considered to have any type paramete
...
 - `A` and `B` are events, and the name and type of `A` and `B` are identical.
-- `A` and `B` are indexers, the type and parameter lists of `A` and `B` are identical, and `A` has the same accessors as `B` (`A` is permitted to have additional accessors if it is not an explicit interface member implementation).
+- `A` and `B` are indexers, the parameter lists of `A` and `B` are identical, `A` has the same accessors as `B` (`A` is permitted to have additional accessors if it is not an explicit interface member implementation), and the type of `A` is convertible to the type of `B` via an identity conversion or, if `A` is a readonly indexer, an implicit reference conversion.
+
+> *Note*: Prior to the introduction of covariant return types, matching required an identity conversion between the return type of `A` and the return type of `B`. Allowing implicit reference conversions changes which member is selected as the implementation when a derived class re-implements an interface with a method that has a more derived return type. *end note*
```
BillWagner: I was ambivalent about adding this note. What do others on the committee think?

BillWagner:
@jskeet This is ready for a first read.

standard/classes.md
```diff
@@ -3937,9 +3942,11 @@ An abstract property declaration specifies that the accessors of the property ar
...
...d it is declared in a different assembly than the assembly containing the override declaration then the override declaration’s declared accessibility shall be protected. The overriding property's type shall be at least as accessible as the overriding property. If the inherited property has only a single accessor (i.e., if the inherited property is read-only or write-only), the overriding property shall include only that accessor. If the inherited property includes both accessors (i.e., if the inherited property is read-write), the overriding property can include either a single accessor or both accessors. There shall be an identity conversion or (if the inherited property is read-only and has a value return) an implicit reference conversion between the type of the overriding and the inherited property. For an override property declared in an interface, there shall also be an identity conversion or (if the inherited property is read-only and has a value return) an implicit reference conversion between the type of the overriding property and the type of every override of the overridden base property that is declared in a (direct or indirect) base interface of the overriding property.
```
jskeet, line 3952: Do we need the same bit about accessibility of the return type as we have for override methods?

standard/expressions.md
```diff
@@ -582,7 +582,7 @@ For each argument in an argument list there has to be a corresponding parameter
 
 The parameter list used in the following is determined as follows:
 
-- For virtual methods and indexers defined in classes, the parameter list is picked from the first declaration or override of the function member found when starting with the static type of the receiver, and searching through its base classes.
+- For virtual methods and indexers defined in classes, the parameter list is picked from the first declaration or override of the function member found when starting with the static type of the receiver, and searching through its base classes. For virtual methods and indexers defined in interfaces, the parameter list is picked from the declaration or override of the function member found in the most derived type among those types containing the declaration of override of the function member. It is a compile-time error if no unique such type exists.
```
jskeet: I'm not sure about this - it feels odd for the compiler to have to know about all implementations of an interface (even derived interfaces). I think I may be missing something.

standard/interfaces.md
```diff
@@ -892,6 +892,8 @@ A *type_parameter_constraints_clause* on an explicit interface method implementa
 
 For an explicit interface member implementation to be valid, the class, struct, or interface shall name an interface in its base class or base interface list that contains a member whose qualified interface member name, type, number of type parameters, and parameter types exactly match those of the explicit interface member implementation. If an interface function member has a parameter array, the corresponding parameter of an associated explicit interface member implementation is allowed, but not required, to have the `params` modifier. If the interface function member does not have a parameter array then an associated explicit interface member implementation shall not have a parameter array.
 
+For an explicit interface member implementation of a method, property, or indexer that has a return type, there shall be an identity conversion or an implicit reference conversion from the return type of the explicit interface member implementation to the return type of every override of the interface member that is declared in a (direct or indirect) base interface.
```
jskeet: Does this need to talk about "value return" as well? (To prevent a return of "ref string" from a method declared as "ref object" for example.)

standard/interfaces.md
```diff
@@ -1099,10 +1101,12 @@ Members of a constructed interface type are considered to have any type paramete
 
 For purposes of interface mapping, a class, interface, or struct member `A` matches an interface member `B` when:
 
-- `A` and `B` are methods, and the name, type, and parameter lists of `A` and `B` are identical.
-- `A` and `B` are properties, the name and type of `A` and `B` are identical, and `A` has the same accessors as `B` (`A` is permitted to have additional accessors if it is not an explicit interface member implementation).
+- `A` and `B` are methods, and the name, type, and parameter lists of `A` and `B` are identical, and the return type of `A` is convertible to the return type of `B` via an identity or implicit reference conversion.
```
jskeet: Again, my guess is that ref is relevant here.

standard/interfaces.md
jskeet: I'd be tempted to move it to inside the "notable implications" part below, and maybe add an example.
(It's not clear to me why those "notable implications" are normative, mind you...)

jskeet (commented):
The parts about searching through classes and interfaces have lost me, but I suspect that's just my lack of concentration rather than a fault.

BillWagner:
- §19.6.2: Added "(if the member has a value return)" qualifier to the explicit interface implementation rule, preventing unsound ref return covariance.
- §19.6.5: Added "if A has a value return" qualifier to the method matching rule, same ref-safety fix.
- §12.6.2.2: Reworded interface parameter list text to say "most derived interface among the static type of the receiver and its direct and indirect base interfaces" — clarifying it refers to the interface hierarchy, not all implementors.
- §15.7.6: Added informative note about the property type accessibility constraint, paralleling the note already in §15.6.5 for methods.
- §19.6.5: Moved the covariant return type note from a standalone *Note* block into the "Notable implications" bullet list, matching jskeet's suggestion.
- §12.8.10.1 : Added interface type-picking text to the returns-by-ref bullet for completeness (identity conversion only, no covariance for ref returns).
-§15.7.6: Updated sealed property text to say "derived class or interface" for consistency.

I still need to investigate bullet 3 in Rex's notes above.

For bullet 4:  This breaking change wasn't implemented. Rather, covariant return types on implicit interface implementations isn't supported (including by derived interfaces).

BillWagner:
Thanks!

BillWagner:
This is ready for a first review at our next meeting

standard/classes.md
```diff
@@ -2755,8 +2757,11 @@ A compile-time error occurs unless all of the following are true for an override
 - There is exactly one such overridden base method. This restriction has effect only if the base class type is a constructed type where the substitution of type arguments makes the signature of two methods the same.
 - The overridden base method is a virtual, abstract, or override method. In other words, the overridden base method cannot be static or non-virtual.
 - The overridden base method is not a sealed method.
-- There is an identity conversion between the return type of the overridden base method and the override method.
+- The override method has a return type that is convertible by an identity conversion or (if the method has a value return) an implicit reference conversion to the return type of the overridden base method.
+- The override method has a return type that is convertible by an identity conversion or (if the method has a value return) an implicit reference conversion to the return type of every override of the overridden base method that is declared in a (direct or indirect) base type of the override method.
```
Nigel-Ecma: Feels like this should be one bullet, or a bullet with two sub-bullets

The use of parenthesis around “if the method has a value return” just seems wrong, maybe commas?

“value return” is potentially confusing as it qualifies the reference conversion – the Standard uses the term return-by-value (§13.10.5) to distinguish from the value of value types.

standard/classes.md
```diff
@@ -2755,8 +2757,11 @@ A compile-time error occurs unless all of the following are true for an override
...
 - The override declaration and the overridden base method have the same declared accessibility. In other words, an override declaration cannot change the accessibility of the virtual method. However, if the overridden base method is protected internal and it is declared in a different assembly than the assembly containing the override declaration then the override declaration’s declared accessibility shall be protected.
+- The override method's return type is at least as accessible as the override method.
+  > *Note*: This constraint permits an override method in a `private` class to have a `private` return type.  However, it requires a `public` override method in a `public` type to have a `public` return type. *end note*
```
Nigel-Ecma: Why is this (just) about the class – “override method in a `private` class” – does it not also apply to a `private` method in a `public` class?

standard/classes.md
Nigel-Ecma, line 2761: This might be better as one bullet, or a bullet with two sub-bullets.

The use of parenthesis around “if the method has a value return” just seems wrong, maybe commas?

“value return” is potentially confusing as it qualifies the reference conversion – the Standard uses the term “return-by-value” (§13.10.5) to distinguish the meaning of “value” from its use in “value types”, or maybe something like “non-ref return”?

standard/classes.md
```diff
@@ -2747,6 +2747,8 @@ When an instance method declaration includes an `override` modifier, the method
...
 
+For an override method `M` declared in an interface `I`, the overridden base method is determined by examining each direct or indirect base interface of `I`, collecting the set of interfaces declaring an accessible method which has the same signature as `M` after substitution of type arguments. If this set of interfaces has a *most derived type*, to which there is an identity or implicit reference conversion from every type in this set, and that type contains a unique such method declaration, then that is the overridden base method.
```
jskeet, line 2750: I think it would be worth linking to 7.6 for "same signature as `M`" as that explicitly says that the return type isn't part of the signature.

standard/classes.md
jskeet: I found returns-by-value (15.6.1) - I think that's better (both the name with the "s" and the link) as it's about the method declaration rather than the return statement itself. But yes, I think that's clearer than just "value return".

However, now I've understood it, I think two bullets is fine (one for the base method, one any other overrides)... but the wording of the bullets themselves could still be clearer. If we can find a form of wording that *does* do it nicely with one bullet, that's fine by me of course.

standard/classes.md
```diff
@@ -3940,11 +3945,15 @@ An abstract property declaration specifies that the accessors of the property ar
...
...d it is declared in a different assembly than the assembly containing the override declaration then the override declaration’s declared accessibility shall be protected. The overriding property's type shall be at least as accessible as the overriding property. If the inherited property has only a single accessor (i.e., if the inherited property is read-only or write-only), the overriding property shall include only that accessor. If the inherited property includes both accessors (i.e., if the inherited property is read-write), the overriding property can include either a single accessor or both accessors. There shall be an identity conversion or (if the inherited property is read-only and has a value return) an implicit reference conversion between the type of the overriding and the inherited property. For an override property declared in an interface, there shall also be an identity conversion or (if the inherited property is read-only and has a value return) an implicit reference conversion between the type of the overriding property and the type of every override of the overridden base property that is declared in a (direct or indirect) base interface of the overriding property.
```
jskeet, line 3952: We've noted before that "read-only" and "write-only" can be tricky for properties due to ref. I suspect that's important here too.

standard/classes.md
jskeet: (It looks like the note at 3954 is along those lines, but the previous restriction around accessibility has just been removed entirely.)

standard/expressions.md
```diff
@@ -1974,8 +1974,8 @@ The optional *argument_list* ([§12.6.2](expressions.md#1262-argument-lists)) pr
...
+- Otherwise, if the *invocation_expression* invokes a returns-by-ref method ([§15.6.1](classes.md#1561-general)) or a returns-by-ref delegate, the result is a variable with an associated type of the return type of the method or delegate. If the invocation is of an instance method, and the receiver is of a class type `T`, the associated type is picked from the first declaration or override of the method found when starting with `T` and searching through its base classes. If the invocation is of an instance method, and the receiver is of an interface type `T`, the associated type is picked from the declaration or override of the method found in the most specific (§19.4.10) interface from among `T` and its direct or indirect base interfaces. It is a compile-time error if no unique such type exists.
```
jskeet, line 1977: Just to clarify, this change isn't needed due to covariant return types (because the bullet is about returns-by-ref) but it picks the override-declared return type rather than the base-method-declared return type which *could* be different based on an identity conversion?

standard/interfaces.md
```diff
@@ -892,6 +892,8 @@ A *type_parameter_constraints_clause* on an explicit interface method implementa
 
 For an explicit interface member implementation to be valid, the class, struct, or interface shall name an interface in its base class or base interface list that contains a member whose qualified interface member name, type, number of type parameters, and parameter types exactly match those of the explicit interface member implementation. If an interface function member has a parameter array, the corresponding parameter of an associated explicit interface member implementation is allowed, but not required, to have the `params` modifier. If the interface function member does not have a parameter array then an associated explicit interface member implementation shall not have a parameter array.
 
+For an explicit interface member implementation of a method, property, or indexer that has a return type, there shall be an identity conversion or (if the member has a value return) an implicit reference conversion from the return type of the explicit interface member implementation to the return type of every override of the interface member that is declared in a (direct or indirect) base interface.
```
jskeet, line 895: Whatever we do with the bullets in classes.md around line 2760 should probably be replicated here.

PR #995, opened 2023-11-20
Add support for Lambda discard parameters

RexJaeschke:

PR #1464, opened 2025-11-11
[Version 9.0] Feature support for lambda discards

BillWagner:
This replaces #995


BillWagner:
@jskeet This is very small, and ready for a first look at our next meeting.

jskeet (commented):
Remarkably simple - slightly odd, but I can see how it gets the desired effect.

BillWagner:
Thanks!

BillWagner:
This is ready for a first review in our September meeting.

standard/expressions.md
```diff
@@ -5364,7 +5364,31 @@ The behavior of *lambda_expression*s and *anonymous_method_expression*s is the s
 
 ### 12.21.2 Anonymous function signatures
 
-The *anonymous_function_signature* of an anonymous function defines the names and optionally the types of the parameters for the anonymous function. The scope of the parameters of the anonymous function is the *anonymous_function_body* ([§7.7](basic-concepts.md#77-scopes)). Together with the parameter list (if given) the anonymous-method-body constitutes a declaration space ([§7.3](basic-concepts.md#73-declarations)). It is thus a compile-time error for the name of a parameter of the anonymous function to match the name of a local variable, local constant or parameter whose scope includes the *anonymous_method_expression* or *lambda_expression*.
+If an *explicit_anonymous_function_parameter_list* or an *implicit_anonymous_function_parameter_list* contains multiple *identifier*s `_`, each of those identifiers denotes a discard ([§9.2.9.2](variables.md#9292-discards)). Otherwise, any single *identifier* `_` denotes a parameter.
```
Nigel-Ecma, line 5367: This is not quite correct. It assumes that any *identifier*s in a *explicit_anonymous_function_parameter_list*/*implicit_anonymous_function_parameter_list* are parameter names, they could however be *type*s. This needs to be written to restrict to the *identifier*s occurring directly in *explicit_anonymous_function_parameter*/*implicit_anonymous_function_parameter*.

A *discard_token* is a subset of *identifier*, so in both these cases the non-terminal is *discard_token* and referring to that is better than using `_`.

See also *deconstructor_element* where *discard_token* is explicitly included when other alternatives might also resolve to *identifier*, along with rules when to treat it as a discard or something else. This might help in deciding how this should be phrased.

Nigel-Ecma (changes requested):
Some changes needed, otherwise looks OK (and an advert for why `_` should no longer be allowed as a name, sometimes a breaking change is worth it, but yes this is MS… 😉)

standard/basic-concepts.md
```diff
@@ -620,8 +621,8 @@ The ***scope*** of a name is the region of program text within which it is possi
 - The scope of a parameter declared in an *indexer_declaration* ([§15.9](classes.md#159-indexers)) is the *indexer_body* of that *indexer_declaration*.
 - The scope of a parameter declared in an *operator_declaration* ([§15.10](classes.md#1510-operators)) is the *operator_body* of that *operator_declaration*.
 - The scope of a parameter declared in a *constructor_declaration* ([§15.11](classes.md#1511-instance-constructors)) is the *constructor_initializer* and *block* of that *constructor_declaration*.
-- The scope of a parameter declared in a *lambda_expression* ([§12.21](expressions.md#1221-anonymous-function-expressions)) is the *lambda_expression_body* of that *lambda_expression*.
-- The scope of a parameter declared in an *anonymous_method_expression* ([§12.21](expressions.md#1221-anonymous-function-expressions)) is the *block* of that *anonymous_method_expression*.
+- With the exception of discard parameters (§12.21.2), the scope of a parameter declared in a *lambda_expression* ([§12.21](expressions.md#1221-anonymous-function-expressions)) is the *lambda_expression_body* of that *lambda_expression*.
```
jskeet, line 624: Section 12.21.2 seems to split things into "discards" and "parameters" - in a way that suggests a discard isn't a parameter, making this change unnecessary. It might still be better for the sake of clarity, but we should probably decide that intentionally.

It's also worth noting that 12.21.2 just calls them discards, not discard parameters, at least sometimes.

Perhaps we should just change 12.21.2 to explicitly specify "non-discard parameters" rather than saying "otherwise they are parameters".

jskeet (commented):
Agreed with Nigel's general comment too. It's pretty gnarly that the meaning changes between one and two occurrences...

jskeet:
We think this is ready for @BillWagner to make the final changes before we get final approval and merge.

jnm2 (approved):

jnm2:
The language team has been talking approvingly of making a hard break in an upcoming version so `_` is a keyword everywhere (no longer a valid identifier), given the successful (i.e. uneventful/slightly happy) community reaction to the breaking changes we unleashed with the `field` keyword. We've been talking about this for years but it keeps coming up.

PR #1465, opened 2025-11-11
[Version 9.0] Feature support for target-typed conditional expression

RexJaeschke:
This PR is my formal version of the corresponding [MS spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/target-typed-conditional-expression.md). Previously, I created PR #1071 to address this, but when Bill went to push that work into the new v9 process, he found it badly compromised. As such, that PR will be retired and this one will be used in its place.

12.18 Conditional operator has an open issue; see the following words in that section: ***placeholder for words somehow referring to “12.6.4.5 Better conversion from expression.”***

Here are the details:

Mail to Bill on 2024-03-18: The MS spec suggests the text for “[12.6.4.5 Better conversion from expression](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#12645-better-conversion-from-expression)” be augmented; OK. However, “[12.18 Conditional operator](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#1218-conditional-operator)“ does *not* reference that section; instead, it points to “[12.6.3.15 Finding the best common type of a set of expressions](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#126315-finding-the-best-common-type-of-a-set-of-expressions).” So, I don’t see how the proposed changes to “better conversion” will be seen in the context of “best common type.” Unless these two are somehow linked and I’m not seeing where.

Bill’s reply on 2024-04-01: I think the conditional expression section should reference both sections, as sometimes the conditional expression applies when there is no best common type.

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -390,6 +390,15 @@ While throw expressions do not have a type, they may be implicitly converted to 
 
 There is an implicit conversion from a *switch_expression* ([§12.11](expressions.md#1211-switch-expression)) to every type `T` for which there exists an implicit conversion from each *switch_expression_arm*’s *switch_expression_arm_expression*’s to `T`.
 
+### §imp-cond-expr-conv Implicit conditional expression conversions
+
+For a *conditional_expression* `c ? e1 : e2`, when
+
+1. there is no common type for `e1` and `e2`, or
+1. for which a common type exists, but one of the expressions `e1` or `e2` has no implicit conversion to that type
+
+an implicit ***conditional expression conversion*** exists that permits an implicit conversion from *conditional_expression* to any type `T` for which there is a conversion-from-expression from `e1` to `T` and also from `e2` to `T`.  It is an error if *conditional_expression* neither has a common type between `e1` and `e2` nor is subject to a conditional expression conversion.
+
 ## 10.3 Explicit conversions
 
 ### 10.3.1 General
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1114,10 +1114,11 @@ Given `int i = 10;`, according to [§12.6.4.2](expressions.md#12642-applicable-f
 
 #### 12.6.4.5 Better conversion from expression
 
-Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if one of the following holds:
+Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if `E` does not exactly match `T₂` and at least one of the following holds:
 
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` ([§12.6.4.6](expressions.md#12646-exactly-matching-expression))
-- `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§12.6.4.7](expressions.md#12647-better-conversion-target))
+- `C₁` is not a conditional expression conversion and `C₂` is a conditional expression conversion.
+- `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§12.6.4.7](expressions.md#12647-better-conversion-target)) and either `C₁` and `C₂` are both conditional expression conversions or neither is a conditional expression conversion.
 - `E` is a method group ([§12.2](expressions.md#122-expression-classifications)), `T₁` is compatible ([§21.4](delegates.md#214-delegate-compatibility)) with the single best method from the method group for conversion `C₁`, and `T₂` is not compatible with the single best method from the method group for conversion `C₂`
 
 #### 12.6.4.6 Exactly matching expression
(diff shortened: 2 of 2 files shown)


standard/conversions.md
```diff
@@ -384,6 +384,15 @@ While throw expressions do not have a type, they may be implicitly converted to
...
 There is an implicit conversion from a *switch_expression* ([§12.11](expressions.md#1211-switch-expression)) to every type `T` for which there exists an implicit conversion from each *switch_expression_arm*’s *switch_expression_arm_expression*’s to `T`.
 
+### §imp-cond-expr-conv Implicit conditional expression conversions
+
+For a *conditional_expression* `c ? e1 : e2`, when
+
+1. there is no common type for `e1` and `e2`, or
+1. for which a common type exists, but one of the expressions `e1` or `e2` has no implicit conversion to that type
+
+an implicit ***conditional expression conversion*** exists that permits an implicit conversion from *conditional_expression* to any type `T` for which there is a conversion-from-expression from `e1` to `T` and also from `e2` to `T`.  It is an error if *conditional_expression* neither has a common type between `e1` and `e2` nor is subject to a conditional expression conversion.
```
jskeet, line 400: So this makes available a conversion to a specific type... and I'm assuming we expect the rules to just drop out for which types are "tested" for this? For example, if I have

```csharp
void M(int x) { ... }
void M(string y) { ... }
```

... then overload resolution of `M(true ? new(new char[0]) : new(new char[1]))` would detect that there's an implicit conversion from both of those target-typed new expressions to `string`, but not to `int`?

standard/expressions.md
```diff
@@ -1110,10 +1110,11 @@ Given `int i = 10;`, according to [§12.6.4.2](expressions.md#12642-applicable-f
...
+Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if `E` does not exactly match `T₂` and at least one of the following holds:
 
 - `E` exactly matches `T₁` and `E` does not exactly match `T₂` ([§12.6.4.6](expressions.md#12646-exactly-matching-expression))
-- `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§12.6.4.7](expressions.md#12647-better-conversion-target))
+- `C₁` is not a conditional expression conversion and `C₂` is a conditional expression conversion.
+- `E` exactly matches both or neither of `T₁` and `T₂`, and `T₁` is a better conversion target than `T₂` ([§12.6.4.7](expressions.md#12647-better-conversion-target)) and either `C₁` and `C₂` are both conditional expression conversions or neither is a conditional expression conversion.
```
jskeet, line 1121: I need to figure out what this is about - but E can't exactly match T2 given line 1113, so it can't match both T1 and T2...

standard/expressions.md
```diff
@@ -3758,7 +3759,7 @@ cast_expression
     ;
 ```
 
-A *cast_expression* of the form `(T)E`, where `T` is a type and `E` is a *unary_expression*, performs an explicit conversion ([§10.3](conversions.md#103-explicit-conversions)) of the value of `E` to type `T`. If no explicit conversion exists from `E` to `T`, a binding-time error occurs. Otherwise, the result is the value produced by the explicit conversion. The result is always classified as a value, even if `E` denotes a variable.
+A *cast_expression* of the form `(T)E`, where `T` is a type and `E` is a *unary_expression*, performs an explicit conversion ([§10.3](conversions.md#103-explicit-conversions)) of the value of `E` to type `T`. In the presence of a conditional expression conversion (§imp-cond-expr-conv) there may be more than one possible conversion from `E` to `T`, in which case, the conditional expression conversion shall only be used as a last resort. If no explicit conversion exists from `E` to `T`, a binding-time error occurs. Otherwise, the result is the value produced by the explicit conversion. The result is always classified as a value, even if `E` denotes a variable.
```
jskeet, line 3742: I'm confused about this and think it would benefit from an example.

standard/expressions.md
```diff
@@ -5228,7 +5229,7 @@ If `ref` is present:
...
 
 - If `x` has type `X` and `y` has type `Y` then,
-  - If an identity conversion exists between `X` and `Y`, then the result is the best common type of a set of expressions ([§12.6.3.16](expressions.md#126316-finding-the-best-common-type-of-a-set-of-expressions)). If either type is `dynamic`, type inference prefers `dynamic` ([§8.7](types.md#87-the-dynamic-type)). If either type is a tuple type ([§8.3.11](types.md#8311-tuple-types)), type inference includes the element names when the element names in the same ordinal position match in both tuples.
+  - If an identity conversion exists between `X` and `Y`, then the result is the best common type of a set of expressions ([§12.6.3.16](expressions.md#126316-finding-the-best-common-type-of-a-set-of-expressions)). ***placeholder for words somehow referring to “12.6.4.5 Better conversion from expression.”*** If either type is `dynamic`, type inference prefers `dynamic` ([§8.7](types.md#87-the-dynamic-type)). If either type is a tuple type ([§8.3.11](types.md#8311-tuple-types)), type inference includes the element names when the element names in the same ordinal position match in both tuples.
```
jskeet, line 5212: I'm hoping @BillWagner can provide motivation for this via an example.

jskeet (commented):
We should probably discuss https://github.com/dotnet/csharpstandard/pull/1469 and ideally get it improved and merged before we do much more with this PR. I definitely need to look at it in more detail, and would probably benefit from a chat with Bill about it.

standard/expressions.md
BillWagner: The examples are in the design notes of the speclet:

https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/target-typed-conditional-expression.md#design-notes

standard/expressions.md
BillWagner: I think this needs to be updated per this text in the speclet: https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/target-typed-conditional-expression.md#better-conversion-from-expression

BillWagner (commented):
Added notes on Jon's comments.

BillWagner:
Thanks!

BillWagner:
During post-meeting rebase/renumber validation I noticed two items in this PR that the author needs to finalize:

1. The anchor `§imp-cond-expr-conv` in `standard/expressions.md` is **unnumbered** (it is not a proper `[§X.Y.Z](expressions.md#anchor)` cross-reference), so the renumber tool cannot resolve/number it.
2. `standard/expressions.md` still contains an author **placeholder**: "placeholder for words somehow referring to 12.6.4.5…".

Please replace the placeholder with the intended normative wording and convert `§imp-cond-expr-conv` into a numbered cross-reference so renumber stays clean. Thanks!

PR #1466, opened 2025-11-11
[Version 9.0] Feature support for attributes on local functions

BillWagner:
Update the C# standard for the feature of applying attributes on local functions.

Note that this also includes specifying `extern` on a static local function.

- [C# LDM implementation spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-9.0/local-function-attributes.md)
- [LDM 2019-04-29](https://github.com/dotnet/csharplang/blob/main/meetings/2019/LDM-2019-04-29.md)
- [LDM 2019-09-18](https://github.com/dotnet/csharplang/blob/main/meetings/2019/LDM-2019-09-18.md)

This contains all commits from #994. Subsequent edits also made.


standard/statements.md
```diff
@@ -506,18 +506,21 @@ local_function_modifier
 
 ref_local_function_modifier
     : 'static'
+    : 'extern'
```
BillWagner:
```suggestion
    | 'extern'
```

BillWagner:
I reviewed the existing edits, made a couple small wording changes.

This is ready for committee review for V9.

standard/attributes.md
```diff
@@ -836,6 +842,8 @@ For invocations that occur within field or event initializers, the member name u
 
 For invocations that occur within declarations of instance constructors, static constructors, finalizers and operators the member name used is implementation-dependent.
 
+For an invocation that occurs within a local function, the name of the member method that calls that local function is used. Consider the following: if member method `M` calls local function `F1`, which in turn calls local function `F2`, and `F2` has a parameter marked with this attribute, the method name passed to `F2` is `M`, because a local function is *not* a function member!
```
jskeet: We should separate out the normative and informative here (and remove the exclamation mark preferrably...) Showing this as actual example code would be useful too.

standard/attributes.md
```diff
@@ -667,6 +667,12 @@ The use of conditional methods in an inheritance chain can be confusing. Calls m
 >
 > *end example*
 
+#### §conditional-local-function Conditional local functions
+
+A static local function may be made conditional in the same sense as a conditional method ([§23.5.3.2](attributes.md#23532-conditional-methods)).
+
+A compile time error ocurrs if a non-static local function is made conditional.
```
jskeet: I'm interested in why this is the case - and maybe readers would be too?

standard/attributes.md
```diff
@@ -513,7 +513,7 @@ A class that is decorated with the `AttributeUsage` attribute shall derive from
 
 #### 23.5.3.1 General
 
-The attribute `Conditional` enables the definition of ***conditional method***s and ***conditional attribute class***es.
+The attribute `Conditional` enables the definition of ***conditional methods***, ***conditional local function***s, and ***conditional attribute class***es.
```
jnm2: Since the original style continues to be followed with the last two items:
```suggestion
The attribute `Conditional` enables the definition of ***conditional method***s, ***conditional local function***s, and ***conditional attribute class***es.
```

standard/attributes.md
jnm2:
```suggestion
A compile time error occurs if a non-static local function is made conditional.
```

standard/attributes.md
jnm2: Lambdas and anonymous methods are in the same boat as local functions.

jnm2 (approved):

standard/attributes.md
RexJaeschke: Agreed! My original PR (https://github.com/dotnet/csharpstandard/pull/994) had a version of this text in an example, but that appears to have been lost. I'll work with Bill to restore it.

RexJaeschke: This is resolved by [PR #1548](https://github.com/dotnet/csharpstandard/pull/1548).

BillWagner: Coming in next commit.

standard/attributes.md
BillWagner: It's a coarse rule that prevents a non-static local function that could modify state by capturing variables in the outer member. (There are cases where there's no correctness issue (reading outer variables for example), but the rule is easy to explain and implement.

History is here:  https://github.com/dotnet/csharplang/issues/1888#issuecomment-737710456 (see referenced discussion as well)

BillWagner: coming in the next commit.

BillWagner:
@jskeet I've addressed all open comments on this one. 

It is small, and Joseph approved.  I'd like to put this on the next agenda for possibly merging it.

standard/attributes.md
```diff
@@ -667,6 +667,14 @@ The use of conditional methods in an inheritance chain can be confusing. Calls m
 >
 > *end example*
 
+#### §conditional-local-function Conditional local functions
+
+A static local function may be made conditional in the same sense as a conditional method ([§23.5.3.2](attributes.md#23532-conditional-methods)).
+
+A compile time error occurs if a non-static local function is made conditional.
+
+> *Note*: This restriction protects against side-effects for non-static local functions that write to captured outer variables. *end note*
```
jskeet: That seems a little odd as a motivation - other conditional methods could have side effects too...

jskeet (commented):
Generally looks ready to go.

standard/variables.md
```diff
@@ -677,7 +677,7 @@ For an expression *expr*, which has subexpressions *expr₁*, *expr₂*, …, *e
 
 #### 9.4.4.24 Invocation expressions and object creation expressions
 
-If the method to be invoked is a partial method that has no implementing partial method declaration, or is a conditional method for which the call is omitted ([§23.5.3.2](attributes.md#23532-conditional-methods)), then the definite-assignment state of *v* after the invocation is the same as the definite-assignment state of *v* before the invocation. Otherwise the following rules apply:
+If the method to be invoked is a partial method that has no implementing partial method declaration, or is a conditional method or conditional local function for which the call is omitted ([§23.5.3.2](attributes.md#23532-conditional-methods), [§conditional-local-function](attributes.md#conditional-local-function)), then the definite-assignment state of *v* after the invocation is the same as the definite-assignment state of *v* before the invocation. Otherwise the following rules apply:
```
BillWagner:
```suggestion
If the method to be invoked is a partial method that has no implementing partial method declaration, or is a conditional method or conditional local function for which the call is omitted ([§23.5.3.2](attributes.md#23532-conditional-methods), §conditional-local-function), then the definite-assignment state of *v* after the invocation is the same as the definite-assignment state of *v* before the invocation. Otherwise the following rules apply:
```

BillWagner:
Thanks!

BillWagner:
@jskeet If we finish the C# 8 work in time, this should be ready to merge for C# 9.

standard/attributes.md
```diff
@@ -836,6 +845,41 @@ For invocations that occur within field or event initializers, the member name u
 
 For invocations that occur within declarations of instance constructors, static constructors, finalizers and operators the member name used is implementation-dependent.
 
+For an invocation that occurs within a local function or an anonymous function, the name of the member method that calls that function is used.
+
+> *Example*: Consider the following:
+>
+> <!-- Example: {template:"standalone-console", name:"CallerMemberName1", inferOutput:true} -->
```
jskeet, line 852: Can we easily extend the example to include a lambda? Presumably this would do the trick, if we've got `Action` declared elsewhere:

```csharp
Action anonymousFunction = () => F2();
anonymousFunction();
```

jskeet (approved):

standard/attributes.md
BillWagner: addressed in the next commit.

BillWagner:
@jskeet If you're happy with the updated sample, this should be ready to go.

jskeet (approved):

PR #1467, opened 2025-11-11
[Version 9.0] Feature support for module initializers

BillWagner:
This PR replaces the work in #992 

There are a number of comments on #992 that aren't addressed yet:

- https://github.com/dotnet/csharpstandard/pull/992#discussion_r1948267752
- https://github.com/dotnet/csharpstandard/pull/992#discussion_r1948219990
- https://github.com/dotnet/csharpstandard/pull/992#issuecomment-2646549518
- https://github.com/dotnet/csharpstandard/pull/992#discussion_r1948265334
- https://github.com/dotnet/csharpstandard/pull/992#discussion_r1948266157

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -496,6 +496,7 @@ A number of attributes affect the language in some way. These attributes include
 - `System.Runtime.CompilerServices.AsyncMethodBuilderAttribute` ([§23.5.5](attributes.md#2355-the-asyncmethodbuilder-attribute)), which is used to establish a task builder for an async method.
 - `System.Runtime.CompilerServices.CallerLineNumberAttribute` ([§23.5.6.2](attributes.md#23562-the-callerlinenumber-attribute)), `System.Runtime.CompilerServices.CallerFilePathAttribute` ([§23.5.6.3](attributes.md#23563-the-callerfilepath-attribute)), and `System.Runtime.CompilerServices.CallerMemberNameAttribute` ([§23.5.6.4](attributes.md#23564-the-callermembername-attribute)), which are used to supply information about the calling context to optional parameters.
 - `System.Runtime.CompilerServices.EnumeratorCancellationAttribute` ([§23.5.8](attributes.md#2358-the-enumeratorcancellation-attribute)), which is used to specify parameter for the cancellation token in an asynchronous iterator.
+- `System.Runtime.CompilerServices.ModuleInitializer` (§module-init-attr), which is used to mark a method as a module initializer.
 
 The Nullable static analysis attributes ([§23.5.7](attributes.md#2357-code-analysis-attributes)) can improve the correctness of warnings generated for nullabilities and null states ([§8.9.5](types.md#895-nullabilities-and-null-states)).
 
@@ -1147,6 +1148,22 @@ The iterator will not have access to the `CancellationToken` argument for `GetAs
 > ```
 >
 > *end example*
+
+### §module-init-attr The ModuleInitializer attribute
+
+The attribute `ModuleInitializer` is used to mark a method as a ***module initializer***. Such a method is called during initialization of the containing module. A module may have multiple initializers, which are called in an implementation-defined order.
+
+There are no limitations on what code is permitted in a module initializer.
+
+A module initializer shall have the following characteristics:
+
+- The *method_modifier* `static`.
+- No *parameter_list*.
+- A *return_type* of `void`.
+- No *type_parameter_list*.
+- Not be declared inside a *class_declaration* having a *type_parameter_list*.
+- Be accessible from the containing module (that is, have an access modifier `internal` or `public`).
+- Not be a local function.
 
 ## 23.6 Attributes for interoperation
 
--- a/standard/standard-library.md
+++ b/standard/standard-library.md
@@ -1388,6 +1388,7 @@ The following library types are referenced in this specification. The full names
 - `global::System.Runtime.CompilerServices.ICriticalNotifyCompletion`
 - `global::System.Runtime.CompilerServices.IndexerNameAttribute`
 - `global::System.Runtime.CompilerServices.INotifyCompletion`
+- `global::System.Runtime.CompilerServices.ModuleInitializerAttribute`
 - `global::System.Runtime.CompilerServices.TaskAwaiter`
 - `global::System.Runtime.CompilerServices.TaskAwaiter<TResult>`
 - `global::System.Runtime.CompilerServices.ValueTaskAwaiter`
--- a/standard/standard-library.md
+++ b/standard/standard-library.md
@@ -829,6 +829,12 @@ namespace System.Runtime.CompilerServices
         void OnCompleted(Action continuation);
     }
 
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
+    public sealed class ModuleInitializerAttribute : Attribute
+    {
+        public ModuleInitializerAttribute() { }
+    }
+
     public readonly struct TaskAwaiter : ICriticalNotifyCompletion,
         INotifyCompletion
     {
--- a/standard/portability-issues.md
+++ b/standard/portability-issues.md
@@ -43,6 +43,7 @@ A conforming implementation is required to document its choice of behavior in ea
 1. The impact of thread termination when a thread has no handler for an exception, and the thread is itself terminated. ([§13.10.6](statements.md#13106-the-throw-statement))
 1. The mechanism by which linkage to an external method is achieved. ([§15.6.8](classes.md#1568-external-methods))
 1. The impact of thread termination when no matching `catch` clause is found for an exception and the code that initially started that thread is reached. ([§22.4](exceptions.md#224-how-exceptions-are-handled)).
+1. The order of execution of module initializers in a module (§module-init-attr).
 1. An execution environment may provide additional attributes that affect the execution of a C# program. ([§23.5.1](attributes.md#2351-general))
 1. The mappings between pointers and integers. ([§24.5.1](unsafe-code.md#2451-general))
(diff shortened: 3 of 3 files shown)


BillWagner:
I attempted to rebase this PR onto the updated `draft-v9` after the latest TC49-TG2 committee meeting but encountered conflicts that need author resolution. Please rebase locally onto `draft-v9` and resolve them. Thanks!

standard/attributes.md
```diff
@@ -496,6 +496,7 @@ A number of attributes affect the language in some way. These attributes include
 - `System.Runtime.CompilerServices.AsyncMethodBuilderAttribute` ([§23.5.5](attributes.md#2355-the-asyncmethodbuilder-attribute)), which is used to establish a task builder for an async method.
 - `System.Runtime.CompilerServices.CallerLineNumberAttribute` ([§23.5.6.2](attributes.md#23562-the-callerlinenumber-attribute)), `System.Runtime.CompilerServices.CallerFilePathAttribute` ([§23.5.6.3](attributes.md#23563-the-callerfilepath-attribute)), and `System.Runtime.CompilerServices.CallerMemberNameAttribute` ([§23.5.6.4](attributes.md#23564-the-callermembername-attribute)), which are used to supply information about the calling context to optional parameters.
 - `System.Runtime.CompilerServices.EnumeratorCancellationAttribute` ([§23.5.8](attributes.md#2358-the-enumeratorcancellation-attribute)), which is used to specify parameter for the cancellation token in an asynchronous iterator.
+- `System.Runtime.CompilerServices.ModuleInitializer` (§module-init-attr), which is used to mark a method as a module initializer.
```
jskeet, line 499:
```suggestion
- `System.Runtime.CompilerServices.ModuleInitializerAttribute` (§module-init-attr), which is used to mark a method as a module initializer.
```

standard/attributes.md
jskeet, line 1109: See https://github.com/dotnet/csharpstandard/issues/1789

jskeet (commented):
I think this is ready for wider review, knowing that the comments in #992 will still need to be addressed in a future PR. I think it would be fine to create issues for those comments as a simpler way of tracking them. (I'm happy to do that if we don't want to try to fix them in this PR.)

standard/attributes.md
jskeet: (See #1789 for other times where we might want to add the suffix.)

Nigel-Ecma (changes requested):
Unless I’m missing it this does not explicitly state when the module init is run. One might intuit that it runs before anything else in the module, however as the module init is a static method in a class within a module we might expect that the class init is run before the module init even though that might appear to be counterintuitive to the module init’s purpose – to run first.

This is what appears to happen on the *current* .NET and a certain compiler – class init, module init.

However I expect (as in I haven't done IL/assembler debugging) this is an implementation artifact – the implementation writes a module init (.cctor) into the module, which is executed first, that init calls the user’s init which is a static method in a class... so that triggers the class init to run before the user’s module init.

Something needs to be specified.

jnm2:
The runtime's 335 augments document codifies the guarantee: <https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md#module-initializer>

Documenting the guarantee was required as part of shipping the C# feature.

Nigel-Ecma:
@jnm2 commented:

> The runtime's 335 augments document codifies the guarantee: https://github.com/dotnet/runtime/blob/main/docs/design/specs/Ecma-335-Augments.md#module-initializer
> 
> Documenting the guarantee was required as part of shipping the C# feature.

Hi Joseph, thanks for the URL. Unfortunately I don’t think CLR/Roslyn is doing this right. Those specs say:

> 3. No method other than those called directly or indirectly from the module initializer will be able to access the types, methods, or data in a module before its initializer completes execution.

Which at the CLR level says that the modules .cctor runs before any .cctor for the types in the module, which is all good.

LDM notes show that for this small feature they didn’t want to add special syntax, so they choose to add an attribute to add to a user written static method defined in a type. At this point an implementation could choose:

- To place the compiled/VM code for this method into the CLR module;
- To place it into the type it is defined in and add a implementation generated .cctor into the CLR module which calls it;
- or something else…

Roslyn has currently picked the second option. So the CLR calls the module’s .cctor, that calls the attributed method in some type, and that call triggers the CLR to call a user-written type initialiser, if any, *before* it executes the attributed method… So the user type initializer runs before the user module initialzer. E.g. try the following class:

```csharp
class C
{
    [ModuleInitializer]
    internal static void ModuleInit() { Console.WriteLine("Module init"); }

    static C() { Console.WriteLine("C class init"); }

    public C() { Console.WriteLine("C instance init"); }
}
```

and create an instance of `C` in your entry-point method, you will see the class init message before the module init one – both before the entry-point’s first statement is executed, then you will see the instance init message. That is intuitively the *wrong* order from a C# user POV, but the CLR is following its design.

The workaround of course is to always declare a module init as the only member of some type not otherwise used, and not to give that type an initializer.

So the question is whether this a Roslyn “bug” or ”C# design feature” – the latter will need to be spec’ed if so.

jnm2:
My money would be on point 3 needing to have an explicit fallout added, "except that the type constructor still runs before the method that is called from the module constructor." It seems likely to me that it was probably the intent but not properly stated. Besides that:

1. It is no surprise that the type initializer runs before the [executable body of the] attributed method. Anything else would be an aberration from the usual guarantees that users can work with. In a sense, a user can successfully work with the mental model "the type constructor is triggered during the start of the module initializer attributed method." I.e. which one is called first is largely a matter of subjective interpretation as long as we're strictly discussing the observable behavior of the program as it runs.

2. This behavior of a call to a method triggering the cctor will not and cannot change. It was added as part of C++/CLI and there is a sense it can never change without risks to that integration. An example I asked for at the time was making it so that reflection over all types in an assembly would trigger a module initializer that had been designed to hook AssemblyResolve to allow types to load. They said no, this additional guarantee could have unintended consequences on the implementation of the C++/CLI runtime.

jnm2:
> At this point an implementation could choose:
>
> - To place the compiled/VM code for this method into the CLR module;
> - To place it into the type it is defined in and add a implementation generated .cctor into the CLR module which calls it;
> - or something else…

I agree it's worth being explicit in the spec. However, an implementation could not choose the first option. There are no additional restrictions on a module-initializer-attributed method accessing private state in its containing type. If code doing such accesses was moved to the module cctor, it would be a violation of ecma335.

(Managing such private state is also often the reason it's so useful to have the type cctor run during the start of the method invocation. The type cctor can assign to readonly static fields, and the module initializer method cannot since it is just an otherwise ordinary method that has been scheduled.)

> Or something else...

I can only think of one "something else" which could have a meaningful effect wrt the type cctor running, which would be to move the module initializer method inside a nested type.  Then I think the cctor of the original type would not run until and unless the module initializer invokes a method or accesses a field of its containing type.

jskeet:
We agree that we want to spec the observed semantics (a static constructor is executed before any module initializer method in that class), and we *may* want to add a note suggesting that module initializers are placed in separate classes to avoid this ordering causing issues.

standard/attributes.md
```diff
@@ -1195,6 +1196,22 @@ The iterator will not have access to the `CancellationToken` argument for `GetAs
 >
 > *end example*
 
+### §module-init-attr The ModuleInitializer attribute
+
+The attribute `ModuleInitializerAttribute` is used to mark a method as a ***module initializer***. Such a method is called during initialization of the containing module. A module may have multiple initializers, which are called in an implementation-defined order.
```
jnm2, line 1201: Do we want to specify a Roslyn guarantee which is that the order is deterministic? (No two compilations with precisely the same inputs will differ in this order)

standard/attributes.md
```diff
@@ -1195,6 +1196,22 @@ The iterator will not have access to the `CancellationToken` argument for `GetAs
...
+
+There are no limitations on what code is permitted in a module initializer.
+
+A module initializer shall have the following characteristics:
+
+- The *method_modifier* `static`.
+- No *parameter_list*.
```
jnm2, line 1208: Since we're naming the grammar constructs, isn't `()` a *parameter_list*?
```suggestion
- An empty *parameter_list*.
```

standard/attributes.md
```diff
@@ -1195,6 +1196,22 @@ The iterator will not have access to the `CancellationToken` argument for `GetAs
...
+- A *return_type* of `void`.
+- No *type_parameter_list*.
+- Not be declared inside a *class_declaration* having a *type_parameter_list*.
```
jnm2, line 1211: Or any type declaration having a type parameter list. It's valid to declare module initializers inside structs, interfaces, and records too.

standard/attributes.md
```diff
@@ -1195,6 +1196,22 @@ The iterator will not have access to the `CancellationToken` argument for `GetAs
...
+- Be accessible from the containing module (that is, have an access modifier `internal` or `public`).
+- Not be a local function.
```
jnm2, line 1213: I feel that local functions are ruled out by the 'accessible' point. Maybe it's worth mentioning as a way of explaining why `[ModuleInitializer]` won't work on a top-level statements method (because it is technically a local function). But then in C# 10 when attributes can be placed on lambda methods, do we need to call them out here too?

```cs
Action m = [ModuleInitializer] () => { };
```

standard/standard-library.md
```diff
@@ -863,6 +863,10 @@ namespace System.Runtime.CompilerServices
         ///    <c>[0, Length)</c>.
         /// </exception>
         object? this[int index] { get; }
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false)]
```
jnm2, line 866: `Inherited = false` should be added because it is not the default. `AllowMultiple = false` could remain for clarity but it does not override a default.

PR #991, opened 2023-11-13
Add support for extending partial methods

RexJaeschke:

PR #1468, opened 2025-11-11
[Version 9.0] Feature support for more partial methods

BillWagner:
This PR replaces #991 

There are several unaddressed comments on #991:

- https://github.com/dotnet/csharpstandard/pull/991#discussion_r1391445329
- https://github.com/dotnet/csharpstandard/pull/991#discussion_r1391449478
- https://github.com/dotnet/csharpstandard/pull/991#discussion_r1391457993
- https://github.com/dotnet/csharpstandard/pull/991#discussion_r1391467261
- https://github.com/dotnet/csharpstandard/pull/991#discussion_r1391470874
- https://github.com/dotnet/csharpstandard/pull/991#discussion_r1391477159


BillWagner:
This is ready for a first look from the committee.

I addressed all the comments from #991. However, https://github.com/dotnet/csharpstandard/pull/991#discussion_r1391445329 is an instance where the spec language is correct, and the roslyn implementation has a parsing error.  See dotnet/roslyn#81947.

Secondly, I made more substantial edits to the original PR. The draft retained a single clause for **Partial methods**. Looking ahead at the future versions (for the standard), I split this into three sections: General, restricted partial methods, and unrestricted partial methods. I think that makes the text easier to read, and provides the framework for future versions where there are more unrestricted partial methods.

Discussion question:  I used the terms in the current PR of "restricted" and "unrestricted". I'm not convinced those are the best terms. It might be better to use "optional partial methods", and "required partial methods", respectively. I'm not sold on those either. Other thoughts?

standard/classes.md
```diff
@@ -3026,21 +3027,53 @@ The mechanism by which linkage to an external method is achieved is implementati
...
+In *method_declaration*, the identifier `partial` is recognized as a contextual keyword ([§6.4.4](lexical-structure.md#644-keywords)) only if it immediately precedes the *return_type*. A partial method cannot explicitly implement interface methods.
 
-There are two kinds of partial method declarations: If the body of the method declaration is a semicolon, the declaration is said to be a ***defining partial method declaration***. If the body is other than a semicolon, the declaration is said to be an ***implementing partial method declaration***. Across the parts of a type declaration, there shall be only one defining partial method declaration with a given signature, and there shall be at most only one implementing partial method declaration with a given signature. If an implementing partial method declaration is given, a corresponding defining partial method declaration shall exist, and the declarations shall match as specified in the following:
+There are classifications of a partial method declaration:
```
```suggestion
Partial method declarations are classified as follows:
```

(Or something else - I'm just not keen on the current wording.)

standard/classes.md
```diff
@@ -3026,21 +3027,53 @@ The mechanism by which linkage to an external method is achieved is implementati
...
 
-- The declarations shall have the same modifiers (although not necessarily in the same order), method name, number of type parameters and number of parameters.
-- Corresponding parameters in the declarations shall have the same modifiers (although not necessarily in the same order) and the same types, or identity convertible types (modulo differences in type parameter names).
+- A method with an *expression-body* or a *block-body* or is declared with the `extern` modifier is said to be an ***implementing partial method declaration***.
+- Otherwise, a method declaration where the body of the method declaration is a semicolon is said to be a ***defining partial method declaration***.
+
+Across the parts of a type declaration, there shall be exactly one defining partial method declaration with a given signature. For an implementing partial method declaration is given, a corresponding defining partial method declaration shall exist, and the declarations shall match as specified in the following:
```
```suggestion
Across the parts of a type declaration, there shall be exactly one defining partial method declaration with a given signature. If an implementing partial method declaration is given, a corresponding defining partial method declaration shall exist, and the declarations shall match as specified in the following:
```

standard/classes.md
```diff
@@ -3026,21 +3027,53 @@ The mechanism by which linkage to an external method is achieved is implementati
...
+
+- The declarations shall have the same method name, number of type parameters, and number of parameters.
+- The declarations shall have the same modifiers except for the `async` and `extern` modifiers. The `async` and `extern` modifiers are allowed only on the implementing partial method declaration.
+- Corresponding parameters in the declarations shall have the same modifiers (although not necessarily in the same order) and the same types (modulo differences in type parameter names).
```
jskeet: Not sure about the type parameter names - do we have an example of this?

I can't remember what we say elsewhere about tuple types and "sameness" - is it valid to have:

```csharp
partial void M((int x, string y) tuple);

partial void M((int a, string b) otherTuple) { ... }
```

?

standard/classes.md
```diff
@@ -3026,21 +3027,53 @@ The mechanism by which linkage to an external method is achieved is implementati
...
 - Corresponding type parameters in the declarations shall have the same constraints (modulo differences in type parameter names).
 
-An implementing partial method declaration can appear in the same part as the corresponding defining partial method declaration.
+There are two versions of partial methods: restricted and unrestricted. A ***restricted partial method*** (§restricted-partial-methods) has no explicit access modifiers, and is implicitly private. An ***unrestricted partial method*** (§unrestricted-partial-methods) is a partial method that includes one or more explicit access modifiers.
```
jskeet: I'd prefer to avoid the word "version" here. Maybe this is another "classification" (as well as defining/implementing)?

standard/classes.md
```diff
@@ -3064,25 +3097,12 @@ Only a defining partial method participates in overload resolution. Thus, whethe
 >
 > *end note*
 
-If no part of a partial type declaration contains an implementing declaration for a given partial method, any expression statement invoking it is simply removed from the combined type declaration. Thus the invocation expression, including any subexpressions, has no effect at run-time. The partial method itself is also removed and will not be a member of the combined type declaration.
-
 If an implementing declaration exists for a given partial method, the invocations of the partial methods are retained. The partial method gives rise to a method declaration similar to the implementing partial method declaration except for the following:
 
-- The `partial` modifier is not included.
-
+- All modifiers except the `partial` modifier are included in the resulting, including any `extern` or `async` modifier declared on the implementing declaration.
```
```suggestion
- All modifiers except the `partial` modifier are included in the resulting method declaration, including any `extern` or `async` modifier declared on the implementing declaration.
```

Separately, it's not clear to me what the impact of this is.

standard/classes.md
```diff
@@ -3026,21 +3027,53 @@ The mechanism by which linkage to an external method is achieved is implementati
...
-- Corresponding parameters in the declarations shall have the same modifiers (although not necessarily in the same order) and the same types, or identity convertible types (modulo differences in type parameter names).
+- A method with an *expression-body* or a *block-body* or is declared with the `extern` modifier is said to be an ***implementing partial method declaration***.
+- Otherwise, a method declaration where the body of the method declaration is a semicolon is said to be a ***defining partial method declaration***.
+
+Across the parts of a type declaration, there shall be exactly one defining partial method declaration with a given signature. For an implementing partial method declaration is given, a corresponding defining partial method declaration shall exist, and the declarations shall match as specified in the following:
+
+- The declarations shall have the same method name, number of type parameters, and number of parameters.
+- The declarations shall have the same modifiers except for the `async` and `extern` modifiers. The `async` and `extern` modifiers are allowed only on the implementing partial method declaration.
```
jskeet, line 3044: How about readonly? I think it would make sense for an implementation to potentially be readonly even if the declaration isn't - and in that case, I'd expect the readonly-ness to be available elsewhere in the code.

standard/classes.md
```diff
@@ -3166,6 +3186,31 @@ class Customer
...
+If a defining declaration but not an implementing declaration is given for a restricted partial method `M`, the following restrictions apply:
+
+- It is a compile-time error to create a delegate from `M` ([§12.8.17.5](expressions.md#128175-delegate-creation-expressions)).
+- It is a compile-time error to refer to `M` inside an anonymous function that is converted to an expression tree type ([§8.6](types.md#86-expression-tree-types)).
+- Expressions occurring as part of an invocation of `M` do not affect the definite assignment state ([§9.4](variables.md#94-definite-assignment)), which can potentially lead to compile-time errors.
+- `M` cannot be the entry point for an application ([§7.1](basic-concepts.md#71-application-startup)).
+
+#### §unrestricted-partial-methods Unrestricted partial methods
+
+An unrestriced partial method declartion incluces an explicit access modifier. There shall be exactly one implementing partial method declaration.
```
```suggestion
An unrestricted partial method declaration incluces an explicit access modifier. There shall be exactly one implementing partial method declaration.
```

standard/classes.md
BillWagner: It might make sense, but it is not allowed. If the `readonly` modifier is on a partial member of a struct type, it must be on both the declaration and the implementation.

That is true for both the defining and implementing declarations.

standard/classes.md
BillWagner: Notes: 

- item names for tuples in both partial members must match.
- Different type parameter names generates a warning.
- Different type parameter constraints is an error (including not stating them on the implementation)

standard/classes.md
```diff
@@ -2151,15 +2151,16 @@ A declaration has a valid combination of modifiers if all of the following are t
...
 
 Methods are classified according to what, if anything, they return:
 
 - If `ref` is present, the method is ***returns-by-ref*** and returns a *variable reference*, that is optionally read-only;
 - Otherwise, if *return_type* is `void`, the method is ***returns-no-value*** and does not return a value;
 - Otherwise, the method is ***returns-by-value*** and returns a value.
 
-The *return_type* of a returns-by-value or returns-no-value method declaration specifies the type of the result, if any, returned by the method. Only a returns-no-value method may include the `partial` modifier ([§15.6.9](classes.md#1569-partial-methods)). If the declaration includes the `async` modifier then *return_type* shall be `void` or the method returns-by-value and the return type is a *task type* ([§15.14.1](classes.md#15141-general)).
+The *return_type* of a returns-by-value or returns-no-value method declaration specifies the type of the result, if any, returned by the method. If the declaration includes the `async` modifier then *return_type* shall be `void` or the method returns-by-value and the return type is a *task type* ([§15.14.1](classes.md#15141-general)).
```
```suggestion
The *return_type* of a returns-by-value or returns-no-value method declaration specifies the type of the result, if any, returned by the method. If the declaration includes the `async` modifier then *return_type* shall be `void` or the method returns-by-value and the return type shall be a *task type* ([§15.14.1](classes.md#15141-general)).
```

standard/classes.md
```diff
@@ -3026,22 +3027,53 @@ The mechanism by which linkage to an external method is achieved is implementati
...
-- Corresponding parameters in the declarations shall have the same modifiers (although not necessarily in the same order) and the same types, or identity convertible types (modulo differences in type parameter names).
-- Corresponding type parameters in the declarations shall have the same constraints (modulo differences in type parameter names).
+- The declarations shall have the same modifiers except for the `async` and `extern` modifiers. The `async` and `extern` modifiers are allowed only on the implementing partial method declaration.
+- Corresponding parameters in the declarations shall have the same modifiers (although not necessarily in the same order) and the same types (modulo differences in type parameter names). Tuple types (§8.3.11) used as parameters or return types shall have the same item names in both the defining and implementing partial method declarations.
+- Corresponding type parameters in the declarations shall have the same constraints. An implementation may choose to issue a warning if the type parameter names are different in the defining and implementing declarations.
```
jskeet: Do we need this bit about warnings? (I think we generally say implementations can issue warnings whenever they want.)

standard/classes.md
```diff
@@ -3026,22 +3027,53 @@ The mechanism by which linkage to an external method is achieved is implementati
...
 
-An implementing partial method declaration can appear in the same part as the corresponding defining partial method declaration.
+There are two classifications of partial methods: required and optional. A ***required partial method*** (§required-partial-methods) has no explicit access modifiers, and is implicitly private. An ***optional partial method*** (§optional-partial-methods) is a partial method that includes one or more explicit access modifiers.
 
-Only a defining partial method participates in overload resolution. Thus, whether or not an implementing declaration is given, invocation expressions may resolve to invocations of the partial method. Because a partial method always returns `void`, such invocation expressions will always be expression statements. Furthermore, because a partial method is implicitly `private`, such statements will always occur within one of the parts of the type declaration within which the partial method is declared.
+For an optional partial method both the definition and implementation shall exist.
```
```suggestion
For a required partial method both the definition and implementation shall exist.
```

I'm hoping, anyway. It's pretty weird if "required" means "there doesn't have to be an implementation" and "optional" means "there does have to be an implementation".

standard/classes.md
```diff
@@ -3026,22 +3027,53 @@ The mechanism by which linkage to an external method is achieved is implementati
...
+- Corresponding parameters in the declarations shall have the same modifiers (although not necessarily in the same order) and the same types (modulo differences in type parameter names). Tuple types (§8.3.11) used as parameters or return types shall have the same item names in both the defining and implementing partial method declarations.
+- Corresponding type parameters in the declarations shall have the same constraints. An implementation may choose to issue a warning if the type parameter names are different in the defining and implementing declarations.
 
-An implementing partial method declaration can appear in the same part as the corresponding defining partial method declaration.
+There are two classifications of partial methods: required and optional. A ***required partial method*** (§required-partial-methods) has no explicit access modifiers, and is implicitly private. An ***optional partial method*** (§optional-partial-methods) is a partial method that includes one or more explicit access modifiers.
```
jskeet: Let's use a different word than "classification" here - given that we've previously said "Partial method declarations are classified as follows" (and then talked about implementing vs defining).

Or we might want to put both "axes" of classification together, to make it clear that they're somewhat orthogonal.

standard/classes.md
```diff
@@ -3167,6 +3186,31 @@ class Customer
 }
 ```
 
+#### §required-partial-methods Required partial methods
+
+A required partial method shall have a `void` return type, and shall not declare out parameters. There shall be zero or one implementing declaration for each defining declaration. If no part implements the partial method, the partial method declaration and all calls to it are removed from the type declaration resulting from the combination of the parts. Whether or not an implementing declaration is given, invocation expressions may resolve to invocations of the partial method.
```
jskeet: Again, this looks like it's the wrong way round. This sounds like it's an optional partial method, given that it's only optionally implemented.

jskeet (commented):
Overall - either optional and required are the wrong way round here, or I think we should find another pair of terms. Sorry to throw a spanner in the works!

standard/classes.md
BillWagner: I tried "variation". I was trying to avoid "type"

standard/classes.md
BillWagner: I switched them when I did the global search / replace. Fixed in next commit.

BillWagner:
Thanks!

BillWagner:
This is ready for a first review at the next committee meeting.

@Nigel-Ecma I think the ANTLR tests are failing because this PR allows new partial declarations.

Nigel-Ecma:
> This is ready for a first review at the next committee meeting.
> 
> @Nigel-Ecma I think the ANTLR tests are failing because this PR allows new partial declarations.

@BillWagner – A quick glance suggests the two tests fail because the *structure* of existing syntax rules has changed, i.e. `'partial'?` has been moved up from *method_modifiers* to *method_declaration* – the recognised syntax for non-ref methods remain the same but the tree-structure changes.

As far as my quick glance goes you’ve done this so that ref-methods can be partial as well. There is nothing wrong with this *per se*, it is a choice, but if you want to achieve the same result without breaking anything just add `'partial'?` to `ref_method_modifers` – adding an extra optional child (`'partial'?`) in a rule doesn’t change its structure, so existing tests should continue without error (but you can add a new test for a partial ref method to make sure partial ref methods are recognised as expected).

HTH

standard/basic-concepts.md
```diff
@@ -12,7 +12,7 @@ A program compiled as an application shall contain at least one method qualifyin
 - It shall be declared in a non-generic type. If the type declaring the method is a nested type, none of its enclosing types may be generic.
 - It may have the `async` modifier provided the method’s return type is `System.Threading.Tasks.Task` or `System.Threading.Tasks.Task<int>`.
 - The return type shall be `void`, `int`, `System.Threading.Tasks.Task`, or `System.Threading.Tasks.Task<int>`.
-- It shall not be a partial method ([§15.6.9](classes.md#1569-partial-methods)) without an implementation.
+- It shall not be an optional partial method (§optional-partial-methods) without an implementing declaration.
```
Nigel-Ecma, line 15: @BillWagner – Just a heads up that #1454 moves this line to ~38.

standard/classes.md
```diff
@@ -2059,13 +2059,13 @@ A ***method*** is a member that implements a computation or action that can be p
 
 ```ANTLR
 method_declaration
-    : attributes? method_modifiers return_type method_header method_body
-    | attributes? ref_method_modifiers ref_kind ref_return_type method_header
+    : attributes? method_modifiers 'partial'? return_type method_header method_body
+    | attributes? ref_method_modifiers 'partial'? ref_kind ref_return_type method_header
```
Nigel-Ecma, line 2063: Group change: 1 of 3, do all or none!

As per [this comment](https://github.com/dotnet/csharpstandard/pull/1468#issuecomment-5334979247) I suggest reverting this change:


```suggestion
    : attributes? method_modifiers return_type method_header method_body
    | attributes? ref_method_modifiers ref_kind ref_return_type method_header
```

standard/classes.md
```diff
@@ -2059,13 +2059,13 @@ A ***method*** is a member that implements a computation or action that can be p
...
       ref_method_body
     ;
 
 method_modifiers
-    : method_modifier* 'partial'?
+    : method_modifier*
```
Nigel-Ecma, line 2068: Group change: 2 of 3, do all or none!

As per [this comment](https://github.com//pull/1468#issuecomment-5334979247) I suggest reverting this change:

```suggestion
    : method_modifier* 'partial'?
```

standard/classes.md
Nigel-Ecma, line 2077: Group change: 3 of 3, do all or none!

As per [this comment](https://github.com//pull/1468#issuecomment-5334979247) I suggest adding this change:


```suggestion
    : ref_method_modifier* 'partial'?
```

standard/classes.md
```diff
@@ -3026,22 +3027,54 @@ The mechanism by which linkage to an external method is achieved is implementati
...
+> // part containing implementing partial method declarations
+> partial class C
+> {
+>     private partial void M2() { ... }
+>     protected partial bool M3() { ... }
+>     public partial void M4(out int i) { ... }
+> }
+> ```
+>
+> *end example*
```
Nigel-Ecma, line 3075: I suggest adding a `M5` which uses `extern`, thus showing an implementing declaration with a semicolon rather than a block.

Nigel-Ecma (changes requested):
Some suggested small changes, including a grammar rewrite previously suggested in [this comment](https://github.com/dotnet/csharpstandard/pull/1468#issuecomment-5334979247)

PR #1469, opened 2025-11-11
[Version 9.0] Feature support for target typed `new`

BillWagner:
This PR replaces #990 

- The description in 990 lists [potential additional work](https://github.com/dotnet/csharpstandard/pull/990#issue-1989557014).

The following comments on 990 haven't been addressed:

- https://github.com/dotnet/csharpstandard/pull/990#discussion_r1390480900
- https://github.com/dotnet/csharpstandard/pull/990#discussion_r1390483550
- https://github.com/dotnet/csharpstandard/pull/990#discussion_r1390487271
-

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -390,6 +390,12 @@ While throw expressions do not have a type, they may be implicitly converted to 
 
 There is an implicit conversion from a *switch_expression* ([§12.11](expressions.md#1211-switch-expression)) to every type `T` for which there exists an implicit conversion from each *switch_expression_arm*’s *switch_expression_arm_expression*’s to `T`.
 
+### §imp-obj-creation-conv Implicit object-creation conversions
+
+There is an implicit ***object-creation conversion*** from a *target_typed_new* expression ([§12.8.17.2](expressions.md#128172-object-creation-expressions)) to every type.
+
+Given a target type `T`, if `T` is an instance of `System.Nullable`, the type `T0` is `T`'s underlying type. Otherwise `T0` is `T`. The meaning of a *target_typed_new* expression that is converted to the type `T` is the same as the meaning of a corresponding *object_creation_expression* that specifies `T0` as the type.
+
 ## 10.3 Explicit conversions
 
 ### 10.3.1 General
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -2524,6 +2524,11 @@ An *object_creation_expression* is used to create a new instance of a *class_typ
 object_creation_expression
     : 'new' type '(' argument_list? ')' object_or_collection_initializer?
     | 'new' type object_or_collection_initializer
+    | target_typed_new
+    ;
+
+target_typed_new
+    : 'new' '(' argument_list? ')' object_or_collection_initializer?
     ;
 
 object_or_collection_initializer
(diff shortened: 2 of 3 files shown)


standard/expressions.md
```diff
@@ -2566,7 +2575,7 @@ Processing of an object creation expression that includes an object initializer
 
 If any of the arguments in the optional *argument_list* has the compile-time type `dynamic` then the *object_creation_expression* is dynamically bound ([§12.3.3](expressions.md#1233-dynamic-binding)) and the following rules are applied at run-time using the run-time type of those arguments of the *argument_list* that have the compile-time type `dynamic`. However, the object creation undergoes a limited compile-time check as described in [§12.6.5](expressions.md#1265-compile-time-checking-of-dynamic-member-invocation).
 
-The binding-time processing of an *object_creation_expression* of the form `new T(A)`, where `T` is a *class_type*, or a *value_type*, and `A` is an optional *argument_list*, consists of the following steps:
+The binding-time processing of an *object_creation_expression* of the form `new T(A)`, where the specified or implied type `T` is a *class_type*, or a *value_type*, and `A` is an optional *argument_list*, consists of the following steps:
```
jskeet, line 2554: A target-typed new expression *isn't* "of the form `new T(A)`... I think this can probably be fixed by saying something along the lines of "or a *target_typed_new* expression with an implied type `T`".

standard/expressions.md
```diff
@@ -2581,7 +2590,7 @@ The binding-time processing of an *object_creation_expression* of the form `new
 
 Even if the *object_creation_expression* is dynamically bound, the compile-time type is still `T`.
 
-The run-time processing of an *object_creation_expression* of the form new `T(A)`, where `T` is *class_type* or a *struct_type* and `A` is an optional *argument_list*, consists of the following steps:
+The run-time processing of an *object_creation_expression* of the form `new T(A)`, where the specified or implied type `T` is *class_type* or a *struct_type* and `A` is an optional *argument_list*, consists of the following steps:
```
jskeet, line 2569: Another example of the same kind of thing.

standard/conversions.md
```diff
@@ -384,6 +384,12 @@ While throw expressions do not have a type, they may be implicitly converted to
 
 There is an implicit conversion from a *switch_expression* ([§12.11](expressions.md#1211-switch-expression)) to every type `T` for which there exists an implicit conversion from each *switch_expression_arm*’s *switch_expression_arm_expression*’s to `T`.
 
+### §imp-obj-creation-conv Implicit object-creation conversions
+
+There is an implicit ***object-creation conversion*** from a *target_typed_new* expression ([§12.8.17.2](expressions.md#128172-object-creation-expressions)) to every type.
```
jskeet, line 395: Does this include pointer types?

standard/expressions.md
jskeet, line 2554: (Existing potential problem) Should this include `object_or_collection_initializer?`? It's *sort of* described earlier, but it's not really clear. We could potentially *make* it clear by saying that processing an expression which contains an object or collection initializer is equivalent to processing an expression without it, and then executing the object/collection initializer. Although I suspect init-only properties might make that tricky.

If it's *not* intended to exclude that case, why use "of the form" at all?

standard/expressions.md
```diff
@@ -2558,6 +2563,10 @@ object_or_collection_initializer
 
 The *type* of an *object_creation_expression* shall be a *class_type*, a *value_type*, or a *type_parameter*. The *type* cannot be a *tuple_type* or an abstract or static *class_type*.
 
+If `type` can be inferred from usage, it can be omitted, as allowed by *target_typed_new*. It is a compile-time error to omit `type` if the type cannot be inferred. A *target_typed_new* expression has no type. However, there is an implicit object-creation conversion (§imp-obj-creation-conv) from a *target_typed_new* expression to every type. It is a compile-time error if a *target_typed_new* is used as an operand of a unary or binary operator, or if it is used where it is not subject to an object-creation conversion.
```
jskeet, line 2542: "If `type` can be inferred from usage" is too vague here.

For example, if I have:

```csharp
void M(int x) { ... }
void M(Guid g) { ... }
```

... then can I write `M(new(new byte[16]))` as `Guid` has a constructor that accepts a byte array, but `int` doesn't?

jskeet: (I strongly suspect that clarifying this will be the biggest change required for this feature. I don't know whether we'll need to consider *everywhere* that target_typed_new can exist...)

jskeet (commented):
I think this is ready for further discussion in the group. I can then take any feedback and work on changes to the PR.

standard/conversions.md
BillWagner: No, because a pointer type can't be used in an object creation expression. See https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#1281721-general

standard/expressions.md
BillWagner: This one will be tough. Your example generates CS0121: The call is ambiguous between `M(Guid)` and `M(int)`. That's weird, because `int y = new(new byte[16]))` correctly doesn't compile. 

The example in the speclet shows a method with several overloads, `XmlReader.Create`. But, in that case, all the overloads have the same type as the 2nd positional parameter, so it's obvious what to create.

Now, if you use this: `M(g: new(new byte[16]));`, it compiles (note the named parameter). If you change the definition of both `M` methods so that the parameter name is the same, it's ambiguous again.

BillWagner (commented):
Initial review / comments.

BillWagner:
Thanks!

PR #1470, opened 2025-11-11
[Version 9.0] Feature support for unconstrained type parameter annotations

BillWagner:
This PR replaces #1326

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -432,7 +432,9 @@ The list of constraints given in a `where` clause can include any of the followi
 
 A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, the ***not null constraint*** `notnull`, the ***unmanaged type constraint*** `unmanaged`, or `default`. The class type and the reference type constraint can include the *nullable_type_annotation*.
 
-It is a compile-time error to use a `default` constraint other than on a method override or explicit implementation. It is a compile-time error to use a `default` constraint when the corresponding type parameter in the overridden or interface method is constrained to a reference type or value type.
+It is a compile-time error to use a `default` constraint other than on a method override or explicit interface member implementation. It is a compile-time error to use a `default` constraint when the corresponding type parameter in the overridden or interface method is constrained to a reference type or non-nullable value type.
+
+> *Note*: A class type constraint such as `where T : Stream` constrains the type parameter to a reference type; the `default` constraint may not be used when such a constraint is inherited. *end note*
 
 A secondary constraint can be an *interface_type* or *type_parameter*, optionally followed by a *nullable_type_annotation*. The presence of the *nullable_type_annotation* indicates that the type argument is allowed to be the nullable reference type that corresponds to a non-nullable reference type that satisfies the constraint.
 
@@ -452,7 +454,7 @@ Except when a type parameter is explicitly constrained to value types, the nulla
 For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`.
 
 > *Note*: On a type parameter, a return type `T?` has the same nullability effect as `[MaybeNull] T`, and a parameter type `T?` has the same nullability effect as `[AllowNull] T`. *end note*
-!-- markdownlint-disable MD028 -->
+<!-- markdownlint-disable MD028 -->
 
 <!-- markdownlint-enable MD028 -->
 > *Example*: The following examples show how the nullability of a type argument impacts the nullability of a declaration of its type parameter:
@@ -539,7 +541,7 @@ The unmanaged type constraint specifies that a type argument used for the type p
 
 Because `unmanaged` is not a keyword, in *primary_constraint* the unmanaged constraint is always syntactically ambiguous with *class_type*. For compatibility reasons, if a name lookup ([§12.8.4](expressions.md#1284-simple-names)) of the name `unmanaged` succeeds it is treated as a `class_type`. Otherwise it is treated as the unmanaged constraint.
 
-The `default` constraint applies to a type parameter whose inherited constraints do not establish it as a reference type or a value type; its effect on `T?` in override methods and explicit interface method implementations is specified in [§15.6.5](classes.md#1565-override-methods) and [§19.6.2](interfaces.md#1962-explicit-interface-member-implementations).
+The `default` constraint may be used for a type parameter whose inherited constraints do not establish it as a reference type or a value type; its effect on `T?` in override methods and explicit interface member implementations is specified in [§15.6.5](classes.md#1565-override-methods) and [§19.6.2](interfaces.md#1962-explicit-interface-member-implementations).
 
 Pointer types are never allowed to be type arguments, and do not satisfy any type constraints, even unmanaged, despite being unmanaged types.
 
@@ -2764,7 +2766,7 @@ A compile-time error occurs unless all of the following are true for an override
 - The overridden base method is not a sealed method.
 - There is an identity conversion between the return type of the overridden base method and the override method.
 - The override declaration and the overridden base method have the same declared accessibility. In other words, an override declaration cannot change the accessibility of the virtual method. However, if the overridden base method is protected internal and it is declared in a different assembly than the assembly containing the override declaration then the override declaration’s declared accessibility shall be protected.
-- A *type_parameter_constraints_clause* may only consist of the `class`, `struct`, or `default` *primary_constraint*s. The `class` and `struct` constraints are applied to *type_parameter*s which are known according to the inherited constraints to be either reference or value types respectively. The `default` constraint is applied to *type_parameter*s that are not constrained to either reference or value types. Any type of the form `T?` in the overriding method’s signature, where `T` is a type parameter, is interpreted as follows:
+- A *type_parameter_constraints_clause* may only consist of the `class`, `struct`, or `default` *primary_constraint*s. The `class` and `struct` constraints may be placed on *type_parameter*s known according to the inherited constraints to be either reference or non-nullable value types respectively. The `default` constraint may be placed on *type_parameter*s that are not constrained to either reference or value types. Any type of the form `T?` in the overriding method's signature, where `T` is a type parameter, is interpreted as follows:
   - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type.
   - If either a `struct` constraint is added, or no constraint is added and the inherited constraint is a value type constraint, for the type parameter `T` then `T?` is a nullable value type.
   - If a `default` constraint is added for type parameter `T` then `T?` represents a nullable instance of the corresponding reference type when `T` is a reference type, and an instance of `T` when `T` is a value type. If `T` is substituted with an annotated type `U?`, then `T?` represents `U?`, not `U??`.
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -434,7 +434,7 @@ A primary constraint can be a class type, the ***reference type constraint*** `c
 
 It is a compile-time error to use a `default` constraint other than on a method override or explicit interface member implementation. It is a compile-time error to use a `default` constraint when the corresponding type parameter in the overridden or interface method is constrained to a reference type or non-nullable value type.
 
-> *Note*: A class type constraint such as `where T : Stream` constrains the type parameter to a reference type; the `default` constraint may not be used when such a constraint is inherited. *end note*
+> *Note*: A class type constraint such as `where T : Stream` constrains the type parameter to a reference type; the `default` constraint may not be used with such a constraint. *end note*
 
 A secondary constraint can be an *interface_type* or *type_parameter*, optionally followed by a *nullable_type_annotation*. The presence of the *nullable_type_annotation* indicates that the type argument is allowed to be the nullable reference type that corresponds to a non-nullable reference type that satisfies the constraint.
 
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -452,7 +452,9 @@ Except when a type parameter is explicitly constrained to value types, the nulla
 For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`.
 
 > *Note*: On a type parameter, a return type `T?` has the same nullability effect as `[MaybeNull] T`, and a parameter type `T?` has the same nullability effect as `[AllowNull] T`. *end note*
+!-- markdownlint-disable MD028 -->
 
+<!-- markdownlint-enable MD028 -->
 > *Example*: The following examples show how the nullability of a type argument impacts the nullability of a declaration of its type parameter:
 >
 > <!-- Example: {template:"standalone-lib-without-using", name:"RepeatedNullable"} -->
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -2769,7 +2769,10 @@ A compile-time error occurs unless all of the following are true for an override
 - A *type_parameter_constraints_clause* may only consist of the `class`, `struct`, or `default` *primary_constraint*s. The `class` and `struct` constraints may be placed on *type_parameter*s known according to the inherited constraints to be either reference or non-nullable value types respectively. The `default` constraint may be placed on *type_parameter*s that are not constrained to either reference or value types. Any type of the form `T?` in the overriding method's signature, where `T` is a type parameter, is interpreted as follows:
   - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type.
   - If either a `struct` constraint is added, or no constraint is added and the inherited constraint is a value type constraint, for the type parameter `T` then `T?` is a nullable value type.
-  - If a `default` constraint is added for type parameter `T` then `T?` represents a nullable instance of the corresponding reference type when `T` is a reference type, and an instance of `T` when `T` is a value type. If `T` is substituted with an annotated type `U?`, then `T?` represents `U?`, not `U??`.
+  - If a `default` constraint is added for type parameter `T`, the interpretation of `T?` depends on the type argument substituted for `T`:
+    - If `T` is substituted with a reference type `C`, `T?` represents the nullable reference type `C?`.
+    - If `T` is substituted with an already-annotated reference type `C?`, `T?` represents `C?`, not `C??`.
+    - If `T` is substituted with a value type, `T?` represents `T` itself, not `Nullable<T>`.
 
 > *Example*: The following demonstrates how the overriding rules work for generic classes:
 >
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -447,11 +447,9 @@ The nullability of the type argument need not match the nullability of the type 
 
 > *Note*: To specify that a type argument is a nullable reference type, do not add the nullable type annotation as a constraint (use `T : class` or `T : BaseClass`), but use `T?` throughout the generic declaration to indicate the corresponding nullable reference type for the type argument. *end note*
 
-<!-- Remove in C# 9, when `?` is allowed on any type parameter. -->
-The nullable type annotation, `?`, can only be used on a type parameter that has the value type constraint, the reference type constraint without the *nullable_type_annotation*, or a class type constraint without the *nullable_type_annotation*.
+Unless a type parameter is explicitly constrained to value types, the nullable type annotation `?` can only be applied to a type parameter within a `#nullable enable` context.
 
-<!-- Add in C# 9, when `?` is allowed on nullable reference type parameters. -->
-<!-- For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`. -->
+For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`.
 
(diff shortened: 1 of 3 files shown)


BillWagner:
I can't add a suggestion here yet, but this text needs to be removed from types.md:

> - Nullable annotations (`?`) aren't allowed on an instance of a type parameter unless that type parameter is constrained to be either a reference type or a value type.  

It's being proposed in #1575.  It's no longer correct for v9.

BillWagner:
Thanks!

BillWagner:
@jskeet If there is time after other agenda items, this is close and could use comments from other committee members.

jskeet:
Yup, I strongly suspect we'll have plenty of time.

jskeet:
Aim is get approval (or lack thereof) of general shape and first draft comments, so @BillWagner can make changes ready for a second round of review before the next meeting. Added all team members to review.

standard/classes.md
```diff
@@ -429,7 +430,9 @@ Each *type_parameter_constraints_clause* consists of the token `where`, followed
 
 The list of constraints given in a `where` clause can include any of the following components, in this order: a single primary constraint, one or more secondary constraints, and the constructor constraint, `new()`.
 
-A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, the ***not null constraint*** `notnull` or the ***unmanaged type constraint*** `unmanaged`. The class type and the reference type constraint can include the *nullable_type_annotation*.
+A primary constraint can be a class type, the ***reference type constraint*** `class`, the ***value type constraint*** `struct`, the ***not null constraint*** `notnull`, the ***unmanaged type constraint*** `unmanaged`, or `default`. The class type and the reference type constraint can include the *nullable_type_annotation*.
+
+It is a compile-time error to use a `default` constraint other than on a method override or explicit implementation. It is a compile-time error to use a `default` constraint when the corresponding type parameter in the overridden or interface method is constrained to a reference type or value type.
```
jskeet, line 435: Perhaps "explicit implementation" => "explicit interface implementation"? Not really required, just a thought.

jskeet, line 435: Just to check, "is constrained to a reference type or value type" includes a class type constraint, right? (So I can't write `where T : Stream, default`.

It's all slightly confusing that there's the reference type constraint which is spelled `class` which is different from a class type constraint which itself constrains it to be a reference type... not that we can do anything about that now.

standard/classes.md
```diff
@@ -444,12 +447,14 @@ The nullability of the type argument need not match the nullability of the type
 
 > *Note*: To specify that a type argument is a nullable reference type, do not add the nullable type annotation as a constraint (use `T : class` or `T : BaseClass`), but use `T?` throughout the generic declaration to indicate the corresponding nullable reference type for the type argument. *end note*
 
-<!-- Remove in C# 9, when `?` is allowed on any type parameter. -->
-The nullable type annotation, `?`, can only be used on a type parameter that has the value type constraint, the reference type constraint without the *nullable_type_annotation*, or a class type constraint without the *nullable_type_annotation*.
+Except when a type parameter is explicitly constrained to value types, the nullable type annotation `?` can only be applied to a type parameter when the nullable annotations flag is enabled ([§6.5.9](lexical-structure.md#659-nullable-directive)).
```
jskeet, line 450: Constrained to non-nullable value types?

standard/classes.md
```diff
@@ -534,6 +539,8 @@ The unmanaged type constraint specifies that a type argument used for the type p
 
 Because `unmanaged` is not a keyword, in *primary_constraint* the unmanaged constraint is always syntactically ambiguous with *class_type*. For compatibility reasons, if a name lookup ([§12.8.4](expressions.md#1284-simple-names)) of the name `unmanaged` succeeds it is treated as a `class_type`. Otherwise it is treated as the unmanaged constraint.
 
+The `default` constraint applies to a type parameter whose inherited constraints do not establish it as a reference type or a value type; its effect on `T?` in override methods and explicit interface method implementations is specified in [§15.6.5](classes.md#1565-override-methods) and [§19.6.2](interfaces.md#1962-explicit-interface-member-implementations).
```
jskeet, line 542: "applies to a type parameter" feels a bit odd here. Is it actually "can be applied to a type parameter" or have I missed something?

jskeet: And possibly remove "method" from "explicit interface method implementations"?

standard/classes.md
```diff
@@ -2757,9 +2764,10 @@ A compile-time error occurs unless all of the following are true for an override
...
-- A *type_parameter_constraints_clause* may only consist of the `class` or `struct` *primary_constraint*s applied to *type_parameter*s which are known according to the inherited constraints to be either reference or value types respectively. Any type of the form `T?` in the overriding method’s signature, where `T` is a type parameter, is interpreted as follows:
-  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
-  - If either there is no added constraint, or a `struct` constraint is added, for the type parameter `T` then `T?` is a nullable value type.
+- A *type_parameter_constraints_clause* may only consist of the `class`, `struct`, or `default` *primary_constraint*s. The `class` and `struct` constraints are applied to *type_parameter*s which are known according to the inherited constraints to be either reference or value types respectively. The `default` constraint is applied to *type_parameter*s that are not constrained to either reference or value types. Any type of the form `T?` in the overriding method’s signature, where `T` is a type parameter, is interpreted as follows:
```
jskeet, line 2767: Again the "are applied" and "is applied" feel slightly confusing to me.

standard/classes.md
```diff
@@ -2757,9 +2764,10 @@ A compile-time error occurs unless all of the following are true for an override
...
+  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type.
+  - If either a `struct` constraint is added, or no constraint is added and the inherited constraint is a value type constraint, for the type parameter `T` then `T?` is a nullable value type.
+  - If a `default` constraint is added for type parameter `T` then `T?` represents a nullable instance of the corresponding reference type when `T` is a reference type, and an instance of `T` when `T` is a value type. If `T` is substituted with an annotated type `U?`, then `T?` represents `U?`, not `U??`.
```
jskeet, line 2770: I don't understand this bullet at the moment.

standard/classes.md
```diff
@@ -2809,6 +2817,28 @@ A compile-time error occurs unless all of the following are true for an override
...
+> }
+>
+> class B2 : A2
+> {
+>     public override void F2<T>(T? t) /*where T : struct*/ { }
+>     public override void F2<T>(T? t) where T : default { }
+> }
+> ```
+>
+> The `default` constraint on `B2.F2` is required to override the unconstrained `A2.F2` with a `T?` parameter. Without it, `T?` in the second override would be interpreted as a nullable value type, so the declaration would not override the unconstrained `A2.F2`. *end example*
```
jskeet, line 2841: This is a really good example - but wow this is a confusing bit of the language. (At least for me...)

I'm hoping this is a corner case that needs to be specified because that's what we do, but which most people never run into.

standard/interfaces.md
```diff
@@ -855,10 +855,11 @@ It is a compile-time error for an explicit interface member implementation to in
 
 An explicit interface method implementation inherits any type parameter constraints from the interface.
 
-A *type_parameter_constraints_clause* on an explicit interface method implementation may only consist of the `class` or `struct` *primary_constraint*s applied to *type_parameter*s which are known according to the inherited constraints to be either reference or value types respectively. Any type of the form `T?` in the signature of the explicit interface method implementation, where `T` is a type parameter, is interpreted as follows:
+A *type_parameter_constraints_clause* on an explicit interface method implementation may only consist of the `class`, `struct`, or `default` *primary_constraint*s. The `class` and `struct` constraints are applied to *type_parameter*s which are known according to the inherited constraints to be either reference or value types respectively. The `default` constraint is applied to *type_parameter*s that are not constrained to either reference or value types. Any type of the form `T?` in the signature of the explicit interface method implementation, where `T` is a type parameter, is interpreted as follows:
```
jskeet, line 858: Again, "is applied" feels confusing to me.

standard/interfaces.md
```diff
@@ -855,10 +855,11 @@ It is a compile-time error for an explicit interface member implementation to in
...
 
-- If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
-- If either there is no added constraint, or a `struct` constraint is added, for the type parameter `T` then `T?` is a nullable value type.
+- If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type.
+- If either a `struct` constraint is added, or no constraint is added and the inherited constraint is a value type constraint, for the type parameter `T` then `T?` is a nullable value type.
+- If a `default` constraint is added for type parameter `T` then `T?` represents a nullable instance of the corresponding reference type when `T` is a reference type, and an instance of `T` when `T` is a value type. If `T` is substituted with an annotated type `U?`, then `T?` represents `U?`, not `U??`.
```
jskeet, line 862: As per classes.md line 2770, I don't understand this bullet point.

standard/types.md
```diff
@@ -737,7 +737,7 @@ Since a type parameter can be instantiated with many different type arguments, t
 > - A type parameter cannot be used anywhere within an attribute ([§23.2.1](attributes.md#2321-general)).
 > - A type parameter cannot be used in a member access ([§12.8.7](expressions.md#1287-member-access)) or type name ([§7.8](basic-concepts.md#78-namespace-and-type-names)) to identify a static member or a nested type.
 > - A type parameter can only be used as an *unmanaged_type* ([§8.8](types.md#88-unmanaged-types)) if the type parameter is constrained by the unmanaged constraint ([§15.2.5](classes.md#1525-type-parameter-constraints)).
-> - Nullable annotations (`?`) aren’t allowed on an instance of a type parameter unless that type parameter is constrained to be either a reference type or a value type ([§15.2.5](classes.md#1525-type-parameter-constraints)).  
+> - Except when a type parameter is explicitly constrained to value types, the nullable type annotation (`?`) can only be applied to an instance of a type parameter when the nullable annotations flag is enabled ([§6.5.9](lexical-structure.md#659-nullable-directive), [§15.2.5](classes.md#1525-type-parameter-constraints)).
```
jskeet, line 740: I think I understand what "instance" means here, but that an alternative would be better. (It means "a situation where it occurs in the source code" right?) I don't have a better word right now, but we should brainstorm it.

jskeet (commented):
Various comments around clarity, but I think the overall shape/approach is fine.

BillWagner:
The two most recent commits address all Jon's comments.

jskeet:
Will review on Wednesday (driving tomorrow)

standard/types.md
jskeet: Thanks, occurrence is much better.

standard/interfaces.md
jskeet: This is much clearer, although it's really weird for the value type case. Here be dragons!

standard/classes.md
jskeet: (Leaving this comment unresolved in case we want to discuss it in meetings.)

standard/classes.md
jskeet: I think this comment may still be relevant.

Nigel-Ecma: @jskeet, yes & no…

The `struct` constraint is called the *value type constraint* and it enforces that the parameter is a non-nullable value type.

So writing “constrained to value types” is, shall we say, imprecise. “constrained by the value type constraint” or, as you suggest, “constrained to non-nullable value type”, maybe even “constrained to non-nullable value type by the value type constraint” for the long winded, but please not “constrained to struct” which I just spotted in §15.6.10 *Extension methods*.

Over to the wordsmiths

standard/classes.md
```diff
@@ -444,12 +449,14 @@ The nullability of the type argument need not match the nullability of the type
...
 
-<!-- Remove in C# 9, when `?` is allowed on any type parameter. -->
-The nullable type annotation, `?`, can only be used on a type parameter that has the value type constraint, the reference type constraint without the *nullable_type_annotation*, or a class type constraint without the *nullable_type_annotation*.
+Except when a type parameter is explicitly constrained to value types, the nullable type annotation `?` can only be applied to a type parameter when the nullable annotations flag is enabled ([§6.5.9](lexical-structure.md#659-nullable-directive)).
 
-<!-- Add in C# 9, when `?` is allowed on nullable reference type parameters. -->
-<!-- For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`. -->
+For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`.
 
+> *Note*: On a type parameter, a return type `T?` has the same nullability effect as `[MaybeNull] T`, and a parameter type `T?` has the same nullability effect as `[AllowNull] T`. *end note*
```
Nigel-Ecma, line 456: It is not my understanding that code analysis attributes (§23.5.7) have a “nullability effect”, being as they are just hints:

> The attributes in this subclause are used to provide additional information to support a compiler that provides nullability and null-state diagnostics ([§8.9.5](types.md#895-nullabilities-and-null-states)). A compiler is not required to perform any null-state diagnostics. The presence or absence of these attributes do not affect the language nor the behavior of a program.

doesn’t read to me as an *effect*.

The source of this note is I assume the *Unconstrained type parameter annotations* proposal where it says:

> For return values, `T?` is equivalent to `[MaybeNull]T`; for argument values, `T?` is equivalent to `[AllowNull]T`. The equivalence is important when overriding or implementing interfaces from an assembly compiled with C#8.

Which raises the question whether this is about a particular implementation or whether this equivalence is required, regardless of implementation, to allow C# 9 to call C# 8 code correctly? But does the Standard even deal with such inter-version issues? If not then maybe this Note is redundant? If yes then is it not normative?

If the content of the Note, whether informative or normative, is not redundant it needs rewriting to be clear what it is saying. (To be clear: the source of the lack of clarity comes from the source material.)

standard/classes.md
```diff
@@ -534,6 +541,8 @@ The unmanaged type constraint specifies that a type argument used for the type p
 
 Because `unmanaged` is not a keyword, in *primary_constraint* the unmanaged constraint is always syntactically ambiguous with *class_type*. For compatibility reasons, if a name lookup ([§12.8.4](expressions.md#1284-simple-names)) of the name `unmanaged` succeeds it is treated as a `class_type`. Otherwise it is treated as the unmanaged constraint.
 
+The `default` constraint may be used for a type parameter whose inherited constraints do not establish it as a reference type or a value type; its effect on `T?` in override methods and explicit interface member implementations is specified in [§15.6.5](classes.md#1565-override-methods) and [§19.6.2](interfaces.md#1962-explicit-interface-member-implementations).
```
Nigel-Ecma, line 544: Maybe “may *only* be used”? I.e.:

```suggestion
The `default` constraint may only be used for a type parameter whose inherited constraints do not establish it as a reference type or a value type; its effect on `T?` in override methods and explicit interface member implementations is specified in [§15.6.5](classes.md#1565-override-methods) and [§19.6.2](interfaces.md#1962-explicit-interface-member-implementations).
```

standard/classes.md
Nigel-Ecma: The last example here states that the `default` constraint is required, but the formal description 2769-75 does not appear to specify this.

All the examples use multiple definitions of the same method name, but these are not required to need `default`:

```csharp
﻿﻿public class Base
{
    public virtual void M<T>(T? value) /* where T : notnull */ { }
}

public sealed class Derived : Base
{
    // Without "default", this is not allowed
    public override void M<T>(T? value) where T : default { }
}
```

Also while the term “unconstrained” is being used in places this is not about no constraints but the lack of constraints which tie the type parameter to being a value xor reference type. The commented `notnull` constraint above still requires the `default`.

@jskeet “confusing bit of the language” – not just for you, adding `default` was intended to add clarity… did it succeed? YMMV. Would `T??` have been better? Hmmm…

standard/classes.md
```diff
@@ -2757,9 +2766,13 @@ A compile-time error occurs unless all of the following are true for an override
...
+  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type.
+  - If either a `struct` constraint is added, or no constraint is added and the inherited constraint is a value type constraint, for the type parameter `T` then `T?` is a nullable value type.
+  - If a `default` constraint is added for type parameter `T`, the interpretation of `T?` depends on the type argument substituted for `T`:
+    - If `T` is substituted with a reference type `C`, `T?` represents the nullable reference type `C?`.
+    - If `T` is substituted with an already-annotated reference type `C?`, `T?` represents `C?`, not `C??`.
+    - If `T` is substituted with a value type, `T?` represents `T` itself, not `Nullable<T>`.
```
Nigel-Ecma, line 2775: Is when a `default` contraint is *required* specified somewhere? See comment on line 2846.

Nigel-Ecma (changes requested):
Changes requested, comments added but in brief:

- When and why `default` is needed is unclear. The examples show methods which are overloaded and overridden but only the latter is required to need `default` in some cases.
- The feature is called “unconstrained” but that is not actually true, it is about type parameters which are not constrained to be value *xor* reference types (e.g. they may be constrained to be `notnull` which is value *or* reference).

To be clear: I think the source material for this feature is, let’s say, fuzzy; which makes the task of specifying this clearly & precisely a challenge.

For some reason “Oh what a tangle web we weave…” came to mind while reviewing this feature ;-)

standard/classes.md
```diff
@@ -444,12 +449,14 @@ The nullability of the type argument need not match the nullability of the type
...
 
-<!-- Remove in C# 9, when `?` is allowed on any type parameter. -->
-The nullable type annotation, `?`, can only be used on a type parameter that has the value type constraint, the reference type constraint without the *nullable_type_annotation*, or a class type constraint without the *nullable_type_annotation*.
+Except when a type parameter is explicitly constrained to value types, the nullable type annotation `?` can only be applied to a type parameter when the nullable annotations flag is enabled ([§6.5.9](lexical-structure.md#659-nullable-directive)).
 
-<!-- Add in C# 9, when `?` is allowed on nullable reference type parameters. -->
-<!-- For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`. -->
+For a type parameter `T` when the type argument is a nullable reference type `C?`, instances of `T?` are interpreted as `C?`, not `C??`.
```
jnm2, line 454: I'm guessing this is also true for any nullable type, whether reference or value?

standard/classes.md
jnm2, line 456: The effect of `T?` is actually to equivalent `[AllowNull, MaybeNull] T` _wherever_ it appears, unless `T` is a struct type or is a `: struct` type parameter. (Also unless the `T?` has `[NotNull]` and/or `[DisallowNull]` on it)

This is true inside the method:

```cs
void M(string? p, [AllowNull] string p2, [AllowNull, MaybeNull] string p3)
{
    p = null; // No warning
    p2 = null; // Warning, null assigned to non-nullable variable (allownull is a *caller* precondition)
    p3 = null; // No warning
}
```

And is also true outside the method if the parameter is byref:

```cs
var p = (string?)null;
var p2 = (string?)null;
var p3 = (string?)null;
M(ref p, ref p2, ref p3);
p.ToString(); // Warning, p may be null
p2.ToString(); // No warning, the NotNull method postcondition for `string` has not been overridden
p3.ToString(); // Warning, p may be null

void M(ref string? p, [AllowNull] ref string p2, [AllowNull, MaybeNull] ref string p3)
{
}
```

jnm2: One way to understand nullability analysis as implemented by Roslyn is that every value has two independent dimensions: preconditions and postconditions. Preconditions can be explicit via `[AllowNull]`/`[DisallowNull]` or they can be inferred from the nullability of the type. Postconditions can be explicit via `[MaybeNull]`/`[DisallowNull]` or they can be inferred from the nullability of the type. Preconditions and postconditions apply more places than method parameters and returns: also to fields and to properties.

standard/classes.md
```diff
@@ -2757,9 +2766,13 @@ A compile-time error occurs unless all of the following are true for an override
...
-  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type; otherwise
-  - If either there is no added constraint, or a `struct` constraint is added, for the type parameter `T` then `T?` is a nullable value type.
+- A *type_parameter_constraints_clause* may only consist of the `class`, `struct`, or `default` *primary_constraint*s. The `class` and `struct` constraints may be placed on *type_parameter*s known according to the inherited constraints to be either reference or non-nullable value types respectively. The `default` constraint may be placed on *type_parameter*s that are not constrained to either reference or value types. Any type of the form `T?` in the overriding method's signature, where `T` is a type parameter, is interpreted as follows:
+  - If a `class` constraint is added for type parameter `T` then `T?` is a nullable reference type.
```
jnm2, line 2770: For symmetry with the following line?
```suggestion
  - If a `class` constraint is added for type parameter `T`, or no constraint is added and the inherited constraint is a reference type constraint, then `T?` is a nullable reference type.
```

standard/classes.md
jnm2, line 2775: Should something be explicitly said about if `T` is substituted with a value type which happens to itself be some `Nullable<S>`?

PR #1486, opened 2025-11-27
Clarify conformance with respect to additional types, members, attribute targets

### LINKED ISSUE #870: Must attribute arguments be exactly as shown in Annex C?
**Describe the bug**

From §5 and §C.1 in the C# 7.x draft, it is not clear whether a conforming implementation has to decorate the standard classes with attributes exactly as specified in Annex C, or whether it can use different values as attribute arguments — for example, adding more AttributeTargets bits to an AttributeUsageAttribute on an attribute class.

**Example**

§C.1 specifies this type:

```csharp
[AttributeUsage(AttributeTargets.Parameter, Inherited = false)]
    public sealed class CallerFilePathAttribute : Attribute
    {
        public CallerFilePathAttribute() { }
    }
```

Could a conforming implementation use e.g. `AttributeTargets.Parameter | AttributeTargets.Field` instead?  If so, would that count as an extension and have to be documented?  (Independently of whether the attribute has any compile-time effect when applied to a field.)

**Expected behavior**

Directly specify whether this is allowed.

**Additional context**

§5 does not mention attributes here:

> A conforming implementation of C# is permitted to provide additional types, values, objects, properties, and methods beyond those described in this specification, provided they do not alter the behavior of any strictly conforming program.

(If that paragraph were modified to allow changing attribute arguments, the "provided that" condition at the end would not be a problem, as MemberInfo.GetCustomAttributes and Attribute.GetCustomAttributes methods are not listed in Annex C and thus cannot be called by a strictly conforming program.)

§C.1 says that additional attributes are allowed, but does not mention modifications of specified attributes.  Also, that text is informative.

Inspired by <https://github.com/dotnet/csharpstandard/issues/858>; if C# 10 specifies AttributeTargets.Method on AsyncMethodBuilderAttribute but C# 7.3 does not, and modifications of attribute arguments are not allowed, then it won't be possible for implementations using the same library to comply with both versions of C#, which would be unfortunate.

jskeet:
Fixes #870

(I had to have a couple of goes at the "additional targets" part - feel free to nitpick the exact wording.)


Nigel-Ecma (approved):
LGTM

KalleOlaviNiemitalo:
Side note.  "provided they do not alter the behavior of any strictly conforming program" seems like it could interfere with `using static`.  Something like

```csharp
using static A;

namespace B {
    using static System.Int32;
    class C {
        static int Main() => MaxValue;
    }
}

class A {
    public const int MaxValue = 42;
}
```

Now if the implementation extends System.Int32 with a `public const int MaxValue` member, then Main() will return that rather than A.MaxValue, thus changing the behaviour.  So the rule seems to pretty much disallow providing additional static members.  Or is this kind of program not strictly conforming?

jnm2 (approved):

jskeet:
@KalleOlaviNiemitalo: I'll raise a separate issue for that.

PR #1487, opened 2025-11-27
Prohibit applying Conditional to property/event accessors

### LINKED ISSUE #912: ConditionalAttribute on property or event accessor
<https://github.com/dotnet/csharpstandard/issues/375> discussed ObsoleteAttribute and ConditionalAttribute on accessors of properties and events.

For ObsoleteAttribute, the conclusion was that the standard does not disallow it even though a C# 7 compiler does, but that it should be allowed in a later version of C# and it is not worth the trouble to change the standard now and revert that change later.  The issue was then closed.

Was something decided about whether the standard should be changed to disallow ConditionalAttribute?

jskeet:
Fixes #912

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -543,13 +543,13 @@ A call to a conditional method is included if one or more of its associated cond
 A conditional method is subject to the following restrictions:
 
 - The conditional method shall be a method in a *class_declaration* or *struct_declaration*. A compile-time error occurs if the `Conditional` attribute is specified on a method in an interface declaration.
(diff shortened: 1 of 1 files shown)


Nigel-Ecma (approved):
LGTM

KalleOlaviNiemitalo:
Does this cover attributes of indexers?

jskeet:
Happy to add indexer accessors - will do that on Monday.

jnm2 (approved):

PR #1489, opened 2025-12-01
Fix typo in 10.5.4 User-defined implicit conversions

jnm2:
Typo introduced in v6.


RexJaeschke (approved):
I remember making that whole set of edits, and it looks like I got one wrong. Good catch!

jnm2:
That section was certainly very useful for me today in tracking down some [odd behaviors](https://github.com/dotnet/csharplang/discussions/8201#discussioncomment-15118618)! It's nice knowing that it was your efforts that paved the way, thank you for that.

PR #1493, opened 2025-12-01
Use the latest action to make a PR

BillWagner:
This SHA matches version 7.0.9. By using the SHA, this remains the same version even if the tag is redefined.

Note that dependabot honors this use and will update the sha if a newer version is tagged.

### DIFF
--- a/.github/workflows/update-on-merge.yaml
+++ b/.github/workflows/update-on-merge.yaml
@@ -56,7 +56,7 @@ jobs:
       shell: bash
 
     - name: Create pull request
-      uses: dotnet/actions-create-pull-request@v4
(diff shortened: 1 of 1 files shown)


jnm2 (approved):
What is the reason to avoid a tag for this action when the other actions use tags?

BillWagner:
> What is the reason to avoid a tag for this action when the other actions use tags?

@jnm2 
The preferred use is by SHA, because that won't change. A tag could be moved to a new commit, which makes it a potential attack vector. We should change the others as well, but I haven't done that yet. This one is the only one that requires write perms, so it's more important than others.

PR #1496, opened 2025-12-04
Add note on scopes and type parameters

### LINKED ISSUE #1435: Clarify nested scope for type parameters from outer scope
See https://github.com/dotnet/csharpstandard/pull/681#discussion_r2424276421

Changes necessary for default interface members may require changing the wording in §7.7.1 (Scopes / General).  In the following, the outer `T` is hidden by the inner `T`.

```csharp
interface IOuter<out T>
{
    interface IInner<T>
    {
        class C { }
    }
}
```

§7.7.1 currently says:

> The ***scope*** of a name is the region of program text within which it is possible to refer to the entity declared by the name without qualification of the name. Scopes can be *nested*, and an inner scope may redeclare the meaning of a name from an outer scope. (This does not, however, remove the restriction imposed by §7.3 that within a nested block it is not possible to declare a local variable or local constant with the same name as a local variable or local constant in an enclosing block.) The name from the outer scope is then said to be ***hidden*** in the region of program text covered by the inner scope, and access to the outer name is only possible by qualifying the name.

Within the definition of `interface IInner<T>` , it is not possible to refer to the `T` type parameter of `IOuter`.  According to this wording then, the definition of `class C` is not in the scope of the outer `T`.

Alternatively, we could include "including any nested scopes" to the restriction here.

BillWagner:
Fixes #1435

Add a note that when a type parameter from an outer scope is hidden by a nested declaration, that hidden outer name may not be accessible because it can't be a qualified name.


jskeet (approved):

Nigel-Ecma (approved):
LGTM

Issue #1053, opened 2024-03-04
“Defensive” copies – standardising a compiler strategy or language semantics?


Nigel-Ecma:
@KalleOlaviNiemitalo uses the following [code](https://github.com/dotnet/csharpstandard/pull/673#discussion_r1336779227) in a comment on PR #673:

```csharp
struct S {
    int P => 0; // not readonly
    
    int this[string s] => 0; // not readonly

    readonly void M() {
        _ = P; // warning CS8656: Call to non-readonly member 'S.P.get' from a 'readonly' member results in an implicit copy of 'this'.
        _ = this[""]; // warning CS8656: Call to non-readonly member 'S.this[string].get' from a 'readonly' member results in an implicit copy of 'this'.
        N(); // warning CS8656: Call to non-readonly member 'S.N()' from a 'readonly' member results in an implicit copy of 'this'.
    }

    void N() {}
}
```

@KalleOlaviNiemitalo is making a point about the warnings Roslyn issues, not about the three copies of `this` the Standard *mandates* as “defensive copies” as being redundant.

Surely a different compiler might choose to analyze `P`, `this[]` & `N`; figure out they do not modify the struct; skip the defensive copies; produce the intended semantics; and yet be non-compliant for doing so…

Are not defensive copies a *compiler strategy* for achieving a particular *semantics*; is the Standard mandating the strategy rather than the semantics; and in doing so limiting compilers (and performance in this case)?

Should not the strategy be left to the compilers, and the semantics to the Standard?

Or is it argued that the strategy here *is* the intended semantics?

[Ducking ;-)] Maybe someone should look into this…

BillWagner:
I'd like to leave this in the backlog for our future selves. A good time is when we cover `ref readonly` parameter in C# 12.

It's a side trip at the moment, but I know the roslyn team has made some improvements so that defensive copies are made in fewer situations. I think more of these come in more recent versions, so what we have should be correct for 8.

Nigel-Ecma:
While I’m not looking at getting into this anytime soon @BillWagner’s remark “the roslyn team has made some improvements so that defensive copies are made in fewer situations” does suggest this is a compiler strategy more than language semantics and maybe the closing remark should be “correct for *roslyn C#* 8”?

[...]

PR #1497, opened 2025-12-04
Add rules for `readonly` members

### LINKED ISSUE #1339: Define rules for "readonly" in `struct` types
See https://github.com/dotnet/csharpstandard/pull/673/files#r1510548259

The language that describes what `readonly` allows and prohibits needs more work.

1. There are instances of the phrase "It is a compile-time error to attempt to modify the state of an instance struct variable via a readonly property declared in that struct." in the standard. This isn't defined as well as it should be.
2. A compiler may allow `readonly` members to invoke non-readonly members on "the same instance" of a `struct` by creating a defensive copy and invoking the mutating member on the defensive copy.
3. A readonly member can't change the *value* of a member, but can change the *fields* that are members of that member. (not specese, but that describes the issue)

Copied from the linked discussion:

under classes modifiers are generally described in subclauses under *§15.3 Class members*.

That same pattern can be repeated here with the current content of *§16.3 Struct members* becoming *§16.3.1 General* and adding *§16.3.2 The readonly modifier*.

In §16.3.2 the rules for readonly members can be given (not proposed text):

- They cannot mutate the state of the struct instance
- They may mutate the static state of the struct
- And the (subjectively) ugly one, a readonly member may call a mutating member (of the same instance only?) but any mutations made must not be observable after the call. This could in terms of the compiler strategy of “defensive copies”, see #1053, or not.

This would provide a single place for defining what a readonly member. The additional per-member differences could follow in §16.3 subclauses or remain as they are in §16.4 subclauses.

In addition to the normative language, an example (as discussed [here](https://github.com/dotnet/csharpstandard/pull/673/files#r1670134847) would be useful as well.

BillWagner:
Fixes #1339

Add a new section for the rules on `readonly` members of structs.

I didn't write a specific rule that `readonly` members can modify `static` fields. There's no rule prohibiting it, so it should be allowed.

I chose language consistent with the future work on #1053 to consider. One popular implementation creates a "defensive copy" of the receiver to invoke a non-readonly member.

Finally, add an example that demonstrates the "direct member of a struct" rule by having a `readonly` property and `readonly` method modify the members of a direct member, which is allowed.

### DIFF
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -178,7 +178,7 @@ An instance member definition or accessor of an instance property, indexer, or e
 > }
 > ```
 >
-> The `readonly` method `AddMessage` can change the state of a message list. The `InitializeMessages` member can clear and re-initialize the list of messages. In the first case, the `readonly` modifier is valid. In the case of `InitializeMessages`, adding the `readonly` modifier is invalid. *end example*
+> The `readonly` method `AddMessage` can change the state of a message list. The `InitializeMessages` member can clear and re-initialize the list of messages. In the case of `AddMessage`, the `readonly` modifier is valid. In the case of `InitializeMessages`, adding the `readonly` modifier is invalid. *end example*
 
 ## 16.4 Class and struct differences
 
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -114,6 +114,8 @@ struct_body
 
 ## 16.3 Struct members
 
+### §struct_general General
+
 The members of a struct consist of the members introduced by its *struct_member_declaration*s and the members inherited from the type `System.ValueType`.
 
 ```ANTLR
@@ -137,6 +139,51 @@ struct_member_declaration
 > *Note*: All kinds of *class_member_declaration*s except *finalizer_declaration* are also *struct_member_declaration*s. *end note*
 
 Except for the differences noted in [§16.4](structs.md#164-class-and-struct-differences), the descriptions of class members provided in [§15.3](classes.md#153-class-members) through [§15.12](classes.md#1512-static-constructors) apply to struct members as well.
+
+### §struct_readonly Readonly members
+
+An instance member definition that includes the `readonly` modifier has the following restrictions:
+
+- The member shall not modify the value of a direct instance member of the receiver.
+- A readonly member may call a non-readonly member using the receiver instance only if it ensures no modifications to the receiver are observable after the readonly member returns.
+
+> *Example*: The following code demonstrates the rule that a readonly member shall not modify the value of a direct instance member, but can modify the members of a direct instance member:
+>
+> <!-- Example: {template:"standalone-lib-without-using", name:"ReadonlyMember" } -->
+> ```csharp
+> public struct S
+> {
+>    private bool[] flags;
+>
+>    public S(int size)
+>    {
+>       flags = new bool[size];
+>    }
+>
+>    public readonly bool this[int index]
+>    {
+>       get => flags[index]; 
+>       set => flags[index] = value;
+>    }
+>
+>    // Not readonly
+>    public void ResetFlags()
+>    {
+>       flags = new bool[flags.Length];
+>    }
+>
+>    // Allowed on readonly member
+>    public readonly void ReadonlyResetFlags()
+>    {
+>       for (int i = 0; i < flags.Length; i++)
+>       {
+>          flags[i] = false;
+>       }
+>    }
+> }
+> ```
+>
+> The `readonly` indexer can change the state of a single member of the `flags` array. The `ReadonlyResetFlags` member can set all flags to false. In both those cases, the direct member, `flags` has not been reassigned. The `ResetFlags` member reassigns the direct member, and therefore cannot include the `readonly` modifier. *end example*
 
 ## 16.4 Class and struct differences
 
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -144,7 +144,7 @@ Except for the differences noted in [§16.4](structs.md#164-class-and-struct-dif
 
 An instance member definition or property accessor of an instance property that includes the `readonly` modifier has the following restrictions:
 
-- The member shall not reassign the value of an instance field of the receiver.
+- The member shall not reassign the value of `this` or an instance field of the receiver.
 - The member shall not reassign the value of an instance field-like event (§15.8.2) of the receiver.
 - A readonly member may call a non-readonly member using the receiver instance only if it ensures no modifications to the receiver are observable after the readonly member returns.
 
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -142,7 +142,7 @@ Except for the differences noted in [§16.4](structs.md#164-class-and-struct-dif
 
 ### §struct_readonly Readonly members
 
-An instance member definition or property accessor of an instance property that includes the `readonly` modifier has the following restrictions:
+An instance member definition or accessor of an instance property, indexer, or event that includes the `readonly` modifier has the following restrictions:
 
 - The member shall not reassign the value of `this` or an instance field of the receiver.
 - The member shall not reassign the value of an instance field-like event (§15.8.2) of the receiver.
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -144,51 +144,41 @@ Except for the differences noted in [§16.4](structs.md#164-class-and-struct-dif
 
 An instance member definition or accessor of an instance property, indexer, or event that includes the `readonly` modifier has the following restrictions:
 
+- The `this` parameter is a `ref readonly` reference.
 - The member shall not reassign the value of `this` or an instance field of the receiver.
 - The member shall not reassign the value of an instance field-like event (§15.8.2) of the receiver.
-- A readonly member may call a non-readonly member using the receiver instance only if it ensures no modifications to the receiver are observable after the readonly member returns.
+- If a readonly member invokes a non-readonly member, the structure referred to by `this` must be copied to use a writable reference for the `this` argument.
 
 > *Note:* Instance fields include the hidden backing field used for automatically implemented properties (§15.7.4). *end note*
 <!-- markdownlint-disable MD028 -->
 
 <!-- markdownlint-enable MD028 -->
-> *Example*: The following code demonstrates the reassigning and modifying an instance field. Obfuscating the intent of `readonly` in this manner is discouraged:
+> *Example*: A readonly member can modify the state of an object referred to by an instance field, even though the readonly member can't reassign that instance member. The following code demonstrates the reassigning and modifying an instance field:
 >
 > <!-- Example: {template:"standalone-lib-without-using", name:"ReadonlyMember" } -->
 > ```csharp
 > public struct S
 > {
->    private bool[] flags;
+>     private List<string> messages;
 >
->    public S(int size)
->    {
->       flags = new bool[size];
->    }
+>     public S(IEnumerable<string> messages) =>
+>         this.messages = new List<string>(messages);
 >
->    public readonly bool this[int index]
->    {
->       get => flags[index]; 
->       set => flags[index] = value;
->    }
+>     public void InitializeMessages() =>
+>         messages = new List<string>();
 >
->    // Not readonly
->    public void ResetFlags()
->    {
->       flags = new bool[flags.Length];
->    }
->
->    // Allowed on readonly member
->    public readonly void ReadonlyResetFlags()
->    {
->       for (int i = 0; i < flags.Length; i++)
->       {
->          flags[i] = false;
->       }
->    }
+>     public readonly void AddMessage(string message)
+>     {
+>         if (messages == null)
+>         {
+>             throw new InvalidOperationException("Messages collection is not initialized.");
+>         }
+>         messages.Add(message);
+>     }
 > }
 > ```
 >
-> The `readonly` indexer can change the state of a single member of the `flags` array. The `ReadonlyResetFlags` member can set all flags to false. In both those cases, the field, `flags` has not been reassigned to a new array. The `ResetFlags` member reassigns the direct member, and therefore cannot include the `readonly` modifier. *end example*
+> The `readonly` method `AddMessage` can change the state of a message list. The `InitializeMessages` member can clear and re-initialize the list of messages. In the first case, the `readonly` modifier is valid. In the case of `InitializeMessages`, adding the `readonly` modifier is invalid. *end example*
 
 ## 16.4 Class and struct differences
 
--- a/standard/structs.md
+++ b/standard/structs.md
@@ -142,12 +142,17 @@ Except for the differences noted in [§16.4](structs.md#164-class-and-struct-dif
 
 ### §struct_readonly Readonly members
 
-An instance member definition that includes the `readonly` modifier has the following restrictions:
+An instance member definition or property accessor of an instance property that includes the `readonly` modifier has the following restrictions:
 
-- The member shall not modify the value of a direct instance member of the receiver.
+- The member shall not reassign the value of an instance field of the receiver.
+- The member shall not reassign the value of an instance field-like event (§15.8.2) of the receiver.
 - A readonly member may call a non-readonly member using the receiver instance only if it ensures no modifications to the receiver are observable after the readonly member returns.
 
-> *Example*: The following code demonstrates the rule that a readonly member shall not modify the value of a direct instance member, but can modify the members of a direct instance member:
+> *Note:* Instance fields include the hidden backing field used for automatically implemented properties (§15.7.4). *end note*
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+> *Example*: The following code demonstrates the reassigning and modifying an instance field. Obfuscating the intent of `readonly` in this manner is discouraged:
 >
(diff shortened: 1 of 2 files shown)


standard/structs.md
```diff
@@ -138,6 +140,51 @@ struct_member_declaration
 
 Except for the differences noted in [§16.4](structs.md#164-class-and-struct-differences), the descriptions of class members provided in [§15.3](classes.md#153-class-members) through [§15.12](classes.md#1512-static-constructors) apply to struct members as well.
 
+### §struct_readonly Readonly members
+
+An instance member definition that includes the `readonly` modifier has the following restrictions:
+
+- The member shall not modify the value of a direct instance member of the receiver.
```
jskeet: Possibly be explicit about "field or property"? Is it okay to call a property setter if that doesn't actually change any instance state?

standard/structs.md
```diff
@@ -138,6 +140,51 @@ struct_member_declaration
...
+- A readonly member may call a non-readonly member using the receiver instance only if it ensures no modifications to the receiver are observable after the readonly member returns.
+
+> *Example*: The following code demonstrates the rule that a readonly member shall not modify the value of a direct instance member, but can modify the members of a direct instance member:
```
jskeet: I think we should say this is strongly discouraged though...

standard/structs.md
jskeet: (I guess we should include field-like events in this too.)

standard/structs.md
Nigel-Ecma: The example suggests to me that the definition of (in)direct should mention arrays

standard/structs.md
Nigel-Ecma: I can find no definition of “direct [instance] member”.

Unless I’m missing them definitions for direct and indirect members are required.

Nigel-Ecma (dismissed):
Missing definition(s)?

standard/structs.md
```diff
@@ -138,6 +140,51 @@ struct_member_declaration
 
 Except for the differences noted in [§16.4](structs.md#164-class-and-struct-differences), the descriptions of class members provided in [§15.3](classes.md#153-class-members) through [§15.12](classes.md#1512-static-constructors) apply to struct members as well.
 
+### §struct_readonly Readonly members
+
+An instance member definition that includes the `readonly` modifier has the following restrictions:
```
jnm2: Are accessors considered members?

```cs
struct S
{
    int Prop
    {
        readonly get { return 1; }
        set { }
    }
}
```

BillWagner: Should be, but I made this more explicit.

standard/structs.md
BillWagner: Updated in next commit

BillWagner: I did finesse my way to avoid the use of "direct" and "indirect" members.

standard/structs.md
BillWagner: Aded text on this one.

BillWagner (commented):
This is now ready for final review @jskeet @Nigel-Ecma

standard/structs.md
```diff
@@ -138,6 +140,56 @@ struct_member_declaration
 
 Except for the differences noted in [§16.4](structs.md#164-class-and-struct-differences), the descriptions of class members provided in [§15.3](classes.md#153-class-members) through [§15.12](classes.md#1512-static-constructors) apply to struct members as well.
 
+### §struct_readonly Readonly members
+
+An instance member definition or property accessor of an instance property that includes the `readonly` modifier has the following restrictions:
```
jnm2:
```suggestion
An instance member definition or accessor of an instance property, indexer, or event that includes the `readonly` modifier has the following restrictions:
```

jnm2: Then below, it says "the member" even if at this point it had been "the accessor." Does the spec have a handy concept of "function member body" or something like that?

standard/structs.md
```diff
@@ -138,6 +140,56 @@ struct_member_declaration
...
+
+- The member shall not reassign the value of an instance field of the receiver.
```
jnm2:
```suggestion
- The member shall not reassign the value of `this` or an instance field of the receiver.
```

standard/structs.md
```diff
@@ -138,6 +140,56 @@ struct_member_declaration
...
+- The member shall not reassign the value of an instance field-like event (§15.8.2) of the receiver.
+- A readonly member may call a non-readonly member using the receiver instance only if it ensures no modifications to the receiver are observable after the readonly member returns.
```
jnm2: What does a readonly member do to ensure this? I would have thought it was an error to call a non-readonly member using the receiver instance.

BillWagner: The roslyn implementation creates a defensive copy, and calls the non-readonly member using that copy as `this`.

There are also instances where static analysis determines that the call doesn't modify state, so the copy is elided.

That's the reason for the language that doesn't require an implementation choice.

jnm2: What do you think about having it spec clearly that the compiler must perform this defensive copy? As written it sounds ambiguous, as though there might be something the language allows the user to do in order to achieve this.

BillWagner: In the meeting, @Nigel-Ecma pointed out that saying it must be a defensive copy is standardizing one implementation choice. While it's an obvious choice, it's not the only possible implementation. I see your point. How about this alternative wording:

```suggestion
- A readonly member may call a non-readonly member using the receiver instance. A conforming implementation must ensure no modifications to the receiver are observable after the readonly member returns.
```

jnm2: It's the only possible implementation that does not result in a compiler error, isn't it? I would think for portability reasons, all compilers should agree on whether or not it's an error. If it's not, then defensive copy is the only possible approach.

jnm2: (Here's the reason why I'm not seeing a third approach, between making this a compile error, or making a defensive copy. A hypothetical third approach that allows the user to make the call and does _not_ make a defensive copy would involve temporary writes and then more writes to return the state. Since writes are actually observably occurring, both observable within the same thread and observable in other threads, I believe this does not satisfy the meaning of `readonly`. Therefore the only remaining approach that satisfies the meaning of readonly is one done on a defensive copy to sequester the writes somewhere safe.)

standard/structs.md
```diff
@@ -138,6 +140,56 @@ struct_member_declaration
 
 Except for the differences noted in [§16.4](structs.md#164-class-and-struct-differences), the descriptions of class members provided in [§15.3](classes.md#153-class-members) through [§15.12](classes.md#1512-static-constructors) apply to struct members as well.
 
+### §struct_readonly Readonly members
+
+An instance member definition or accessor of an instance property, indexer, or event that includes the `readonly` modifier has the following restrictions:
+
+- The member shall not reassign the value of `this` or an instance field of the receiver.
+- The member shall not reassign the value of an instance field-like event (§15.8.2) of the receiver.
```
jnm2, line 149: I just remembered refs!
```suggestion
- The member shall not reassign the value of or take a writable reference to `this`, to an instance field of the receiver, or to an instance field-like event (§15.8.2) of the receiver.
```

Example (and same for field or field-like event as for `this` itself):
```cs
struct S
{
    readonly void M()
    {
        ref readonly S s = ref this; // Allowed
        ref S s = ref this; // Not allowed

        M2(ref this); // Not allowed
        M3(in this); // Allowed
    }

    void M2(ref S p) { }
    void M3(in S p) { }
}
```

standard/structs.md
BillWagner: I want to leave a lot of wiggle room because a theoritical compiler could improve static analysis in many places where a developer left off `readonly` *on members that don't modify state*. Yes, it's an optimization, but if the standard *requires* a defensive copy, that optimization would be prohibited.

I admit I can't think of another general implementation either. But, why prohibit one?

standard/structs.md
```diff
@@ -138,6 +140,56 @@ struct_member_declaration
...
+- A readonly member may call a non-readonly member using the receiver instance only if it ensures no modifications to the receiver are observable after the readonly member returns.
+
+> *Note:* Instance fields include the hidden backing field used for automatically implemented properties (§15.7.4). *end note*
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+> *Example*: The following code demonstrates the reassigning and modifying an instance field. Obfuscating the intent of `readonly` in this manner is discouraged:
```
jnm2: What does "obfuscating the intent" mean in this context, and what is being discouraged? Looking at this from my own context, both approaches seem (individually) appropriate to me. I know when I'd use one and I know when I'd use the other.

BillWagner: I'll defer to @jskeet and @Nigel-Ecma Both thought the example demonstrated a practice they wanted "strongly discouraged".

standard/structs.md
Nigel-Ecma: @jnm2 – The underlying reason “defensive copy” should not be used here is because it a language Standard defines the semantics of a language and not how those semantics are implemented on this or that architecture. Implementation should be left to implementers (for whom “how to” documentation may well be written if a language's semantics are novel enough).

A second reason is we’ve been here before, it was argued that it was *vital* for “defensive copies” to be specified, for reasons like yours – "we can’t see any other way this can be achieved” – but as @BillWagner has pointed out they since have and are now technically non-conforming to a requirement that should never have existed…

There is an issue, to be followed when there is time, on eradicating the phrase from the Standard altogether.

Stuff likes this can seep in as the Standard often has to be based on implementation notes etc. which have to redone as language specification.

Nigel-Ecma: @jnm2 & @BillWagner why not just:

```suggestion
- A readonly member may call a non-readonly member using the receiver instance; however no modifications to the receiver are observable after the readonly member returns.
```

Which just sticks to the required language semantics – which implementations are required to provide.

> *Note*: I followed @BillWagner’s suggestion for the semantics, which are presumably this way as there is no concern over with possible observation between the call & return – if that presumption is incorrect then the wording will need modifying appropriately, e.g. by dropping “after the readonly member returns”.

standard/structs.md
Nigel-Ecma: @jnm2 – I can’t speak for @jskeet, for myself:

This is very much subjective. The usual understanding of `readonly` is, unsurprisingly, that state is not altered – this example alters what is clearly *private* state. It would be better without the `set`, as it is the readonly feels a bit like a false flag operation 😉 (you may collectively groan)

It feels a bit worse here as, depending on language, arrays may be value or reference types but are often thought of as more like values regardless – here `flags` certainly has the flavour of local value state of the struct.

It’s a subjective boundary – if the struct stored a reference to, say, a GUI window then not being able to change what window was reference, but being able to change the window’s content would likely not be viewed negatively.

standard/structs.md
jnm2: Okay, I'm following and I agree. I like Nigel's suggestion. The main things I wanted the spec to be clear on is that it is never a compiler error to make that call, while simultaneously no writes are being made to the current instance.

jnm2: Okay, I'm following and I agree. I like Nigel's suggestion. The main things I wanted the spec to be clear on is that it is never a compiler error to make that call, while simultaneously no writes are being made to the current instance.

jnm2: I do have one quibble about "after it returns"- I want a guarantee that there are no writes before it returns, either. I think I can make an example with a callback that shows the difference being observable.

standard/structs.md
```diff
@@ -138,6 +140,56 @@ struct_member_declaration
 
 Except for the differences noted in [§16.4](structs.md#164-class-and-struct-differences), the descriptions of class members provided in [§15.3](classes.md#153-class-members) through [§15.12](classes.md#1512-static-constructors) apply to struct members as well.
 
+### §struct_readonly Readonly members
+
+An instance member definition or accessor of an instance property, indexer, or event that includes the `readonly` modifier has the following restrictions:
+
+- The member shall not reassign the value of `this` or an instance field of the receiver.
+- The member shall not reassign the value of an instance field-like event (§15.8.2) of the receiver.
+- A readonly member may call a non-readonly member using the receiver instance only if it ensures no modifications to the receiver are observable after the readonly member returns.
```
BillWagner: How about this slight modification to Nigel's suggestion:

```suggestion
- A readonly member may call a non-readonly member using the receiver instance; however no modifications to the receiver are observable outside the scope of the readonly member.
```

jskeet: It feels a little odd in terms of scope usually being about the meaning of names, but I'm ambivalent.

Nigel-Ecma: Seeing @jskeet’s above comment in my email I spotted that all three of the original lines 147-149 appear to have the wrong actor. Back here on github I see proposed modifications to 149 do not, but that leaves 147 & 148. It is not “The member shall not reassign” – the member is not the actor, the language semantics are by preventing such an assignment. So all three lines need rewording, I have **not** checked whether the same issue appears elsewhere in the PR.

standard/structs.md
jnm2: A reader should never gather more from the `readonly` modifier than its technical meaning, which is "the struct instance will not have changed due to this call." It does not imply anything more broad, such as that no mutations are being made in private state through an indirection.

For example, CancellationToken.Register is `readonly` (the whole struct is readonly), but no one would imagine that this means that no mutations are being done at all within indirect private state during the Register call.

standard/structs.md
jnm2: Here is the example. It's not sufficient to say that readonly means that no writes are observable after the method returns. No writes should happen at all, within the memory of the `this` instance, if the method is readonly. I believe "observable" is also tricky to define and I would leave it out. If we keep "observable," I'd want to carefully define it, since at the level of subtle threading behaviors and such, there can be observable differences if you do even a "no-op" write, differences which go beyond observing the eventual _value_.

```cs
using System;

class Program
{
    static void Main()
    {
        var storageLocation = new S();

        Console.WriteLine(storageLocation.SomeField); // 0

        storageLocation.M(() =>
        {
            Console.WriteLine(storageLocation.SomeField); // 1
        });

        Console.WriteLine(storageLocation.SomeField); // 0
    }
}

struct S
{
    public int SomeField;

    public void M(Action someCallback)
    {
        // No writes are observable *after* M returns, yet crucially,
        // writes are observable. Readonly should disallow this scenario.

        SomeField++;
        someCallback();
        SomeField--;
    }
}
```

standard/structs.md
jnm2: I'd rather not mention observability because I think that is a thorny topic (see <https://github.com/dotnet/csharpstandard/pull/1497#discussion_r2674460577>). I believe the fact that a write has happened can be observable through other effects besides reading the eventual value.

jnm2: Actually, we can be far more definite. A readonly member is passed a `ref readonly this`, and therefore it cannot write to fields of `this`, nor can it call a non-readonly member on `this` because that would require passing `ref this` which it does not have. It must create a copy in order to do so, and this is exactly the concept of "defensive copy" which is demanded by the fact that a readonly member never gains a writeable reference to `this`.

I have more details now too on how any mutation that a readonly member somehow managed to do could cause access violations, since the `this` reference may be within readonly memory when calling a readonly member, but luckily we don't have to discuss that (and other side effects of mutations) because mutations are not legal through a `readonly ref this`.

standard/structs.md
jskeet: I think this is worth discussing more. I would be happy with softening it to something like "used with care and with clear documentation" rather than outright discouragement. `CancellationToken` is a good example to discuss.

standard/structs.md
jskeet: We decided that we should say a copy is taken, and the non-readonly member is invoked on the copy. We don't need to call it "defensive" and we don't need to explicitly say that an implementation can optimize the copy away.

standard/structs.md
BillWagner: Much more clear language came from adding the change that a `readonly` member has a `ref readonly` `this` parameter.

standard/structs.md
BillWagner: I replaced the sample with one that's more obvious and clear.

BillWagner:
@jskeet I've updated per comments and last month's discussion. This should be ready for the next meeting and merging.

standard/structs.md
```diff
@@ -138,6 +140,46 @@ struct_member_declaration
...
+>         if (messages == null)
+>         {
+>             throw new InvalidOperationException("Messages collection is not initialized.");
+>         }
+>         messages.Add(message);
+>     }
+> }
+> ```
+>
+> The `readonly` method `AddMessage` can change the state of a message list. The `InitializeMessages` member can clear and re-initialize the list of messages. In the first case, the `readonly` modifier is valid. In the case of `InitializeMessages`, adding the `readonly` modifier is invalid. *end example*
```
jskeet: Sorry to nitpick at this late stage, and we can fix this later - but can we refer to the methods in the same order they appear in the code? Possibly just move `InitializeMessages` to after `AddMessage`.

jskeet (approved):
I've somewhat forgotten the context of the discussion, but this looks good to me with one nit.

standard/structs.md
BillWagner:
```suggestion
> The `readonly` method `AddMessage` can change the state of a message list. The `InitializeMessages` member can clear and re-initialize the list of messages. In the case of `AddMessage`, the `readonly` modifier is valid. In the case of `InitializeMessages`, adding the `readonly` modifier is invalid. *end example*
```

PR #1515, opened 2025-12-19
Address remaining nullable reference type issues

### LINKED ISSUE #1228: Nullable Spec Issue List
These are things that should be spelled out in the [nullable reference types feature specification](https://github.com/dotnet/csharplang/blob/master/proposals/csharp-8.0/nullable-reference-types-specification.md):

- [ ] The handling of a lifted operator (if a comparison operator, the return type is still `bool`, but other operators have value types turned into `Nullable` of that value type, and reference types are made nullable.  If either operand might be null, then the result might be null if the resulting return type is capable containing null.

- [ ] In order to preserve the safety in the face of unconstrained type parameters, we warn immediately when a null value of such a type is introduced.  This is a safety warning.  The contexts in which such a warning are given are
  - `default` and `default(T)`
  - `null` conversion to `T` (when `T` is known to be ref type)
  - `e?.M()` when the return type is `T`
  - dynamic conversion or cast to `T` when the dynamic might be null
  - explicit conversion to `T`
  - `e as T` when there is not an implicit conversion from the type of `e` to `T`
  - a call to a method like `FirstOrDefault()` that is annotated with `[MaybeNull] (not yet implemented)

BillWagner:
Fixes #1228 

- For item 2: Add that a lifted conversion may return a null reference type, which can then be converted to a non-nullable reference type, which generates a warnings.
- For item 3 & 4: Add bullet points for the `default` (and `default(T)`) expression, and the `null` literal expression.

### DIFF
--- a/standard/types.md
+++ b/standard/types.md
@@ -871,11 +871,11 @@ The ***default null state*** of an expression is determined by its type, and the
 
 - The default state of the `null` literal (§6.4.5.7) is maybe null.
 - The default state of a default value expression (§12.8.21) is not null when the type is a non-nullable value type, otherwise maybe null.
-- For all other expressions
-  - The default null state of a nullable reference type is:
+- For all other expressions:
+  - The default null state of an expression whose type is a nullable reference type is:
     - Maybe null when its declaration is in text where the annotations flag is enabled.
     - Not null when its declaration is in text where the annotations flag is disabled.
-  - The default null state of a non-nullable reference type is not null.
+  - The default null state of an expression whose type is a non-nullable reference type is not null.
 
 > *Note:* The *maybe default* state is used with unconstrained type parameters when the type is a non-nullable type, such as `string` and the expression `default(T)` is the null value. Because null is not in the domain for the non-nullable type, the state is maybe default. *end note*
 
--- a/standard/types.md
+++ b/standard/types.md
@@ -1168,4 +1168,24 @@ A compiler may issue a warning when nullability differs in either direction in t
 >
 > *end example*
 
+A compiler may issue a warning when the result of a lifted conversion operator (§10.6.2) is the `null` reference and that value is assigned to a non-nullable reference type.
+
+> <!-- Example: {template:"code-in-class-lib", name:"LiftedOperator", ignoredWarnings:["CS8600"]} -->
+> ```csharp
+> #nullable enable
+> class C
+> {
+>     void M()
+>     {
+>         C? a = (int?)null; // No warning
+>         C b = (int?)null; // Warning: Possible assignment to null
+>         C c = 1; // No warning
+>     }
+> 
+>     public static implicit operator C(int v) => new C();
+> }
+> ```
+>
+> *end example*
+
 **End of conditionally normative text**
--- a/standard/types.md
+++ b/standard/types.md
@@ -869,10 +869,13 @@ Every expression has one of three ***null state***s:
 
 The ***default null state*** of an expression is determined by its type, and the state of the annotations flag when it is declared:
 
-- The default null state of a nullable reference type is:
-  - Maybe null when its declaration is in text where the annotations flag is enabled.
-  - Not null when its declaration is in text where the annotations flag is disabled.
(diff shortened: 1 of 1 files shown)


standard/types.md
```diff
@@ -869,10 +869,13 @@ Every expression has one of three ***null state***s:
 
 The ***default null state*** of an expression is determined by its type, and the state of the annotations flag when it is declared:
 
-- The default null state of a nullable reference type is:
-  - Maybe null when its declaration is in text where the annotations flag is enabled.
-  - Not null when its declaration is in text where the annotations flag is disabled.
-- The default null state of a non-nullable reference type is not null.
+- The default state of the `null` literal (§6.4.5.7) is maybe null.
+- The default state of a default value expression (§12.8.21) is not null when the type is a non-nullable value type, otherwise maybe null.
+- For all other expressions
```
jnm2: Is a colon needed?
```suggestion
- For all other expressions:
```

standard/types.md
```diff
@@ -1168,4 +1171,24 @@ A compiler may issue a warning when nullability differs in either direction in t
 >
 > *end example*
 
+A compiler may issue a warning when the result of a lifted conversion operator (§10.6.2) is the `null` reference and that value is assigned to a non-nullable reference type.
```
jnm2: What about something like this, because the compiler doesn't know it _is_ the null reference, it just knows it is _maybe_ the null reference?
```suggestion
A compiler may issue a warning when the result of a lifted conversion operator (§10.6.2) is a reference type and is converted to a non-nullable reference type, since the result of a lifted conversion operator is maybe-null.
```
Or alternatively we say "the default null state of a lifted conversion operator that returns a reference type is maybe-null" and skip talking about the "assigned to a non-nullable reference type" part of this, since the warning-on-assignment part of this is the same for all maybe-null expressions no matter how they came to be maybe-null.

standard/types.md
BillWagner: I defer to @RexJaeschke  I think so.

standard/types.md
```diff
@@ -869,10 +869,13 @@ Every expression has one of three ***null state***s:
...
 The ***default null state*** of an expression is determined by its type, and the state of the annotations flag when it is declared:
 
-- The default null state of a nullable reference type is:
-  - Maybe null when its declaration is in text where the annotations flag is enabled.
-  - Not null when its declaration is in text where the annotations flag is disabled.
-- The default null state of a non-nullable reference type is not null.
+- The default state of the `null` literal (§6.4.5.7) is maybe null.
+- The default state of a default value expression (§12.8.21) is not null when the type is a non-nullable value type, otherwise maybe null.
+- For all other expressions
+  - The default null state of a nullable reference type is:
```
```suggestion
  - The default null state of an expression whose type is a nullable reference type is:
```

jnm2 (approved):

standard/types.md
RexJaeschke: Yes!

standard/types.md
```diff
@@ -869,10 +869,13 @@ Every expression has one of three ***null state***s:
...
+    - Maybe null when its declaration is in text where the annotations flag is enabled.
+    - Not null when its declaration is in text where the annotations flag is disabled.
+  - The default null state of a non-nullable reference type is not null.
```
```suggestion
  - The default null state of an expression whose type is a non-nullable reference type is not null.
```

standard/types.md
```diff
@@ -869,10 +869,13 @@ Every expression has one of three ***null state***s:
 
 The ***default null state*** of an expression is determined by its type, and the state of the annotations flag when it is declared:
 
-- The default null state of a nullable reference type is:
-  - Maybe null when its declaration is in text where the annotations flag is enabled.
-  - Not null when its declaration is in text where the annotations flag is disabled.
-- The default null state of a non-nullable reference type is not null.
+- The default state of the `null` literal (§6.4.5.7) is maybe null.
```
jskeet, line 872: @RexJaeschke Please file an issue (assign it to me) for expressions without types, referring to this PR.

jskeet (approved):

PR #1521, opened 2025-12-27
Add an exception to the requirement that partial method declarations have the same modifiers

jnm2:
This code is accepted just fine:
```cs
partial class C
{
    async partial void M() { }

    partial void M();
}
```

And this is by design. It's not _allowed_ for the `async` modifier to be on both parts. Nor would that make intuitive sense, since `async` is a keyword that only affects the body. It does not affect the signature.

```cs
partial class C
{
    async partial void M() { }

    // ❌ CS1994: The 'async' modifier can only be used in methods that have a body.
    async partial void M();
}
```

When C# 9 comes along, `extern` will have to be added as another exception for the same reason.

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -3034,7 +3034,8 @@ Partial methods shall not define access modifiers; they are implicitly private. 
 
 There are two kinds of partial method declarations: If the body of the method declaration is a semicolon, the declaration is said to be a ***defining partial method declaration***. If the body is other than a semicolon, the declaration is said to be an ***implementing partial method declaration***. Across the parts of a type declaration, there shall be only one defining partial method declaration with a given signature, and there shall be at most only one implementing partial method declaration with a given signature. If an implementing partial method declaration is given, a corresponding defining partial method declaration shall exist, and the declarations shall match as specified in the following:
 
-- The declarations shall have the same modifiers (although not necessarily in the same order), method name, number of type parameters and number of parameters.
+- The declarations shall have the same method name, number of type parameters, and number of parameters.
+- The declarations shall have the same modifiers (although not necessarily in the same order), with the exception of the `async` modifier which may only appear on an implementing part.
 - Corresponding parameters in the declarations shall have the same modifiers (although not necessarily in the same order) and the same types, or identity convertible types (modulo differences in type parameter names).
 - Corresponding type parameters in the declarations shall have the same constraints (modulo differences in type parameter names).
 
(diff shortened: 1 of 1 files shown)


jnm2:
@BillWagner @jskeet This looks like a candidate for merging without a meeting discussion.

Nigel-Ecma:
> Nor would that make intuitive sense, since `async` is a keyword that only affects the body. It does not affect the signature.

Could not an implementation choose to invoke an `async` method differently than a non-async one? And user code surely does? So why does it only “affect the body”?

While I do not doubt that a certain compiler made this choice intentionally, I’m unsure that an implementation which choose to allow it should be deemed to be in error.

jnm2:
> Could not an implementation choose to invoke an async method differently than a non-async one? And user code surely does? So why does it only “affect the body”?

User code does not invoke an async method differently from a non-async one. Nor would an implementation invoke it differently. The effect of the `async` keyword is invisible to callers of the method.

What the `async` keyword does in a nutshell is rewrite the body of the method to use a state machine to produce the `Task` object (or some _task-like_ object) which is returned by the method, rather than having the user manually return some `Task` object. The effect of `async` from a language user perspective is not "up to the implementation" but is rather a semantic mode switch that the user understands.

Either way, the caller invokes the method the same way (a direct method call), and it sees only the `Task` object which is returned the same way (from that direct method call). The signature is entirely unaffected by the `async` modifier by design. It can be swapped in and out to suit the method implementer.

jnm2:
And then, given that the `async` keyword affects only the body, it would be nonsensical to allow the keyword on a method which has no body. So an implementation which ignored this nonsensical placement should be in error. The thing which the keyword refers to is not there.

Nigel-Ecma:
> What the async keyword does in a nutshell is rewrite the body of the method to use a state machine to produce the Task object (or some task-like object) which is returned by the method, rather than having the user manually return some Task object.

Where in your code:

> `async partial void M() { }`

is this “Task object” you mention specified, the signature seems to say it returns `void`.

Is it perhaps indicated by the `async`, which thereby informs that calling this method should be by way of something like an `await` and it returns a `Task` object?

So is this method perhaps not called like one taking no arguments and returning nothing as its “signature” might suggest?

Anyway, neither of us should be here today! Happy New Year, see you in 2026 🙂

jnm2:
Oops, I forgot that part. With `async void` all the same things hold true about its lack of effect on the signature:

- `async` enables the `await` keyword within the body and rewrites the body as a state machine, for example marshaling exceptions elsewhere
- `async` may be added or removed without affecting the caller in any way- the caller does not react to its presence

But the caller cannot observe when any asynchronous work is completed due to the return type being void. The method _is_ called like one taking no arguments and returning nothing, since that is what it is.

I appreciate that ☺️ Happy 2026 to you as well!

standard/classes.md
```diff
@@ -3034,7 +3034,8 @@ Partial methods shall not define access modifiers; they are implicitly private.
 
 There are two kinds of partial method declarations: If the body of the method declaration is a semicolon, the declaration is said to be a ***defining partial method declaration***. If the body is other than a semicolon, the declaration is said to be an ***implementing partial method declaration***. Across the parts of a type declaration, there shall be only one defining partial method declaration with a given signature, and there shall be at most only one implementing partial method declaration with a given signature. If an implementing partial method declaration is given, a corresponding defining partial method declaration shall exist, and the declarations shall match as specified in the following:
 
-- The declarations shall have the same modifiers (although not necessarily in the same order), method name, number of type parameters and number of parameters.
+- The declarations shall have the same method name, number of type parameters, and number of parameters.
+- The declarations shall have the same modifiers (although not necessarily in the same order), with the exception of the `async` modifier which may only appear on an implementing part.
```
```suggestion
- The declarations shall have the same modifiers (although not necessarily in the same order), with the exception of the `async` modifier which shall not appear on a defining part.
```

jskeet (approved):

PR #1535, opened 2026-01-19
[Version 10.0] Feature support for CallerArgumentExpression attribute

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/caller-argument-expression.md).

FYI, during testing, I discovered a feature that is *not* documented in the MS proposal: Not only is leading and trailing whitespace removed from an argument, but so too are any outermost grouping parens, with these two removal processes done repeatedly, as necessary, so an argument token set of ` ( ( (123) + 0) ) ` is reduced to `(123) + 0`.

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -800,13 +800,13 @@ Caller information is only substituted when a function is explicitly invoked in 
 
 One exception is query expressions. These are considered syntactic expansions, and if the calls they expand to omit optional parameters with caller-info attributes, caller information will be substituted. The location used is the location of the query clause which the call was generated from.
 
-If more than one caller-info attribute is specified on a given parameter, they are recognized in the following order: `CallerLineNumber`, `CallerFilePath`, `CallerMemberName`. Consider the following parameter declaration:
+If more than one caller-info attribute is specified on a given parameter, they are recognized in the following order: `CallerLineNumber`, `CallerFilePath`, `CallerMemberName`, `CallerArgumentExpression`. Consider the following parameter declaration:
 
 ```csharp
 [CallerMemberName, CallerFilePath, CallerLineNumber] object p = ...
 ```
 
-`CallerLineNumber` takes precedence, and the other two attributes are ignored. If `CallerLineNumber` were omitted, `CallerFilePath` would take precedence, and `CallerMemberName` would be ignored. The lexical ordering of these attributes is irrelevant.
+`CallerLineNumber` takes precedence, and the other three attributes are ignored. If `CallerLineNumber` were omitted, `CallerFilePath` would take precedence, and `CallerMemberName` and `CallerArgumentExpression` would be ignored. The lexical ordering of these attributes is irrelevant.
 
 #### 23.5.6.2 The CallerLineNumber attribute
 
@@ -886,9 +886,9 @@ For an invocation that occurs within a local function or an anonymous function, 
 
 #### §callargexpattr The CallerArgumentExpression attribute
 
-The attribute `System.Runtime.CompilerServices.CallerArgumentExpression` is applied to a *target parameter*, and can result in the capture of the source-code text of a sibling parameter’s argument as a string, referred to here as the *captured string*.
+The attribute `System.Runtime.CompilerServices.CallerArgumentExpressionAttribute` is applied to a *target parameter*, and can result in the capture of the source-code text of a sibling parameter’s argument as a string, referred to here as the *captured string*.
 
-Except when it is the first parameter in an extension method, the target parameter shall have a *default_argument*.
+Except when it is the first parameter in an extension method, the target parameter shall have a *default_argument*. When applied to the first parameter of an extension method, the captured string is the source text of the receiver expression in an extension method invocation. If the method is invoked using static method syntax, the captured string is the argument corresponding to the first parameter.
 
 Consider the following method declaration:
 
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -888,7 +888,7 @@ For an invocation that occurs within a local function or an anonymous function, 
 
 The attribute `System.Runtime.CompilerServices.CallerArgumentExpression` is applied to a *target parameter*, and can result in the capture of the source-code text of a sibling parameter’s argument as a string, referred to here as the *captured string*.
 
-The target parameter shall have a *default_argument*.
+Except when it is the first parameter in an extension method, the target parameter shall have a *default_argument*.
 
 Consider the following method declaration:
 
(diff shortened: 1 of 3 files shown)


standard/attributes.md
```diff
@@ -845,6 +845,79 @@ For invocations that occur within declarations of instance constructors, static
...
+The attribute constructor takes an argument of type `string`. That string
+
+- Shall contain the name of a sibling parameter; otherwise, the attribute is ignored.
+- Shall omit the leading `@` from a parameter name having that prefix.
+
+A *parameter_list* may contain multiple target parameters.
+
+The type of the target parameter shall have a standard conversion from `string`.
+
+> *Note:* This means no user-defined conversions from `string` are allowed, and in practice means the type of such a parameter must be `string`, `object`, or an interface implemented by `string`. *end note*
```
KalleOlaviNiemitalo, line 920: Or `dynamic`.

standard/attributes.md
```diff
@@ -845,6 +845,79 @@ For invocations that occur within declarations of instance constructors, static
...
+
+If an explicit argument is passed for the target parameter, no string is captured, and that parameter takes on that argument’s value. Otherwise, the text for the argument corresponding to the sibling parameter is converted to a captured string, according to the following rules:
+
+- Leading and trailing white space is removed both before and after any outermost grouping parentheses are removed.
+- All outermost grouping parentheses are removed both before and after any leading and trailing white space is removed.
+- All other *input_element*s are retained verbatim (including white space, comments, *Unicode_Escape_Sequence*s, and `@` prefixes on identifiers).
```
KalleOlaviNiemitalo, line 926: Roslyn deletes some comments here:

```csharp

using System;
#nullable enable
class Test
{
    public static void M(int val = 0, [CallerArgumentExpression("val")] string? text = null)
    {
        Console.WriteLine($"val = {val}, text = <{text}>");
        M(/*a*/ ( /*b*/ 1 /*c*/ + /*d*/ 2 /*e*/ ) /*f*/);
    }
}
```

lowered to:

```csharp
      M(3, "1 /*c*/ + /*d*/ 2");
```

KalleOlaviNiemitalo:
I'm seeing odd interactions with `params`, may be a compiler bug:

```csharp
#nullable enable
class Test
{
    string M(
        int first,
        [CallerArgumentExpression("vals")] string text = "N/A",
        params int[] vals)
    {
        return text;
    }

    static void N()
    {
        string a = new Test().M(1);
        string b = new Test().M(2, vals: new[] { 3 });
        string c = new Test().M(4, vals: 5);
    }
}
```

Lowered to:

```csharp
    static void N()
    {
        // OK: the default value of text is used, because there is no argument for vals.
        string a = new Test().M(
            first: 1,
            text: "N/A",
            vals: System.Array.Empty<int>());

        // OK: the argument expression for vals is used.
        string b = new Test().M(
            first: 2,
            text: "new[] { 3 }",
            vals: new int[1] { 3 });

        // What on earth?
        // I expected text: "5".
        string c = new Test().M(
            first: 4,
            text: "new Test().M(4, vals: 5)",
            vals: new int[1] { 5 });
    }
```

Likewise in an attribute:

```csharp
using System;

class TestAttribute: Attribute
{
    public TestAttribute(
        [CallerArgumentExpression("val")] string text = "N/A",
        params int[] val)
    {
    }
    
    public char C { get; set; }
}

// [Test("Test(val: 1, C = 'x')", new int[] { 1 })]
[Test(val: 1, C = 'x')]
class Test
{
}
```

standard/attributes.md
KalleOlaviNiemitalo, line 926: *input_element* does not include *PP_Directive* but AFAICT those are also retained, if they occur within the argument expression.

```suggestion
- All other *input_element*s and *PP_Directive*s are retained verbatim (including white space, comments, *Unicode_Escape_Sequence*s, and `@` prefixes on identifiers).
```

KalleOlaviNiemitalo (changes requested):

standard/attributes.md
```diff
@@ -845,6 +845,79 @@ For invocations that occur within declarations of instance constructors, static
...
+A *parameter_list* may contain multiple target parameters.
+
+The type of the target parameter shall have a standard conversion from `string`.
+
+> *Note:* This means no user-defined conversions from `string` are allowed, and in practice means the type of such a parameter must be `string`, `object`, or an interface implemented by `string`. *end note*
+
+If an explicit argument is passed for the target parameter, no string is captured, and that parameter takes on that argument’s value. Otherwise, the text for the argument corresponding to the sibling parameter is converted to a captured string, according to the following rules:
+
+- Leading and trailing white space is removed both before and after any outermost grouping parentheses are removed.
+- All outermost grouping parentheses are removed both before and after any leading and trailing white space is removed.
```
KalleOlaviNiemitalo, line 925: I suppose "grouping parentheses" here means that the parentheses are part of the same *parenthesized_expression*, rather than a *tuple_expression*:

```csharp
#nullable enable
class Test
{
    public static void M(
        (int, int) val,
        [CallerArgumentExpression("val")] string? text = null)
    {
    }
    
    static void N()
    {
        // text: "(1, 2)" rather than "1, 2"
        M(((1, 2)));
    }
}
```

(In the grammar, a *deconstruction_tuple* likewise has parentheses, but that cannot be an *argument_value*.)

standard/attributes.md
```diff
@@ -845,6 +845,79 @@ For invocations that occur within declarations of instance constructors, static
...
+- Shall contain the name of a sibling parameter; otherwise, the attribute is ignored.
+- Shall omit the leading `@` from a parameter name having that prefix.
+
+A *parameter_list* may contain multiple target parameters.
+
+The type of the target parameter shall have a standard conversion from `string`.
+
+> *Note:* This means no user-defined conversions from `string` are allowed, and in practice means the type of such a parameter must be `string`, `object`, or an interface implemented by `string`. *end note*
+
+If an explicit argument is passed for the target parameter, no string is captured, and that parameter takes on that argument’s value. Otherwise, the text for the argument corresponding to the sibling parameter is converted to a captured string, according to the following rules:
```
KalleOlaviNiemitalo, line 922: Apparently, "the text for the argument" means the *expression* or *variable_reference* part of *argument_value*, or the *attribute_argument_expression*.  It would be good to have an example that shows the `in`/`out`/`ref` part of *argument_value* is not included in the text.

standard/attributes.md
KalleOlaviNiemitalo: The "First-class Span Types" proposal will also add a standard implicit conversion from `string` to `ReadOnlySpan<char>`.

standard/attributes.md
RexJaeschke: @KalleOlaviNiemitalo At a glance, this suggests that comments before the first token (redundant parens excluded) and after the last token are treated as whitespace---so are removed---while those between tokens are preserved. This follows the rule I wrote at Line 887 above.

KalleOlaviNiemitalo: White space is defined in 6.3.4 and does not include comments.

BillWagner:
Thanks!

PR #1538, opened 2026-01-21
Update “tuple” handling

Nigel-Ecma:
Addresses issue #1155
Replaces PR #1366

A “surgical” fix for issue 1155 was outlined and later tried in PR 1366 (spun off from an earlier PR), unfortunately it did not work (mea culpa). This PR is the less surgical fix…

- Fixes a number of issues with the description of “tuples” existing prior to issue 1155, and the issue itself
- Split “tuples” into tuple_literal, local_deconstructing_declaration, and deconstructing_assignment to enable fixes and improve clarity
- Split assignment into 4 subrules (rather than just four alternatives) to support above and improve clarity
- Update samples as needed
- Embed HTML comments covering how to update the additions for C#10.


jskeet:
> * Fixes a number of issues with the description of “tuples” existing prior to issue 1155, and the issue itself

Do we have a concrete list of those issues somewhere? Given the size of this change, it would be nice to have a really easy-to-reference list of issues that it fixes.

Nigel-Ecma:
Here is an overview of the PR (this list is not guaranteed to be complete):

- The syntax pattern `( <item>, <item>, … )` occurs in multiple places in C# with different meanings, e.g. parameter lists and tuples. The original presentation of “tuples” covered three different meanings using a single set of grammar rules and associated text. Unfortunately certain restrictions, such as where a *declaration_expression* is valid, where omitted in this compound description. By separating out the three meanings into *tuple_literal*, *local_deconstructing_declaration*, and *deconstructing_assignment* the PR addresses these issues using a combination of grammar and text.
- Unfortunately it was not always clear when the described manipulation of tuple constructions was occurring at compile or execution time. The PR addresses this, for an example see §12.7.
- Tuples are (currently?) unique among the language specified compound types in requiring a certain underlying implementation using a family of `ValueTuple` types (whose underlying implementation is left to the the implementation). The description of this requirement was distributed and incomplete; in particular while all tuples have a unique `ValueTuple` representation the reverse is *not* true and how to handle a `ValueTuple` value which is not equivalent to a tuple was unfortunately omitted. The PR concentrates the relationship into one clause and covers the missing cases.
- The conversions available for tuples was unfortunately incomplete, an issue raised by #1155. This PR addresses this,  and some reorganisation and clarification of the material was required to enable this.
- In destructions it was unfortunately not clear what the required order of evaluation of elements was, or whether they even needed to be evaluated.
- For discards it was not clear that they must have a type, e.g. it is invalid as part of a destruction to match a discard to a `null` as, though the value is unused, the discard has no type.
- The PR adds details on when implementations may elide intermediate tuples while maintaining the defined semantics of the language. For example `(a, b) = (b, a);` does not *require* tuple construction and deconstruction.
- Over time the number of uses & meanings of assignment in the language has increased and the text, and particularly the grammar, was unfortunately no longer as clear was one would wish. Tuples and destruction only added to this so as well as adding the new grammar and text required the PR does some reorganisation to clarify the grammar and make the text clearer.
- Overall the PR uses more precise grammar and reorganisaton of the text to improve the clarity of the Standard in this area.

In addition to these other issues were addressed as they came to light during development.

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
 > The tuple type for `pair4` is valid because the names `Item1` and `Item2` match their positions, whereas the tuple type for `pair5` is disallowed, because the names `Item2` and `Item123` do not.
 >
-> The declarations for `pair6` and `pair7` demonstrate that tuple types are interchangeable with constructed types of the form `ValueTuple<...>`, and that the `new` operator is allowed with the latter syntax.
->
 >The last line shows that tuple elements can be accessed by the `Item` name corresponding to their position, as well as by the corresponding tuple element name, if present in the type.
 > *end example*
 
+#### §eliding-tuples Eliding intermediate tuple creation
+
+If the result of constructing a tuple (§12.8.6) is not required outside of the context in which it is constructed implementations are explicitly allowed to elide the construction as an optimisation provide all other semantic requirements are met.
```
```suggestion
If the result of constructing a tuple (§12.8.6) is not required outside of the context in which it is constructed implementations are explicitly allowed to elide the construction as an optimisation provided all other semantic requirements are met.
```

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+> *Example*: Such a situation may commonly arise from deconstructing assignments ($deconstructing-assignment), deconstructing declarations (§local-deconstructing-declarations), and switch statements (§13.8.3). Consider the deconstructing assignment:
+>
+> ```csharp
+>    (a, b) = (b, a);
+> ```
+>
+> The right hand side `(b, a)` constructs a tuple containing the values of `b` & `a`. The left hand side deconstructor `(a, b)` then, in order, selects the first item of that tuple and assigns it to `a`, followed by assigning the second item to `b`. The overall result is the the values in `a` & `b` are exchanged, while the tuple created during this process is discarded. The explicit allowance granted here to elide such intermediate tuple construction allows an implementation to exchange the two values in whatever ways it chooses provide it evaluates `b` before `a` to meet the left-to-right evaluation order of tuple literal elements. In the code:
+>
+> ```csharp
+>    (a, b, _) = (b, a, thing.expensiveMethod(x));
```
```suggestion
>    (a, b, _) = (b, a, thing.ExpensiveMethod(x));
```

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+> ```
+>
+> An implementation can also choose to exchange the two values without constructing the tuple provided the tuple elements are evaluated in order: `b`, `a` and `thing.expensiveMethod(x)`; before doing so. *end example*
```
```suggestion
> An implementation can also choose to exchange the two values without constructing the tuple provided the tuple elements are evaluated in order: `b`, `a` and `thing.ExpensiveMethod(x)`; before doing so. *end example*
```

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+> *Note*: If an implementation elides an intermediate tuple it may also be able to elide now “redundant” (no effect) expressions. For example if an intermediate tuple is the result of an implicit tuple conversion, those implicit conversions have no side effects, the intermediate tuple is subject to deconstruction where some elements are discarded, then it may be possible to elide the implicit conversion of those discarded elements. *end note*
```
```suggestion
> *Note*: If an implementation elides an intermediate tuple it may also be able to elide now “redundant” (no effect) expressions. For example if an intermediate tuple is the result of an implicit tuple conversion, those implicit conversions have no side effects, and the intermediate tuple is subject to deconstruction where some elements are discarded, then it may be possible to elide the implicit conversion of those discarded elements. *end note*
```

(I think, anyway.)

standard/expressions.md
```diff
@@ -1260,18 +1261,33 @@ In these situations, the boxed instance is considered to contain a variable of t
...
+  > } 
+  >```
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+  > *Note*: “is semantically equivalent to” – there is no requirement, or even expectation, that an implementation will actually implement this deconstruction by defining a method and calling it, just that the observable semantics follow this sample. (Inlining a method is not “observable” in this sense.) *end note*
+
+If none of the above hold `E` cannot be deconstructed, which is a compile-time error.
+
+> *Note*: When deconstruction is used in a context; such as part of a deconstructing assignment or local deconstructing declaration; where only individual elements of the *tuple-literal* are used then an implementation is explicitly allowed to elide creating a tuple value from the *tuple-literal* provided the observable semantics remain the same, as specified in *Eliding tuples* (§eliding-tuples). *end note*
```
```suggestion
> *Note*: When deconstruction is used in a context where only individual elements of the *tuple-literal* are used (such as part of a deconstructing assignment or local deconstructing declaration) then an implementation is explicitly allowed to elide creating a tuple value from the *tuple-literal* provided the observable semantics remain the same, as specified in *Eliding tuples* (§eliding-tuples). *end note*
```

standard/statements.md
```diff
@@ -481,6 +482,124 @@ The scope of a local constant is the block in which the declaration occurs. It i
 
 A local constant declaration that declares multiple constants is equivalent to multiple declarations of single constants with the same type.
 
+### §local-deconstructing-declarations Local deconstructing declarations
```
jskeet, line 485: Was this just not valid at all before? Were we missing the feature? Or is there some subtlety of the change which means what is now a local deconstructing declaration was previously something else? I'd like to understand this more.

standard/types.md
Nigel-Ecma: Thanks, in next commit.

standard/types.md
Nigel-Ecma: Thanks, in next commit.

standard/types.md
Nigel-Ecma: Thanks, in next commit.

standard/statements.md
jskeet: I've figured this out now - it was in tuple_expression

standard/types.md
Nigel-Ecma: You’re right, I’ll put it in next commit.

standard/expressions.md
Nigel-Ecma: Thanks, in next commit

standard/statements.md
Nigel-Ecma: @jskeet – Deconstructing declarations & assignment were effectively combined in the old text, however the restriction that a *declaration_expression* could only occur in a deconstruction if it was the first in a statement was missing. E.g. something like:

```csharp
(a, _) = (_, int b) = (42, 24)
```

is invalid but was allowed by the old text. If you need to do this the order must be swapped:

```csharp
(_, int b) = (a, _) = (42, 24)
```

which is deconstructing declaration initialised from a deconstructing assignment.

In C#10 this example can become:

```csharp
(a, int b) = (42, 24)
```

(the changes to support this are in this PR as HTML comments). However this doesn’t remove the restriction that a deconstructing assignment cannot contain a *declaration_expression*.

Hope that explains it, let me know if it doesn’t!

standard/expressions.md
```diff
@@ -6841,25 +6811,25 @@ When a property or indexer declared in a *struct_type* is the target of an assig
 >
 > struct Rectangle
 > {
->     Point a, b;
```
jskeet, line 6844: This aligns with Point above - but it's Point that's wrong in using a 3-space padding, I believe.
(I'm creating a PR to apply all typo/whitespace changes so that *this* PR can have a smaller diff.)

standard/statements.md
jskeet: Yup, it absolutely does :)

standard/conversions.md
```diff
@@ -345,10 +345,16 @@ In all cases, the rules ensure that a conversion is executed as a boxing convers
...
+An implicit conversion exists from a tuple literal `e`, of the form `(e₁, ..., eₙ)`, to a tuple value of type `T = (T₁, ..., Tₙ)` if there is an implicit conversion for each element `eᵢ` to the corresponding element type `Tᵢ`. The result of the conversion is the tuple value `((T₁)e₁, ..., (Tₙ)eₙ)` of type `T`.
 
-If an element name in the tuple expression does not match a corresponding element name in the tuple type, a warning shall be issued.
+If an element name in the tuple literal does not match a corresponding element name in the tuple type, a warning shall be issued.
 
+An implicit conversion exists from the tuple type `T = (T₁, ..., Tₙ)` to the tuple type `S = (S₁, ..., Sₙ)` if there is an implicit conversion from each `Tᵢ` to the corresponding `Sᵢ`. The result of this conversion when applied to a tuple `t` with value `(t₁, ..., tₙ)` and of type `T` is the tuple value `((S₁)t₁, ..., (Sₙ)tₙ)` of type `S`.
```
jskeet, line 352: If we were to take this single change, wouldn't that fix #1155 in a very simple way?

I completely understand that it wouldn't fix other aspects, but I'm trying to chip away at this. Would there be any problems caused by merging this one change in isolation?

cc @BillWagner

standard/statements.md
jskeet: Actually, I'm now confused again.

Leaving discards aside, tuple_expression didn't include what I thought it did. Misreading the current spec, I'd thought that `(int x, int y)` was a tuple expression, when it isn't.

I don't know how:

```csharp
(int x, int y) = (3, 5);
```

... is parsed at the moment. 12.23.2 states:

> A simple assignment where the left operand is classified as a tuple is also called a deconstructing assignment. If any of the tuple elements of the left operand has an element name, a compile-time error occurs. If any of the tuple elements of the left operand is a declaration_expression and any other element is not a declaration_expression or a simple discard, a compile-time error occurs.

... but it's not clear to me how that works with the above example. I don't *think `(int x, int y)` is a tuple_expression, but is it something else that is "classified as a tuple"?

jskeet: Ah, no, it *is* a *tuple_expression* because it's fine for a *tuple_element* just to be an *expression*, and a *declaration_expression* is an *expression*.

standard/expressions.md
```diff
@@ -1576,97 +1592,61 @@ parenthesized_expression
...
 
-### 12.8.6 Tuple expressions
+### 12.8.6 Tuple literals
 
-A *tuple_expression* represents a tuple, and consists of two or more comma-separated and optionally-named *expression*s enclosed in parentheses. A *deconstruction_expression* is a shorthand syntax for a tuple containing implicitly typed declaration expressions.
+A *tuple_literal* represents a tuple, and consists of two or more comma-separated and optionally-named *expression*s enclosed in parentheses.
 
 ```ANTLR
-tuple_expression
+tuple_literal
```
jskeet, line 1600: This still allows `(int x, int y)` to be a tuple_literal. Is that intentional?

standard/types.md
```diff
@@ -403,22 +403,24 @@ An enumeration type is a distinct type with named constants. Every enumeration t
...
-There does not need to exist a `System.ValueTuple<...>` declaration that directly matches the arity of any tuple type with a corresponding number of type parameters. Instead, tuples with an arity greater than seven are represented with a generic struct type `System.ValueTuple<T1, ..., T7, TRest>` that in addition to tuple elements has a `Rest` field containing a nested value of the remaining elements, using another `System.ValueTuple<...>` type. Such nesting may be observable in various ways, e.g. via the presence of a `Rest` field. Where only a single additional field is required, the generic struct type `System.ValueTuple<T1>` is used; this type is not considered a tuple type in itself. Where more than seven additional fields are required, `System.ValueTuple<T1, ..., T7, TRest>` is used recursively.
+Element names within a tuple type shall be distinct. A tuple element name of the form `ItemX`, where `X` is any sequence of decimal digits with no leading zeros, is only permitted at the position denoted by `X`.
```
BillWagner, line 410: Should add that these positions start at 1, not 0:

```suggestion
Element names within a tuple type shall be distinct. A tuple element name of the form `ItemX`, where `X` is any sequence of decimal digits with no leading zeros, is only permitted at the position denoted by `X`. The position indexes start at 1.
```

standard/types.md
```diff
@@ -403,22 +403,24 @@ An enumeration type is a distinct type with named constants. Every enumeration t
...
 
-Element names within a tuple type shall be distinct. A tuple element name of the form `ItemX`, where `X` is any sequence of non-`0`-initiated decimal digits that could represent the position of a tuple element, is only permitted at the position denoted by `X`.
+> *Note* This restriction on element names avoids any confusion between them and tuple field names, e.g. where element name `ItemX` is associated with field `ItemY` where `X ≠ Y`. *end note*
+<!-- markdownlint-disable MD028 -->
 
-The optional element names are not represented in the `ValueTuple<...>` types, and are not stored in the runtime representation of a tuple value. Identity conversions ([§10.2.2](conversions.md#1022-identity-conversion)) exist between tuples with identity-convertible sequences of element types.
+<!-- markdownlint-enable MD028 -->
+> *Note* The optional element names are not represented in the runtime representation (§tuple-runtime-representation) of a tuple value.
```
BillWagner, line 416: Even as a note, this could use a rewording. The optional element names need to be stored in metadata in some fashion. It's a requirement that a public API that returns a tuple, or has a tuple as a parameter includes the optional names in the metadata read from a referenced assembly. For example:

```csharp
public class Weather
{
    (DateTime time, double temperature) GetTimeAndTemperature((double latitude, double longitude) location)
    { ... }
    }
```

The member names of the tuples should be available to tools that reference the contained assembly.

standard/types.md
```diff
@@ -403,22 +403,24 @@ An enumeration type is a distinct type with named constants. Every enumeration t
...
 
-The `new` operator [§12.8.17.2](expressions.md#128172-object-creation-expressions) cannot be applied with the tuple type syntax `new (T1, ..., Tn)`. Tuple values can be created from tuple expressions ([§12.8.6](expressions.md#1286-tuple-expressions)), or by applying the `new` operator directly to a type constructed from `ValueTuple<...>`.
+Identity conversions ([§10.2.2](conversions.md#1022-identity-conversion)) exist between tuples with identity-convertible sequences of element types.
```
BillWagner: This doesn't mention arity. It probably should.

standard/types.md
```diff
@@ -403,22 +403,24 @@ An enumeration type is a distinct type with named constants. Every enumeration t
...
 
-Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can be accessed via a member access on a tuple value ([§12.8.7](expressions.md#1287-member-access). Additionally, if the tuple type has a name for a given element, that name can be used to access the element in question.
+Tuple values can be created from tuple literals ([§12.8.6](expressions.md#1286-tuple-literals)), or by creating a value using the underlying runtime representation (§tuple-runtime-representation) directly. The tuple type syntax `(T1, ..., Tn)` cannot be used with the `new` operator [§12.8.17.2](expressions.md#128172-object-creation-expressions).
+
+Tuple elements are public fields with the names `Item1` … `ItemN`, where `N` is the tuple arity and the numbers have no leading zeros, and can be accessed via a member access on a tuple value ([§12.8.7](expressions.md#1287-member-access). Additionally, if the tuple type has a name for a given element, that name can be used to access the element in question.
```
BillWagner, line 422: I'd like to reverse the emphasis here: The purpose of the tuple syntax is to provide a lightweight data container type. Without the optional names, no syntax is needed! System.ValueTuple. Perhaps something like:

```suggestion
If the tuple type has a name for a given element, that name can be used to access the element in question. Tuple elements are public fields with the names `Item1` … `ItemN`, where `N` is the tuple arity and the numbers have no leading zeros, and can be accessed via a member access on a tuple value ([§12.8.7](expressions.md#1287-member-access).
```

standard/expressions.md
```diff
@@ -1576,97 +1592,61 @@ parenthesized_expression
...
->
-> <!-- Example: {template:"standalone-console-without-using", name:"DiscardExpressions3",expectedErrors:["CS8130","CS8130", "CS8130"]} -->
-> ```csharp
-> (int a, string? b) = (42, null); //OK
-> var (c, d) = (42, null); // Invalid as type of d cannot be inferred
-> (int e, var f) = (42, null); // Invalid as type of f cannot be inferred
-> ```
->
-> *end example*
+A *tuple_literal* is classified as a tuple.
```
BillWagner: Should this be a "tuple type" and link to 8.3.11 (or a sub-clause)?

standard/expressions.md
```diff
@@ -1576,97 +1592,61 @@ parenthesized_expression
...
+A tuple literal has a type if and only if each of its element expressions `Eᵢ` has a type `Tᵢ`. The type is a tuple type of the same arity as the tuple literal, where each element is given by the following:
 
-- If the tuple element in the corresponding position has a name `Ni`, then the tuple type element shall be `Ti Ni`.
-- Otherwise, if `Ei` is of the form `Ni` or `E.Ni` or `E?.Ni` then the tuple type element shall be `Ti Ni`, *unless* any of the following holds:
-  - Another element of the tuple expression has the name `Ni`, or
-  - Another tuple element without a name has a tuple element expression of the form `Ni` or `E.Ni` or `E?.Ni`, or
-  - `Ni` is of the form `ItemX`, where `X` is a sequence of non-`0`-initiated decimal digits that could represent the position of a tuple element, and `X` does not represent the position of the element.
-- Otherwise, the tuple type element shall be `Ti`.
+- If the tuple element in the corresponding position has a name `Nᵢ`, then the tuple type element shall be `Tᵢ Nᵢ`.
```
BillWagner: should `Ni` be a simple name instead of a name?

standard/expressions.md
```diff
@@ -1576,97 +1592,61 @@ parenthesized_expression
...
+> var t3 = (i: 3, null);              // Error: no type
 > ```
 >
-> In this example, all four tuple expressions are valid. The first two, `t1` and `t2`, do not use the type of the tuple expression, but instead apply an implicit tuple conversion. In the case of `t2`, the implicit tuple conversion relies on the implicit conversions from `2` to `long` and from `null` to `string`. The third tuple expression has a type `(int i, string)`, and can therefore be reclassified as a value of that type. The declaration of `t4`, on the other hand, is an error: The tuple expression has no type because its second element has no type.
+> In this example of three declaration statements, the initialising tuple literals themselves are all valid. In the case of `t1`, the implicit tuple conversion relies on the implicit conversions from `2` to `long` and from `null` to `string`. The second tuple literal has a type `(int i, string)`, and can therefore be reclassified as a value of that type. The declaration of `t3`, on the other hand, is an error: the tuple literal has no type because its second element has no type and reclassifying at as a value therefore fails.
```
BillWagner: nit:

```suggestion
> In this example of three declaration statements, the initialising tuple literals themselves are all valid. In the case of `t1`, the implicit tuple conversion relies on the implicit conversions from `1` to `long` and from `null` to `string`. The second tuple literal has a type `(int i, string)`, and can therefore be reclassified as a value of that type. The declaration of `t3`, on the other hand, is an error: the tuple literal has no type because its second element has no type and reclassifying at as a value therefore fails.
```

standard/expressions.md
```diff
@@ -6890,11 +6860,74 @@ When a property or indexer declared in a *struct_type* is the target of an assig
...
+  - otherwise `dᵢ` and `eᵢ` are incompatible, and a compile-time error results.
+
+The run-time processing of a deconstructing assignment, now `d = e`, proceeds as follows:
+
+1. In the following steps each and every `dᵢ` and `eᵢ` must be evaluated exactly once in order left to right. For the avoidance any confusion this means all `dᵢ` are evaluated before any `eᵢ`; and if any nested *deconstructor*s or *tuple_literal*s are present the order is depth-first.
+2. The tuple value `t`, of the form `(t₁, ..., tₙ)`, is created by converting `e` to `T` using an implicit tuple conversion (§10.2.13).
+3. For each non-discard `dᵢ` in order from left to right:
+   - if `dᵢ` is a variable reference the assignment `dᵢ = tᵢ` is performed;
+   - otherwise `dᵢ` is a nested *deconstructor* and this step (3) is recursively applied to the elements of `dᵢ` and `tᵢ`.
+4. The type and value of the whole *deconstructing_assignment* expression are respectively `T` and `t`.
```
BillWagner, line 6912: Markdown nit:  If all these use "1. ", the markdown processor numbers them, and we won't have extra diffs in the future if we add new steps in the middle somewhere.

```suggestion
1. In the following steps each and every `dᵢ` and `eᵢ` must be evaluated exactly once in order left to right. For the avoidance any confusion this means all `dᵢ` are evaluated before any `eᵢ`; and if any nested *deconstructor*s or *tuple_literal*s are present the order is depth-first.
1. The tuple value `t`, of the form `(t₁, ..., tₙ)`, is created by converting `e` to `T` using an implicit tuple conversion (§10.2.13).
1. For each non-discard `dᵢ` in order from left to right:
   - if `dᵢ` is a variable reference the assignment `dᵢ = tᵢ` is performed;
   - otherwise `dᵢ` is a nested *deconstructor* and this step (3) is recursively applied to the elements of `dᵢ` and `tᵢ`.
1. The type and value of the whole *deconstructing_assignment* expression are respectively `T` and `t`.
```

standard/statements.md
```diff
@@ -481,6 +482,124 @@ The scope of a local constant is the block in which the declaration occurs. It i
...
+    | discard_token
+    | declaration_deconstructor
+    ;
+```
+
+If the input can be recognised as both a *local_deconstructing_declaration* and a *deconstructing_assignment* (§deconstructing-assignment) then the latter shall be chosen.
+
+> *Note*: Put another way a *local_deconstructing_declaration* must contain at least one *declaration_expression* to be recognised as a declaration, otherwise it is recognised as a *deconstructing_assignment*. This follows the same split as for declaration vs. assignment for other types. *end note*
+
+If the input can be syntatically recognized as a *declaration_deconstructor* and as a *variable_reference* then the former shall be chosen. Likewise for *discard_token* and *variable_reference*.
```
BillWagner: I'd prefer it just repeated. When I read it first, I thought it meant 3 repeated rules, where *discard_token* and *variable_reference* were substituted in turn for *local_deconstructing_declaration*. That didn't process well, but it took me longer than I'd like to come to that conclusion.

```suggestion
If the input can be syntatically recognized as a *declaration_deconstructor* and as a *variable_reference* then the former shall be chosen. If the input can be syntatically recognized as a *discard_token* and as a *variable_reference* then the former shall be chosen.
```

BillWagner (commented):
I gave it a good first read. I like the direction.

standard/expressions.md
Nigel-Ecma: You’ve missed that *declaration_expression* is not an *expression* any more and *tuple_literal* does not therefore allow them.

The original grammar allowed *declaration_expression*s all over the place and the text describing them (§12.19) had semantic rules restricting them to 2½† of those places. This was an implementation grammar approach that slipped into our (reference) grammar and was on my list to fix, and as it is key to tuples/deconstructors now was the time. It is fixed simply by including the rule directly in the places it was valid – *argument_value* and *declaration_deconstructor_element* – and minor edits to §12.19 converting the semantic rules into just a list of where it is used.

(If in future it is allowed in more places it can be added, just like for other constructs.)

---
† The grammar rule *declaration_expression* was only valid in 2 places, the ½ comes from when a discard is classified as a declaration expression.

standard/conversions.md
Nigel-Ecma: The short answer is it didn’t. I tried it at the start of this second attempt to fix the issue and it failed for some reason; so then accepted that I had to address the other issues from the ground up, and when I arrived back at this one I fixed it.

standard/expressions.md
Nigel-Ecma: OK, I will post a commit of the changes from the reviews so far, and then look at the new PR/rebase.

standard/types.md
Nigel-Ecma: I’m going to disagree with this one. Here we are talking about element names already denoted by `I1...In` associated with types `T1...Tn`, the subscript denoting their position. There is no zeroth position. At first I considered adding it would be harmless, but had second thoughts over combining “position” and “index” as that could be confusing. So I remain to be persuaded on this one.

standard/types.md
Nigel-Ecma: This was not in the original text. I would expect any implementation to make available the names in its debugger and other tools, but that is surely out of scope here?

standard/types.md
Nigel-Ecma: I’ll put:

> Identity conversions ([§10.2.2](conversions.md#1022-identity-conversion)) exist between tuples of the same arity with identity-convertible sequences of element types.


in the next commit.

standard/types.md
Nigel-Ecma: Apologies but I don’t follow this one. The tuple type & literal *syntax* is required or neither could be written, the names are *optional* and hence surely secondary?

standard/expressions.md
Nigel-Ecma: It is a reference to §12.2.1 (Expression classifications -> General), I'll add it in the next commit.

standard/expressions.md
Nigel-Ecma: @BillWagner  – Bar for the proper subscript and expression/literal changes this is the original text. It has used “name” to mean “element name”, which are of course *identifier*s; I honestly did not consider changing this, it read fine to me, but from your comment I can see it might be misread.

If I change all the Nᵢ to Iᵢ and change “a name” to ”an element name” would that be better?

> - If the tuple element in the corresponding position has an element name `Iᵢ`, then the tuple type element shall be `Tᵢ Iᵢ`.
> - Otherwise, if `Eᵢ` is of the form `Iᵢ` or `E.Iᵢ` or `E?.Iᵢ`, where `Iᵢ` is an *identifier*,  then the tuple type element shall be `Tᵢ Iᵢ`, *unless* any of the following holds:
>   - Another element of the tuple literal has the element name `Iᵢ`, or
>   - Another tuple element without an element name has a tuple element expression of the form `Iᵢ` or `E.Iᵢ` or `E?.Iᵢ`, or
>   - `Iᵢ` is of the form `ItemX`, where `X` is a sequence of decimal digits, without any leading zeros, that could represent the position of a tuple element, and `X` does not represent the position of the element.
> - Otherwise, the tuple type element shall be `Tᵢ`.

In the first bullet I didn’t specify that `Iᵢ` is an identifier as that comes from the definition, but I added the qualification in the second bullet. (The `Iᵢ` are only used in this list and not referenced elsewhere in the text so just changing the list is sufficient.)

I’ll put this in the next commit and you (and others) can raise any concerns you still have or request it is reverted.

standard/expressions.md
Nigel-Ecma: Well spotted! Thanks, in next commit

standard/expressions.md
Nigel-Ecma: @BillWagner – I see your point but in this case I’ll disagree. Numbered lists allow the points to be referenced, if you don’t know the number that is a bit tricky. In this case (3) has two sub-bullets and the second one references “this step” which is qualified with (3) so it is clear it means the top-level bullet and is not a self-reference. In a system more powerful than MD we might do better.

standard/statements.md
Nigel-Ecma: Fair enough, too much economy of words! In the next commit.

Nigel-Ecma:
@jskeet & @BillWagner – I’ve pushed a commit with changes based on your reviews. I’ve gone through and marked all as resolved to remove clutter, but if you’ve missed any of the responses or want to follow up do go ahead and unresolve threads.

Coming next: the rebase to Jon’s PR…

jskeet:
@Nigel-Ecma Are you sufficiently happy that #1560 has all the text from this that we can close this?

Nigel-Ecma:
Replaced by #1560

PR #1540, opened 2026-01-22
[Version 10.0] Feature support for file-scoped namespaces

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/file-scoped-namespaces.md).

This feature allows

```csharp
namespace ns;
...
```

as an abbreviation for 

```csharp
namespace ns {
    ...
}
```

As we now have two ways to declare a namespace, different compilation units can use different ways of declaring the same namespace, and we have to make sure the spec allows this. I've attempted to do that, but please read my edits carefully.

Also, I have introduced the grammar rule *compilation_unit_body* and I have attempted to use it instead of older text, but it's possible it could be used in more cases to make things simpler.


standard/basic-concepts.md
```diff
@@ -772,9 +772,9 @@ The ***scope*** of a name is the region of program text within which it is possi
...
+- The scope of a name defined or imported by a *using_directive* ([§14.5](namespaces.md#145-using-directives)) extends over the *global_attributes* and *compilation_unit_body* of the *compilation_unit* or *namespace_body* in which the *using_directive* occurs. A *using_directive* may make zero or more namespace or type names available within a particular *compilation_unit*, *namespace_body*, and *file_scoped_namespace_declaration* but does not contribute any new members to the underlying declaration space. In other words, a *using_directive* is not transitive but rather affects only the *compilation_unit*, *namespace_body*, and *file_scoped_namespace_declaration* in which it occurs.
```
RexJaeschke: I expect that eventually there will be a conflict generated by these bullet items. These edits are built on top of the edits for the V9 feature "Top-Level Statements" (see PR #[1454](https://github.com/dotnet/csharpstandard/pull/1454)). And just an hour ago, I tweaked that PR to include "*statement_list*s" in two places in these items. Ultimately, this PR will need to remove those bits of text as their inclusion is subsumed by this PR.

standard/namespaces.md
```diff
@@ -126,7 +175,7 @@ extern_alias_directive
     ;
 ```
 
-The scope of an *extern_alias_directive* extends over the *using_directive*s, *global_attributes* and *namespace_member_declaration*s of its immediately containing *compilation_unit* or *namespace_body*.
+The scope of an *extern_alias_directive* is described in [§7.7.1]( basic-concepts.md#771-general).
```
RexJaeschke, line 178: Here, I invoked the spec-writing rule, "Don't say the same thing in two different places!"

standard/namespaces.md
```diff
@@ -175,7 +224,7 @@ A *using_namespace_directive* ([§14.5.3](namespaces.md#1453-using-namespace-dir
 
 A *using_static_directive* ([§14.5.4](namespaces.md#1454-using-static-directives)) imports the nested types and static members of a type.
 
-The scope of a *using_directive* extends over the *namespace_member_declarations* of its immediately containing compilation unit or namespace body. The scope of a *using_directive* specifically does not include its peer *using_directive*s. Thus, peer *using_directive*s do not affect each other, and the order in which they are written is insignificant. In contrast, the scope of an *extern_alias_directive* includes the *using_directive*s defined in the same compilation unit or namespace body.
+The scope of a *using_directive* is described in [§7.7.1]( basic-concepts.md#771-general).
```
RexJaeschke, line 227: Here, I invoked the spec-writing rule, "Don't say the same thing in two different places!"

BillWagner:
Thanks!

PR #1543, opened 2026-01-23
[Version 10.0] Feature support for async method builder attribute

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/async-method-builders.md).

### DIFF
--- a/standard/classes.md
+++ b/standard/classes.md
@@ -5912,77 +5912,79 @@ If the return type of the async function is `void`, evaluation differs from the 
 
 This allows the context to keep track of how many `void`-returning async functions are running under it, and to decide how to propagate exceptions coming out of them.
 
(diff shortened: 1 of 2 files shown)


standard/classes.md
```diff
@@ -5886,6 +5886,78 @@ When a function is implemented using an iterator block, it is a compile-time err
...
+```
+
+However, the following code fragment:
+
+```csharp
+[AsyncMethodBuilder(typeof(PoolingAsyncValueTaskMethodBuilder<>))]
+static async ValueTask<int> ExampleAsync() { … }
+```
+
+in which the attribute `AsyncMethodBuilder ` is applied to that method, would instead be compiled to something like:
```
BillWagner: Extra space:


```suggestion
in which the attribute `AsyncMethodBuilder` is applied to that method, would instead be compiled to something like:
```

BillWagner:
Thanks!

PR #1544, opened 2026-01-23
[Version 10.0] Feature support for assignment and declaration in same deconstruction

RexJaeschke:
There is no corresponding MS proposal.

Prior to V10, the tuple elements in a deconstructing assignment all had to be *declaration_expression*s that designated new local variables, possibly with discards, **OR** they all had to be identifiers designating existing variables, possibly with discards, but not a combination of old and new variable designators.

V10 allows a combination, as in 

```csharp
int z;
(int x, _, z) = (2, 4, 6);
```

This is achieved simply by removing the restriction on allowing such a combination.

### DIFF
--- a/standard/README.md
+++ b/standard/README.md
@@ -606,6 +606,9 @@
     - [§15.6.7](classes.md#1567-abstract-methods)  Abstract methods
     - [§15.6.8](classes.md#1568-external-methods)  External methods
     - [§15.6.9](classes.md#1569-partial-methods)  Partial methods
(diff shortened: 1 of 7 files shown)


Nigel-Ecma:
The changes required for this are covered by #1538 (in review at the time or writing). That PR is for v8 but contains the changes required for v10 in HTML comments tagged with [TODO]. If #1538 is merged this issue becomes “apply the HTML comments for v10”.

PR #1547, opened 2026-01-24
Update V9 Foreword

RexJaeschke:
Bill, I simply replaced the v8 Foreword contents with the feature list from the v9 Admin status file.


BillWagner (approved):
This LGTM @RexJaeschke 

I'll :shipit: now.

PR #1548, opened 2026-01-24
Restore CallerMemberName example

RexJaeschke:
Jon, this addresses your [comment](https://github.com/dotnet/csharpstandard/pull/1466#discussion_r2690124552).

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -842,7 +842,38 @@ For invocations that occur within field or event initializers, the member name u
 
 For invocations that occur within declarations of instance constructors, static constructors, finalizers and operators the member name used is implementation-dependent.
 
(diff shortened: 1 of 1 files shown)


KalleOlaviNiemitalo:
Filed <https://github.com/dotnet/csharpstandard/issues/1549>.  But also:

```csharp
using System;

public class MAttribute : Attribute {
    public MAttribute(
        [CallerMemberName]string? name = null) {
    }
}

class C { 
    void Fun1() {
        void Fun2<[M]T>(){}
    }
}
```

SharpLab shows this being lowered to `void Fun2<[M("Fun2")]T>(){}` but the draft seems to require `void Fun2<[M("Fun1")]T>(){}` because Fun2 as a local function is not a function member.  Now is that a compiler bug or a spec bug?

jskeet:
@KalleOlaviNiemitalo That looks like a compiler bug to me, if the spec change here is otherwise correct. Please could you file another issue for that? And thanks for #1549 - I *think* we've previously discussed that and decided not to specify the behavior, but I could be wrong, and it's fine for us to discuss it in a meeting.

jskeet (approved):

BillWagner (approved):
This LGTM as well.

BillWagner (approved):
This LGTM

PR #1552, opened 2026-01-25
[Version 10.0] Feature support for improved interpolated strings

RexJaeschke:
This is Rex's adaptation of the following: [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/constant_interpolated_strings.md) and [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/improved-interpolated-strings.md). As the first is trivial and defines a term that is used by the second, Rex combined the two into this one PR.

Rex decided that the minimum set of members a robust handler must have are, as follows: one constructor having two `int` parameters, method `AppendLiteral`, a `ToString` method to get at the built string, and three `AppendFormatted` methods, so that handler can be called using all interpolated string formats a programmer might throw at it *without* requiring extra arguments to be passed to the constructor. That is, it can correctly handle formats with or without alignment. I’ve used the default handler member set in this decision. Is this OK?

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -1248,7 +1248,7 @@ val = {  FF}; 2 * val = 510.
 
 In the call to `LogMessage`, the target of the interpolated string expression argument is parameter `msg`, which has type `string`. As such, according to [§12.8.3](expressions.md#1283-interpolated-string-expressions), the default interpolated string expression handler is invoked. The following subclause (§custInterpStrExpCustHandling) shows how to use a custom handler.
 
-In order to provide custom processing to the program shown in §custInterpStrExpDefHandling, a *custom interpolated string expression handler* is needed. Here then is the message logger with a custom handler added (which although it does nothing more than behave like the default handler, it provides the hooks for customization):
+In order to provide custom processing to the program above, a *custom interpolated string expression handler* is needed. Here then is the message logger with a custom handler added (which although it does nothing more than behave like the default handler, it provides the hooks for customization):
 
 <!-- Example: {template:"standalone-console-without-using", name:"DeclCustomHandler2", inferOutput:true} -->
 ```csharp
@@ -1313,7 +1313,7 @@ The output produced is, as follows:
 val = {  FF}; 2 * val = 510.
 ```
 
-A type having the attribute `System.Runtime.CompilerServices.InterpolatedStringHandlerAttribute` is said to be an *applicable interpolated string handler type*. 
+A type having the attribute `System.Runtime.CompilerServices.InterpolatedStringHandlerAttribute` is said to be an *applicable interpolated string handler type*.
 
 To qualify as a custom interpolated string expression handler, a class or struct type shall have the following characteristics:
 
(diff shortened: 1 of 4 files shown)


standard/expressions.md
```diff
@@ -1571,6 +1594,14 @@ Then:
 
 *end example*
 
+A *constant interpolated string* is an *interpolated_string_expression* that contains
+  - no interpolations, or
+  - interpolations whose *expression*s are constant expressions of type `string`, and these interpolations have no *interpolation_minimum_width*, *Regular_Interpolation_Format*, or *Verbatim_Interpolation_Format* specifiers.
+
+For example, $"Hello", $"{cs1}, world!" (given `const string cs1 = $"Hello";`), $"{"Hello," + $"world!"}", and $"xxx{(true ? $"{"X"}" : $"{$"{"Y"}"}")}yyy" are all constant interpolated strings. However, $"{123}" and $"{"abc"}{123.45}" are not.
+
+For simplicity, the term *ISE* is used throughout this specification to mean either an *interpolated_string_expression* or an *additive_expression* composed entirely of *interpolated_string_expression*s and binary `+` operators.
```
RexJaeschke, line 1618: Is there a better name than *ISE*? I refer to it in a number of places.

standard/standard-library.md
```diff
@@ -784,6 +784,31 @@ namespace System.Linq.Expressions
...
+            int formattedCount);
+        public DefaultInterpolatedStringHandler(int literalLength,
+            int formattedCount, IFormatProvider? provider);
+        public DefaultInterpolatedStringHandler(int literalLength,
+            int formattedCount, IFormatProvider? provider, Span<char> initialBuffer);
+        public void AppendFormatted(scoped ReadOnlySpan<char> value);
+        public void AppendFormatted(string? value);
+        public void AppendFormatted(object? value, int alignment = 0,
+            string? format = default);
+        public void AppendFormatted(scoped ReadOnlySpan<char> value,
```
RexJaeschke, line 794: These signatures come from the .NET 10 docs; however, `scoped` isn't supported until C# V11. Is it simply a matter of removing that keyword from here and then restoring it in V11?

standard/attributes.md
```diff
@@ -1173,6 +1174,187 @@ A module initializer shall have the following characteristics:
...
+- Have an accessible constructor whose first two parameters have type `int`. (Other parameters may follow, which are used to pass information to/from the handler. These are discussed in §custInterpStrExpPassInfo. An optional final parameter may be declared to inhibit the handler from processing the interpolated string. This is discussed in §custInterpStrExpInhibCustHandler).
+
+When the compiler-generated code calls the constructor, the first parameter is set to the sum of the lengths of the interpolated string expression segments ([§12.8.3](expressions.md#1283-interpolated-string-expressions)) in the interpolated string expression, and the second parameter is set to the number of interpolations. (For `($"val = {{{val,4:X}}}; 2 * val = {2 * val}."`, these values are 21 and 2, respectively.)
+
+- Have an accessible method with the signature `void AppendLiteral(string s)`, which is called to process a single interpolated string expression literal segment.
+- Have a set of accessible overloaded methods called `AppendFormatted`, one of which is called to process a single interpolation, based on that interpolation’s content. Their signatures are, as follows:
```
RexJaeschke, line 1326: Re the handler constructor and the `AppendLiteral` and `AppendFormatted` methods, I’ve required that they be accessible rather than public. Is that OK?

BillWagner:
Thanks!

PR #1554, opened 2026-01-27
Describe the impact of ? on arrays of arrays

### LINKED ISSUE #1385: Arrays of nullable references

jskeet:
Alternative to #1386.

Fixes #1385.


jskeet:
I'm fully expecting there to be some more work here, but I'm hoping it's at least *close*.

jskeet:
@Nigel-Ecma: I *believe* the grammar test failure is now a matter of changing the expected tokens (due to the introduction of non_array_non_nullable_type) - let me know if this is something I can fix myself, or whether it needs your help :)

standard/types.md
```diff
@@ -732,7 +737,9 @@ There are two forms of nullability for reference types:
...
+The syntactic distinction between a *nullable reference type* and its corresponding *non-nullable reference type* enables a compiler to generate diagnostics. A compiler must allow the *nullable_type_annotation* as defined in [§8.2.1](types.md#821-general). The diagnostics must be limited to warnings. Other than in the meaning of array types, neither the presence or absence of nullable annotations, nor the state of the nullable context can change the compile time or runtime behavior of a program except for changes in any diagnostic messages generated at compile time.
+
+The meaning of array types is significantly impacted by the presence of *nullable_type_annotation* within an *array_type*, as described in [§17.2.1](arrays.md#1721-general).
```
BillWagner, line 849: I'd consider a different wording. The impact is that the memory layout changes. That's "significant" only if a program does anything that depends on the layout or order of multi-dimensional arrays. (Not an edge case, but not everything either).

One sugestion:

```suggestion
The memory layout of array types is impacted by the presence of *nullable_type_annotation* within an *array_type*, as described in [§17.2.1](arrays.md#1721-general).
```

standard/arrays.md
```diff
@@ -31,20 +31,64 @@ Every array type is a reference type ([§8.2](types.md#82-reference-types)). The
...
+The rank of an array type is given by the leftmost *rank_specifier* in the *array_type*: A *rank_specifier* indicates that the array is an array with a rank of one plus the number of “`,`” tokens in the *rank_specifier*.
 
-- An array type of the form `T[R]` is an array with rank `R` and a non-array element type `T`.
-- An array type of the form `T[R][R₁]...[Rₓ]` is an array with rank `R` and an element type `T[R₁]...[Rₓ]`.
+The element type of an array type is the type that results from deleting the leftmost *rank_specifier*.
 
 In effect, the *rank_specifier*s are read from left to right *before* the final non-array element type.
 
-> *Example*: The type in `T[][,,][,]` is a single-dimensional array of three-dimensional arrays of two-dimensional arrays of `int`. *end example*
+> *Example*: The following code shows several variable declarations, including a mixture of single-dimensional arrays, multi-dimensional arrays, and arrays of arrays, with some using nullable reference types. In each case, the rank and element type is described, and then demonstrated with a second variable declaration which is initialized using an element access expression.
```
BillWagner, line 44: Per my other comment, consider expanding the example to show the memory order of a `foreach` across the multiple dimensions of these arrays.

BillWagner (commented):
I like this approach. I have a couple comments to consider and discuss.

standard/types.md
jskeet: It's more than just the memory layout - it's what the element type of the array is. Memory layout makes it sound like too much like an implementation detail IMO.

I think the difference between declarations `string[][,] x` and `string[,][] y` is significant - which means the difference between `string[][,] x` and `string[]?[,] y` is also significant.

standard/arrays.md
jskeet: Let's chat about the first comment before making a change here :) (I suspect we might want a separate example, keeping this one just on "lots of examples with rank and element type".

standard/arrays.md
```diff
@@ -31,20 +31,64 @@ Every array type is a reference type ([§8.2](types.md#82-reference-types)). The
 
 The grammar productions for array types are provided in [§8.2.1](types.md#821-general).
 
-An array type is written as a *non_array_type* followed by one or more *rank_specifier*s.
+An array type is written as a *non_array_type* followed by one or more *rank_specifier*s, or an *array_type* followed by a *nullable_type_annotation* followed by one or more *rank_specifier*s. The latter production is used to represent array types where the element type is includes a nullable array type.
```
Nigel-Ecma: Extra word? Maybe the intention was:

```suggestion
An array type is written as a *non_array_type* followed by one or more *rank_specifier*s, or an *array_type* followed by a *nullable_type_annotation* followed by one or more *rank_specifier*s. The latter production is used to represent array types where the element type is a nullable array type.
```

What might be taken from a reading of this is an array element type is either a non-array or a nullable array, but not an array, but that has been a common fiction – see comment on line 40.

standard/arrays.md
```diff
@@ -31,20 +31,64 @@ Every array type is a reference type ([§8.2](types.md#82-reference-types)). The
...
 
 A *non_array_type* is any *type* that is not itself an *array_type*.
 
-The rank of an array type is given by the leftmost *rank_specifier* in the *array_type*: A *rank_specifier* indicates that the array is an array with a rank of one plus the number of “`,`” tokens in the *rank_specifier*.
+When determining the rank and element type of array type as specified below, only the *rank_specifier*s in the top-most production are considered, so in the production `array_type nullable_type_annotation rank_specifier+`, any `rank_specifier` within the `array_type` is ignored.
```
Nigel-Ecma: First I think the usual term would be outer-most rather than top-most. But maybe this could be written as the number of ranks ignoring any in the element type – as it is an array of element type after all!

standard/arrays.md
```diff
@@ -31,20 +31,64 @@ Every array type is a reference type ([§8.2](types.md#82-reference-types)). The
...
 
-The element type of an array type is the type that results from deleting the leftmost *rank_specifier*:
+The rank of an array type is given by the leftmost *rank_specifier* in the *array_type*: A *rank_specifier* indicates that the array is an array with a rank of one plus the number of “`,`” tokens in the *rank_specifier*.
```
Nigel-Ecma, line 40: I *think* you have:
- An array has *one* `[...]` – called a *rank specifier*;
  - i.e. there are __no__ *multiple rank specifier* arrays (they are arrays of arrays)
- A `[...]` has zero or more `,`s (commas)
  - The number of “gaps” (i.e. the number of commas + 1) is called the *rank*; and
  - and the things the *rank* is counting are called *dimensions* e.g. `[x,y]` has a rank of 2, has 2 dimensions, and is commonly referred to as a 2-dimensional array.

Is that the model you are describing?

standard/arrays.md
```diff
@@ -31,20 +31,64 @@ Every array type is a reference type ([§8.2](types.md#82-reference-types)). The
...
 
-- An array type of the form `T[R]` is an array with rank `R` and a non-array element type `T`.
-- An array type of the form `T[R][R₁]...[Rₓ]` is an array with rank `R` and an element type `T[R₁]...[Rₓ]`.
+The element type of an array type is the type that results from deleting the leftmost *rank_specifier*.
```
Nigel-Ecma: I don’t think this is clear, though offhand I’ve little idea how to fix it. By leftmost you mean the first one *after* the element type, which by line 38 you are ignoring. But here a straight reading of “The element type on an array ” does not ignore the element type.

Maybe you could stick with the previous `T[R][R₁]...[Rₓ]` where `T` is a *non_array_type* or *array_type nullable_type_annotation* but that sounds complicated…

standard/arrays.md
Nigel-Ecma: As per my comment on the other comment if anything is done here it would be the indexing, not memory, order.

standard/types.md
```diff
@@ -54,17 +54,22 @@ interface_type
     ;
 
 array_type
-    : non_array_type rank_specifier+
+    : array_type nullable_type_annotation rank_specifier+
+    | non_array_type rank_specifier+
     ;
 
 non_array_type
```
Nigel-Ecma, line 61: Is now the time to mention that *non_array_type* includes `System.Array` (a type required by C#) which is the supertype of all arrays?

Probably not ;-)

standard/types.md
Nigel-Ecma: I agree with @jskeet that memory layout is not the significant impact here, I expect few users even consider it per se. The most obvious impact is that the order in which a multi-rank array[1] is indexed is changed – add a nullable annotation to a multi-rank array type and every index operation needs to be changed, I’d call that significant! And probably a sufficiently error prone undertaking to recommend avoiding it.

So I would start with Jon’s original wording. Does it need to say that this holds even if an implementation does no nullable analysis and therefore effectively ignores all nullable annotations? I think it probably needs to say that *somewhere* and here seems a reasonable place.

---
[1] Yes, under the model being described there are (I think) no multi-rank arrays and so it is technically index operations which index multiple single rank arrays, but I guess you all knew that! :-)

standard/types.md
```diff
@@ -756,6 +763,8 @@ Throughout this specification, all C# code that does not contain nullable direct
 
 > *Note:* A nullable context where both flags are disabled matches the previous standard behavior for reference types. *end note*
 
+The rank and element of an array type declared using *nullable_type_annotation* is not affected by the nullable context.
```
Nigel-Ecma: Maybe add a suitable x-ref?

Nigel-Ecma (dismissed):
I’ve made a few comments but obviously there is a fair bit to go yet – e.g. conversions. Some of it is still confusing, but it seems a good start to dressing this up ;-)

Nigel-Ecma:
> @Nigel-Ecma: I _believe_ the grammar test failure is now a matter of changing the expected tokens (due to the introduction of non_array_non_nullable_type) - let me know if this is something I can fix myself, or whether it needs your help :)

With the proviso that I haven’t downloaded this and run the failing tests it does look like the errors reported are valid changes in the expect parse due to the revised grammar.

If this is the case then it will just need a few updates to the “reference” parse files (the ones that are machine generated and ignored in PR reviews). You can generate them yourself but you might prefer one of us with the tools already downloaded to run and produce them does it! It can wait until the PR is approved and *before* it is merged.

standard/types.md
jskeet: Resolving after discussion.

standard/types.md
```suggestion
The rank and element of an array type declared using *nullable_type_annotation* is not affected by the nullable context ([§8.9.4](types.md#894-nullable-context)).
```

standard/types.md
jskeet: Agreed, let's not do that here :)

standard/arrays.md
jskeet: Resolving after discussion.

standard/arrays.md
```diff
@@ -31,20 +31,64 @@ Every array type is a reference type ([§8.2](types.md#82-reference-types)). The
 
 The grammar productions for array types are provided in [§8.2.1](types.md#821-general).
 
-An array type is written as a *non_array_type* followed by one or more *rank_specifier*s.
+An array type is written as a *non_array_type* followed by one or more *rank_specifier*s, or an *array_type* followed by a *nullable_type_annotation* followed by one or more *rank_specifier*s. The latter production is used to represent array types where the element type is a nullable array type.
```
jskeet:
```suggestion
An array type is written as a *non_array_type* followed by one or more *rank_specifier*s, or an *array_type* followed by a *nullable_type_annotation* followed by one or more *rank_specifier*s. The latter production is used to represent array types where the element type is an array type which is nullable ([§8.9](types.md#89-reference-types-and-nullability)).
```

standard/arrays.md
```diff
@@ -31,20 +31,64 @@ Every array type is a reference type ([§8.2](types.md#82-reference-types)). The
...
 
 A *non_array_type* is any *type* that is not itself an *array_type*.
 
-The rank of an array type is given by the leftmost *rank_specifier* in the *array_type*: A *rank_specifier* indicates that the array is an array with a rank of one plus the number of “`,`” tokens in the *rank_specifier*.
+When determining the rank and element type of array type as specified below, only the *rank_specifier*s in the top-most production are considered, so in the production `array_type nullable_type_annotation rank_specifier+`, any `rank_specifier` within the `array_type` is ignored.
```
jskeet:
```suggestion
When determining the rank and element type of array type as specified below, only the *rank_specifier*s in the outer-most production are considered, so in the production `array_type nullable_type_annotation rank_specifier+`, any `rank_specifier` within the `array_type` is ignored.
```

standard/arrays.md
jskeet: Resolving after discussion.

standard/arrays.md
```diff
@@ -31,20 +31,64 @@ Every array type is a reference type ([§8.2](types.md#82-reference-types)). The
...
+When determining the rank and element type of array type as specified below, only the *rank_specifier*s in the outer-most production are considered, so in the production `array_type nullable_type_annotation rank_specifier+`, any `rank_specifier` within the `array_type` is ignored.
 
-The element type of an array type is the type that results from deleting the leftmost *rank_specifier*:
+The rank of an array type is given by the leftmost *rank_specifier* in the *array_type*: A *rank_specifier* indicates that the array is an array with a rank of one plus the number of “`,`” tokens in the *rank_specifier*.
 
-- An array type of the form `T[R]` is an array with rank `R` and a non-array element type `T`.
-- An array type of the form `T[R][R₁]...[Rₓ]` is an array with rank `R` and an element type `T[R₁]...[Rₓ]`.
+The element type of an array type is the type that results from deleting the leftmost *rank_specifier*.
 
 In effect, the *rank_specifier*s are read from left to right *before* the final non-array element type.
```
jskeet:
```suggestion
```

(Removing the text as we don't understand it.)

standard/arrays.md
```diff
@@ -31,20 +31,62 @@ Every array type is a reference type ([§8.2](types.md#82-reference-types)). The
...
-
-The element type of an array type is the type that results from deleting the leftmost *rank_specifier*:
+When determining the rank and element type of array type as specified below, only the *rank_specifier*s in the outer-most production are considered, so in the production `array_type nullable_type_annotation rank_specifier+`, any `rank_specifier` within the `array_type` is ignored.
 
-- An array type of the form `T[R]` is an array with rank `R` and a non-array element type `T`.
-- An array type of the form `T[R][R₁]...[Rₓ]` is an array with rank `R` and an element type `T[R₁]...[Rₓ]`.
+The rank of an array type is given by the leftmost *rank_specifier* in the *array_type*: A *rank_specifier* indicates that the array is an array with a rank of one plus the number of “`,`” tokens in the *rank_specifier*.
 
-In effect, the *rank_specifier*s are read from left to right *before* the final non-array element type.
+The element type of an array type is the type that results from deleting the leftmost *rank_specifier*.
```
jskeet:
```suggestion
The element type of an array type is the type that results from deleting the leftmost *rank_specifier* from the *array_type*.
```

(To clarify a bit, without going back to `T[R][R₁]...[Rₓ]` which I think is confusing.)

jskeet:
Next steps:

- Nigel to email Jon the files to add to this PR to get the grammar checker passing
- Jon will update this PR, then merge.
- Jon to file a new issue around mapping between nullable and non-nullable arrays, see #1386, to also include conversions - and anything else folks think is missing
- Jon will hopefully create a new PR to address that

BillWagner (approved):
Let's :shipit: based on our discussion in the meeting.

jnm2 (approved):
I love this approach!

standard/types.md
```diff
@@ -54,17 +54,22 @@ interface_type
...
+    | non_array_type rank_specifier+
     ;
 
 non_array_type
+    : non_array_non_nullable_type nullable_type_annotation?
+    | pointer_type      // unsafe code support
+    ;
+
+non_array_non_nullable_type
     : value_type
```
Nigel-Ecma: The existing grammar supported `T?[]` **iff** `T` was a value type, this PR intends to extend it to all `T`. The existing grammar supported this as *value_type* covers both nullable & non-nullable value types – the only nullable types that existed at the time. The revised grammar here provides two ways that *non_array_type* can parse `V?` as shown below – text in brackets, e.g. [`?`], following a rule indicates the input token(s) it is matching:

```
non_array_type ➜ non_array_non_nullable_type nullable_type_annotation? [`?`]
    non_array_non_nullable_type ➜ value_type
                                ➜ non_nullable_value_type  [`V`]
```

or

```
non_array_type ➜ non_array_non_nullable_type nullable_type_annotation?
    non_array_non_nullable_type ➜ value_type
                                ➜ nullable_value_type
                                ➜ non_nullable_value_type [`V`] nullable_type_annotation [`?`]
```

As there are two parses the new grammar is ambiguous, but I suspect ambiguity was not intended and this needs to be fixed in the grammar and not with a prose disambiguation rule.

Changing line 67 to reference `non_nullable_value_type` would seem to fixed the immediate issue, but I have not worked that through the PR’s total grammar changes myself.

```suggestion
    : non_nullable_value_type
```

jskeet: That didn't take me as long to understand as I feared - and thank you for the suggestion, which I've applied. Fortunately the name of the rule is pretty clear in terms of what it's trying to convey, too.

@BillWagner please could you check that you're happy with the change on this line as well? (`value_type` => `non_nullable_value_type`). I suspect we'll still need Nigel to run the grammar checker etc, but hopefully this will have cleared up the ambiguity.

jskeet:
This is still waiting for another PR from @Nigel-Ecma that will affect the grammar, then we'll come back to it.

jskeet:
All dependent PRs are now merged, so @Nigel-Ecma should be able to regenerate the necessary files for the GitHub checks to pass.

jskeet:
Tests have now passed - merging.

PR #1555, opened 2026-01-28
Apply non-semantic changes from #1538

jskeet:
The hope is that this can reduce the diff in #1538, making it easier to review.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1639,12 +1639,12 @@ A *deconstruction_expression* `var (e1, ..., en)` is shorthand for the *tuple_ex
 
 A tuple expression has a type if and only if each of its element expressions `Ei` has a type `Ti`. The type shall be a tuple type of the same arity as the tuple expression, where each element is given by the following:
 
-- If the tuple element in the corresponding position has a name `Ni`, then the tuple type element shall be `Ti Ni`.
-- Otherwise, if `Ei` is of the form `Ni` or `E.Ni` or `E?.Ni` then the tuple type element shall be `Ti Ni`, *unless* any of the following holds:
-  - Another element of the tuple expression has the name `Ni`, or
-  - Another tuple element without a name has a tuple element expression of the form `Ni` or `E.Ni` or `E?.Ni`, or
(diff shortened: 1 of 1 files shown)


standard/expressions.md
```diff
@@ -4422,7 +4422,7 @@ equality_expression
 <!-- markdownlint-disable MD028 -->
 
 <!-- markdownlint-enable MD028 -->
-> *Note*: There is a grammar ambiguity between *type* and *constant_pattern* in a `relational_expression` on the right-hand-side of `is`; either might be a valid parse of a qualified identifier. In such a case, only if it fails to bind as a type (for compatibility with previous versions of the language), is it resolved to be the first thing found (which must be either a constant or a type). This ambiguity is only present on the right-hand side of such an expression.
+> *Note*: There is a grammar ambiguity between *type* and *constant_pattern* in a `relational_expression` on the right-hand-side of `is`; either might be a valid parse of a qualified identifier. In such a case, only if it fails to bind as a type (for compatibility with previous versions of the language), is it resolved to be the first thing found (which must be either a constant or a type). This ambiguity is only present on the right-hand side of such an expression. *end note*
```
jskeet, line 4425: I'm sure that when reviewing #1538 I saw more changes like this - will double check for them...

jskeet: Nope, can't see them now...

jskeet:
@Nigel-Ecma The hope is that after this is merged, you can rebase your PR and there'll be a smaller diff. Two bits where you'll have some work to do:

- Undoing your Rectangle whitespace changes as I've changed Point instead
- In two cases (I think) there were subscript changes and other changes in the same diff; I've only take the subscript changes. I think you should be able to just say "accept my changes" and it'll do the right thing.

BillWagner (approved):
This LGTM. Once Nigel has a look, let's see how well this works.

Nigel-Ecma (approved):
Well each red/green pair looks like it has the expected changes, so I guess we try Jon’s idea out…

Nigel-Ecma:
@BillWagner – pushed Merge, but I’ll leave Delete branch to you!

PR #1557, opened 2026-01-28
[Version 10.0] update test templates

RexJaeschke:
@BillWagner This PR updates the csproj files for all of the test templates, to .Net 10 and C# 10. Once you merge this, I can see what, if any, errors result from my V10 tests (all of which current pass on my system, which has these updates).

I'm getting the following (see below) Using launch settings from ExampleTester/Properties/launchSettings.json...
Example tester-ExampleTester::file=tools/arrays.md:14-21,line=14::Testing PascalArrayDeclarations from arrays.md:14-21
Unhandled exception: System.Collections.Generic.KeyNotFoundException: **The given key 'net10.0' was not present in the dictionary.**

Perhaps something in Jon's testing harness files needs to be changed as well.

### DIFF
--- a/standard/attributes.md
+++ b/standard/attributes.md
@@ -20,7 +20,7 @@ A generic class declaration shall not use `System.Attribute` as a direct or indi
 
 > *Example*:
 >
-> <!-- Example: {template:"standalone-lib", name:"AttributeCantBeGeneric", expectedErrors:["CS8773"], ignoredWarnings:["CS0169"]} -->
+> <!-- Example: {template:"standalone-lib", name:"AttributeCantBeGeneric", expectedErrors:["CS8936"], ignoredWarnings:["CS0169"]} -->
 > ```csharp
 > public class B : Attribute {}
 > public class C<T> : B {} // Error – generic cannot be an attribute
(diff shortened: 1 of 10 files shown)


RexJaeschke:
@jskeet, @BillWagner Looking further at the failing check messages I see two steps, "Setup .NET 6.0 and 8.0" and "Post Setup .NET 6.0 and 8.0," which suggests the process is not yet capable of handling a target of .NET 10.0. Given where we are trying to get the v8 spec out the door, if resolving this failure is non-trivial/distracting, can I fall back to using a target of .NET 8 or even go back to 6? If 8 or 6, I can simply have a new commit to restore the target but keep the new langversion.

BillWagner:
This might be going in the wrong direction. C# 10 would be .NET 6. So, I'm not sure why you need to use .NET 10.

That said, if we do need it, look here: https://github.com/dotnet/csharpstandard/blob/draft-10/.github/workflows/test-examples.yaml#L45

That's where the versions of .NET needed for the test-examples workflow are instaled.

RexJaeschke:
Thanks, Bill. For my local testing, I'm using VS2026, which by default targets net10.0, and I used that for the csproj updates, with langversion 10. However, given your feedback, I'll revert the templates to net6.0 but keep langversion 10, which sounds like it should work.

PR #1566, opened 2026-02-08
[Version 10.0] Feature support for lambda improvements

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md).

The MS spec proposal ends with an "Open issue" section, which does *not* say anything about the eventual disposition of those issues. Someone needs to check if more work was done to address them.

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -63,6 +63,7 @@ The following conversions are classified as implicit conversions:
 - Implicit tuple conversions ([§10.2.13](conversions.md#10213-implicit-tuple-conversions))
 - Default literal conversions ([§10.2.16](conversions.md#10216-default-literal-conversions))
 - Implicit throw conversions ([§10.2.17](conversions.md#10217-implicit-throw-conversions))
+- Anonymous function type conversions (§anon-func-type-conversion)
 
 Implicit conversions can occur in a variety of situations, including function member invocations ([§12.6.6](expressions.md#1266-function-member-invocation)), cast expressions ([§12.9.8](expressions.md#1298-cast-expressions)), and assignments ([§12.24](expressions.md#1224-assignment-operators)).
 
@@ -385,7 +386,7 @@ A user-defined implicit conversion consists of an optional standard implicit con
 
 ### 10.2.15 Anonymous function conversions and method group conversions
 
-Anonymous functions and method groups do not have types in and of themselves, but they may be implicitly converted to delegate types. Additionally, some lambda expressions may be implicitly converted to expression tree types. Anonymous function conversions are described in more detail in [§10.7](conversions.md#107-anonymous-function-conversions) and method group conversions in [§10.8](conversions.md#108-method-group-conversions).
+Anonymous functions and method groups do not have types in and of themselves, but they may have a natural type (§anon-func-type). They may be implicitly converted to delegate types. Additionally, some lambda expressions may be implicitly converted to expression tree types. Anonymous function conversions are described in more detail in [§10.7](conversions.md#107-anonymous-function-conversions) and method group conversions in [§10.8](conversions.md#108-method-group-conversions).
 
 ### 10.2.16 Default literal conversions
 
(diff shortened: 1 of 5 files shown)


standard/expressions.md
```diff
@@ -819,7 +819,7 @@ An *input type inference* is made *from* an expression `E` *to* a type `T` in th
 An *output type inference* is made *from* an expression `E` *to* a type `T` in the following way:
 
 - If `E` is a tuple expression with arity `N` and elements `Eᵢ`, and `T` is a tuple type with arity `N` with corresponding element types `Tₑ` or `T` is a nullable value type `T0?` and `T0` is a tuple type with arity `N` that has a corresponding element type `Tₑ`, then for each `Eᵢ` an output type inference is made from `Eᵢ` to `Tₑ`.
-- If `E` is an anonymous function with inferred return type `U` ([§12.6.3.14](expressions.md#126314-inferred-return-type)) and `T` is a delegate type or expression tree type with return type `Tₓ`, then a *lower-bound inference* ([§12.6.3.11](expressions.md#126311-lower-bound-inferences)) is made *from* `U` *to* `Tₓ`.
+- **TBD** If `E` is an anonymous function with inferred return type `U` ([§12.6.3.14](expressions.md#126314-inferred-return-type)) and `T` is a delegate type or expression tree type with return type `Tₓ`, then a *lower-bound inference* ([§12.6.3.11](expressions.md#126311-lower-bound-inferences)) is made *from* `U` *to* `Tₓ`.
```
RexJaeschke: Should anything be added here? The MS spec replacement text was written before we created the separate input and output type inference sections. And I added the proposed text from the MS spec to the input inference section only.

standard/expressions.md
```diff
@@ -1014,14 +1020,16 @@ Instead, all `Xᵢ` are considered *unfixed*, and a *lower-bound inference* is
 
 #### 12.6.3.16 Finding the best common type of a set of expressions
 
+**TBD**
```
RexJaeschke: The MS spec says, “Best common type: Best common type is defined in terms of type inference so the type inference changes above apply to best common type as well."

Someone needs to figure out what changes go in this section.

standard/expressions.md
```diff
@@ -1151,6 +1161,7 @@ Given `int i = 10;`, according to [§12.6.4.2](expressions.md#12642-applicable-f
 
 Given an implicit conversion `C₁` that converts from an expression `E` to a type `T₁`, and an implicit conversion `C₂` that converts from an expression `E` to a type `T₂`, `C₁` is a ***better conversion*** than `C₂` if `E` does not exactly match `T₂` and at least one of the following holds:
 
+- `C1` is not a _function_type_conversion_ and `C2` is a _function_type_conversion_, or
```
RexJaeschke: This list has been changed since the MS spec was written, and the proposed changes do not match the new organization. I have added the proposed new item at the front of the list, but someone will need to see if that is now the correct text and is in the right place.

standard/expressions.md
```diff
@@ -5379,11 +5390,13 @@ The run-time processing of a conditional expression of the form `b ? x : y`
...
 
-> *Note*: For historical reasons, there are two syntactic flavors of anonymous functions, namely *lambda_expression*s and *anonymous_method_expression*s. For almost all purposes, *lambda_expression*s are more concise and expressive than *anonymous_method_expression*s, which remain in the language for backwards compatibility. *end note*
+> *Note*: For historical reasons, there are two syntactic flavors of anonymous functions, namely *lambda_expression* and *anonymous_method_expression*. For almost all purposes, *lambda_expression* is more concise and expressive than *anonymous_method_expression*s, which remain in the language for backwards compatibility. *end note*
 
 ```ANTLR
 lambda_expression
```
RexJaeschke, line 5417: The grammar in the [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md) is quite different from what we actually have in V8, as in V7 we added ref returns. The MS proposal is also missing grammar edits for an explicit return type (see below). It also appears to support default values on parameters, something that was **not** added until V12.

My edits to support an optional return type need close review, esp. now that we have ref and non-ref returns. I modelled these edits on the analogous *method_declaration* grammar rule (§15.6.1), which uses 2 subrules, but in each requires a return type be specified, which is **not** the case with lambdas where that is optional. Also, that pair of rules uses different bodies (*method_body* and *ref_method_body*) whereas I have not for *lambda-expression*, but perhaps we’ll need to add something like *anonymous_ref_function_body* as well, and possibly change the existing definition of *anonymous_function_body*. 

At a glance, it seems that rule *'ref' variable_reference* should **not** be allowed in *anonymous_function_body* if that rule is reached from a *lambda_expression* having *return_type*, but what if no return type is specified at all, and is simply inferred?

I think it likely that my proposed grammar changes will need to be tweaked and/or possibly augmented with constraints in the narrative.

standard/expressions.md
```diff
@@ -5886,6 +5907,68 @@ Separate non-`static` anonymous functions can capture the same instance of an ou
 
 An anonymous function `F` shall always be converted to a delegate type `D` or an expression-tree type `E`, either directly or through the execution of a delegate creation expression `new D(F)`. This conversion determines the result of the anonymous function, as described in [§10.7](conversions.md#107-anonymous-function-conversions).
 
+### §anon-func-type Anonymous function type
```
RexJaeschke, line 5946: The [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md) introduces the term *function_type*. I have two objections to that spelling: First, the separating underscore doesn’t seem right; it makes the name look like a grammar rule name. Second, “function” is a very general term, so I called the new term “anonymous function type” instead, as it applies only in that narrow context.

standard/conversions.md
```diff
@@ -405,6 +405,47 @@ For a *conditional_expression* `c ? e1 : e2`, when
 
 an implicit ***conditional expression conversion*** exists that permits an implicit conversion from *conditional_expression* to any type `T` for which there is a conversion-from-expression from `e1` to `T` and also from `e2` to `T`.  It is an error if *conditional_expression* neither has a common type between `e1` and `e2` nor is subject to a conditional expression conversion.
 
+### §anon-func-type-conversion Anonymous function type conversion
+
+The following conversions are permitted from an anonymous function type `F`(§anon-func-type):
+
+- To an anonymous function type `G` if the parameters and return types of `F` are variance-convertible to the parameters and return type of `G`.
+- To `System.Delegate` or its base classes or interfaces.
```
RexJaeschke, line 437: The [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-10.0/lambda-improvements.md) mentions `System.MulticastDelegate`. However, as we don’t have that type in our spec, I replaced all such mentions with `System.Delegate`.

BillWagner:
Thanks!

PR #1570, opened 2026-02-12
Add tuple conversion-from-type and put ValueTuple text in one place

jskeet:
This is largely copied from #1560, but reverting the "tuple expression" to "tuple literal" change, and adding another example in the implicit conversion section.

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -349,7 +349,7 @@ An implicit conversion exists from a tuple expression `e`, of the form `(e₁, .
 
 An implicit conversion exists from the tuple type `T = (T₁, ..., Tₙ)` to the tuple type `S = (S₁, ..., Sₙ)` if there is an implicit conversion from each `Tᵢ` to the corresponding `Sᵢ`. The result of this conversion when applied to a tuple `t` with value `(t₁, ..., tₙ)` and of type `T` is the tuple value `((S₁)t₁, ..., (Sₙ)tₙ)` of type `S`.
 
-> *Note*: When a conversion is applied to a tuple value, as opposed to a tuple literal, no warnings are required if element names do not match. *end note*
+> *Note*: When a conversion is applied to a tuple value, as opposed to a tuple expression, no warnings are required if element names do not match. *end note*
 <!-- markdownlint-disable MD028 -->
 
 <!-- markdownlint-enable MD028 -->
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -345,9 +345,14 @@ In all cases, the rules ensure that a conversion is executed as a boxing convers
 
 ### 10.2.13 Implicit tuple conversions
 
-An implicit conversion exists from a tuple expression `E` to a tuple type `T` if `E` has the same arity as `T` and an implicit conversion exists from each element in `E` to the corresponding element type in `T`. The conversion is performed by creating an instance of `T`’s corresponding `System.ValueTuple<...>` type, and initializing each of its fields in order from left to right by evaluating the corresponding tuple element expression of `E`, converting it to the corresponding element type of `T` using the implicit conversion found, and initializing the field with the result.
+An implicit conversion exists from a tuple expression `e`, of the form `(e₁, ..., eₙ)`, to a tuple value of type `T = (T₁, ..., Tₙ)` if there is an implicit conversion for each element `eᵢ` to the corresponding element type `Tᵢ`. The result of the conversion is the tuple value `((T₁)e₁, ..., (Tₙ)eₙ)` of type `T`. If an element name in the tuple expression does not match a corresponding element name in the tuple type, a warning shall be issued.
 
-If an element name in the tuple expression does not match a corresponding element name in the tuple type, a warning shall be issued.
+An implicit conversion exists from the tuple type `T = (T₁, ..., Tₙ)` to the tuple type `S = (S₁, ..., Sₙ)` if there is an implicit conversion from each `Tᵢ` to the corresponding `Sᵢ`. The result of this conversion when applied to a tuple `t` with value `(t₁, ..., tₙ)` and of type `T` is the tuple value `((S₁)t₁, ..., (Sₙ)tₙ)` of type `S`.
+
+> *Note*: When a conversion is applied to a tuple value, as opposed to a tuple literal, no warnings are required if element names do not match. *end note*
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
 
 > *Example*:
 >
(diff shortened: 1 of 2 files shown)


standard/types.md
```diff
@@ -403,22 +403,24 @@ An enumeration type is a distinct type with named constants. Every enumeration t
...
+Identity conversions ([§10.2.2](conversions.md#1022-identity-conversion)) exist between tuples of the same arity with identity-convertible sequences of element types.
 
-Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can be accessed via a member access on a tuple value ([§12.8.7](expressions.md#1287-member-access). Additionally, if the tuple type has a name for a given element, that name can be used to access the element in question.
+Tuple values can be created from tuple expressions ([§12.8.6](expressions.md#1286-tuple-expressions)), or by creating a value using the underlying runtime representation (§tuple-runtime-representation) directly. The tuple type syntax `(T1, ..., Tn)` cannot be used with the `new` operator [§12.8.17.2](expressions.md#128172-object-creation-expressions).
+
+Tuple elements are public fields with the names `Item1` … `ItemN`, where `N` is the tuple arity and the numbers have no leading zeros, and can be accessed via a member access on a tuple value ([§12.8.7](expressions.md#1287-member-access). Additionally, if the tuple type has a name for a given element, that name can be used to access the element in question.
```
jskeet, line 422: @BillWagner I know you had some concerns over emphasis of the ItemX names vs the user-specified names - happy to take suggestions here.

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
-> The declarations for `pair6` and `pair7` demonstrate that tuple types are interchangeable with constructed types of the form `ValueTuple<...>`, and that the `new` operator is allowed with the latter syntax.
->
 >The last line shows that tuple elements can be accessed by the `Item` name corresponding to their position, as well as by the corresponding tuple element name, if present in the type.
 > *end example*
 
+#### §eliding-tuples Eliding intermediate tuple creation
+
+If the result of constructing a tuple (§12.8.6) is not required outside of the context in which it is constructed implementations are explicitly allowed to elide the construction as an optimisation provided all other semantic requirements are met.
+
+> *Example*: Such a situation may commonly arise from assignments and switch statements (§13.8.3). Consider the assignment:
```
jskeet, line 447:
```suggestion
> *Example*: Such a situation may commonly arise from assignments and switch statements (§13.8.3). Consider the deconstructing assignment:
```

I now see that "deconstructing assignment" is already present in the standard (currently described *within* simple assignment).

standard/conversions.md
```diff
@@ -345,9 +345,14 @@ In all cases, the rules ensure that a conversion is executed as a boxing convers
...
+An implicit conversion exists from a tuple expression `e`, of the form `(e₁, ..., eₙ)`, to a tuple value of type `T = (T₁, ..., Tₙ)` if there is an implicit conversion for each element `eᵢ` to the corresponding element type `Tᵢ`. The result of the conversion is the tuple value `((T₁)e₁, ..., (Tₙ)eₙ)` of type `T`. If an element name in the tuple expression does not match a corresponding element name in the tuple type, a warning shall be issued.
 
-If an element name in the tuple expression does not match a corresponding element name in the tuple type, a warning shall be issued.
+An implicit conversion exists from the tuple type `T = (T₁, ..., Tₙ)` to the tuple type `S = (S₁, ..., Sₙ)` if there is an implicit conversion from each `Tᵢ` to the corresponding `Sᵢ`. The result of this conversion when applied to a tuple `t` with value `(t₁, ..., tₙ)` and of type `T` is the tuple value `((S₁)t₁, ..., (Sₙ)tₙ)` of type `S`.
+
+> *Note*: When a conversion is applied to a tuple value, as opposed to a tuple literal, no warnings are required if element names do not match. *end note*
```
BillWagner: Should this instance of "tuple literal" be reverted to "tuple expression"?


```suggestion
> *Note*: When a conversion is applied to a tuple value, as opposed to a tuple expression, no warnings are required if element names do not match. *end note*
```

standard/conversions.md
```diff
@@ -502,7 +508,11 @@ For an explicit reference conversion to succeed at run-time, the value of the so
...
+An explicit conversion exists from a tuple expression `e`, of the form `(e₁, ..., eₙ)`, to a tuple value of type `T = (T₁, ..., Tₙ)` if there is an explicit conversion for each element `eᵢ` to the corresponding element type `Tᵢ`. The result of the conversion is the tuple value `((T₁)e₁, ..., (Tₙ)eₙ)` of type `T`. If an element name in the tuple expression does not match a corresponding element name in the tuple type, a warning shall be issued.
+
+An explicit conversion exists from the tuple type `T = (T₁, ..., Tₙ)` to the tuple type `S = (S₁, ..., Sₙ)` if there is an explicit conversion from each `Tᵢ` to the corresponding `Sᵢ`. The result of this conversion when applied to a tuple value `(t₁, ..., tₙ)` of type `T` is the tuple value `((S₁)t₁, ..., (Sₙ)tₙ)` of type `S`.
+
+> *Note*: When a conversion is applied to a tuple value, as opposed to a tuple literal, no warnings are required if element names do not match. *end note*
```
BillWagner: Same note as before, should this be "tuple expression"?


```suggestion
> *Note*: When a conversion is applied to a tuple value, as opposed to a tuple expression, no warnings are required if element names do not match. *end note*
```

standard/types.md
BillWagner: I do, but @Nigel-Ecma 's comment on the previous PR convinced me it's not a normative change. I think it warrants a note something like:

> *Note:* It is expected that most tuple usage will include semantic tuple names for the elements. The `ItemN` names are those used by the underlying representation. *end note*

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+>
+> ```csharp
+>    (a, b) = (b, a);
+> ```
+>
+> The right hand side `(b, a)` constructs a tuple containing the values of `b` & `a`. The left hand side tuple expression `(a, b)` then, in order, selects the first item of that tuple and assigns it to `a`, followed by assigning the second item to `b`. The overall result is the the values in `a` & `b` are exchanged, while the tuple created during this process is discarded. The explicit allowance granted here to elide such intermediate tuple construction allows an implementation to exchange the two values in whatever ways it chooses provide it evaluates `b` before `a` to meet the left-to-right evaluation order of tuple expression elements. In the code:
```
BillWagner: nit: remove repeated word:


```suggestion
> The right hand side `(b, a)` constructs a tuple containing the values of `b` & `a`. The left hand side tuple expression `(a, b)` then, in order, selects the first item of that tuple and assigns it to `a`, followed by assigning the second item to `b`. The overall result is the values in `a` & `b` are exchanged, while the tuple created during this process is discarded. The explicit allowance granted here to elide such intermediate tuple construction allows an implementation to exchange the two values in whatever ways it chooses provide it evaluates `b` before `a` to meet the left-to-right evaluation order of tuple expression elements. In the code:
```

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+
+<!-- markdownlint-enable MD028 -->
+> *Example*: If the runtime representation of a tuple uses instances of `System.ValueTuple<T1, ..., T7, TRest>` then the `Rest` field is accessible. The use of this provides different ways to reference items in large tuples. Given:
+>
+> <!-- Example: {template:"standalone-console", name:"TupleTypes4", ignoredWarnings:["CS0219"]} -->
+> ```csharp
+> var squares = (1, 4, 9, 16, 25, 36, 49, 64, 81, 100, 121, 144, 169, 196, 225);
+> ```
+>
+> Then the 15th square (`225`) can be addressed as `squares.Item15`, `squares.Rest.Item8` and `squares.Rest.Rest.Item1`.
```
BillWagner, line 533: This may be a better location for the note I suggested on line 422 (or similar language). It's a good example on why using the names is much more convenient in C#.

BillWagner (approved):
I had a few comments, but overall this is good.

I did compare with #1560 and it looks like all the described portions have been added.

standard/conversions.md
jskeet: Yes, thank you!

standard/types.md
jskeet: Filed https://github.com/dotnet/csharpstandard/issues/1574 for this.

jskeet:
Will leave this for a few more days in case @MadsTorgersen or @Nigel-Ecma want to request changes, but otherwise I'll merge next week.

MadsTorgersen (approved):

Nigel-Ecma (changes requested):
There seem to be issues with the attempt to slot the missing conversion into the existing text without addressing the issues with that existing text, as detailed in offline email. Until that is addressed I cannot say this PR is a valid step towards fixing tuple construction and the deconstruction of tuple-ish things.

jskeet:
Closing in favor of #1608

PR #1573, opened 2026-02-12
[Version 11.0] Feature support for auto default structs

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/auto-default-structs.md).


BillWagner:
Thanks!

PR #1576, opened 2026-02-18
`stackalloc` can initialize a ROS

BillWagner:
The `ReadOnlySpan<T>` is also a valid type for a `stackalloc` in safe code.

Found while investigating #1232 

While this doesn't address #1232, it was an omission.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -3435,11 +3435,12 @@ When a *stackalloc_initializer* is present:
 
 Each *stackalloc_element_initializer* shall have an implicit conversion to *unmanaged_type* ([§10.2](conversions.md#102-implicit-conversions)). The *stackalloc_element_initializer*s initialize elements in the allocated memory in increasing order, starting with the element at index zero. In the absence of a *stackalloc_initializer*, the content of the newly allocated memory is undefined.
 
-If a *stackalloc_expression* occurs directly as the initializing expression of a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)), where the *local_variable_type* is either a pointer type ([§24.3](unsafe-code.md#243-pointer-types)) or inferred (`var`), then the result of the *stackalloc_expression* is a pointer of type `T*` ([§24.9](unsafe-code.md#249-stack-allocation)). In this case the *stackalloc_expression* must appear in unsafe code. Otherwise the result of a *stackalloc_expression* is an instance of type `Span<T>`, where `T` is the *unmanaged_type*:
+If a *stackalloc_expression* occurs directly as the initializing expression of a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)), where the *local_variable_type* is either a pointer type ([§24.3](unsafe-code.md#243-pointer-types)) or inferred (`var`), then the result of the *stackalloc_expression* is a pointer of type `T*` ([§24.9](unsafe-code.md#249-stack-allocation)). In this case the *stackalloc_expression* must appear in unsafe code. Otherwise the result of a *stackalloc_expression* is an instance of type `Span<T>` or `ReadOnlySpan<T>`, where `T` is the *unmanaged_type*:
 
-- `Span<T>` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)) is a ref struct type ([§16.2.3](structs.md#1623-ref-modifier)), which presents a block of memory, here the block allocated by the *stackalloc_expression*, as an indexable collection of typed (`T`) items.
+- `Span<T>` and `ReadOnlySpan<T>` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)) are ref struct types ([§16.2.3](structs.md#1623-ref-modifier)), which presents a block of memory, here the block allocated by the *stackalloc_expression*, as an indexable collection of typed (`T`) items.
 - The result’s `Length` property returns the number of items allocated.
-- The result’s indexer ([§15.9](classes.md#159-indexers)) returns a *variable_reference* ([§9.5](variables.md#95-variable-references)) to an item of the allocated block and is range checked.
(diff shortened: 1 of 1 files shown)


standard/expressions.md
```diff
@@ -3435,11 +3435,12 @@ When a *stackalloc_initializer* is present:
...
+If a *stackalloc_expression* occurs directly as the initializing expression of a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)), where the *local_variable_type* is either a pointer type ([§24.3](unsafe-code.md#243-pointer-types)) or inferred (`var`), then the result of the *stackalloc_expression* is a pointer of type `T*` ([§24.9](unsafe-code.md#249-stack-allocation)). In this case the *stackalloc_expression* must appear in unsafe code. Otherwise the result of a *stackalloc_expression* is an instance of type `Span<T>` or `ReadOnlySpan<T>`, where `T` is the *unmanaged_type*:
```
KalleOlaviNiemitalo, line 3438: I don't think this change is right.

1. Consider the following ([SharpLab](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLoN7LpGYZQAs6AsgBQCU+hxTuAxugLzoDOALgIYsA1nwA2IgPZsAlgDseAbTgBdANyMiAX3UlMcAGzopAWwAOIqSyk904kwFMATnx7iHOagGUTfGQB5ZPAB86AAe9OzBMnYA7u60akgaQA)):

   ```csharp
   using System;
   public class C {
       public void M() {
           C c = stackalloc int[1];
       }
       public static implicit operator C(Span<int> x) => new C();
   }
   ```

   If ReadOnlySpan\<T\> has a similar implicit conversion operator from Span\<T\>, then that can be used for the initialization, and the type of the stackalloc expression need not be ReadOnlySpan\<T\> directly.

2. If the type of stackalloc expression could be either  Span\<T\> or ReadOnlySpan\<T\>, then you'd likely need a disambiguation rule for when both types are feasible, e.g. when calling an overloaded or generic method.

KalleOlaviNiemitalo (changes requested):

standard/expressions.md
Nigel-Ecma: I think @KalleOlaviNiemitalo is correct here, it looks like `Span<T>` has an implicit conversion to `ReadOnlySpan<T>`. If `Span` is changed to `ReadOnlySpan` in the example it fails as it requires two implicit conversions to be inserted, but use an intermediate variable and it is valid as each line only requires the insertion of one conversion:

```csharp
    public void M()
    {
        ReadOnlySpan<int> x = stackalloc int[1];
        C c = x;
    }
```

KalleOlaviNiemitalo: The implicit conversion is required by *C.3 Standard Library Types not defined in ISO/IEC 23271*:

<https://github.com/dotnet/csharpstandard/blob/a0995e579c0f3149dc952092a9b095a8d73f260e/standard/standard-library.md?plain=1#L578-L583>

So I don't think any normative text for `stackalloc` should be changed for this.  A note could be added though.

C.3 seems to be the only mention of ReadOnlySpan\<\> in the spec.  AFAICT the semantics of the implicit conversion are unspecified and a conforming implementation could make it always return `default(ReadOnlySpan<T>)`.  Users would be unlikely to choose such an implementation though.  But any note added in the `stackalloc` description should be carefully written not to claim any semantics that the standard does not actually require.

Nigel-Ecma: 👍’ed Kalle’s comment but I don’t think adding a note is required.

standard/expressions.md
```diff
@@ -3435,11 +3435,12 @@ When a *stackalloc_initializer* is present:
...
 
-- `Span<T>` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)) is a ref struct type ([§16.2.3](structs.md#1623-ref-modifier)), which presents a block of memory, here the block allocated by the *stackalloc_expression*, as an indexable collection of typed (`T`) items.
+- `Span<T>` and `ReadOnlySpan<T>` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)) are ref struct types ([§16.2.3](structs.md#1623-ref-modifier)), which presents a block of memory, here the block allocated by the *stackalloc_expression*, as an indexable collection of typed (`T`) items.
```
jskeet, line 3440:
```suggestion
- `Span<T>` and `ReadOnlySpan<T>` ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)) are ref struct types ([§16.2.3](structs.md#1623-ref-modifier)), which present a block of memory, here the block allocated by the *stackalloc_expression*, as an indexable collection of typed (`T`) items.
```

standard/expressions.md
jskeet: Avoiding double conversions and implicit conversions, I think there's still potentially an issue. Consider this code:

```csharp
class Program
{
    static void Main()
    {
        int count = M(stackalloc int[5]);
    }

    static int M(ReadOnlySpan<int> x) => x.Length;
    static int M(Span<int> x) => x.Length;
}
```

It's not clear (at least to me) which overload of `M` is invoked here. Both are valid (if you comment out either one, the code still compiles).

Nigel-Ecma: @jskeet wrote:

>     static int M(ReadOnlySpan<int> x) => x.Length;
>     static int M(Span<int> x) => x.Length;

Type of `stackalloc` is `Span<T>`, `Span<T>` has an implicit conversion to `ReadOnlySpan<T>` but the reverse is not true, so in overload resolution `Span<T>` is the better conversion target.

So the second `M` should be selected (I have not checked what a certain compiler does).

jskeet:
Discussed in today's meeting. Closing without merging - the implicit conversion from `Span<T>` to `ReadOnlySpan<T>` is already present in annex C3, and that covers everything that's required *without* rules needing to mention `ReadOnlySpan<T>` with reference to stackalloc.

PR #1597, opened 2026-03-07
[Version 11.0] Feature support for required members

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/required-members.md).

## Open Issue: Combining `required` with `ref` or `ref readonly`

The MS spec states that,

> `required` cannot be combined with the following modifiers: …, ref, ref readonly, …

I have confirmed that while this is true for properties, it is *not* true for fields. The following compiles:

```csharp
public ref struct S2
{
    [SetsRequiredMembers]
    public S2() { }
    public static int x = 555;
 
    public          ref int r1a = ref x;
    public required ref int r1b = ref x;
    public          ref readonly int r2a = ref x;
    public required ref readonly int r2b = ref x;
}

    var s2 = new S2();
    Console.WriteLine($"S2.x   = {S2.x}");
    Console.WriteLine($"s2.r1a = {s2.r1a}");
    Console.WriteLine($"s2.r1b = {s2.r1b}");
    Console.WriteLine($"s2.r2a = {s2.r2a}");
    Console.WriteLine($"s2.r2b = {s2.r2b}");
```

Is the MS spec wrong/out-of-date, or am I missing something?

Bill Wagner stated: I don’t see any notes in LDM meetings. But, I think the spec was a bit vague, and was originally meant to focus on properties.  I’m hoping Fred can shed more light on it.

Fred Silberberg stated: Required members supports both fields and properties, but I'm not sure whether we specifically considered the intersection given the late-breaking nature of the way ref fields were developed. Likely we should block this.


BillWagner:
Thanks!

PR #1598, opened 2026-03-07
[Version 11.0] Feature support for Numeric IntPtr

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/numeric-intptr.md).

## Issue 1

Since .NET 5, `IntPtr` and `UIntrPtr` have had properties `MinValue` and `MaxValue`. Shall we admit that and add these members to the library annex as new members of an existing type?

## Issue 2

Should we add any new operators (such as addition and subtraction) to `IntPtr`/`UIntPtr` in the library annex?

## Issue 3

The following text was added to 8.3.5 Simple types in Rex's Draft V9 PR #1457. Please review it to see which text should be kept, which should be deleted, and which should be changed.

Although `nint` and `nuint` shall be represented by the types `System.IntPtr` and `System.UIntPtr`, respectively, `nint` and `nuint` are *not* aliases for those types. As such, not all members of the corresponding `System` types are defined for `nint` and `nuint`. Instead, the compiler shall make available additional conversions and operations for the types `System.IntPtr` and `System.UIntPtr`, when used in the context of native integer types.
Consider the following:

```csharp
nint a1 = 1;           // OK
System.IntPtr a2 = 1;  // Error: no implicit conversion
```

While the implementation provides operations and conversions for `nint` and `nuint` that are appropriate for integer types, those operations and conversions are not available on the `System` type counterparts. Similarly,

```csharp
M((nint)1);

static void M(dynamic d)
{
    var v = d >> 2; // RuntimeBinderException: '>>' cannot be applied to operands
                    // of type System.IntPtr/System.UIntPtr and int
}
```

The only constructor for `nint` or `nuint` is the parameter-less constructor.

The following members of `System.IntPtr` and `System.UIntPtr` are explicitly excluded from `nint` or `nuint`:

```csharp
// constructors
// arithmetic operators
// implicit and explicit conversions
public static readonly IntPtr Zero; // use 0 instead
public static int Size { get; }     // use sizeof() instead
public static IntPtr Add(IntPtr pointer, int offset);
public static IntPtr Subtract(IntPtr pointer, int offset);
public int ToInt32();
public long ToInt64();
public void* ToPointer();
```

The remaining members of `System.IntPtr` and `System.UIntPtr` are implicitly included in `nint` and `nuint`. These are:

```csharp
public override bool Equals(object obj);
public override int GetHashCode();
public override string ToString();
public string ToString(string format);
```
Interfaces implemented by `System.IntPtr` and `System.UIntPtr` are implicitly included in `nint` and `nuint`, with occurrences of the underlying types replaced by the corresponding native integer types. For example, if `IntPtr` implements `ISerializable`, `IEquatable<IntPtr>`, and `IComparable<IntPtr>`, then `nint` implements `ISerializable`, `IEquatable<nint>`, and `IComparable<nint>`.

`nint` and `System.IntPtr`, and `nuint` and `System.UIntPtr`, are considered equivalent for overriding, hiding, and implementing, however.

Overloads cannot differ by `nint` and `System.IntPtr`, and `nuint` and `System.UIntPtr`, alone. However, overrides and implementations may differ by `nint` and `System.IntPtr`, or `nuint` and `System.UIntPtr`, alone.

Methods hide other methods that differ by `nint` and `System.IntPtr`, or `nuint` and `System.UIntPtr`, alone.

`typeof(nint)` is `typeof(System.IntPtr)`, and `typeof(nuint)` is `typeof(System.UIntPtr)`.[

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -117,8 +117,6 @@ All identity conversions are symmetric. If an identity conversion exists from `T
 
 In most cases, an identity conversion has no effect at runtime. However, since floating point operations may be performed at higher precision than prescribed by their type ([§8.3.7](types.md#837-floating-point-types)), assignment of their results may result in a loss of precision, and explicit casts are guaranteed to reduce precision to what is prescribed by the type ([§12.9.8](expressions.md#1298-cast-expressions)).
 
-There is an identity conversion between `nint` and `System.IntPtr`, and between `nuint` and `System.UIntPtr`.
-
 For the compound types array, nullable type, constructed type, and tuple, there is an identity conversion between native integers ([§8.3.6](types.md#836-integral-types)) and their underlying types.
 
 ### 10.2.3 Implicit numeric conversions
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -287,6 +287,8 @@ As an example of numeric promotion, consider the predefined implementations of t
 ```csharp
 int operator *(int x, int y);
 uint operator *(uint x, uint y);
+nint operator *(nint x, nint y);
+nuint operator *(nuint x, nuint y);
 long operator *(long x, long y);
 ulong operator *(ulong x, ulong y);
 float operator *(float x, float y);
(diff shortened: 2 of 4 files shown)


BillWagner:
Thanks!

PR #1601, opened 2026-03-10
Expand all contractions except those used as examples.

### LINKED ISSUE #1596: To contract or not to contract, that is the question ;-)
A recent PR contained a very obvious “arn’t” which just jumped off the page.

There is a general recommendation in style guides to avoid contractions in formal and/or technical documents, and the Standard is a formal technical document…

So time for some research, first contractions used which have a single possible expansion:

<table>
<tr align=left><th>Contraction</th><th>Count</th><th>Expansion</th><th>Count</th><tr>
<tr><td>aren’t</td><td>2</td><td>are not</td><td>109</td><tr>
<tr><td>can’t</td><td>9</td><td>cannot</td><td>161 [‡]</td><tr>
<tr><td>doesn’t</td><td>23</td><td>does not</td><td>179</td><tr>
<tr><td>don’t</td><td>4</td><td>do not</td><td>75</td><tr>
<tr><td>hasn’t</td><td>1</td><td>has not</td><td>1</td><tr>
<tr><td>isn’t</td><td>22</td><td>is not</td><td>437</td><tr>
<tr><td>shouldn’t</td><td>1</td><td>should not</td><td>4</td><tr>
<tr><td>wasn’t</td><td>2</td><td>was not</td><td>6</td><tr>
<tr><td>won’t</td><td>4</td><td>will not</td><td>12</td><tr>
</table>

[‡] There is also one “can not” and style guides tend to recommend “cannot”

The above contractions could all be expanded with find & replace or automatically by, say, expanding the Smarten tool.

Next contractions used which have possible multiple expansions:

<table>
<tr align=left><th>Contraction</th><th>Count</th><th>Expansion</th><th>Count</th><tr>
<tr><td>it’s</td><td>4</td><td>it is [†], it has</td><td>464</td><tr>
<tr><td>there’s</td><td>2</td><td>there is [†], there has</td><td>135</td><tr>
<tr><td>what’s</td><td>1</td><td>what is [†], what has, what does</td><td>5</td><tr>
</table>

[†] Meanings in use

As these contractions have possible multiple expansions, though current uses are all the “is” variant, automatic handling by Smarten would be a challenge; however manual find & replace is easy given the low numbers.

The numbers show that the Standard tends to avoid contractions.

Recommended Actions

- Make it policy to avoid contractions.
- Produce a PR to expand the current uses.
- Consider an update to Smarten, at least for the single expansion cases, to handle this automatically going forward.

Please up/down vote if you think this should/should not be done. I’ll do the suggested PR if the vote indicates.

Nigel-Ecma:
Fixes #1596

While the “vote” on 1596 is 100% for this, it is hardly representative! ;-) However there is a meeting shortly so I’m clearing my desk and pushing this out.


Nigel-Ecma:
Note that since #1596 I found one case where a contraction is used as an example, see `lexical-structure.md` lines 1439 & 1444, so a simple F&R addition to Smarten to do this automatically wouldn’t be so simple and likely fragile (if it looked for the exception). When I first wrote Smarten I did consider whether a “don’t smarten” tag might be needed in the text on occasions, I decided not at the time… Maybe checking for stray contractions will need to be a pre-publishing manual check.

jskeet (approved):
Upvoted 1596, and these changes look good to me.

jskeet:
Any reason not to update the example as well? For the example you've given, I think we could easily change the message.

One place which *could* be tricky is where we've taken text from other documentation, e.g. in the standard library. It looks like this PR doesn't change any of those files - is that because there were no changes to make, or because you excluded them from your F&R?

BillWagner (approved):
This LGTM as well.

Nigel-Ecma:
@jskeet wrote:

> Any reason not to update the example as well? For the example you've given, I think we could easily change the message.

Reason: I was just trying to fix the language and not overload the PR ;-)

> One place which _could_ be tricky is where we've taken text from other documentation, e.g. in the standard library. It looks like this PR doesn't change any of those files - is that because there were no changes to make, or because you excluded them from your F&R?

They were not excluded. I did consider excluding text in comments, as they might be seen as informal, but decided against it. I suggest an attributed in-text quote should remain true to the source, but would not be concerned over text we derive from library documentation as we invariably edit that anyway.

PR #1608, opened 2026-03-16
Update “tuple” handling – Alternative “2-way” version of #1560

### LINKED ISSUE #1155: Tuple conversions are incorrectly specified

Nigel-Ecma:
Fixes: #1155
Replaces: PR #1366, #1538, #1570
Alternative to: #1560

This replaces #PR1560’s 3-way division of tuples & deconstruction; into *tuple_literal*, *deconstructing_assignment* and *local_deconstructing_declaration*; with a 2-way division; into *tuple_literal* and *deconstructing_assignment*.

The distinction between a deconstruction assigning to *existing* vs. *new* variables is not syntactic in this version. Instead a textual list of restrictions is specified which; by governing what a *deconstructor* can contain based on its position within the code; distinguish the two cases.

Other changes since PR1560 include:

- Renamed abridged_* rules per suggestion.
- Inclusion of *declaration_expression* as an option in a *for_initializer* (this was an accidental omission in PR1560).
- Other editorial changes.

This topic is not yet complete: The deconstructing foreach stattement, which was missing in v7, is not yet included. This will either be added in an update to this PR, or as a new PR building on this one. For this reason this PR is marked as draft.


standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+>
+> An implementation can also choose to exchange the two values without constructing the tuple provided the tuple elements are evaluated in order: `b`, `a` and `thing.ExpensiveMethod(x)`; before doing so. *end example*
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+> *Note*: If an implementation elides an intermediate tuple it may also be able to elide now “redundant” (no effect) expressions. For example if an intermediate tuple is the result of an implicit tuple conversion, those implicit conversions have no side effects, and the intermediate tuple is subject to deconstruction where some elements are discarded, then it may be possible to elide the implicit conversion of those discarded elements. *end note*
+
+#### §tuple-runtime-representation Runtime representation
+
+> *Note*: Unlike other types such as arrays, the runtime representation of tuple types is specified in terms of a set of generic value types, and a tuple may be directly referenced in terms of this representation. However the runtime representation of these generic value types remains implementation defined. *end note*
```
BillWagner, line 467: I'm not sure what this note is conveying. We've said (normatively) that a tuple's runtime representation is `System.ValueTuple<...>`. What is "implementation defined" about that?

standard/expressions.md
```diff
@@ -1260,18 +1261,31 @@ In these situations, the boxed instance is considered to contain a variable of t
...
+
+  >```csharp
+  > E.Deconstruct(out T1 v1, ..., out TN vn) andThen (v1, ..., vn);
+  >```
+
+  Where `andThen` is a pseudo C# operation which performs its left-hand operand and then returns its right-operand as the result.
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+  >> *Note*: `andThen` is the equivalent of C & C++’s comma operator. *end note*
```
BillWagner: I think we want all these lines to be indented, as they are part of the bullet item on 1272. (I apologize if they already are, that's just hard are to see.

standard/expressions.md
```diff
@@ -6724,68 +6700,59 @@ The methods above use the generic delegate types `Func<T1, R>` and `Func<T1, T2,
...
+> *Note*: ANTLR grammar semantics enforce this requirement due to the ordering of the alternatives. *Semantically* there is no overlap between the four alternatives, this is a syntactic disambiguation.
 
-The `+=` and `-=` operators with an event access expression as the left operand are called the ***event assignment operator***s. No other assignment operator is valid with an event access as the left operand. The event assignment operators are described in [§12.23.5](expressions.md#12235-event-assignment).
+The *simple_assignment* and *compound_assignment* expressions assign a new value to a variable, a property, or an indexer element. Event assignment ([§12.23.5](expressions.md#12235-event-assignment)), a subset of *compound_assignment*, assigns a new value to an event. The *ref_assignment* expression assigns a variable reference ([§9.5](variables.md#95-variable-references)) to a reference variable ([§9.7](variables.md#97-reference-variables-and-returns)). The *deconstructing_assignment* assigns new values to one or more *variable_reference*s.
```
BillWagner: nit: Would *deconstructing_assignment* require "two or more *variable_reference*s?

standard/expressions.md
```diff
@@ -6926,27 +7038,44 @@ The ref assignment operator shall not read the storage location referenced by th
...
 
 ### 12.23.4 Compound assignment
 
-If the left operand of a compound assignment is of the form `E.P` or `E[Ei]` where `E` has the compile-time type `dynamic`, then the assignment is dynamically bound ([§12.3.3](expressions.md#1233-dynamic-binding)). In this case, the compile-time type of the assignment expression is `dynamic`, and the resolution described below will take place at run-time based on the run-time type of `E`. If the left operand is of the form `E[Ei]` where at least one element of `Ei` has the compile-time type `dynamic`, and the compile-time type of `E` is not an array, the resulting indexer access is dynamically bound, but with limited compile-time checking ([§12.6.5](expressions.md#1265-compile-time-checking-of-dynamic-member-invocation)).
+```ANTLR
```
BillWagner, line 7057: Everything here is correct, but I'm wondering why this is shown as an addition. I'm curious where it was moved from?

BillWagner (commented):
This is great @Nigel-Ecma 

I had a few small questions / nits in text to consider. One may even be just the result of my failing eyesight.

standard/expressions.md
```diff
@@ -1260,18 +1261,31 @@ In these situations, the boxed instance is considered to contain a variable of t
...
-- Otherwise, `E` cannot be deconstructed.
+- If `E` is a *tuple-literal* the result of deconstruction is the expression `E` itself.
 
-Here, `__v` and `__v1, ..., __vn` refer to otherwise invisible and inaccessible temporary variables.  
+- Otherwise, if `E` has a tuple type `(T₁, ..., Tₙ)`, then the result of deconstruction is semantically equivalent to the expression `(E.Item1, ..., E.Itemn)` except `E` is evaluated only once.
 
-> *Note*: An expression of type `dynamic` cannot be deconstructed. *end note*
+- Otherwise if there is a unique instance or extension method `S.Deconstruct`; with `n ≥ 2` output parameters, with types `T₁` to `Tₙ`, and no other parameters; then `E` can be deconstructed.
+
+  The result of the deconstruction is the *tuple-literal* formed from the values returned via the out parameters of a call to `E.Deconstruct(...)`. The result is semantically equivalent to replacing `E` with the following pseudo-code:
```
jskeet: As per email, I think this is where it would be good to re-emphasize in the first sentence that this is effectively a source-to-source transform. That would mitigate my concerns around using tuple-literal.

standard/types.md
Nigel-Ecma: For most types that are part of the language; arrays, class, structs, etc.; the runtime representation of them is up to the implementation – there is nothing in the Standard about how they should be implemented. Tuples are different, they are language types but an implementation requirement is they be implemented using another language type – structs. Those structs are the ones whose runtime representation remains implementation defined. The Note was meant to make this connection clear: tuples -> structs -> implementation defined representation.

standard/expressions.md
Nigel-Ecma: Correct the bullet on 1272 is a multi-paragraph one and also contains one code and one note “display”. They mostly look correctly indented to me, the exception being the Note on 1284 which GitHub is showing with the wrong indentation and two left side bars. My editor shows the Note correctly indented (i.e. indented the same as the code `E.Deconstruct…`), but to achieve that I had to insert a double `>>` which seems logically wrong… But then what is logical about markdown markup… 😉

I think we need a *GitHub markdown whisperer* on this one, @RexJaeschke or @jskeet can either of you suggest the required incantation?

standard/expressions.md
Nigel-Ecma: No, but we’re both wrong! There are two errors in the line:

- It is *zero* or more – a deconstructor can contain only discards (`_`) (and nested deconstructors with only discards…). Thinking two as you did makes sense, tuples must have at least two elements, why I wrote one for the 3-way I cannot remember offhand but I’m “sure” there was a good reason that applies in the 3-way case… 😉
- The line hasn’t been changed from the 3-way where it does indeed assign “new values” to existing variables (or *variable_references*). However in the 2-way it can also declare new variables and assign them initial values…

The next commit will contain:

> The *simple_assignment* and *compound_assignment* expressions assign a new value to a variable, a property, or an indexer element. Event assignment ([§12.23.5](expressions.md#12235-event-assignment)), a subset of *compound_assignment*, assigns a new value to an event. The *ref_assignment* expression assigns a variable reference ([§9.5](variables.md#95-variable-references)) to a reference variable ([§9.7](variables.md#97-reference-variables-and-returns)). The *deconstructing_assignment* assigns values to zero or more *variable_reference*s.

This is only an overview para so I skipped the distinction between existing and newly declared variables, that is already covered in the full text.

standard/expressions.md
Nigel-Ecma: The clause has been edited in the next commit, the changes are not large but hopefully sufficient.

standard/expressions.md
```diff
@@ -1260,18 +1261,29 @@ In these situations, the boxed instance is considered to contain a variable of t
 
 ## 12.7 Deconstruction
 
-Deconstruction is a process whereby an expression gets turned into a tuple of individual expressions. Deconstruction is used when the target of a simple assignment is a tuple expression, in order to obtain values to assign to each of that tuple’s elements.
+Deconstruction is a compile-time transformation whereby an expression is replaced by a *tuple-literal* of individual expressions. Deconstruction is used in deconstructing assignment (§deconstructing-assignment).
```
jskeet, line 1264: This is great - thanks so much for the rewording.

standard/expressions.md
```diff
@@ -1260,18 +1261,29 @@ In these situations, the boxed instance is considered to contain a variable of t
...
+- Otherwise, if `E` has a tuple type `(T₁, ..., Tₙ)`, then the result of deconstruction is semantically equivalent to the expression `(E.Item1, ..., E.Itemn)` except `E` is evaluated only once.
 
-> *Note*: An expression of type `dynamic` cannot be deconstructed. *end note*
+- Otherwise if there is a unique instance or extension method `S.Deconstruct`; with `n ≥ 2` output parameters, with types `T₁` to `Tₙ`, and no other parameters; then `E` can be deconstructed. The result of the deconstruction is semantically equivalent to replacing `E` with the following pseudo-code:
+
+  >```csharp
+  > E.Deconstruct(out T1 v1, ..., out TN vn) andThen (v1, ..., vn);
+  >```
+
+  Where `andThen` is a pseudo C# operation which performs its left-hand operand and then returns its right-operand as the result.
```
jskeet, line 1278: I like this as a way of explaining the operation. It may be slightly less formal than most of the spec, but it's really clear.

standard/expressions.md
```diff
@@ -1260,18 +1261,29 @@ In these situations, the boxed instance is considered to contain a variable of t
...
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+  >> *Note*: `andThen` is equivalent to C & C++’s comma operator. *end note*
```
jskeet, line 1282: This still looks weird in GitHub, but I'd propose that we don't block this PR on it. (We can file a new issue and fix it separately. We'll need to see what it looks like in Word too.)

standard/expressions.md
```diff
@@ -6724,68 +6698,59 @@ The methods above use the generic delegate types `Func<T1, R>` and `Func<T1, T2,
...
-    | right_shift_assignment
+    : deconstructing_assignment
+    | simple_assignment
+    | compound_assignment
+    | ref_assignment
     ;
 ```
 
-The left operand of an assignment shall be an expression classified as a variable, or, except for `= ref`, a property access, an indexer access, an event access or a tuple. A declaration expression is not directly permitted as a left operand, but may occur as a step in the evaluation of a deconstructing assignment.
+If the input can be syntactically recognised as both a *deconstructing_assignment* and any of the other alternatives then the *deconnstructing_assignment* shall be chosen.
```
```suggestion
If the input can be syntactically recognised as both a *deconstructing_assignment* and any of the other alternatives then the *deconstructing_assignment* shall be chosen.
```

standard/expressions.md
```diff
@@ -6891,11 +6856,156 @@ When a property or indexer declared in a *struct_type* is the target of an assig
...
+[TODO] For C#10 the second of these restrictions is removed. This requires minor text changes here and the removal of the corresponding semantic check in the grammar checker
+-->
+There are restrictions on which *deconstructor_element*s are valid in a given context which are not expressed in the grammar:
+
+- a *declaration_expression* can only occur if the containing *deconstructor* is at the start of a *statement* or a member of a *for_initializer*; and
+- a *variable_reference* can only occur if the containing *deconstructor*:
+  - **is not** at the start of a statement, or
+  - **is** at the start of a statement and contains no *declaration_expression*
+
+For these restrictions:
```
jskeet, line 6911: I think we should have a follow-up PR with examples of these restrictions. We can file an issue for that before merging this PR.

standard/expressions.md
jskeet: `assignment_operator` used to just inline all the options, I believe. There was no simple_assignment, compound_assignment etc in terms of the grammar. (I like the fact that they're now separate grammar rules.)

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+The runtime representation of a tuple `(T1, ..., Tn)` is constructed from `System.ValueTuple<...>`  ([§C.3](standard-library.md#c3-standard-library-types-not-defined-in-isoiec-23271)) instances which are a set of generic struct types for representing tuple types of aritys two to seven. Tuples with an arity greater than seven are represented with the generic struct type `System.ValueTuple<T1, ..., T7, TRest>` that in addition to tuple elements has a `Rest` field containing a nested `System.ValueTuple` of the remaining elements. Where only a single additional field is required, the generic struct type `System.ValueTuple<T1>` is used; this type is not considered a tuple type in itself. Where more than seven additional fields are required further `System.ValueTuple<T1, ..., T7, TRest>` instances are nested.
+
+> *Example*:
+>
+> `(T1, T2)` is represented by `ValueTuple<T1, T2>`<br>
+> `(T1, ..., T15)` is represented by `ValueTuple<T1, ..., T7, ValueTuple<T8, ..., T14, ValueTuple<T15>>>`
+>
+> *end example*
+
+The runtime representation of tuples is directly accessible, and tuple & `System.ValueType<...>` types may be used interchangeably subject to the following:
```
```suggestion
The runtime representation of tuples is directly accessible, and tuple & `System.ValueTuple<...>` types may be used interchangeably subject to the following:
```

(Not sure about the "&" here, but I can see why it's being used. We could potentially split this into two sentences instead, and use "and" instead of "&".)

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+
+- Any value of type `(T1, ..., Tn)` may be treated as the equivalent `System.ValueType<...>` value.
```
```suggestion
- Any value of type `(T1, ..., Tn)` may be treated as the equivalent `System.ValueTuple<...>` value.
```

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+- Any value of type `System.ValueType<T1, T2>` through `System.ValueType<T1, T2, T3, T4, T5, T6, T7>` may be treated as the equivalent `(T1, T2)` through `(T1, T2, T3, T4, T5, T6, T7)` tuple value.
```
```suggestion
- Any value of type `System.ValueTuple<T1, T2>` through `System.ValueTuple<T1, T2, T3, T4, T5, T6, T7>` may be treated as the equivalent `(T1, T2)` through `(T1, T2, T3, T4, T5, T6, T7)` tuple value.
```

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+- A value of type `System.ValueTuple<T1, ..., T7, TRest>` may only be treated as a tuple if `TRest` is a tuple or any `System.ValueTuple<...>` type, the latter including `System.ValueType<T1>`.
```
```suggestion
- A value of type `System.ValueTuple<T1, ..., T7, TRest>` may only be treated as a tuple if `TRest` is a tuple or any `System.ValueTuple<...>` type, the latter including `System.ValueTuple<T1>`.
```

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+- Any other value of type `System.ValueType<T1>` may not be treated as a tuple.
```
```suggestion
- Any other value of type `System.ValueTuple<T1>` may not be treated as a tuple.
```

standard/types.md
```diff
@@ -427,22 +429,126 @@ Tuple elements are public fields with the names `Item1`, `Item2`, etc., and can
...
+> <!-- Example: {template:"standalone-console", name:"TupleTypes4", ignoredWarnings:["CS0219"]} -->
+> ```csharp
+> var squares = (1, 4, 9, 16, 25, 36, 49, 64, 81, 100, 121, 144, 169, 196, 225);
+> ```
+>
+> Then the 15th square (`225`) can be addressed as `squares.Item15`, `squares.Rest.Item8` and `squares.Rest.Rest.Item1`.
+>
+> *end example*
+
+Though tuple and `System.ValueType<...>` values may be treated as equivalent, subject to the above, there is an important semantic difference between tuple and `System.ValueType<...>` types – only the former support tuple element names (§tuple-types-general).
```
```suggestion
Though tuple and `System.ValueTuple<...>` values may be treated as equivalent, subject to the above, there is an important semantic difference between tuple and `System.ValueTuple<...>` types – only the former support tuple element names (§tuple-types-general).
```

jskeet (approved):
Various comments and a few typo fixes (which I'm sure Nigel only left in so that I could feel useful) but I hope we're able to merge this in our meeting, creating follow-up issues for minor formatting and examples.

standard/expressions.md
jskeet: I'll file an issue tomorrow.

standard/expressions.md
jskeet: I'll file an issue in the morning.

jskeet (approved):
After discussion, we'll make this ready to review, and merge it. We'll file follow-up issues for:

- deconstructing foreach
- assignment (simple or deconstructing) to discards - search for "zero or more"
- examples of deconstructing assignment restrictions
- markdown handling of nested notes

jskeet:
Merging despite the failing test, which is about our tools rather than the standard.

BillWagner (approved):
This LGTM.

PR #1609, opened 2026-03-17
[Version 11.0] Feature support for static abstract members in interfaces

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/static-abstracts-in-interfaces.md).

## **Issue**

The [MS spec](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/static-abstracts-in-interfaces.md#pattern-matching) contains a section called “Pattern matching.” In there, a change to constant patterns is proposed. However, I *cannot* find the text below they propose augmenting, anywhere in the spec

> A constant pattern tests the value of an expression against a constant value. The constant may be any constant expression, such as a literal, the name of a declared `const` variable, or an enumeration constant. When the input value is not an open type, the constant expression is implicitly converted to the type of the matched expression; if the type of the input value is not *pattern-compatible* with the type of the constant expression, the pattern-matching operation is an error. **If the constant expression being matched against is a numeric value, the input value is a type that inherits from `System.Numerics.INumberBase<T>`, and there is no constant conversion from the constant expression to the type of the input value, the pattern-matching operation is an error.**

Note that the proposed new text refers to `System.Numerics.INumberBase<T>`, which I mention elsewhere in the PR as needing attention.


standard/patterns.md
```diff
@@ -447,9 +447,9 @@ relational_pattern
...
+When the input value has a type for which a suitable built-in binary relational operator is defined, the evaluation of that operator is taken as the meaning of the relational pattern.  Otherwise, the input value is converted to the type of *constant_expression* using an explicit nullable or unboxing conversion.  It is a compile-time error if no such conversion exists. It is a compile-time error if the input type is a type parameter constrained to, or a type inheriting, from `System.Numerics.INumberBase<T>` and the input type has no suitable built-in binary relational operator defined.  The pattern is considered to not match if the conversion fails.  If the conversion succeeds, the result of the pattern-matching operation is the result of evaluating the expression `e «op» v` where `e` is the converted input, «op» is the relational operator, and `v` is the *constant_expression*.
```
RexJaeschke: This is the first mention of the type `System.Numerics.INumberBase<T>`  in the spec. Unless we are going to add all the machinery about that type and how built-in numeric types now have that as a base interface, we'll need some alternative wording here.

BillWagner:
Thanks!

PR #1610, opened 2026-03-18
[Version 11.0] Feature support for UTF-8 string literals

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/utf8-string-literals.md).

## Note 1 re overlap with v11 raw string literal PR

Some spec changes required to support this feature touch the same section(s) as does PR #1599 for V11’s raw string literals. We need to make sure they coexist, depending on which of the two gets merged first.

What is missing from this PR and needs to be added once the raw string PR has been added is the addition of the two Utf8_Suffixes, as shown below:

```ANTLR
fragment Single_Line_Raw_String_Literal
    : Raw_String_Literal_Delimiter  Raw_String_Literal Content
      Raw_String_Literal_Delimiter >>>Utf8_Suffix?<<<
    ;

fragment Multi_Line_Raw_String_Literal
    : Raw_String_Literal_Delimiter Whitespace* New_Line
      (Raw_String_Literal Content | New_Line)* New_Line
      Whitespace* Raw_String_Literal_Delimiter >>>Utf8_Suffix?<<<
    ;
```

## Note 2 re Lowering

Re the [“Lowering” section of the MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/utf8-string-literals.md#lowering), that appears to be about implementation details. In any event, Rex did *not* add anything from there to the spec.

## Note 3 re “Unresolved questions”

Re the [“Unresolved questions” section of the MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-11.0/utf8-string-literals.md#unresolved-questions), no text was added w.r.t implicit conversion from `string`. Should something be added?


standard/lexical-structure.md
```diff
@@ -990,7 +996,26 @@ fragment Quote_Escape_Sequence
...
+
+A *String_Literal* that contains a *Utf8_Suffix* is a ***UTF-8 string literal***, whose type is `System.ReadOnlySpan<byte>` (an indexable collection type), and whose value contains a UTF-8 byte representation of the string. A null terminator (a byte with value zero) is placed beyond the last byte in memory (and outside the length of the `ReadOnlySpan<byte>`) in order to support scenarios that expect null-terminated byte strings. A UTF-8 string literal is not a constant. A UTF-8 string literal without its *Utf8_Suffix* shall be valid UTF-16. (For example, `"\uDC00\uDD00"u8` is ill-formed as one low surrogate cannot be followed by another.)
+
+> *Note*: While every UTF-8 string literal is a `ReadOnlySpan<byte>`, not every `ReadOnlySpan<byte>` represents a UTF-8 string literal. See the description of UTF-8 string concatenation in [§12.13.5](expressions.md#12135-addition-operator). *end note*
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+> *Note*: As `ReadOnlySpan<byte>` is a ref struct type, a UTF-8 string literal cannot be converted to `object` or used as a type parameter ([§16.2.3]( structs.md#1623-ref-modifier)). *end note*
```
KalleOlaviNiemitalo: `allows ref struct` is added in C# 13, and this PR is for C# 11, so that part is OK…

But the wording is imprecise.  A string literal is syntax rather than a type so it couldn't be used as a type parameter (or rather a type argument) anyway.

standard/expressions.md
```diff
@@ -4372,7 +4372,28 @@ The predefined addition operators are listed below. For numeric and enumeration
...
+  The result of the operator is a `string` that consists of the characters of the left operand followed by the characters of the right operand. The string concatenation operator never returns a `null` value. A `System.OutOfMemoryException` may be thrown if there is not enough memory available to allocate the resulting string.
+- UTF-8 string concatenation:
+
+  ```csharp
+  ReadOnlySpan<byte> operator +(ReadOnlySpan<byte> x, ReadOnlySpan<byte> y);
+  ```
+
+  This overload of the binary `+` operator performs concatenation of UTF-8 string literals and the concatenated results thereof (which is much more restrictive than for UTF-16 string concatenation). The operands shall be UTF-8-encoded values.
+  The result of the operator is a ReadOnlySpan<byte> that consists of the bytes of the left operand followed by the bytes of the right operand. The result may be used directly as an operand to the UTF-8 string concatenation operator.
```
KalleOlaviNiemitalo: From this normative text, it is not clear whether redundant parentheses are allowed.

```csharp
using System;
public class C {
    public void M() {
        ReadOnlySpan<byte> a
            = (( "A"u8 + "a"u8 ))
            + (( "B"u8 + "b"u8 ));
    }
}
```

standard/lexical-structure.md
```diff
@@ -990,7 +996,26 @@ fragment Quote_Escape_Sequence
 <!-- markdownlint-enable MD028 -->
 > *Note*: Since a hexadecimal escape sequence can have a variable number of hex digits, the string literal `"\x123"` contains a single character with hex value `123`. To create a string containing the character with hex value `12` followed by the character `3`, one could write `"\x00123"` or `"\x12"` + `"3"` instead. *end note*
 
-The type of a *String_Literal* is `string`.
+A *String_Literal* that does not contain a *Utf8_Suffix* is a ***UTF-16 string literal***, whose type is `string`.
+
+A *String_Literal* that contains a *Utf8_Suffix* is a ***UTF-8 string literal***, whose type is `System.ReadOnlySpan<byte>` (an indexable collection type), and whose value contains a UTF-8 byte representation of the string. A null terminator (a byte with value zero) is placed beyond the last byte in memory (and outside the length of the `ReadOnlySpan<byte>`) in order to support scenarios that expect null-terminated byte strings. A UTF-8 string literal is not a constant. A UTF-8 string literal without its *Utf8_Suffix* shall be valid UTF-16. (For example, `"\uDC00\uDD00"u8` is ill-formed as one low surrogate cannot be followed by another.)
```
KalleOlaviNiemitalo, line 985: Should add a new subclause within 16.5.15 Safe context constraint, saying that when a UTF-8 string literal has a `ReadOnlySpan<byte>` value, the safe-context of that value is caller-context.  That would allow it to be returned out of the function, like so:

```csharp
using System;
public class C {
    public ReadOnlySpan<byte> M() {
        return "xyz"u8;
    }
}
```

Concatenation of UTF-8 string literals would then follow the safe-context rule in 16.5.15.5 Operators.  As the safe-context of both operands would be caller-context, the safe-context of the result would be the same, without having to be specified separately for this case.

KalleOlaviNiemitalo (changes requested):

standard/lexical-structure.md
KalleOlaviNiemitalo: The following would then likewise be valid:

```csharp
public class C {
    public ref readonly byte M() {
        return ref "xyz"u8[0];
    }
}
```

BillWagner:
Thanks!

PR #1620, opened 2026-04-06
[Version 12.0] Feature support for inline arrays

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/inline-arrays.md).

### DIFF
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -709,6 +709,7 @@ The following implicit conversions are classified as standard implicit conversio
 - Boxing conversions ([§10.2.9](conversions.md#1029-boxing-conversions))
 - Implicit constant expression conversions ([§10.2.11](conversions.md#10211-implicit-constant-expression-conversions))
 - Implicit conversions involving type parameters ([§10.2.12](conversions.md#10212-implicit-conversions-involving-type-parameters))
+- Implicit inline array conversions (§ImplicitInlineArrayConversions)
 
 The standard implicit conversions specifically exclude user-defined implicit conversions.
 
--- a/standard/conversions.md
+++ b/standard/conversions.md
@@ -455,6 +455,16 @@ Although an implicit conversion to `object` is permitted, a warning shall be iss
 >
 > *end example*
 
+### §ImplicitInlineArrayConversions Implicit inline array conversions
+
+The implicit inline array (§InlineArray) conversions are:
+
+- From an expression designating a writable inline array with element type `T` to `System.Span<T>`
+- From an expression designating a writable inline array with element type `T` to `System.ReadonlySpan<T>`
+- From an expression designating a readonly inline array with element type `T` to `System.ReadonlySpan<T>`
+
+The conversion of an inline array to a `System.Span` or `System.ReadonlySpan` ignores any declared operators in the inline array type that might otherwise appear to be applicable. See §InlineArray for more information.] 
+
 ## 10.3 Explicit conversions
 
 ### 10.3.1 General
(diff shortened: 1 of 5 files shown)


standard/conversions.md
```diff
@@ -455,6 +455,16 @@ Although an implicit conversion to `object` is permitted, a warning shall be iss
...
 
+### §ImplicitInlineArrayConversions Implicit inline array conversions
+
+The implicit inline array (§InlineArray) conversions are:
+
+- From an expression designating a writable inline array with element type `T` to `System.Span<T>`
+- From an expression designating a writable inline array with element type `T` to `System.ReadonlySpan<T>`
+- From an expression designating a readonly inline array with element type `T` to `System.ReadonlySpan<T>`
+
+The conversion of an inline array to a `System.Span` or `System.ReadonlySpan` ignores any declared operators in the inline array type that might otherwise appear to be applicable. See §InlineArray for more information.
```
KalleOlaviNiemitalo, line 466: Should be `ReadOnlySpan` with capital O.

standard/structs.md
```diff
@@ -463,6 +463,95 @@ constructor declaration. The body of the method assigns each parameter of the De
...
+
+There are a number of restrictions on the instance field’s declaration:
+
+- It shall not contain the *field_modifier*s `readonly`, `ref`, `required`, or `volatile`.
+- It shall not be declared as a *fixed_size_buffer_declaration* ([§24.8](unsafe-code.md#248-fixed-size-buffers)).
+- *type* shall be a type valid as a type argument.
+
+`buffer` is an instance of `Buffer`, and contains an array of five `int` with element `buffer[0]` occupying the same memory as `buffer._element0`. However, unlike an array type, which has a public `Length` property, `buffer` is not derived from `System.Array`, and has no such member.
+
+> *Note*: One could add a public `Length` property to `Buffer` (which returns `sizeof(Buffer)/sizeof(int)`, but that property would have to be an unsafe member. Alternatively, one could perform an implicit conversion to `System.Span<int>` or `System.ReadOnlySpan<int>` and use their `Length` property. *end note*
```
KalleOlaviNiemitalo, line 521: Missing closing parenthesis at the end of `` (which returns `sizeof(Buffer)/sizeof(int)` ``.

The wording "that property would have to be an unsafe member" feels a bit unclear when the unsafety is only due to `sizeof` in the the suggested implementation, rather than required by the property definition in general.

The referenced alternative using "an implicit conversion" seems to mean:

```csharp
[System.Runtime.CompilerServices.InlineArray(5)]
public struct Buffer {
    private int _element0;

    public readonly int Length
    {
        get
        {
            System.ReadOnlySpan<int> span = this;
            return span.Length;
        }
    }
}
```

but there is also a simpler way to implement the property:

```csharp
[System.Runtime.CompilerServices.InlineArray(constLength)]
public struct Buffer {
    private const int constLength = 5;

    private int _element0;

    public readonly int Length => constLength;
}
```

KalleOlaviNiemitalo:
Trying this out at <https://sharplab.io/>

```
System.Console.WriteLine(((dynamic)new Buffer())[2]);

[System.Runtime.CompilerServices.InlineArray(5)]
public struct Buffer {
    private int _element0;
    
    public int this[int index] => 42;
}
```

I get 42 rather than 0 so it seems the `dynamic` part of the MS proposal has not been implemented in that environment.

standard/structs.md
```diff
@@ -477,6 +477,95 @@ constructor declaration. The body of the method assigns each parameter of the De
...
+
+A list pattern ([§11.2.11](patterns.md#11211-list-pattern)) shall not be used in the context of an inline array.
+
+An inline array type is a valid constructible collection target type for a collection literal.
+
+Any indexers or `Slice` methods declared for an inline array type that have signatures matching those defined for non-inline array processing, shall not be used during inline array element access.
+> *Example*: Consider the following:
+>
+> <!-- Example: {template:"standalone-console-without-using", name:"InlineArrays4", ignoredWarnings:["CS9181"]} -->
+```csharp
```
BillWagner: Add block quote:


````suggestion
> ```csharp
````

PR #1621, opened 2026-04-07
V12 set compilation mode for testing to v12

RexJaeschke:
@Bill This PR changes the LangVersion from 11 to 12 for all v12 testing. You might have to ignore/bypass some of the GitHub tests, as their failures are unrelated to this PR, so should not block it.


BillWagner (approved):
This LGTM @RexJaeschke 

I'll :shipit: now.

PR #1624, opened 2026-04-08
[Version 12.0] Feature support for ref readonly parameters

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-12.0/ref-readonly-parameters.md).


KalleOlaviNiemitalo:
The MS proposal says that a warning would be issued when the argument corresponding to a `ref readonly` parameter has no `ref` or `in` annotation.  I didn't find that warning in the diff of this PR.

PR #1632, opened 2026-04-09
Incorporate using_declaration

### LINKED ISSUE #1606: EBNF error with declaration_statement or embedded_statement (draft-v8).
**Describe the bug**
Consider this example:
```
using System.IO;
class Program
{
    static void Main()
    {
            using var r = new StringReader("hello");
    }
}
```
This code compiles and runs perfectly fine with version 8 of the Roslyn compiler. (It does not for version 7.)

This input does not parse using the scraped grammar, using the latest on "draft-v8" branch. This is because there is an error with the statement rules.

https://github.com/dotnet/csharpstandard/blob/089fae96f3774f90cb25e2df730ca7c5a4d47da9/standard/statements.md?plain=1#L10-L31

And, more critically,

https://github.com/dotnet/csharpstandard/blob/089fae96f3774f90cb25e2df730ca7c5a4d47da9/standard/statements.md?plain=1#L284-L288

The Roslyn compiler parser performs the test in ParseStatementStartingWithUsing():

https://github.com/dotnet/roslyn/blob/f881263cdd42dc2526afe797d678c644414f7b3f/src/Compilers/CSharp/Portable/Parser/LanguageParser.cs#L8475-L8476

The most natural location for "| using_declaration" would be in "declaration_statement".

```
 declaration_statement
      : local_variable_declaration ';'
      | local_constant_declaration ';'
      | local_function_declaration
      | using_declaration
      ;
```

You'll need to decide what may be more natural point in the EBNF to make a change, perhaps local_variable_declaration but that would require a bit more refactoring.

In any case, the grammar as is does not work.

**Example**

[v8test-embedded-statement.zip](https://github.com/user-attachments/files/25970331/v8test-embedded-statement.zip)

BillWagner:
Fixes #1606.

Add the using_declaration to the ANTLR grammar for *declaration_statement*, In addition, add a sentence in the clause that points to the correct clause for the declaration.

### DIFF
--- a/standard/statements.md
+++ b/standard/statements.md
@@ -285,10 +285,11 @@ declaration_statement
     : local_variable_declaration ';'
     | local_constant_declaration ';'
     | local_function_declaration
+    | using_declaration
     ;
 ```
 
-A local variable is declared using a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)). A local constant is declared using a *local_constant_declaration* ([§13.6.3](statements.md#1363-local-constant-declarations)). A local function is declared using a *local_function_declaration* ([§13.6.4](statements.md#1364-local-function-declarations)).
+A local variable is declared using a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)). A local constant is declared using a *local_constant_declaration* ([§13.6.3](statements.md#1363-local-constant-declarations)). A local function is declared using a *local_function_declaration* ([§13.6.4](statements.md#1364-local-function-declarations)). A using declaration is declared using a *using_declaration* (§13.14.2).
 
 The declared names are introduced into the nearest enclosing declaration space ([§7.3](basic-concepts.md#73-declarations)).
 
--- a/standard/statements.md
+++ b/standard/statements.md
@@ -278,20 +278,20 @@ In addition to the reachability provided by normal flow of control, a labeled st
 
 ### 13.6.1 General
 
-A *declaration_statement* declares one or more local variables, one or more local constants, a local function, or a local using scope ([§13.14.2](statements.md#13142-using-declaration)). Declaration statements are permitted in blocks and switch blocks, but are not permitted as embedded statements. A *using_declaration* differs from the other forms in that it shall not appear directly in a *switch_block*, but may appear in a block nested within a *switch_block*.
+A *declaration_statement* declares one or more local variables, one or more local constants, a local function, or a local using scope ([§13.14.2](statements.md#13142-using-declaration)). Declaration statements are permitted in blocks and switch blocks, but are not permitted as embedded statements. A *local_using_declaration* differs from the other forms in that it shall not appear directly in a *switch_block*, but may appear in a block nested within a *switch_block*.
 
 ```ANTLR
 declaration_statement
     : local_variable_declaration ';'
     | local_constant_declaration ';'
     | local_function_declaration
-    | using_declaration
+    | local_using_declaration
     ;
 ```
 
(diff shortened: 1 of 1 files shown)


standard/statements.md
```diff
@@ -285,10 +285,11 @@ declaration_statement
     : local_variable_declaration ';'
     | local_constant_declaration ';'
     | local_function_declaration
+    | using_declaration
```
Nigel-Ecma: I suggest changing the rule name to `local_using_declaration`, this follows the pattern of the others and IIRC I saw it used in some of the source material.

standard/statements.md
Nigel-Ecma: With the introduction of using declarations this is no longer correct – a using declaration is not allowed directly in a switch block, it has to be inside a block in a switch block.

The first sentence should also include something about a local using scope or something.

Also see suggestion below about adding x-refs to each item in the list.

standard/statements.md
```diff
@@ -285,10 +285,11 @@ declaration_statement
...
     ;
 ```
 
-A local variable is declared using a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)). A local constant is declared using a *local_constant_declaration* ([§13.6.3](statements.md#1363-local-constant-declarations)). A local function is declared using a *local_function_declaration* ([§13.6.4](statements.md#1364-local-function-declarations)).
+A local variable is declared using a *local_variable_declaration* ([§13.6.2](statements.md#1362-local-variable-declarations)). A local constant is declared using a *local_constant_declaration* ([§13.6.3](statements.md#1363-local-constant-declarations)). A local function is declared using a *local_function_declaration* ([§13.6.4](statements.md#1364-local-function-declarations)). A using declaration is declared using a *using_declaration* (§13.14.2).
```
Nigel-Ecma: The original para was bad enough, it is just stating the blindly obvious. (I am no fan of all the prose paras which just re-state grammar, bet that surprises no one! ;-)). In this case why not just add the x-refs to the first sentence of the para on line 281 and drop this para?

standard/statements.md
```diff
@@ -285,10 +285,11 @@ declaration_statement
...
 
 The declared names are introduced into the nearest enclosing declaration space ([§7.3](basic-concepts.md#73-declarations)).
```
Nigel-Ecma: This is no longer correct, a using declaration introduces a *new* declaration space/scope/lifetime. A using declaration, §13.14.2, is syntactic sugar for a using statement, §13.4.1, whose definition introduces a block, §13.3, which introduces a new declaration space & scope, §7.3 & §13.3.1…

Nigel-Ecma (changes requested):
The grammar is fine (I have suggested a cosmetic rule name change), and this is a good place to link this in.

Unfortunately the using declaration differs from the others wrt to declaration spaces so adding it requires more prose work. The existing text for using statements & declarations in §13.14 itself does not directly discuss declaration spaces, instead relying on the reader to deduce this detail from the semantic equivalent code – that choice should be reviewed in light of this PR.

standard/statements.md
BillWagner: Done in latest commit.

standard/statements.md
BillWagner: Done in the latest commit.

standard/statements.md
BillWagner: Updated in the latest commit.

standard/statements.md
BillWagner: Updated in the last commit.

jskeet (approved):
Looks good with the additional prose changes

Nigel-Ecma (approved):

PR #1639, opened 2026-04-16
Clarify LHS as discard behavior

### LINKED ISSUE #1631: null literal conversions and implicit reference conversions

BillWagner:
Fixes #1631

"zero or more variable_references" → "two or more targets" (grammar requires 2+ elements; targets can be variable_references or discards) 
Adds: compound_assignment LHS shall not be a discard
Adds: simple_assignment and ref_assignment LHS may be a discard
Adds: RHS is evaluated even when target is a discard, but no value is stored

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -6711,7 +6711,9 @@ If the input can be syntactically recognised as both a *deconstructing_assignmen
 
 > *Note*: ANTLR grammar semantics enforce this requirement due to the ordering of the alternatives. *Semantically* there is no overlap between the four alternatives, this is a syntactic disambiguation.
 
-The *simple_assignment* and *compound_assignment* expressions assign a new value to a variable, a property, or an indexer element. Event assignment ([§12.23.6](expressions.md#12236-event-assignment)), a subset of *compound_assignment*, assigns a new value to an event. The *ref_assignment* expression assigns a variable reference ([§9.5](variables.md#95-variable-references)) to a reference variable ([§9.7](variables.md#97-reference-variables-and-returns)). The *deconstructing_assignment* assigns values to two or more targets, where each target is either a *variable_reference* or a discard. The left operand of a *simple_assignment* or *ref_assignment* may also be a discard. The left operand of a *compound_assignment* shall not be a discard. When the left operand of a *simple_assignment* or *ref_assignment*, or a target of a *deconstructing_assignment*, is a discard, the corresponding right-side expression is evaluated but no value is stored.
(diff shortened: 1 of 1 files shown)


standard/expressions.md
```diff
@@ -6711,7 +6711,7 @@ If the input can be syntactically recognised as both a *deconstructing_assignmen
...
+The *simple_assignment* and *compound_assignment* expressions assign a new value to a variable, a property, or an indexer element. Event assignment ([§12.23.6](expressions.md#12236-event-assignment)), a subset of *compound_assignment*, assigns a new value to an event. The *ref_assignment* expression assigns a variable reference ([§9.5](variables.md#95-variable-references)) to a reference variable ([§9.7](variables.md#97-reference-variables-and-returns)). The *deconstructing_assignment* assigns values to two or more targets, where each target is either a *variable_reference* or a discard. The left operand of a *simple_assignment* or *ref_assignment* may also be a discard. The left operand of a *compound_assignment* shall not be a discard. When the left operand of a *simple_assignment* or *ref_assignment*, or a target of a *deconstructing_assignment*, is a discard, the corresponding right-side expression is evaluated but no value is stored.
```
jskeet: Do we want something for the simple assignment at the start of this line too?

Nigel-Ecma: “The *deconstructing_assignment* assigns values to **two or more** targets … is a discard, the corresponding right-side expression is evaluated but **no value is stored**”

Reads a little contradictory… As a discard is defined as a local variable, §9.2.9.2, the same could be said for “assign a new value to a variable” for simple & ref assignment. And as a discard is a local variable saying the LHS “may also be a discard” when it a;ready can be a variable…

Summary paras can be hard!

Currently the compound assignment clause doesn’t mention discards directly, leaving the reader to infer that detail from the equivalent code it is treated as; and of course there a get/set references which omit the later ref get but that is hopefully being addressed in another PR…

I suggest leaving the para as is, dropping the introduction of *target*, and adding a new second para to just summarise what the subclauses say regarding discards; and also revising the compound subclause to include that discards are not allowed (or else how can the summary para summarise something that isn't there).

Nigel-Ecma (changes requested):

standard/expressions.md
BillWagner: Next commit contains updates based on this conversation.

jskeet (approved):
I like this latest version.

Nigel-Ecma (approved):

PR #1678, opened 2026-05-07
Fix consistency in description of nullability

### LINKED ISSUE #1664: 8.9.1 inconsistency in punctuation and "type-ness"
Nit found while reviewing #1637

8.9.1 includes (quoted here; not a block quote in the original):

> There are two forms of nullability for reference types:
> 
> - *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
> - *non-nullable*: A *non-nullable reference* should not be assigned a `null` value. Its default null state is *not-null*.

There are two inconsistencies here:

- Are we discussing the references, or the types? I *think* it should be references
- Should we have hyphens or not between "reference" and "type"?

Basically I *think* the first bullet should be:

> - *nullable*: A *nullable reference* can be assigned `null`. Its default null state is *maybe-null*.

(Even the idea of a reference itself being assigned feels a bit odd. We could add "variable" in both places, but I suspect that's a bit too specific.)

jskeet:
Fixes #1664

### DIFF
--- a/standard/types.md
+++ b/standard/types.md
@@ -834,7 +834,7 @@ A *nullable reference type* is denoted by appending a *nullable_type_annotation*
 
 There are two forms of nullability for reference types:
 
-- *nullable*: A *nullable-reference-type* can be assigned `null`. Its default null state is *maybe-null*.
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM. I'll :shipit:

PR #1680, opened 2026-05-08
Patterns: address precedence and disambiguation

### LINKED ISSUE #1285: Issues with is-operator (§12.12.12, §12.12.1, §6.2.5)
The final para of §6.2.5 is:

> When recognising a *relational_expression* ([§12.12.1](expressions.md#12121-general)) if both the “*relational_expression* `is` *type*” and “*relational_expression* `is` *pattern*” alternatives are applicable, and *type* resolves to an accessible type, then the “*relational_expression* `is` *type*” alternative shall be chosen.

The rest of §6.2.5 is about disambiguation *type_argument_list*, and this para is just about
lost after the disambiguation examples. This information should be relocated to §12.12.1 (*relational_expression*) and/or §12.12.12 (`is` operator).

In §12.12.1 following the grammar for *relational_expression* & *equality_expression* there
is:

> *Note*: Lookup for the right operand of the `is` operator must first test as a *type*, then as an *expression* which may span multiple tokens. In the case where the operand is an *expression*, the pattern expression must have precedence at least as high as *shift_expression*. *end note*

“must have” sounds rather normative for a non-normative note, and the statement “precedence at least as 
high as *shift_expression*” is rather curious without further explanation – if precedence prevents
the writing of what is required then using parenthesis is usually indicated, which is supported
by *expression*, and something the Standard doesn’t repeat all over the place…

In §12.12.12 and its two subclauses on the is-type & is-pattern operators there is no mention of
disambiguation between them or of any precedence requirements.

The current text in §6.2.5 probably needs to be incorporated into §12.12.1/12, and that in §12.12.1 at minimum revised.

Note that `is` will change in v9, including the is-type/is-pattern distinction,
which should be kept in mind in any fix-up now. Marking this as v8 for now…

### LINKED ISSUE #1428: Precedence of expression-pattern
# Precedence of expression-pattern

The language grammar currently contains these productions:

```ANTLR
constant_pattern
    : constant_expression
    ;

constant_expression
    : expression
    ;
```

These imply that a pattern can be, syntactically, any expression.
However, because patterns appear _within_ expressions, ambiguity can arise.
This document describes how we resolve these ambiguities.

> Note: The grammar in grammar.md does not contain the full grammar for patterns.

In fact, we do not accept expressions looser than _conditional_expression_; none of them make sense for patterns.
So the basic grammar should be

```ANTLR
constant_pattern
    : conditional_expression // restricted to be a valid constant expression
    ;
```

## Expanded patterns grammar

We expand the pattern grammar to exclude expressions (i.e. the constant pattern),
so that we can explicitly include expressions at an appropriate precedence for the context.
In the Roslyn compiler, which uses a recursive descent parser,
we accomplish this by passing a precedence level to the function that parses a pattern.
We can express this in `ANTLR`:

```ANTLR
// any pattern except a constant pattern
pattern_no_expression
    : declaration_pattern
    | var_pattern
    | positional_pattern
    | property_pattern
    | discard_pattern
    ;

// any pattern
pattern
    : pattern_no_expression
    | conditional_expression // restricted to be a valid constant expression
    ;
```

### The _is pattern_ expression

The grammar is currently written

```ANTLR
relational_expression
    : relational_expression 'is' pattern
    ;
```

The _is pattern_ expression would introduce an ambiguity in cases such as

```csharp
    _ = A is B is C;
```

This could be interpreted syntactically as `(A is B) is C` or as `A is (B is C)`. To eliminate this ambiguity, we restrict the pattern on the right-hand-side to an appropriate precedence:

```ANTLR
relational_expression
    : relational_expression 'is' pattern_no_expression
    | relational_expression 'is' relational_expression  // RHS restricted to be a valid constant expression
    ;
```

### switch expression arms

The grammar currently in the spec for switch expressions contains

```ANTLR
switch_expression_arm
    : pattern case_guard? '=>' switch_expression_arm_expression
    ;
```

However, this can introduce a syntactic ambiguity if the _pattern_ is a lambda expression:

```csharp
    _ = A switch {
        a => b => c
    }
```

Is this a case trying to match the value of the constant expression `a => b` and yielding the value of `c`, or is it trying to match the constant expression `a` and yield the value `b => c`? Both are syntactically valid parses given the draft grammar. So we modify the grammar to eliminate the ambiguity:

```ANTLR
switch_expression_arm
    : pattern_no_expression case_guard? '=>' switch_expression_arm_expression
    | null_coalescing_expression case_guard? '=>' switch_expression_arm_expression
    ;
```

Nigel-Ecma:
Resolves #1285, resolves #1428

1285:
- text incorrectly placed in §6.2.5 removed
- text added to §12.14.1 to cover disambiguation is-type/is-pattern
- the comment regarding “precedence at least as high" is covered under 1428, see HTML comment bullet below

1428:
- The suggested grammar changes reduce to changing one rule, *constant_expression*
- This may seem surprising but stems from the Standard using ANTLR grammar semantics, particularly in how left recursion (used extensively to model precedence in the grammar) is translated to iteration and ANTLR’s LL(*) lookahead.
> Note: The new sample, which like other samples contains an SVG of the parse structure, can be compared with the syntax tree produced by Roslyn by installing the Syntax Visualizer in Visual Studio and having it parse the sample (allowing for any semantic errors that result) – the former uses the grammar in this PR and the latter uses a modified grammar along the lines proposed in 1428. Observe that two produce the same resolutions for operator evaluation order.
- Inserted HTML comment [TODO] note in §12.14.1 regarding reviewing decision on “is pattern” to build on ANTLR semantics rather than add additional pattern rules.

New sample “Pattern precedence” to check relational_expression, and lambdas in conjunction with patterns, have correct precedence.


BillWagner (approved):
Thanks @Nigel-Ecma 

This LGTM.

I'm happy merging before the next meeting, but I'd like to confirm with @jskeet

standard/expressions.md
```diff
@@ -7151,7 +7150,7 @@ A constant expression is an expression that shall be fully evaluated at compile-
 
 ```ANTLR
 constant_expression
-    : expression
+    : conditional_expression
```
jskeet, line 7153: I've been trying to follow how this works, and failing.

`conditional_expression` is

```text
conditional_expression
    : null_coalescing_expression
    | null_coalescing_expression '?' expression ':' expression
    | null_coalescing_expression '?' 'ref' variable_reference ':'
      'ref' variable_reference
    ;
```

How does the constant `true` for example meet any of those? Given the lack of other tokens, the only possible option is `null_coalescing_expression`, which is:

```text
null_coalescing_expression
    : conditional_or_expression
    | conditional_or_expression '??' null_coalescing_expression
    | throw_expression
    ;
```

Again, only the first option can possibly apply, so let's find `condition_or_expression`:

```text
conditional_and_expression
    : inclusive_or_expression
    | conditional_and_expression '&&' inclusive_or_expression
    ;

conditional_or_expression
    : conditional_and_expression
    | conditional_or_expression '||' conditional_and_expression
    ;
```

... leading us to `inclusive_or_expression`:

```text
and_expression
    : equality_expression
    | and_expression '&' equality_expression
    ;

exclusive_or_expression
    : and_expression
    | exclusive_or_expression '^' and_expression
    ;

inclusive_or_expression
    : exclusive_or_expression
    | inclusive_or_expression '|' exclusive_or_expression
    ;
```

At this point, I've stopped going down the rabbit-hole.

There's nothing conditional about constants such as `"text"`, `5` or `false` so I think it's confusing to use `conditional_expression` here, at least without further explanation. If it's really what we want - which I can believe, but will want further explanation - we should add some informative text.

Nigel-Ecma: @jskeet – You searched in the wrong direction ;-)


The  *constant_expression* rule you are looking at was:

```ANTLR
constant_expression
    : expression
    ;
```

and this is the grammar for *expression*:

```ANTLR
expression
    : non_assignment_expression
    | assignment
    ;

non_assignment_expression
    : conditional_expression
    | lambda_expression
    | query_expression
    ;
```

*assignment*, *lambda_expression* and *query_expression* are not valid as constant expressions so the change here changes the definition of *constant_expression* to *conditional_expression* which is where the semantic requirement of being a constant gets you anyway – this simplifies the grammar/parser, and therefore reduces the possibility for unexpected grammar ambiguities, as it reduces the syntactic options which can be accepted only to be rejected based on semantics.

jskeet: I can see that it's excluding appropriate things. I'm still confused as to how to get down from `conditional_expression` to a constant though. (I strongly suspect our grammar names aren't as helpful as they might be, too.)

I *think* I've got it now - via a shocking number of levels.

I think it goes like this, where each of these can be "just the next one down":

- `conditional_expression`
- `null_coalescing_expression`
- `conditional_or_expression`
- `conditional_and_expression`
- `inclusive_or_expression`
- `exclusive_or_expression`
- `and_expression`
- `equality_expression`
- `relational_expression`
- `shift_expression`
- `additive_expression`
- `multiplicative_expression`
- `switch_expression`
- `range_expression`
- `unary_expression`
- `primary_expression`
- `literal`

I would never have guessed that `true` would count as a valid `null_coalescing_expression` for example. I don't know what shocks me more - the depth of all of the above, or that I've managed to be on this technical group for so long without realizing all of this...

jskeet (approved):

jskeet:
I think it would be good to wait until the meeting before merging this, just so we can check that my understanding is correct :)

PR #1685, opened 2026-05-11
Fix issues with Positional Patterns

### LINKED ISSUE #1647: Address issues with positional pattern description
Need to check whether/how much of this was done

---

The cases this covers:

- if the input value is a tuple type: the number of subpatterns must statically match the tuple arity, subpatterns & tuple elements are matched pairwise, order not specified, match can fail without checking all subpatterns.
- if the input value has a `Deconstruct` method whose number out parameters match the number of subpatterns, and the types of each parameter is compatible with the corresponding subpatterns matched pairwise then the values from invoking `Deconstruct` are matched pairwise, order not specified, match can fail without checking all subpatterns.
- if the input value implements `ITuple` then the number of elements in the `ITuple` value is checked at runtime to be equal to the number of subpatterns. Then the elements of the `ITuple` value are matched pairwise… etc.

I’ve listed these cases in the probable order of priority, i.e. if the input value is a tuple the first case is picked even though tuples implement `ITuple` (if that is added to the Standard) and `Deconstruct` (if that is added to the Standard); next if the type of the input value implements a suitable `Deconstruct` then case 2 is chosen; finally, if applicable, case 3 with dynamic arity check is chosen, and a compile time error occurs if it is not applicable.

Is this ordering correct, note it does not match the current text?

Whatever the ordering it needs to be clearly specified.

_Originally posted by @Nigel-Ecma in https://github.com/dotnet/csharpstandard/pull/873#discussion_r2377971015_

BillWagner:
Fixes #1647

Rewrite §11.2.5 Positional pattern in patterns.md so the three matching strategies and their priority order are unambiguous, and add System.Runtime.CompilerServices.ITuple to the standard library annex. Roslyn-verified: named subpatterns are rejected when matching goes through ITuple, so that prohibition stays in the spec.

I had to add the `System.Runtime.CompilerServices.ITuple` interface to Annex C.3. I followed the practice we agreed to in the last meeting to add `///` comments that describe the necessary semantics for the type.


BillWagner:
The changes look good, so I'll be using this to figure out what's wrong in the tools.

BillWagner:
See #1687 for tools fixes.

RexJaeschke (approved):
Bill, it looks good from an editorial pov.

standard/patterns.md
```diff
@@ -196,7 +196,7 @@ If *designation* is a *tuple_designation*, the pattern is equivalent to a *posit
 
 ### 11.2.5 Positional pattern
 
-A *positional_pattern* checks that the input value is not `null`, invokes an appropriate `Deconstruct` method ([§12.7](expressions.md#127-deconstruction)), and performs further pattern matching on the resulting values.  It also supports a tuple-like pattern syntax (without the type being provided) when the type of the input value is the same as the type containing `Deconstruct`, or if the type of the input value is a tuple type, or if the type of the input value is `object` or `System.ITuple` and the runtime type of the expression implements `System.ITuple`.
+A *positional_pattern* checks that the input value is not `null`, extracts a sequence of values from it, and matches each extracted value against a corresponding *subpattern*. The values are extracted in one of three ways: by treating the input as a tuple, by invoking a `Deconstruct` method ([§12.7](expressions.md#127-deconstruction)), or by indexing the input through `System.Runtime.CompilerServices.ITuple`.
```
Nigel-Ecma: §12.7 is not a suitable reference here – it defines deconstruction in terms of a source-level transformation, that may involve `Deconstruct`, which results in a *tuple-literal*. This is not the way `Deconstruct` is used in patterns.

standard/patterns.md
```diff
@@ -211,14 +211,14 @@ subpattern
     ;
 ```
 
-Given a match of an input value to the pattern *type* `(` *subpatterns* `)`, a method is selected by searching in *type* for accessible declarations of `Deconstruct` and selecting one among them using the same rules as for the deconstruction declaration.
-It is an error if a *positional_pattern* omits the type, has a single *subpattern* without an *identifier*, has no *property_subpattern* and has no *simple_designation*. This disambiguates between a *constant_pattern* that is parenthesized and a *positional_pattern*.
-In order to extract the values to match against the patterns in the list,
+It is an error if a *positional_pattern* omits the *type*, has a single *subpattern* without an *identifier*, has no *property_subpattern* and has no *simple_designation*. This disambiguates between a *constant_pattern* that is parenthesized and a *positional_pattern*.
```
Nigel-Ecma: Requiring disambiguation is not an error, its a feature ;-) Disambiguation rules tend to be written along the lines of “if the input can be recognised as A or B then A shall be chosen” and placed after the grammar rule containing A & B. E.g. after *assignment* in §12.23.1 we have:

> If the input can be syntactically recognised as both a *deconstructing_assignment* and any of the other alternatives then the *deconstructing_assignment* shall be chosen.

So a similar disambiguation should be added after *pattern* in §12.2.1 and this para can go.

standard/patterns.md
```diff
@@ -211,14 +211,14 @@ subpattern
...
+2. **Deconstruct form.** Otherwise, if either *type* is present, or *type* is omitted and the static type of the input value contains an accessible `Deconstruct` method, then a `Deconstruct` method is selected from that type using the same rules as for a deconstruction declaration ([§12.7](expressions.md#127-deconstruction)), with the additional requirement that its number of `out` parameters is equal to *n*; it is a compile-time error if no such method exists. If *type* is present, it is a compile-time error if the static type of the input value is not pattern-compatible with *type*; at runtime the input value is tested against *type* and, if that test fails, the positional pattern match fails. Otherwise, the input value is converted to the type containing the selected `Deconstruct` method and that method is invoked with fresh variables receiving its `out` parameters. Each received value is matched against the corresponding *subpattern*, and the match succeeds if all of these succeed. If any *subpattern* has an *identifier*, that *identifier* shall name the parameter at the corresponding position of `Deconstruct`.
```
Nigel-Ecma: I got lost parsing this, is it meant to say that `Deconstruct` selected from the *type* or failing that the static type of the input value? Also a second ref to §12.7.

Also does the use of compile-time error in this list may mean “move to next bullet” and there is only a compile-time error if bullet 4 is reached? Or does it mean there is a point in each bullet when the choice is made and after that there can be a compile-time error?

standard/patterns.md
```diff
@@ -211,14 +211,14 @@ subpattern
...
-- Otherwise, if *type* is omitted, and the input value is of type `object` or some type that can be converted to `System.ITuple` by an implicit reference conversion, and no *identifier* appears among the subpatterns, then the match uses `System.ITuple`.  
-- Otherwise, the pattern is a compile-time error.
+Let *n* be the number of *subpattern*s appearing between the parentheses. The matching strategy is selected at compile time by applying the following cases in order; the first case whose conditions are satisfied is used, and the remaining cases are not considered.
+
+1. **Tuple form.** If *type* is omitted and the static type of the input value is a tuple type ([§8.3.11](types.md#8311-tuple-types)), then it is a compile-time error if *n* is not equal to the arity of that tuple type. At runtime, each tuple element is matched against the corresponding *subpattern*; the match succeeds if all of these succeed. If any *subpattern* has an *identifier*, that *identifier* shall name the tuple element at the corresponding position in the tuple type.
```
Nigel-Ecma: It is unclear to me where an input value that is a *tuple-literal* without a type (e.g. `(1, null)`) is covered, under `is` maybe?

Nigel-Ecma (changes requested):
A few suggestions made.

Deconstruction and pattern matching are different but involve similar things: `Deconstruct`; tuples, which introduces the possibility of no type; etc.(?)

I’m not convinced the similarities/differences are fully defined, this may involve changing text in the deconstruction areas as well as in the pattern areas.

standard/patterns.md
BillWagner: addressed in 4th commit.

standard/patterns.md
BillWagner: Addressed in 4th commit.

standard/patterns.md
BillWagner: addressed in 4th commit.

standard/patterns.md
BillWagner: addressed in 4th commit.

standard/patterns.md
```diff
@@ -30,6 +30,8 @@ pattern
     ;
 ```
 
+If the input can be syntactically recognised as both a *constant_pattern* and a *positional_pattern* with a single *subpattern* that lacks an *identifier*, no *property_subpattern*, and no *simple_designation*, then the *constant_pattern* shall be chosen.
```
Nigel-Ecma: You’re helping too much ;-) There is no need to spell out why a *positional_pattern* can be recognised, and we don’t generally do it, so this could be shortened to:

> If the input can be syntactically recognised as both a *constant_pattern* and a *positional_pattern* then the *constant_pattern* shall be chosen.

But spell it out if you like.

standard/patterns.md
```diff
@@ -196,7 +198,9 @@ If *designation* is a *tuple_designation*, the pattern is equivalent to a *posit
...
+A *positional_pattern* checks that the input value is not `null`, extracts a sequence of values from it, and matches each extracted value against a corresponding *subpattern*. The values are extracted in one of three ways: by treating the input as a tuple, by invoking a `Deconstruct` method, or by indexing the input through `System.Runtime.CompilerServices.ITuple`.
+
+> *Note*: The use of `Deconstruct` here is distinct from the source-level deconstruction transformation defined in [§12.7](expressions.md#127-deconstruction); the `Deconstruct` method is invoked directly to obtain the values matched against the *subpattern*s. *end note*
```
Nigel-Ecma: I think the method ends up invoked directly in either scenario, I suggest you just note the its not the same.

```suggestion
> *Note*: The use of `Deconstruct` here is distinct from the source-level deconstruction transformation defined in [§12.7](expressions.md#127-deconstruction). *end note*
```

standard/patterns.md
```diff
@@ -196,7 +198,9 @@ If *designation* is a *tuple_designation*, the pattern is equivalent to a *posit
 
 ### 11.2.5 Positional pattern
 
-A *positional_pattern* checks that the input value is not `null`, invokes an appropriate `Deconstruct` method ([§12.7](expressions.md#127-deconstruction)), and performs further pattern matching on the resulting values.  It also supports a tuple-like pattern syntax (without the type being provided) when the type of the input value is the same as the type containing `Deconstruct`, or if the type of the input value is a tuple type, or if the type of the input value is `object` or `System.ITuple` and the runtime type of the expression implements `System.ITuple`.
+A *positional_pattern* checks that the input value is not `null`, extracts a sequence of values from it, and matches each extracted value against a corresponding *subpattern*. The values are extracted in one of three ways: by treating the input as a tuple, by invoking a `Deconstruct` method, or by indexing the input through `System.Runtime.CompilerServices.ITuple`.
```
Nigel-Ecma, line 205: Just a comment, I wonder whether “sequence of values” may be a source of confusion at some point – it is a sequence in the English sense not in any C# sense of a data structure. However I’ve no idea for something else.

Nigel-Ecma (approved):
A couple of small change suggestions, none crucial so take or leave.

standard/patterns.md
jskeet: It's a sequence in that it's "multiple items with an order that matters" (because the corresponding subpatterns need to be selected). I suggest we just go with it :)

standard/patterns.md
```diff
@@ -211,14 +215,12 @@ subpattern
     ;
 ```
 
-Given a match of an input value to the pattern *type* `(` *subpatterns* `)`, a method is selected by searching in *type* for accessible declarations of `Deconstruct` and selecting one among them using the same rules as for the deconstruction declaration.
-It is an error if a *positional_pattern* omits the type, has a single *subpattern* without an *identifier*, has no *property_subpattern* and has no *simple_designation*. This disambiguates between a *constant_pattern* that is parenthesized and a *positional_pattern*.
-In order to extract the values to match against the patterns in the list,
+Let *n* be the number of *subpattern*s appearing between the parentheses. The matching strategy is selected at compile time by applying the following cases in order; the first case whose conditions are satisfied is used, and the remaining cases are not considered. Once a case is selected, that strategy is committed: any "compile-time error" stated within that case is reported, and matching does not fall through to a subsequent case.
```
jskeet: Do we need the quotes round "compile-time error" here?

standard/patterns.md
```diff
@@ -211,14 +215,12 @@ subpattern
...
-- Otherwise, if *type* is omitted, and the input value is of type `object` or some type that can be converted to `System.ITuple` by an implicit reference conversion, and no *identifier* appears among the subpatterns, then the match uses `System.ITuple`.  
-- Otherwise, the pattern is a compile-time error.
+1. **Tuple form.** If *type* is omitted and the static type of the input value is a tuple type ([§8.3.11](types.md#8311-tuple-types)) or a tuple literal (§12.8.6), then this case applies. It is a compile-time error if *n* is not equal to the arity of that tuple type. At runtime, each tuple element is matched against the corresponding *subpattern*; the match succeeds if all of these succeed. If any *subpattern* has an *identifier*, that *identifier* shall name the tuple element at the corresponding position in the tuple type.
```
jskeet: This makes it sound like a tuple literal is a type. Perhaps "or if the input value is a tuple literal"?

(Not sure whether "that tuple type" needs spelling out in the tuple literal case; I think it's pretty obvious.)

standard/patterns.md
```diff
@@ -211,14 +215,12 @@ subpattern
...
+2. **Deconstruct form.** Otherwise, if either *type* is present, or *type* is omitted and the static type of the input value contains an accessible `Deconstruct` method, then this case applies. Let *D* be *type* if *type* is present; otherwise let *D* be the static type of the input value. A `Deconstruct` method is selected from *D* using the same overload-resolution rules as for a deconstruction declaration, with the additional requirement that its number of `out` parameters is equal to *n*; it is a compile-time error if no such method exists. If *type* is present, it is a compile-time error if the static type of the input value is not pattern-compatible with *type*; at runtime the input value is tested against *type* and, if that test fails, the positional pattern match fails. Otherwise, the input value is converted to *D* and the selected `Deconstruct` method is invoked with fresh variables receiving its `out` parameters. Each received value is matched against the corresponding *subpattern*, and the match succeeds if all of these succeed. If any *subpattern* has an *identifier*, that *identifier* shall name the parameter at the corresponding position of `Deconstruct`.
```
jskeet: Link to deconstruction declaration and maybe pattern-compatible?

jskeet (approved):
Happy overall, just a few nits.

standard/patterns.md
BillWagner: See commit 6

standard/patterns.md
BillWagner: I tried different wordings, and I couldn't come up with anything. I suggest leaving as it is, and consider revisiting when we do list patterns, where there could be some confusion. 

The best I could manage is:


```suggestion
A *positional_pattern* checks that the input value is not `null`, extracts the ordered values from it, and matches each extracted value against a corresponding *subpattern*. The values are extracted in one of three ways: by treating the input as a tuple, by invoking a `Deconstruct` method, or by indexing the input through `System.Runtime.CompilerServices.ITuple`.
```

standard/patterns.md
BillWagner: No. See commit 6

standard/patterns.md
BillWagner: See commit 6

standard/patterns.md
BillWagner: Links added in commit 6. Also, note that the definition is for "pattern compatible", not "pattern-compatible". I made that change globally as well (one other instance in this file).

BillWagner:
@jskeet @Nigel-Ecma I've made all the updates based on your comments. There's one suggestion to consider for the nit. Then, let's merge this one as well.

jskeet (approved):

PR #1687, opened 2026-05-11
Don't add links in codefences

BillWagner:
If a code fenced sample or declaration contains a section reference, don't add a link to the markdown. Keep it a raw xref.

The markdown links won't get replaced, so they'll show up in the finished spec


Nigel-Ecma (approved):
Another reason to dislike MD ;-(

Nigel-Ecma:
It would be nice if §x.y.z was updated in code blocks, or at least in comments in code blocks – I just wanted to xref a sibling clause in such a block and could not without the risk of rot over time :-(

However I just blame MD for its lack of smarts.

jskeet (approved):

PR #1688, opened 2026-05-12
Clarify discard patterns

### LINKED ISSUE #1649: Resolve restriction on *discard_pattern* when there is a defn for `_`
In this case, for compatibility it isn't a discard pattern. I will work on specifying that or add an issue for it.

_Originally posted by @gafter in https://github.com/dotnet/csharpstandard/pull/873#discussion_r2347041797_

BillWagner:
Fixes #1649 

Specify that when the lone token `_` could parse as a `discard_pattern` but `_` also resolves to an accessible declaration in scope (a constant or a type), `_` is interpreted as a reference to that declaration (i.e. a `constant_pattern` or a type reference for the `is` operator) rather than as a `discard_pattern`. This preserves backward compatibility with code that defined `_` (typically a type) prior to discard support.

1. **Edit `standard/patterns.md` §11.2.7 (Discard pattern)** — after the grammar block, before/around the existing compile-time error sentence.
2. **Add a `*Note*`** under §11.2.7 contrasting with §11.2.2's `discard_designation` rule: the `_` in `Type _` is *always* a `discard_designation` (the in-scope `_` is hidden), whereas a standalone `_` pattern *is* hidden by an in-scope `_` for compatibility.
3. **Add a small `*Example*`** under §11.2.7 showing a `switch_expression` with an in-scope type/constant named `_`, demonstrating that the arm `_ => …` is interpreted against that name (e.g. as a type or constant pattern), not as a discard. Use `// compile-time warning expected` comment if applicable; pick a template such as `standalone-console` consistent with the surrounding examples. Keep it minimal.
4. **Cross-reference touch-up in §11.2.2**: append a brief parenthetical or `*Note*` stating "Contrast with §11.2.7, where a standalone `_` pattern is interpreted as a reference to an in-scope declaration named `_` if one exists."


standard/patterns.md
```diff
@@ -369,7 +371,11 @@ discard_pattern
     ;
 ```
 
-It is a compile-time error to use a discard pattern in a *relational_expression* of the form *relational_expression* `is` *pattern* or as the pattern of a *switch_label*.  
+Where the syntactic context permits a *pattern*, if the token `_` would resolve as a *simple_name* ([§12.8.4](expressions.md#1284-simple-names)) to an accessible constant or to a type, then `_` is *not* treated as a *discard_pattern*; instead, it is interpreted as a *constant_pattern* ([§11.2.3](patterns.md#1123-constant-pattern)) when `_` is an accessible constant, or as a reference to the type named `_` (so that the surrounding form is matched per the rules for that referent — for example, as a *constant_pattern* whose constant expression names the type, or as the *type* of a *declaration_pattern*). This rule preserves backward compatibility with code that defined `_` prior to the introduction of the discard pattern.
```
jskeet: A constant pattern for a constant, but presumably the "is-type" operator for a type? I haven't checked how the grammar works for `x is y` - how is that determined to be a `constant_pattern` or an is-type operator?

standard/patterns.md
```diff
@@ -369,7 +371,11 @@ discard_pattern
...
+
+> *Note*: This is analogous to the rule for `var` in §11.2.4, but for `_` an in-scope declaration *demotes* the token to a reference to that declaration rather than producing an error. *end note*
```
jskeet: I'm not sure what "demotes" means in this context.

standard/patterns.md
```diff
@@ -369,7 +371,11 @@ discard_pattern
...
+
+After applying the preceding rule, it is a compile-time error to use a *discard_pattern* in a *relational_expression* of the form *relational_expression* `is` *pattern* or as the pattern of a *switch_label*.
```
jskeet: Just to be clear, it's a compile-time error to use a *discard_pattern* in those situations *when a constant or type `_` is in scope*? What about something that isn't a constant or type?

standard/patterns.md
```diff
@@ -406,6 +412,36 @@ It is a compile-time error to use a discard pattern in a *relational_expression*
...
 > *end example*
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
+> *Example*: The following illustrates the demotion rule. In the second `switch_expression`, an in-scope constant named `_` causes the arm `_ => …` to be interpreted as a *constant_pattern* matching only `0`, rather than as a *discard_pattern* matching every value.
+>
+> ```csharp
+> int n = 1;
+>
+> // No `_` in scope: the `_` arm is a discard pattern matching any value.
```
jskeet: I wonder whether it would be worth putting both of these in methods accepting a parameter `n`, so that there isn't the difference in indentation. (The method could return a string, so we don't need s1 and s2, potentially...)

jskeet (commented):
Thanks for tackling this - sorry to have more questions...

standard/patterns.md
BillWagner: Addressed in 2nd commit.

standard/patterns.md
BillWagner: Addressed in 2nd commit. Change the terminology so it doesn't use terms that haven't been defined.

standard/patterns.md
BillWagner: Refined in the 2nd commit. Explain the rules where a variable identifier is in scope for `_`

standard/patterns.md
BillWagner: updated in 2nd commit.

standard/patterns.md
```diff
@@ -369,7 +371,16 @@ discard_pattern
...
+
+This rule preserves backward compatibility with code that defined `_` as a type or identifier prior to the introduction of the discard pattern. If `_` resolves to anything other than an accessible constant or type (for example, a local variable, parameter, field, or method), the rule does not apply and `_` remains a *discard_pattern*.
+
+> *Note*: This is analogous to the rule for `var` in §11.2.4, except that for `_` an in-scope constant or type causes `_` to be interpreted as a reference to that declaration rather than producing an error. *end note*
+
+If, after applying the preceding rule, the token `_` is still a *discard_pattern*, it is a compile-time error for that *discard_pattern* to appear in a *relational_expression* of the form *relational_expression* `is` *pattern* or as the pattern of a *switch_label*.
```
Nigel-Ecma: This isn’t quite correct, while `... is _` is invalid the use of discards within the pattern is allowed, e.g. `... is (int, _)` tests for a pair with a first element type of `int`, and even `... is (_, _)` is valid to test for some pair.

Nigel-Ecma (changes requested):
I *think* these rules are mostly correct, spotted just one change being needed

standard/patterns.md
BillWagner: @Nigel-Ecma Check the most recent commit. If I have this correct, `... is (int, _)` or `... is (_, _)` is a *postional_pattern*, with one (or more) of its *subpattern*s as `discard_patterns`. Another interesting case I found is `... is (_)` which is also valid. In this case, it's a `parenthesized_pattern* that contains a *discard_pattern*. Also valid.

Nigel-Ecma (approved):
The change covers it.

standard/patterns.md
Nigel-Ecma: @BillWagner  – Exactly, but the distinction between `_` and `(_)` is a bit “unfortunate”, c'est la vie.

standard/patterns.md
```diff
@@ -369,7 +371,16 @@ discard_pattern
...
+
+This rule preserves backward compatibility with code that defined `_` as a type or identifier prior to the introduction of the discard pattern. If `_` resolves to anything other than an accessible constant or type (for example, a local variable, parameter, field, or method), the rule does not apply and `_` remains a *discard_pattern*.
+
+> *Note*: This is analogous to the rule for `var` in §11.2.4, except that for `_` an in-scope constant or type causes `_` to be interpreted as a reference to that declaration rather than producing an error. *end note*
+
+If, after applying the preceding rule, the token `_` is still a *discard_pattern*, it is a compile-time error for that *discard_pattern* to appear as the entire *pattern* of a *relational_expression* of the form *relational_expression* `is` *pattern*, or as the entire *pattern* of a *switch_label*. A *discard_pattern* may, however, appear as a *subpattern* of an enclosing pattern (for example, as a *subpattern* of a *positional_pattern* or *property_pattern*).
```
jskeet, line 385: Just to clarify, this is saying that you can't have:

```csharp
switch (x)
{
    case _:
        Console.WriteLine("Yes");
        break;
}
```

... because you'd just use default for that. Am I right, or have I missed something?

(And likewise the discard pattern is pointless with `is` because it's not actually testing anything?)

jskeet (approved):
Fine assuming I've understood it correctly...

standard/patterns.md
Nigel-Ecma: @jskeet – You’re right (and likewise)

PR #1689, opened 2026-05-13
Complete #1560 by adding deconstructing foreach

### LINKED ISSUE #1602: EBNF for the `foreach_statement` is incorrect. (draft-v8)
**Describe the bug**
I am a maintainer for https://github.com/antlr/grammars-v4 and I am in the process of bringing the grammar in grammars-v4 up to date for C#. I am also scraping and validating the grammars over here and in the published ECMA 334 .pdf.

The rule for the [foreach_statement](https://github.com/dotnet/csharpstandard/blob/0346abf0132ec71716229380f4be340b7931d61c/standard/statements.md?plain=1#L1109) is incorrect, starting in version 7.

https://github.com/dotnet/csharpstandard/blob/0346abf0132ec71716229380f4be340b7931d61c/standard/statements.md?plain=1#L1109-L1112

Unless it really is not part of the spec, the rule should have an alternative for `deconstruction_expression`.

```
// Source: §13.9.5.1 General
foreach_statement
    : 'await'? 'foreach' '(' ref_kind? local_variable_type identifier
      'in' expression ')' embedded_statement
    | 'await'? 'foreach' '(' deconstruction_expression
      'in' expression ')' embedded_statement
    ;
```

Without the fix, the spec EBNF does not parse several input in the Roslyn v7 compiler itself. But, below is a simpler example that does not parse. (There are other places the Roslyn compiler diverges, so I don't know what is intended.) 

**Example**
```csharp

  class Program
  {
      static void Main()
      {
          var map = new Dictionary<string, int> { ["a"] = 1 };
          foreach (var (key, value) in map) { }
      }
  }
```
Attached is a zip file containing a full program with the example. It builds and runs with compiler version 7.

[v7test.zip](https://github.com/user-attachments/files/25901719/v7test.zip)

**Additional context**
The current means of generating and testing the Antlr4 grammar is archaic. The "BuildGrammar" tool should generate a C#-target port, not a Java one, and definitely not using the ancient Antlr 4.9.2. There are better tools nowadays for analyzing the EBNF.

Nigel-Ecma:
Closes #1602

- Complete #1560 by covering *foreach_statement* using a *deconstructor*
- Add sample “Deconstructing foreach”
- Change subclause title from “await foreach” to “Asynchronous foreach” for consistency with siblings
- Also cleans up some whitespace


standard/statements.md
```diff
@@ -1356,19 +1365,21 @@ await foreach (T item in enumerable) «embedded_statement»
 is semantically equivalent to:
 
 ```csharp
-var enumerator = enumerable.GetAsyncEnumerator();
-try
 {
-    while (await enumerator.MoveNextAsync())
+    var enumerator = enumerable.GetAsyncEnumerator();
```
BillWagner, line 1369: It looks like the only change here is indenting. Correct?

standard/statements.md
```diff
@@ -1426,6 +1437,77 @@ The local variable `d` is not visible to or accessible to any user code. In par
 
 > *Note*: An `await foreach` is not required to dispose of `e` synchronously if an asynchronous dispose mechanism is not available. *end note*
 
+#### §deconstructing-foreach Deconstructing foreach
+
+A deconstructing foreach replaces the declaration and initialisation of a single iteration variable per iteration, in the synchronous and asynchronous foreach statements, with a collection of zero or more iteration variables declared per iteration within the *deconstructor*([§12.23.3](expressions.md#12233-deconstructing-assignment)) of a *deconstructing_assignment*.
```
BillWagner, line 1442: Similar to the changes in 1560, this should be "2 or more":


```suggestion
A deconstructing foreach replaces the declaration and initialisation of a single iteration variable per iteration, in the synchronous and asynchronous foreach statements, with a collection of two or more iteration variables declared per iteration within the *deconstructor*([§12.23.3](expressions.md#12233-deconstructing-assignment)) of a *deconstructing_assignment*.
```

standard/statements.md
```diff
@@ -1426,6 +1437,77 @@ The local variable `d` is not visible to or accessible to any user code. In par
...
+        }
+    }
+    finally
+    {
+        ... // dispose of enumerator as for synchronous foreach
+    }
+}
+```
+
+this follows the behavior of synchronous foreach ([§13.9.5.2](statements.md#13952-synchronous-foreach)), differing by replacing the delaration and initialisation of a single iteration variable with a *deconstructing_assignment* which declares and assigns zero or more initialisation variables:
```
BillWagner: Again, 2 or more, and we should add text about how discards are handled:


```suggestion
this follows the behavior of synchronous foreach ([§13.9.5.2](statements.md#13952-synchronous-foreach)), differing by replacing the delaration and initialisation of a single iteration variable with a *deconstructing_assignment* which declares and assigns two or more initialisation variables:
```

standard/statements.md
```diff
@@ -1426,6 +1437,77 @@ The local variable `d` is not visible to or accessible to any user code. In par
 
 > *Note*: An `await foreach` is not required to dispose of `e` synchronously if an asynchronous dispose mechanism is not available. *end note*
 
+#### §deconstructing-foreach Deconstructing foreach
```
BillWagner, line 1440: I ran a couple tests, and we need to specify discards in a deconstructing foreach. I think the language should be similar to what was added in #1639.

What's interesting here is that a synchronous and asynchronous `foreach` don't allow the iteration variable to be a discard. However, in a deconstructing foreach, any number of the iteration variables may be a discard (from 0 to all).

BillWagner (commented):
Great start. It looks like the specter of discards in a deconstructing assignment has made its appearance again.

standard/statements.md
Nigel-Ecma: The expansions for the others where all wrapped in an enclosing block and this one was missed/different. So yes, the whole thing was indented and unindented { & } added to wrap it – not that the diff makes that easy to spot!

standard/statements.md
Nigel-Ecma: @BillWagner – see reply on line 1442

standard/statements.md
Nigel-Ecma: @BillWagner – I had completely forgotten that #1639 introduced targets and counted them rather than assigned named variables, so I just wrote this doing the same as that is what is important – discards are discarded afterall :-)

Your proposed change here however makes it “two or more *iteration variables*” so more wordsmithing would be required to go the “target” route. I think targets would not fit well here, but then I went with the consensus on §12.23 so that is not too surprising.  [Note that #1639 actually contradicts itself, it starts by stating two or more *targets* are assigned, then later that discard targets are not assigned (“but no value is stored”).]

That discards may be present is covered just below in line 1444. I did consider whether to list the allowed, which would have surfaced discards, or disallowed alternatives, and decided the key issue was that existing variables could not be assigned to so highlighted that by just listed the disallowed *variable_reference*.

Suggestion: We leave this PR as is on this topic and then you & @jskeet raise an issue to align the text in §12.23 to either use targets (which are not all assigned) in both, neither, or go with variety is the spice of life :-) Mark it pre-C#9 so its not a blocker – what exists here & in §12.23 are not wrong per se.

standard/statements.md
Nigel-Ecma: See comment on line 1442.

standard/statements.md
Nigel-Ecma: @BillWagner PS: On your “What’s interesting” – are you using “interesting” as (apocryphally?) used in a Chinese curse? ;-) BTW a certain compiler rejects `... is _` but accepts `... is (_)` – can that be justified? Maybe. Is it confusing? Probably. We “live in interesting times”.

BillWagner: @Nigel-Ecma This comment helped me understand the final changes needed for #1688.  `... is _` is a *discard_pattern*, but `... is (_)` is a *parenthesised_pattern* that encloses a *discard_pattern*. Because the *discard_pattern* is a subpattern, it's allowed.

standard/statements.md
jskeet: That suggestion is fine by me.

standard/statements.md
```diff
@@ -1426,6 +1437,77 @@ The local variable `d` is not visible to or accessible to any user code. In par
...
+
+A deconstructing foreach replaces the declaration and initialisation of a single iteration variable per iteration, in the synchronous and asynchronous foreach statements, with a collection of zero or more iteration variables declared per iteration within the *deconstructor*([§12.23.3](expressions.md#12233-deconstructing-assignment)) of a *deconstructing_assignment*.
+
+All variables assigned to by the *deconstructor* must be declared within the *deconstructor*, it is a compile time error for any *deconstructor_element* to be a *variable_reference*.
+
+For a `foreach` statement of the form:
```
```suggestion
A foreach statement of the form:
```

standard/statements.md
```suggestion
This follows the behavior of synchronous foreach ([§13.9.5.2](statements.md#13952-synchronous-foreach)), differing by replacing the delaration and initialisation of a single iteration variable with a *deconstructing_assignment* which declares and assigns zero or more initialisation variables:
```

standard/statements.md
```diff
@@ -1426,6 +1437,77 @@ The local variable `d` is not visible to or accessible to any user code. In par
...
+        }
+    }
+    finally
+    {
+        // dispose of enumerator as for asynchronous foreach
+    }
+}
+```
+
+this follows the behavior of asynchronous foreach ([§13.9.5.3](statements.md#13953-asynchronous-foreach)), differing by replacing the delaration and initialisation of a single iteration variable with a *deconstructing_assignment* which declares and assigns zero or more initialisation variables:
```
```suggestion
This follows the behavior of asynchronous foreach ([§13.9.5.3](statements.md#13953-asynchronous-foreach)), differing by replacing the delaration and initialisation of a single iteration variable with a *deconstructing_assignment* which declares and assigns zero or more initialisation variables:
```

jskeet (approved):

PR #1693, opened 2026-05-14
Clarify rules for redundant arms

### LINKED ISSUE #1656: Fix description around it being an error if a *switch_expression_arm* is redundant
Quoted comment is start of thread, whole thread relevant.

---

Is it not that some previous arm will always match but that the sum of all the previous arms match the whole domain, i.e. an arm is *subsumed* (§11.3) as with the switch statement (§13.8.3)?

Or is this not subsumption as it includes the guards?

Or does pattern subsumption (§11.3) need to be expanded to include guards?

Even if it is not subsumption I think it needs to be clear that it is the sum of *all* preceding arms.

_Originally posted by @Nigel-Ecma in https://github.com/dotnet/csharpstandard/pull/873#discussion_r2361637033_

BillWagner:
Fixes #1656

- Make sure that redundancy is defined in terms of all preceding switch arms, not an arm.
- Define "unguarded" in statements.
- Add example to explain the rule that only unguarded switch arms apply to subsumption detection.

### DIFF
--- a/standard/patterns.md
+++ b/standard/patterns.md
@@ -325,9 +325,9 @@ The *property_pattern* may be used to pattern-match with anonymous types.
 > ```
 >
 > *end example*
-> <!-- markdownlint-disable MD028 -->
->
-> <!-- markdownlint-enable MD028 -->
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
 > *Example*: A run-time type check and a variable declaration can be added to a property pattern, as follow:
 >
 > <!-- Example: {template:"standalone-console", name:"PropertyPattern3", inferOutput:true} -->
@@ -422,7 +422,9 @@ A set of patterns `Q` *subsumes* a pattern `P` if any of the following condition
 - `P` is a declaration pattern with type `T` and the set of patterns `Q` is *exhaustive* for the type `T` ([§11.4](patterns.md#114-pattern-exhaustiveness)).
 
 > *Note*: Only *unguarded* cases or arms ([§13.8.3](statements.md#1383-the-switch-statement)) contribute to subsumption; a *case_guard* whose expression is not a constant expression with the value `true` cannot in general be evaluated at compile time, so the corresponding case or arm is conservatively assumed not to match every input value to which its pattern is applicable. *end note*
+<!-- markdownlint-disable MD028 -->
 
+<!-- markdownlint-enable MD028 -->
 > *Example*: In the following switch expression, no arm is subsumed even though arms 1, 2, and 3 share the same pattern:
 >
 > <!-- Example: {template:"code-in-main", name:"SwitchExprUnguardedSubsumption"} -->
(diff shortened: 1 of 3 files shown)


jskeet (approved):
Looks good! I may be missing something, of course - but I'd personally be fine to ship it.

standard/patterns.md
```diff
@@ -409,8 +409,8 @@ It is a compile-time error to use a discard pattern in a *relational_expression*
 
 ## 11.3 Pattern subsumption
 
-In a switch statement, it is an error if a case’s pattern is *subsumed* by the preceding set of unguarded cases ([§13.8.3](statements.md#1383-the-switch-statement)).
-Informally, this means that any input value would have been matched by one of the previous cases.
+In a switch statement ([§13.8.3](statements.md#1383-the-switch-statement)), it is an error if a case’s pattern is *subsumed* by the preceding set of *unguarded* ([§13.8.3](statements.md#1383-the-switch-statement)) cases. In a switch expression ([§12.11](expressions.md#1211-switch-expression)), it is an error if a *switch_expression_arm*’s pattern is *subsumed* by the preceding set of *unguarded* *switch_expression_arm*s’ patterns.
+Informally, this means that any input value would have been matched by one of the previous cases or arms.
```
Nigel-Ecma: As it is informal should it not be a Note, something like:

```suggestion
> *Note*: This means that any input value would have been matched by one of the previous cases or arms. *end note*
```

standard/patterns.md
```diff
@@ -421,6 +421,27 @@ A set of patterns `Q` *subsumes* a pattern `P` if any of the following condition
 - `P` is a var pattern and the set of patterns `Q` is *exhaustive* ([§11.4](patterns.md#114-pattern-exhaustiveness)) for the type of the pattern input value ([§11.1](patterns.md#111-general)), and either the pattern input value is not of a nullable type or some pattern in `Q` would match `null`.
 - `P` is a declaration pattern with type `T` and the set of patterns `Q` is *exhaustive* for the type `T` ([§11.4](patterns.md#114-pattern-exhaustiveness)).
 
+> *Note*: Only *unguarded* cases or arms ([§13.8.3](statements.md#1383-the-switch-statement)) contribute to subsumption; a *case_guard* whose expression is not a constant expression with the value `true` cannot in general be evaluated at compile time, so the corresponding case or arm is conservatively assumed not to match every input value to which its pattern is applicable. *end note*
```
Nigel-Ecma: I’m a little wary about explaining the restriction. Consider the example below: 

- `b` is a simple variable and so evaluating it has no side-effect, the uses on lines 436 & 437 will return the same value for `b`;
- now consider the situation of changing the guard on line 437 to `!b`, the arm becomes unreachable (as line 436 always takes precedence) and yet is not “subsumed”…

An implementation could catch that, indeed there is already a requirement that if the guard is a constant expression evaluating to `true` then the arm must be treated as unguarded and included in subsumption analysis. The step from there to the guard having no-side effect and repeated on more than one arm is a fairly small step.

Something to feed into the discussion on #1691 maybe?

Nigel-Ecma (commented):
A suggestion that an “Informally…” statement become a *Note*, a comment on another *Note*, otherwise looks good.

standard/patterns.md
```diff
@@ -421,6 +421,27 @@ A set of patterns `Q` *subsumes* a pattern `P` if any of the following condition
...
+<!-- markdownlint-disable MD028 -->
+
+<!-- markdownlint-enable MD028 -->
```
BillWagner: Removing per discussion at the meeting:


```suggestion
```

PR #1695, opened 2026-05-15
Define pattern compatible

### LINKED ISSUE #1660: Fix “considered incompatible” but ”are incompatible” et al
There is no “considered incompatible” but ”are incompatible”.

Further “Certain combinations” seems wrong to me, I think the order is: define pattern compatible, define applicable to (though why are two terms required?), state that if the type of the declaration pattern is not compatible with/applicable to the static type of the input value then a compile time error shall be thrown.

Apologies, no wordsmithing suggestion right now, maybe I’ll come up with one when I resume reviewing later…

_Originally posted by @Nigel-Ecma in https://github.com/dotnet/csharpstandard/pull/873#discussion_r2364774347_

BillWagner:
Fixes #1660

This follows the general strategy in the issue and the comment thread. In addition, note that we used "pattern compatible" and "pattern-compatible" in this clause. Standardize on "pattern compatible".


Nigel-Ecma (approved):
I think that captures it all.

standard/patterns.md
```diff
@@ -90,7 +90,7 @@ The runtime type of the value is tested against the *type* in the pattern using
...
+A value of static type `E` is said to be ***pattern compatible*** with the type `T` if there exists an identity conversion, an implicit or explicit reference conversion, a boxing conversion, an unboxing conversion, or an implicit or explicit nullable value type conversion from `E` to `T`, or if either `E` or `T` is an open type ([§8.4.3](types.md#843-open-and-closed-types)). A declaration pattern naming a type `T` is *applicable to* ([§11.2.1](patterns.md#1121-general)) every type `E` for which `E` is pattern compatible with `T`. It is a compile-time error if a declaration pattern naming a type `T` is used to match a pattern input value ([§11.1](patterns.md#111-general)) whose static type `E` is not pattern compatible with `T`.
```
jskeet: Very nitty, but the first sentence says that pattern compatibility applies to a *value* - but later we say "for which `E` is pattern compatible with `T`".

Perhaps pattern compatibility should just be specified in terms of types rather than values? (I think we can just remove "value of" from the first sentence.)

standard/patterns.md
```diff
@@ -310,7 +310,7 @@ It is a compile-time error if the *type* is a nullable value type ([§8.3.12](ty
 >
 > *end note*
 
-Given a match of an expression *e* to the pattern *type* `{` *subpatterns* `}`, it is a compile-time error if the expression *e* is not pattern-compatible with the type *T* designated by *type*. If the type is absent, the type is assumed to be the static type of *e*. Each of the identifiers appearing on the left-hand-side of its *subpatterns* shall designate an accessible readable property or field of *T*. If the *simple_designation* of the *property_pattern* is present, it declares a pattern variable of type *T*.  
+Given a match of an expression *e* to the pattern *type* `{` *subpatterns* `}`, it is a compile-time error if the static type of the expression *e* is not pattern compatible ([§11.2.2](patterns.md#1122-declaration-pattern)) with the type *T* designated by *type*. If the type is absent, the type is assumed to be the static type of *e*. Each of the identifiers appearing on the left-hand-side of its *subpatterns* shall designate an accessible readable property or field of *T*. If the *simple_designation* of the *property_pattern* is present, it declares a pattern variable of type *T*.
```
jskeet: If the type is absent, presumably it isn't a match to the pattern *type* `{` *subpatterns* `}`? Do we need to say "the pattern *type* `{` *subpatterns* `}` or `{` *subpatterns* `}`" perhaps?

jskeet (commented):
Two very nit-picky comments, but fine with the general approach.

standard/patterns.md
BillWagner: Good catch. I made a change, but I think it should be "An expression of static type `E`..." rather than a type. The input value is an expression, and that expression has a type.

standard/patterns.md
BillWagner: I made a small editorial change here.

standard/patterns.md
jskeet: But we still say "for which `E` is pattern compatible with `T`" (where E is a type).

To put it in code terms, we seem to declare `bool IsPatternCompatible(Pattern p, Expression e)` but then call it as if it's declared as `bool IsPatternCompatible(Pattern p, Type t)` :)

The same happens later: "it is a compile-time error if the static type of the expression *e* is not pattern compatible ..."

So it looks like we want to use it as if it's a "pattern + type" test, not a "pattern + expression" test. Is there a reason we want to declare it as if it's "pattern + expression"? I don't *think* it can be for the normal reason of handling expressions without types, because we immedaitely talk about "of static type `E`".

Will approve anyway and we could merge and file a new nitpick issue that we may never get round to if you'd prefer... I'm certainly not going to claim this is the most pressing concern ;)

jskeet (approved):

standard/patterns.md
```diff
@@ -90,7 +90,7 @@ The runtime type of the value is tested against the *type* in the pattern using
...
+An expression of static type `E` is said to be ***pattern compatible*** with the type `T` if there exists an identity conversion, an implicit or explicit reference conversion, a boxing conversion, an unboxing conversion, or an implicit or explicit nullable value type conversion from `E` to `T`, or if either `E` or `T` is an open type ([§8.4.3](types.md#843-open-and-closed-types)). A declaration pattern naming a type `T` is *applicable to* ([§11.2.1](patterns.md#1121-general)) every type `E` for which `E` is pattern compatible with `T`. It is a compile-time error if a declaration pattern naming a type `T` is used to match a pattern input value ([§11.1](patterns.md#111-general)) whose static type `E` is not pattern compatible with `T`.
```
```suggestion
A type `E` is said to be ***pattern compatible*** with the type `T` if there exists an identity conversion, an implicit or explicit reference conversion, a boxing conversion, an unboxing conversion, or an implicit or explicit nullable value type conversion from `E` to `T`, or if either `E` or `T` is an open type ([§8.4.3](types.md#843-open-and-closed-types)). A declaration pattern naming a type `T` is *applicable to* ([§11.2.1](patterns.md#1121-general)) every type `E` for which `E` is pattern compatible with `T`. It is a compile-time error if a declaration pattern naming a type `T` is used to match a pattern input value ([§11.1](patterns.md#111-general)) whose static type `E` is not pattern compatible with `T`.
```

jskeet (approved):

standard/patterns.md
jskeet: Resolved now.

PR #1721, opened 2026-06-30
Clarify rules on safety and sizeof

BillWagner:
This is a small focused change, but fixes what I (at least) misinterpreted at least twice.

In version 7.3 we correctly updated the rules for in 8.8 for Unmanaged types to include struct types where the fields are all unmanaged types. This includes generic types where the type parameter has the `unmanaged` constraint.

However, we didn't restate the restrictions for `sizeof` in 12.8.19 to prevent extending the allowed types in safe context to include those types.

I updated the language in 12.8.19 to make those restrictions clear.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -3201,7 +3201,7 @@ The `typeof` operator can be used on a type parameter. It is a compile time erro
 
 ### 12.8.19 The sizeof operator
 
-The `sizeof` operator returns the number of 8-bit bytes occupied by a variable of a given type. The type specified as an operand to sizeof shall be an *unmanaged_type* ([§8.8](types.md#88-unmanaged-types)).
+The `sizeof` operator returns the number of 8-bit bytes occupied by a variable of a given type.
 
 ```ANTLR
 sizeof_expression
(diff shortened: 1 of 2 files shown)


Nigel-Ecma (approved):
LGTM

§24.6.9 is as undefined as ever, but that is by design.

standard/expressions.md
```diff
@@ -3227,7 +3227,7 @@ For certain predefined types the `sizeof` operator yields a constant `int` value
 |`sizeof(bool)`     | 1          |
 |`sizeof(decimal)`  | 16         |
 
-For an enum type `T`, the result of the expression `sizeof(T)` is a constant value equal to the size of its underlying type, as given above. For all other operand types, the `sizeof` operator is specified in [§24.6.9](unsafe-code.md#2469-the-sizeof-operator).
+For an enum type `T`, the result of the expression `sizeof(T)` is a constant value equal to the size of its underlying type, as given above. For all other operand types, the `sizeof` operator is specified in [§24.6.9](unsafe-code.md#2469-the-sizeof-operator), and is allowed only in an unsafe context.
```
jskeet: Might it be worth moving the second sentence to a separate paragraph? In its current location it feels somewhat "anchored" to enums, making it easy to ignore that "all other" excludes the types in the table above.

jskeet:
Adding as a comment here so we can discuss it... 24.6.9 starts with:

> For certain predefined types ([§12.8.19](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#12819-the-sizeof-operator)), the sizeof operator yields a constant int value. For all other types, the result of the sizeof operator is implementation-defined and is classified as a value, not a constant.

It's not clear from that sentence that sizeof yields a constant int value for enums. Should we update that as well? (I'm happy to create a PR for that.)


Separately, 24.6.9 doesn't have a restriction on sizeof only being used for unmanaged types, so there's *nothing* to stop `sizeof(MyClassType)` being valid in an unsafe context - it's implementation-specific. Is that the intention?

jskeet:
Bill will remove "predefined" from the first sentence in 24.6.9 and add the line break as suggested, then we can merge.

PR #1727, opened 2026-07-13
Avoid double-quoting for a note in a list

### LINKED ISSUE #1628: Fix Markdown for "andThen" note
In expressions.md, search for `*Note*: `andThen` is the equivalent of C & C++’s comma operator. *end note*`

GitHub renders the Markdown in a way we might be able to improve on - and we should check the Word version too.

(We're not looking for perfection here - if it's okay in Word already and we can't make it *much* nicer in GitHub, let's not worry too much. Worth trying though.)

jskeet:
This means that in the Word document, the note isn't indented further than the text preceding it - but it looks entirely reasonable to me.

Fixes #1628

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -1279,10 +1279,8 @@ An expression `E`, with a type `S` other than `dynamic`, can be ***deconstructed
   >```
 
   Where `andThen` is a pseudo C# operation which performs its left-hand operand and then returns its right-operand as the result.
-<!-- markdownlint-disable MD028 -->
 
-<!-- markdownlint-enable MD028 -->
-  >> *Note*: `andThen` is equivalent to C & C++’s comma operator. *end note*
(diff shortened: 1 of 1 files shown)


Nigel-Ecma (approved):
This fixes the issue in GitHub – as @KalleOlaviNiemitalo [pointed out](https://github.com/dotnet/csharpstandard/issues/1628#issuecomment-4215964340) it was my non-indented (and unnecessary as it turns out) mdlint comments which threw off GitHub and required the double-indent.

However according to @jskeet this has uncovered an issue in the Word conversion – it apparently does not indent correctly within list items. I suggest an issue is raised for that.

jskeet:
Will merge this now, and file an issue in the morning. Although to be honest I'm not sure it's really a significant issue.

jskeet:
Filed #1731 for this.

PR #1728, opened 2026-07-13
Provide additional commentary on property pattern

### LINKED ISSUE #1653: Expand example for Property pattern
It may be worth noting the difference between:

```csharp
if (s is {} x2)
```

and

```csharp
if (s is var x3)
```

... the latter only matches on non-null values, whereas `var` would match on null (and the type of `x3` is `string?` instead of `string` in a nullable context). I only mention this as a reader might wonder why `{} x2` is useful when `var` exists.

_Originally posted by @jskeet in https://github.com/dotnet/csharpstandard/pull/873#discussion_r1959591408_

jskeet:
`if (s is var x2)` and `if (s is {} x2)` look similar; this change explicitly notes the difference.

Definitely open to suggestions of changing where this occurs or how it's expressed. This felt like a minimalist change, which I *suspect* is all that's required.

Fixes #1653

### DIFF
--- a/standard/patterns.md
+++ b/standard/patterns.md
@@ -320,6 +320,7 @@ It is a compile-time error if the *type* is a nullable value type ([§8.3.12](ty
 > if (s is {}) ...
 > ```
 >
+> The example declaring `x2` is similar to `if (s is var x2)` in terms of inferring the variable type, but the property pattern guarantees that `x2` is non-null.
 > *end note*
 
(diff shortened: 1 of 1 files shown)


BillWagner (approved):
This LGTM. I agree that it's a minimal change that fixes the issue.

jskeet:
Merging with just one review as it's so trivial.

PR #1729, opened 2026-07-13
Clarify the relationship between declaration patterns and is-type

### LINKED ISSUE #1661: Review defn of semantics of Declaration pattern
Quote is from a thread, all of thread relevant – in particular the last comment might be explaining why the differences exist but it appears the text has not be revised to match this; does it need to be?

---

This states the match follows the rules of `is` (§12.13.12.1), then goes on to “If the test succeeds”.

That test works for nullable value types, but here a following sentence makes that a compile time error.

That test returns false for a null value, this is re-iterated in the last sentence here.

That test never considers user-defined conversions, that is not re-iterated here.

The description of that test carries a *Note* (informative) that as the test is performed at runtime there are no open types, line 92 below says open types are allowed…

This section and §12.13.12.1 need to be aligned, the tests for the two are currently documented to be different but also stated to be the same.

_Originally posted by @Nigel-Ecma in https://github.com/dotnet/csharpstandard/pull/873#discussion_r2366385098_

jskeet:
Fixes #1661

### DIFF
--- a/standard/patterns.md
+++ b/standard/patterns.md
@@ -94,9 +94,11 @@ When recognising a *simple_designation* if both the *discard_designation* and *s
 
 > *Note*: ANTLR makes the specified choice automatically due to the ordering of the alternatives of *simple_designation*. *end note*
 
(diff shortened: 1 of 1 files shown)


jskeet:
It's entirely possible that there are subtleties that I've missed here... I'm really *hoping* there aren't cases where both `e is T` and `e is T _` are valid but they're handled differently...

BillWagner (commented):
(Commenting, because I'd like Nigel to look as well).

I'd be ready to merge this. This looks great.

Nigel-Ecma (approved):
Looks OK

PR #1733, opened 2026-07-16
Improve example of pattern-based deconstruction

### LINKED ISSUE #1667: Poor example?
This seems a poor example as it is longhand for:

```csharp
var (sum, count) = SumAndCount(numbers);
Console.WriteLine($"Sum of [{string.Join(" ", numbers)}] is {sum}");
```

and introduces an `if` suggesting that the match might fail.

_Originally posted by @Nigel-Ecma in https://github.com/dotnet/csharpstandard/pull/873#discussion_r2370306726_

jskeet:
Fixes #1667


BillWagner (approved):
This LGTM.

PR #1734, opened 2026-07-17
Clarify that a switch expression may have no type

### LINKED ISSUE #1651: Determine whether the specification describes target-typed switch expressions
This specification does not describe target-typed switch expressions. See
https://github.com/dotnet/csharplang/issues/2389
https://github.com/dotnet/csharplang/blob/main/proposals/csharp-8.0/patterns.md

The type of the switch_expression is the best common type ([§12.6.3.15](https://github.com/dotnet/csharpstandard/blob/draft-v8/standard/expressions.md#126315-finding-the-best-common-type-of-a-set-of-expressions)) of the expressions appearing to the right of the => tokens of the switch_expression_arms if such a type exists and the expression in every arm of the switch expression can be implicitly converted to that type. In addition, we add a new switch expression conversion, which is a predefined implicit conversion from a switch expression to every type T for which there exists an implicit conversion from each arm's expression to T.

We have to be careful defining when the common type must exist (when there is no target type) and when it doesn't (the switch is target-typed). The spec must say that when there is no target type, it is an error if there is no common type.

_Originally posted by @gafter in https://github.com/dotnet/csharpstandard/pull/873#discussion_r2241208762_

jskeet:
Fixes #1651.

I don't think we need to talk about "if a switch expression is not subject to..." - we just say that the expression has no type, but can be converted to some types, and let the rest of the language rules handle it.

We *may* want to add examples, e.g.:

- A switch expression with a type being used to initialize a `var` variable
- A switch expression without a type being used to initialize an explicitly-typed variable
- A switch expression without a type failing to be used to initialize a `var` variable

I don't know whether those examples are really needed though.

### DIFF
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -3880,6 +3880,7 @@ switch_expression_arm_expression
 ```
 
 The type of a *switch_expression* is the best common type [§12.6.3.16](expressions.md#126316-finding-the-best-common-type-of-a-set-of-expressions)) of the *switch_expression_arm_expression*s of the *switch_expression_arm*s, if such a type exists, and if each *switch_expression_arm_expression* can be implicitly converted to that type. Otherwise, the *switch_expression* has no type, but may still be subject to *switch expression conversions* ([§10.2.18](conversions.md#10218-switch-expression-conversion)).
+<-- The best common type is the "natural type" when we add natural types for lambda expressions and other typeless expressions -->
 
 It is an error if the pattern of any *switch_expression_arm* is *subsumed* by ([§11.1](patterns.md#111-general)) the set of patterns of earlier *unguarded* ([§13.8.3](statements.md#1383-the-switch-statement)) *switch_expression_arm*s of the switch expression.
 
--- a/standard/expressions.md
+++ b/standard/expressions.md
@@ -3879,10 +3879,7 @@ switch_expression_arm_expression
     ;
 ```
 
-There is a *switch expression conversion* ([§10.2.18](conversions.md#10218-switch-expression-conversion)) from a switch expression to a type `T`
-if there is an implicit conversion from every *switch_expression_arm_expression* of each of the switch expression’s *switch_expression_arm*s to `T`.
-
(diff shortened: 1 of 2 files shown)


Nigel-Ecma (approved):
I think this approach fits into the existing approach to ”no type”.

I would probably myself drop lines 3882-3 as they just repeat what is said in the xref and instead add the xref to line 3885 – more succinct, but that is subjective!

jskeet:
> I would probably myself drop lines 3882-3 as they just repeat what is said in the xref and instead add the xref to line 3885 – more succinct, but that is subjective!

Makes sense - have done that in another commit (as well as reverting an accidental incomplete change for a different issue).

standard/expressions.md
```diff
@@ -3879,13 +3879,7 @@ switch_expression_arm_expression
...
-if there is an implicit conversion from every *switch_expression_arm_expression* of each of the switch expression’s *switch_expression_arm*s to `T`.
-
-If a switch expression is not subject to a *switch expression conversion*, then
-
-- The type of the *switch_expression* is the best common type [§12.6.3.16](expressions.md#126316-finding-the-best-common-type-of-a-set-of-expressions)) of the *switch_expression_arm_expression*s of the *switch_expression_arm*s, if such a type exists, and each *switch_expression_arm_expression* can be implicitly converted to that type.
-- It is an error if no such type exists.
+The type of a *switch_expression* is the best common type [§12.6.3.16](expressions.md#126316-finding-the-best-common-type-of-a-set-of-expressions)) of the *switch_expression_arm_expression*s of the *switch_expression_arm*s, if such a type exists, and if each *switch_expression_arm_expression* can be implicitly converted to that type. Otherwise, the *switch_expression* has no type, but may still be subject to *switch expression conversions* ([§10.2.18](conversions.md#10218-switch-expression-conversion)).
```
BillWagner, line 3882: Let's add a comment that when we add more "natural type" conversions, we should define that term:


```suggestion
The type of a *switch_expression* is the best common type [§12.6.3.16](expressions.md#126316-finding-the-best-common-type-of-a-set-of-expressions)) of the *switch_expression_arm_expression*s of the *switch_expression_arm*s, if such a type exists, and if each *switch_expression_arm_expression* can be implicitly converted to that type. Otherwise, the *switch_expression* has no type, but may still be subject to *switch expression conversions* ([§10.2.18](conversions.md#10218-switch-expression-conversion)).
<-- The best common type is the "natural type" when we add natural types for lambda expressions and other typeless expressions -->
```

BillWagner (approved):
This LGTM as well.

PR #1754, opened 2026-07-24
Update agent prompts after this month's propagation cycle

BillWagner:
Updates the post-meeting agent prompts based on the propagation-cycle failures and review feedback.

The revised workflow now:
- discovers actual draft/alpha branch names and dynamically reports the feature-PR count;
- defines the A1 → B → A2 chronology and responsibilities;
- pins feature PR numbers, head SHAs, patches, and cutoff times, with movement checks;
- requires raw symbolic-reference audits and semantic three-way comparisons;
- applies surgical feature deltas without replaying synchronization merges or losing later-version language; and
- treats finalized alphas as risk reduction, with downstream verification still required.

The existing workflow change also renames the automated renumbering branch so it does not match the protected-branch pattern.

### DIFF
--- a/.github/prompts/post-meeting-propagate.prompt.md
+++ b/.github/prompts/post-meeting-propagate.prompt.md
@@ -63,6 +63,29 @@ head-movement checks defined below.
 - Auto-PRs may show a `BLOCKED` merge state due to branch protection rules requiring review approval — this is normal and not a CI failure. Wait for checks to pass, then merge.
 - You have `git` and `gh` available, and push permission to this repo.
 
+## Authoritative remote (set and validate once)
+
+Use one remote for discovery, baselines, merges, comparisons, and pushes.
+This workflow intentionally does not support separate read and write remotes.
+
+```bash
+REMOTE="${REMOTE:-upstream}"
+REPO="$(gh repo view --json nameWithOwner -q .nameWithOwner)"
+FETCH_REPO="$(gh repo view "$(git remote get-url "$REMOTE")" \
+  --json nameWithOwner -q .nameWithOwner)"
+PUSH_REPO="$(gh repo view "$(git remote get-url --push "$REMOTE")" \
+  --json nameWithOwner -q .nameWithOwner)"
+printf 'Authoritative remote: %s (fetch=%s, push=%s, gh=%s)\n' \
+  "$REMOTE" "$FETCH_REPO" "$PUSH_REPO" "$REPO"
+test "$FETCH_REPO" = "$REPO" && test "$PUSH_REPO" = "$REPO"
+```
+
+The final `test` is mandatory. Stop if it fails, if either URL cannot be
+resolved, or if the selected remote does not expose every branch in the
+propagation chain. Keep `REMOTE` unchanged for the entire A1 → B → A2 run.
+The Phase A manifest must record the same remote name and repository; stop
+if it does not match.
+
 ## Discover the propagation chain
 
 Do not rely on a fixed list or assume every alpha is named `alpha-vN`.
@@ -70,8 +93,8 @@ Do not rely on a fixed list or assume every alpha is named `alpha-vN`.
 1. Fetch and list the actual remote branch names:
 
    ```bash
-   git fetch upstream --prune
-   git for-each-ref refs/remotes/upstream \
+   git fetch "$REMOTE" --prune
+   git for-each-ref "refs/remotes/$REMOTE" \
      --format='%(refname:strip=3)' \
      | grep -E '^(draft-v[0-9]+|alpha-v[0-9]+|v[0-9]+-alpha)$' \
      | sort -V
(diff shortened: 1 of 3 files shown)


jskeet:
Have attempted to read on my mobile, and got lost almost immediately. Will look properly early next week.

.github/prompts/post-meeting-propagate.prompt.md
```diff
@@ -1,12 +1,45 @@
...
 
 You are propagating the changes merged into the current "starting" draft branch (after a TC49-TG2 committee meeting) forward through every future draft and alpha branch. See `admin/branch-diagram.md` for the rationale.
 
+## Phase B of the two-phase post-meeting workflow
+
+The monthly post-meeting work is split into two phases (see
+`post-meeting-rebase-prs.prompt.md` for **Phase A**):
+
+- **Phase A — parallel PR rebase (runs first / in parallel).** Rebase all
+  ~54 open feature PRs onto their own base `draft-vN`. Independent per
```
jskeet: Is having the number of PRs here brittle, even if it's only rough? Do you think it's helping the AI to "check" which PRs it's rebasing?

.github/prompts/post-meeting-propagate.prompt.md
```diff
@@ -1,12 +1,45 @@
...
+  PR/version → embarrassingly parallel. Produces a **fresh `.patch` per PR**
+  that this phase consumes.
+- **Phase B — this prompt: a single serial sweep up the chain, atomic per
+  version.** For each version N in chain order, (1) propagate the
+  committee-merged changes into `draft-vN`, (2) real-renumber `draft-vN`
+  (watch the concept-drift trap), (3) cheap re-rebase only the vN PRs whose
+  base moved in step 1, (4) **rebuild `alpha-vN`** by surgically applying the
+  fresh vN patches onto the EXISTING alpha structure (Option 1 — never a
```
jskeet: I don't quite follow the "option 1" here and elsewhere, as I haven't seen other numbered options. Have I missed the other options, or is this setting up for future possibilities?

.github/prompts/post-meeting-propagate.prompt.md
```diff
@@ -187,15 +234,73 @@ Persist the accumulated list across the whole run. Include it in the final repor
...
+
+1. **Confirm the fresh per-PR patches for version N are current.** Phase A
+   produced them and Step B.8 refreshed any whose base moved. If in doubt,
+   regenerate them (`git format-patch upstream/draft-vN..<headRefName>`).
+2. **Apply each fresh vN patch surgically onto the EXISTING `alpha-vN`
+   structure** — wording-only / subtractive hunks. **Never** rebuild
+   `alpha-vN` from scratch or reorganize its structure (Option 1; see the
+   v10→v11 structural-divergence trap). A from-scratch structural reorg
+   amplifies churn into every downstream branch.
+3. **Watch the renumber concept-drift trap.** The renumber tool keeps a
```
jskeet: I'm fine with letting AI sort this, but are there tool changes which would help too?

.github/prompts/post-meeting-rebase-prs.prompt.md
```diff
@@ -1,18 +1,42 @@
...
+> **Ordering note.** After Phase B propagates the committee changes into
+> `draft-vN`, it does a *cheap re-rebase* of only the vN PRs whose base moved
+> and **refreshes their patches** (propagate prompt, Step B.8). So the patches
+> Phase B ultimately applies are the freshest post-propagation ones; the
+> patches produced here are the starting point.
+
+### Notification / routing steps depend on Phase B output
+
+Steps 0 and 0.5 below (alpha-drift detection and vNext-comment routing)
+consume artifacts that **Phase B produces**:
```
jskeet, line 46: I'm confused at this point - is this "phase A after meeting N+1 consumes phase B output after meeting N" or something else? Or is this not actually about phase A at this point?

If this *is* separate, I think it would be clearer for human readers at least to have three docs:

- A summary doc which refers to phase A and phase B without giving details, and then what happens to the phase B output
- Phase A in detail
- Phase B in detail

But I may be misreading this.

jskeet (approved):
I strongly suspect this is one of those things that if you haven't actually done it, it's really hard to really review. I believe I "somewhat" understand what we're trying to achieve, but I can't say I've checked every detail - and as you'll see by the comments, I'm a bit confused by the structure of the docs, but that may well just be me.

I'm happy for you to merge as-is, or we could potentially meet to go through bits I find confusing that way?

.github/prompts/post-meeting-propagate.prompt.md
BillWagner: You had not missed another option. I removed the unexplained label and replaced it with an explicit surgical alpha-reconstruction policy: apply only the minimal feature delta, do not replay synchronization merges wholesale, and preserve later-version branch language.

.github/prompts/post-meeting-propagate.prompt.md
BillWagner: Agreed. The fixed estimate is gone. Phase A1 now dynamically enumerates and de-duplicates open feature PRs across the discovered draft branches, reports the resulting count, and records it in the pinned manifest; Phase B verifies the count again before reconstruction.

.github/prompts/post-meeting-propagate.prompt.md
BillWagner: Yes. The prompt now requires a raw rg audit for symbolic placeholders across standard/*.md because StandardAnchorTags/TOC002 previously missed retained placeholders. It also adds semantic three-way review for high-conflict normative sections; that part cannot be inferred from a successful patch or renumber run. A future validator enhancement could automate more of the placeholder check, but this workflow no longer relies on the existing tool alone.

.github/prompts/post-meeting-rebase-prs.prompt.md
BillWagner: The chronology is now explicit: A1 prepares and pins PR inputs, B performs the serial propagation/alpha reconstruction, and A2 runs the notification pass using B output. I also reordered the rebase prompt into separate A1 and A2 sections. Phase B may invoke only the narrow rebase/patch machinery when it moves a base. I kept two prompts because A1 and A2 share the same PR-processing context, but the overview and responsibilities are now separated at the top.

PR #1755, opened 2026-07-25
Highlight that contextual keywords cannot contain escapes.

Nigel-Ecma:
While reviewing #1696 I noticed that contextual keywords were not called out in prose as not allowing escapes. This does follow from the grammar as it does not allow them in *contextual_keyword* so calling them out in the prose isn't required. However maybe the list here reads as comprehensive even though its just an “e.g.” (not an “i.e.”), and calling it out seems harmless…

Tagged as C#8 as it may help support PR #1696, even though it is not a *required* change.

### DIFF
--- a/standard/lexical-structure.md
+++ b/standard/lexical-structure.md
@@ -359,7 +359,7 @@ token
 
 ### 6.4.2 Unicode character escape sequences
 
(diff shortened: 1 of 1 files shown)


RexJaeschke (approved):
LGTM, Nigel. BTW, at the very end of 6.4.4 Keywords, in the coverage of contextual keywords, we have the following Note:

> *Note*: When used as contextual keywords, these identifiers cannot contain Unicode_Escape_Sequences. *end note*

BillWagner (approved):
I agree that this is a good change. It reinforces the rule.

jskeet (approved):
Yup, lgtm.

PR #1762, opened 2026-08-06
Fix stale section references on draft-v10

BillWagner:
Repairs the 18 stale symbolic references in `standard/classes.md` using the reviewed numbered destinations.

Validation:
- Raw symbolic-reference scan: clean
- `StandardAnchorTags --dryrun`: passed
- `git diff --check`: passed
- Markdownlint: one pre-existing MD028 failure in the nullable example block


jskeet (approved):
Not sure how we've ended up with a lint check failure on existing text, but presumably that can easily be fixed by an HTML comment

RexJaeschke (approved):

PR #1770, opened 2026-08-11
[Version 13.0] Feature support for Lock objects

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-13.0/lock-object.md).


KalleOlaviNiemitalo:
A loophole occurs to me:

```csharp
class C {
    void M() {
        var l = new System.Threading.Lock();
        lock (Launder(l)) {
        }
    }
    static object Launder<T>(T v) => v;
}
```

But that seems unlikely to happen by accident, and anyone doing it deliberately could just disable the warning instead.

PR #1781, opened 2026-08-13
[Version 13.0] Feature support for ref and unsafe in iterators and async

RexJaeschke:
This is Rex's adaptation of the corresponding [MS proposal](https://github.com/dotnet/csharplang/blob/main/proposals/csharp-13.0/ref-unsafe-in-iterators-async.md).

### DIFF
--- a/standard/statements.md
+++ b/standard/statements.md
@@ -162,10 +162,11 @@ The statement list of a block is reachable if the block itself is reachable.
 
 The end point of a block is reachable if the block is empty or if the end point of the statement list is reachable.
 
-A *block* that contains one or more `yield` statements ([§13.15](statements.md#1315-the-yield-statement)) is called an iterator block. Iterator blocks are used to implement function members as iterators ([§15.15](classes.md#1515-synchronous-and-asynchronous-iterators)). Some additional restrictions apply to iterator blocks:
+A *block* that contains one or more `yield` statements ([§13.15](statements.md#1315-the-yield-statement)) is called an iterator block, even if those `yield` statements are contained only indirectly in nested blocks (excluding nested lambdas and local functions). Iterator blocks are used to implement function members as iterators ([§15.15](classes.md#1515-synchronous-and-asynchronous-iterators)).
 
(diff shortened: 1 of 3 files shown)


standard/statements.md
```diff
@@ -477,7 +478,7 @@ The initializing *variable_reference* shall have type *type* and meet the same r
 
 If *ref_kind* is `ref readonly`, the *identifier*s being declared are references to variables that are treated as read-only. Otherwise, if *ref_kind* is `ref`, the *identifier*s being declared are references to variables that shall be writable.
 
-It is a compile-time error to declare a ref local variable, or a variable of a `ref struct` type, within a method declared with the *method_modifier* `async`, or within an iterator ([§15.15](classes.md#1515-synchronous-and-asynchronous-iterators)).
+It is a compile-time error to declare and use (even implicitly in compiler-synthesized code) a ref local variable, or a variable of a `ref struct` type across `await` expressions or `yield return` statements. More precisely, the error is driven by the following mechanism: after an `await` expression  or a `yield return` statement, all ref local variables and variables of a `ref struct` type in scope are considered definitely unassigned.
```
KalleOlaviNiemitalo, line 481: OK, that means it is permissible to read a ref local variable in unreachable code.

```csharp

public class C {
    public IEnumerator<int> M() {
        int i = 0;
        ref int r = ref i;

        yield return r; // OK
        // r is definitely unassigned, because of the yield return above.

        // yield return r; // error

        yield break;
        // r is again definitely assigned, because this point is unreachable.

        yield return r; // OK
    }
}
```

KalleOlaviNiemitalo, line 481: It is not clear to me that this disallows the following:

```csharp
using System.Threading.Tasks;

public class C {
    public static async Task M(Task<int> t) {
        int i = 0;
        ref int r = ref i;
        K1(ref r, await t); // error
    }
    
    private static void K1(ref int r, int i) {}
}
```

One could claim that `ref r` is evaluated before the `await` expression and `r` is still definitely assigned at that point.  So this depends on the "compiler-synthesized code" for `await` expressions but I'm not sure that is specified rigorously enough to say whether the above is allowed.

PR #1798, opened 2026-09-03
Update v14 test tooling

RexJaeschke:
Bill, I updated each project file to net10.0 per our conversation this morning.


BillWagner (approved):
This is a step in the right direction.

PR #1806, opened 2026-09-13
Restrict switch_expression conversions

### LINKED ISSUE #1739: Check that switch expressions with a natural type do not cause additional conversions
See https://github.com/dotnet/csharpstandard/issues/1651#issuecomment-5020579576 for an example

jskeet:
This prevents two different implicit conversions existing when a switch_expression already has a type.

Fixes #1739


standard/conversions.md
```diff
@@ -388,7 +388,7 @@ While throw expressions do not have a type, they may be implicitly converted to
 
 ### 10.2.18 Switch expression conversion
 
-There is an implicit conversion from a *switch_expression* ([§12.11](expressions.md#1211-switch-expression)) to every type `T` for which there exists an implicit conversion from each *switch_expression_arm*’s *switch_expression_arm_expression*’s to `T`.
+There is an implicit conversion from a *switch_expression* ([§12.11](expressions.md#1211-switch-expression)) which does not have a type, to every type `T` for which there exists an implicit conversion from each *switch_expression_arm*’s *switch_expression_arm_expression*’s to `T`.
```
Nigel-Ecma: Just one *switch_expression_arm_expression* per *switch_expression_arm*:


```suggestion
There is an implicit conversion from a *switch_expression* ([§12.11](expressions.md#1211-switch-expression)) which does not have a type, to every type `T` for which there exists an implicit conversion from each *switch_expression_arm*’s *switch_expression_arm_expression* to `T`.
```

Nigel-Ecma (approved):
Small typo, otherwise LGTM

BillWagner (approved):
I agree that this is ready, with one typo correct from Nigel.

PR #1807, opened 2026-09-13
Add identity conversion for arrays based on element type

### LINKED ISSUE #1792: Specific conversions with arrays-of-arrays and nullable reference types
[This comment](https://github.com/dotnet/csharpstandard/pull/1386#discussion_r2485050637) still needs addressing (or at least probably does).

Examples:

- What are the conversions between `string[][]`, `string[][]?`, `string[]?[]`, `string[]?[]?`, `string?[][]`, `string?[][]?`, `string?[]?[]` and `string?[]?[]?` (if any)?
- What are the conversions between `int[][]`, `int[][]?`, `int[]?[]`, `int[]?[]?`, `int?[][]`, `int?[][]?`, `int?[]?[]` and `int?[]?[]?` (if any)?

@Nigel-Ecma If we can answer those specific examples, do you think that covers us, or do you think there are other aspects we'll need to figure out?

jskeet:
Also fixes a textual reference to the tuple rule, which wasn't updated when we added the nullable reference type identity conversion.

Fixes #1792


Nigel-Ecma (approved):

BillWagner (approved):
This also looks great. I'll :shipit: now.

PR #1808, opened 2026-09-13
Add the "out declaration_expression" form to argument value

### LINKED ISSUE #1786: 12.6.2 Argument lists doesn't handle "out declaration_expression"
12.6.2 (expressions.md, "Argument lists") shows that an *argument_value* can take one of *five* forms... but it only then describes four of them.

Found while trying to figure out the differences for #1662. I may end up fixing it in the same PR.

jskeet:
Note that the "respectively" part following the list is slightly awkward. An alternative would be to have a single bullet for the "out" keyword, with two sub-bullets... although that would remove the natural mapping from the grammar to the list.

Fixes #1786


Nigel-Ecma (approved):

BillWagner (approved):
This LGTM as well. I prefer the "respectively" ordering rather than nested bullets here, with only the two options.

PR #1809, opened 2026-09-13
Make Attribute suffix use consistent

### LINKED ISSUE #1789: Ensure consistency when talking about attributes
In attributes.md, we sometimes refer to attributes without the `Attribute` suffix. An example is for `EnumeratorCancellation`.

My personal take:

- When we're spelling out the full name including namespace, it should *always* be the actual type name, including the suffix
- When we've put the attribute name in backticks (code font) I think we *should* make it the type name including the suffix
- When we're using it without backticks, it's reasonable to omit the suffix

We should discuss this, then once we've got a policy it should be reasonably easy to apply.

jskeet:
- Never use the suffix outside backticks
- Always use the suffix inside backticks
- Tweak the text to avoid stuttering (so don't have the suffix in backticks immediately followed by the word "attribute" outside backticks)

Fixes #1789


BillWagner (approved):
This LGTM @jskeet 

Let's :shipit:

PR #1811, opened 2026-09-14
Fix Attribute name usage

RexJaeschke:
@Jon, here's one more occurrence. Separately, I'll go through all the Draft PR's for V8 and later versions and apply your rules to all new relevant attribute-related text.

@Jon, hold off approving this just yet as I may have more tweaks.

For example, 23.5.6.2 The CallerLineNumber attribute (among other specific attribute subclauses) uses a fully qualified name for the attribute type, as in

> The attribute `System.Runtime.CompilerServices.CallerLineNumberAttribute` is ...

As we state the fully qualified name back in 23.5.1 General, I suggest we omit it from all other uses in narrative. Do you agree? If so, I'll fix those as part of this PR.


jskeet:
> As we state the fully qualified name back in 23.5.1 General, I suggest we omit it from all other uses in narrative. Do you agree? If so, I'll fix those as part of this PR.

By "omit it" do you mean the namespace? If so, I agree. If it's the `Attribute` suffix, then no - but I'm hoping it's the former :)

RexJaeschke:
@jskeet  Yes, I meant "omit the namespace." I did that for the Attribute types in attributes.md. However, I did *not* do that for such types mentioned in classes.md, a clause that uses the namespace prefix `System.` a lot. You can process this PR now.

jskeet (approved):
Yup, this looks fine to me, given that we've got the full namespace elsewhere (and it's a real mouthful).

